PR #214, opened 2026-02-09
fix invoke session id threading

AlexanderRichey:
The PR does two things:

1. It fixes a bug where, when specifying the session ID when calling `invoke` manually, the session ID was not threaded through to the actual API call, causing new sessions to be created on every call.
2. It removes a call to `stopRuntimeSession` in the TUI layer had occurred after every successful invoke, which caused multi-turn and other "sessionful" features to look like they were failing. We don't have to worry about not stopping these sessions—they'll timeout and stop on their own.


notgitika (approved):
LGTM, thanks for fixing!

PR #221, opened 2026-02-10
fix(remove): pass --name flag to TUI flow for direct resource removal

### LINKED ISSUE #171: [P0][CLI] Remove agent command doesn't find specified agent with --name flag
## Description
Specifying which agent to remove does not seem to find the agent that the user wants to remove.

## Suggested Fix
Do not enter interactive mode when --name Agent is specified.

## Category
Functional Bug

## Flow
Flow 4: LangChain Agent Add/Remove Cycle

## Test Path
CLI

## Priority
P0

## Reporter
eshkuma

tejaskash:
## Summary

- Fixes the `--name` flag being ignored when running `agentcore remove agent --name MyAgent` without `--json`
- Previously, the TUI would show the agent selection screen instead of directly removing the named agent

## Changes

- Add `name` parameter to `TUIRemoveOptions` interface and pass it to `RemoveFlow` component
- Add `initialResourceName` prop to `RemoveFlow` that triggers auto-selection via `useEffect`
- Prevent selection screens from rendering while waiting for auto-selection to complete
- Add test case for TUI mode removal with `--name` and `--force` flags

## Test plan

- [x] Unit tests pass: `npm test -- src/cli/commands/remove/__tests__/remove.test.ts`
- [x] Manual test: `agentcore remove agent --name TestAgent --force` removes agent directly without showing selection screen

Closes #171

### DIFF
--- a/src/cli/commands/remove/__tests__/remove.test.ts
+++ b/src/cli/commands/remove/__tests__/remove.test.ts
@@ -96,6 +96,38 @@ describe('remove command', () => {
       expect(json.success).toBe(false);
     });
 
+    it('removes existing agent with --name and --force (TUI mode)', async () => {
+      // Add another agent for this test
+      const addResult = await runCLI(
+        [
+          'add',
+          'agent',
+          '--name',
+          'TUITestAgent',
+          '--language',
+          'Python',
+          '--framework',
+          'Strands',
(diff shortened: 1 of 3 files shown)


notgitika (approved):
nice 🚀

PR #270, opened 2026-02-11
fix: `agentcore create` launching CLI instead of TUI

aidandaly24:
## Description

Running agentcore create without any flags was incorrectly showing --name is required error instead of launching the interactive TUI.

The issue was that options.language was being defaulted to 'Python' before the hasAnyFlag check. Since hasAnyFlag includes options.language in its boolean check, it always
evaluated to true, causing CLI mode to run instead of TUI mode.

Fix moves the language default assignment inside the if (hasAnyFlag) block so it only applies when CLI mode is already determined.

## Related Issues

None

## Documentation PR

N/A

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I ran npm run test:all
- [x] I ran npm run typecheck
- [x] I ran npm run lint
- [ ] If I modified src/assets/, I ran npm run test:update-snapshots and committed the updated snapshots

Added regression test that verifies agentcore create without flags launches TUI (checks for "AgentCore Create" header in output).

## Checklist

### DIFF
--- a/src/cli/commands/create/__tests__/create.test.ts
+++ b/src/cli/commands/create/__tests__/create.test.ts
@@ -206,4 +206,13 @@ describe('create command', () => {
       expect(json.error).toContain('already exists');
     });
   });
+
+  describe('no flags', () => {
+    it('launches TUI when no flags provided', async () => {
+      const result = await runCLI(['create'], testDir);
+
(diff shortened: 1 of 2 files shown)


jesseturner21 (approved):
lgtm

PR #272, opened 2026-02-11
fix: include APP_DIR in agent path for TUI add agent venv setup

tejaskash:
## Summary
- Fixed missing `APP_DIR` ('app/') in the path calculation for `setupPythonProject` in the TUI add agent flow
- The venv was being created in `projectRoot/<agentName>` instead of `projectRoot/app/<agentName>`

## Root Cause
In `useAddAgent.ts`, the `handleCreatePath` function calculated the agent path as:
```typescript
const agentPath = join(projectRoot, config.name);  // Wrong
```

While the CLI flags path in `actions.ts` correctly used:
```typescript
const agentPath = join(projectRoot, APP_DIR, options.name);  // Correct
```

## Test plan
- [ ] Run `agentcore add agent` via TUI and verify `.venv` is created in `app/<agentName>/`
- [ ] Compare behavior with `agentcore create` which should work correctly


aidandaly24 (approved):
looks good to me, thanks for making the fix

PR #291, opened 2026-02-11
fix: add publishConfig for GitHub Packages registry

tejaskash:
## Summary

Fixes npm publish failing with `ENEEDAUTH` error by adding `publishConfig` to specify the GitHub Packages registry.

### Changes
- Added `publishConfig.registry` pointing to `https://npm.pkg.github.com`

### Problem
Without this config, npm defaults to publishing to `npmjs.org` which requires separate authentication.


jesseturner21 (approved):

notgitika (approved):
maybe we also need a github token?

PR #334, opened 2026-02-18
Add container agent support

tejaskash:
## Summary
- Adds Container as a build type for agents across the full lifecycle: create, add, dev, deploy, invoke, package
- Container dev server with Docker build, volume mount, and uvicorn hot-reload
- Dockerfile and dockerignore templates for Python container agents
- Docker prerequisite detection and validation
- Wires `userId` (default: `default-user`) through invoke for container runtime auth (`runtimeUserId` header)
- Upgrades vended `@aws/agentcore-cdk` to `^0.1.0-alpha.2`
- Fixes eslint `require-await` errors in codezip-dev-server and container packaging

## Documentation

PR: #340 

## Test plan
- [x] `agentcore create` with Container build type generates correct project structure
- [x] `agentcore add agent --build Container` adds container agent config
- [x] `agentcore dev` builds and runs container with hot-reload
- [x] `agentcore deploy` builds container image via CodeBuild, pushes to ECR, creates runtime
- [x] `agentcore invoke` works for container agents (userId sent as `default-user`)
- [x] `agentcore package` packages container source correctly
- [x] CodeZip agents continue to work unchanged
- [x] All unit tests pass (1235 passed, 21 skipped)

### DIFF
--- a/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
+++ b/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
@@ -1555,11 +1555,17 @@ def add_numbers(a: int, b: int) -> int:
     return a + b
 
 
-# Set environment variables for model authentication
-load_model()
-
 # Get MCP Toolset
 mcp_toolset = [get_streamable_http_mcp_client()]
+
+_credentials_loaded = False
+
+def ensure_credentials_loaded():
+    global _credentials_loaded
+    if not _credentials_loaded:
+        load_model()
+        _credentials_loaded = True
+
 
 # Agent Definition
 agent = Agent(
@@ -1573,6 +1579,7 @@ agent = Agent(
 
 # Session and Runner
 async def setup_session_and_runner(user_id, session_id):
+    ensure_credentials_loaded()
     session_service = InMemorySessionService()
     session = await session_service.create_session(
         app_name=APP_NAME, user_id=user_id, session_id=session_id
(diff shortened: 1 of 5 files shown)


src/cli/external-requirements/checks.ts
```diff
@@ -122,10 +131,22 @@ export async function checkDependencyVersions(projectSpec: AgentCoreProjectSpec)
     }
   }
 
+  // Check container runtime only if there are Container agents (warn only, not error)
+  let containerRuntimeAvailable = true;
+  if (requiresContainerRuntime(projectSpec)) {
+    const info = await detectContainerRuntime();
+    containerRuntimeAvailable = info !== null;
```
notgitika: shouldnt this be `info.runtime !== null`?

src/cli/operations/dev/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+    server.listen(port, host, () => {
+      server.close(() => resolve(true));
+    });
+    server.on('error', () => resolve(false));
+  });
+}
+
+/** Check if a port is available on both localhost and all interfaces. */
+async function isPortAvailable(port: number): Promise<boolean> {
+  // Check both: container runtimes bind on 0.0.0.0, local servers on 127.0.0.1
```
notgitika: interesting... 

seems like this is why the integ test is failing becuase the old isPortAvailable() only checked `127.0.01`. in this new case, since `0.0.0.0` includes localhost, the test is timing out...

what do you think about changing it to

```
  const loopback = await checkPort(port, '127.0.0.1');
  if (!loopback) return false;
  const allInterfaces = await checkPort(port, '0.0.0.0');
  return allInterfaces;
```

PR #462, opened 2026-02-26
feat: add npm cache ownership preflight check

siwachabhi:
## Summary

- Adds a preflight check that detects root-owned `~/.npm` cache directories before running `npm install`
- A previous `sudo npm install` leaves root-owned files in the cache, causing cryptic EACCES permission errors on subsequent installs
- The check runs in both the **create** (`checkCreateDependencies`) and **deploy** (`checkDependencyVersions`) preflight paths
- On failure, displays the exact fix command: `sudo chown -R $(whoami) ~/.npm`
- Skipped on Windows where file ownership semantics differ

## Motivation

Reported by a user running `agentcore create` who had previously run `sudo npm install` on their machine. The resulting EACCES error in npm output was difficult to diagnose. This check catches it early with a clear, actionable message.

## Changes

- `checks.ts`: Added `checkNpmCacheOwnership()` and `formatNpmCacheError()`, wired into both `checkDependencyVersions()` and `checkCreateDependencies()`
- `index.ts`: Exported new function and types
- `checks-extended.test.ts`: Added unit tests for the new check and error formatter

## Test plan

- [x] `npm run build` passes
- [x] All 64 unit tests pass (`vitest run --project unit src/cli/external-requirements/__tests__/`)
- [x] Pre-commit hooks pass (prettier, eslint, typecheck, secretlint)
- [ ] Manual: verify error message displays correctly when `~/.npm` is root-owned


notgitika (approved):
This PR LGTM. Thanks for adding this!

One small comment: it assumes that the npm cache directory is at `~/.npm`. It seems like that is something users can choose to change.

I'm not sure of a better graceful detection for that though, so the current state is fine with me, given we note this limitation.

notgitika:
Maybe a workaround could be adding a subprocess call that runs `npm config get cache` and captures the right directory there (ref: https://www.geeksforgeeks.org/node-js/npm-cache-command/). 

But I think this adds some complexity and overhead that should not be present for 99% of the users so I'm good to merge the current implementation. Thanks once again!

PR #469, opened 2026-02-27
fix: require gateway when creating a gateway target

tejaskash:
## Summary
- A gateway target must always be attached to a gateway. Previously users could create "unassigned" targets by selecting "Skip for now" in the TUI or omitting `--gateway` in non-interactive mode
- Validation now requires `--gateway`, verifies the specified gateway exists, and errors early if no gateways are configured — with a helpful message directing users to `agentcore add gateway` first
- The TUI gateway selection step no longer offers a "Skip for now" option; the existing `NoGatewaysMessage` blocks the flow when no gateways exist
- The operations layer (`createExternalGatewayTarget`) rejects a missing gateway, removing the `unassignedTargets` code path

## Test plan
- [x] `validate.test.ts` — 3 new tests: missing `--gateway`, no gateways exist, specified gateway not found
- [x] `create-mcp.test.ts` — replaced unassigned-target tests with "throws when gateway not provided"
- [x] All existing gateway target tests updated to include `gateway` field and pass
- [x] Manual: run `agentcore add gateway-target` without a gateway configured → should error
- [x] Manual: run `agentcore add gateway-target --gateway nonexistent` → should list available gateways
- [x] Manual: TUI flow no longer shows "Skip for now" in gateway selection

### DIFF
--- a/src/cli/commands/add/validate.ts
+++ b/src/cli/commands/add/validate.ts
@@ -202,7 +202,8 @@ export async function validateAddGatewayTargetOptions(options: AddGatewayTargetO
   if (!options.gateway) {
     return {
       valid: false,
-      error: "--gateway is required. A gateway target must be attached to a gateway. Create a gateway first with 'agentcore add gateway'.",
+      error:
+        "--gateway is required. A gateway target must be attached to a gateway. Create a gateway first with 'agentcore add gateway'.",
     };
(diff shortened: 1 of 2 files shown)


aidandaly24 (approved):
approved very clean change thank you

PR #593, opened 2026-03-20
feat: add advanced settings gate to agent creation wizard

tejaskash:
## Summary

- Adds a "Configure advanced settings?" Yes/No gate before network configuration in the agent creation wizard (both Create and BYO paths)
- Defaults to "No, use defaults" (Public network, no VPC), letting most users skip network config with one Enter press
- Selecting "Yes, customize" expands into the existing network mode, subnet, and security group steps

### Advanced gate (defaults to "No")
Advanced gate default

### Default path: skips to Confirm with PUBLIC network
Confirm with PUBLIC

### "Yes, customize" expands network config
Yes customize network

## Test plan

- [x] Default path: Enter on Advanced → Confirm shows "Network: PUBLIC"
- [x] Advanced Yes: Yes → Network → VPC → Subnets → SGs → Confirm reflects VPC config
- [x] Back navigation: Esc from Network → Advanced; Esc from Advanced → Memory
- [x] State cleanup: Yes → VPC config → back → No → Confirm shows PUBLIC, no VPC
- [x] MCP protocol: Select MCP → Advanced step still appears (skips Framework/Model as expected)
- [x] BYO path: Bedrock → Advanced gate → defaults → Confirm shows PUBLIC
- [x] Unit tests: 2510/2510 passing (2 pre-existing PTY failures unrelated)
- [x] E2E: container-strands-bedrock test fails — pre-existing on main, unrelated to this change

### DIFF
--- a/src/cli/tui/screens/generate/__tests__/useGenerateWizard.test.tsx
+++ b/src/cli/tui/screens/generate/__tests__/useGenerateWizard.test.tsx
@@ -1,0 +1,274 @@
+import { useGenerateWizard } from '../useGenerateWizard';
+import { Text } from 'ink';
+import { render } from 'ink-testing-library';
+import React, { act, useImperativeHandle } from 'react';
+import { describe, expect, it } from 'vitest';
+
+// ---------------------------------------------------------------------------
+// Imperative harness — exposes wizard methods via ref for act()-based tests
+// ---------------------------------------------------------------------------
+
+type WizardReturn = ReturnType<typeof useGenerateWizard>;
+
+interface HarnessHandle {
+  wizard: WizardReturn;
+}
+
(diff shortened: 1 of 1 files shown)


notgitika (changes requested):
Can we add unit tests for this new step logic? Also, I believe "Protocol" should also fall under Advanced... lmk what you think.

Other concerns:
State cleanup inconsistency between Generate and BYO paths

useGenerateWizard.ts clears VPC config with undefined, but AddAgentScreen.tsx clears with empty strings ''. If downstream code checks !== undefined or serializes to JSON, behavior will diverge. Should use the same sentinel value in both paths.

ADVANCED_OPTIONS has description: undefined for "Yes, customize"

Other option arrays use actual strings for all descriptions. If WizardSelect renders description without a null check, this could render "undefined" as text. Should be description: ' ' or verified against the component's rendering logic.

  ---

PR #630, opened 2026-03-25
feat: support external agent evaluation and trace file evaluation

notgitika:
## Summary

Adds support for evaluating agents **not hosted on AgentCore Runtime** and for evaluating **local trace files** without querying CloudWatch.

### Online Evals (`agentcore add online-eval`)
- Made `agent` optional in the online eval config schema
- Added `customLogGroupName` and `customServiceName` fields with Zod validation (must provide either `agent` OR both custom fields)
- TUI wizard branches between "Project Agent" and "External Agent" log source paths
- CDK construct uses custom fields directly in `dataSourceConfig` when present

### On-Demand Evals (`agentcore run evals`)
- Added `--custom-service-name` and `--custom-log-group-name` CLI flags
- CloudWatch Insights queries filter by `resource.attributes.service.name` instead of `cloud.resource_id` when custom mode is active
- Bypasses `requireProject()` in custom mode (like ARN mode)

### Local Trace Evaluation (`agentcore run evals --input-path`)
- Added `--input-path <path>` flag to evaluate local trace files without querying CloudWatch
- Accepts a single JSON file or a directory of JSON files
- Supports both `traces get` output format (`@message` wrapper) and raw OTel span arrays
- Spans are automatically grouped by `session.id`
- Requires `--region` and `Builtin.*` or `--evaluator-arn` evaluators

```bash
# Single trace file (e.g., from `agentcore traces get`)
agentcore run evals --input-path ./my-trace.json -e Builtin.GoalSuccessRate --region us-east-1

# Directory of trace files
agentcore run evals --input-path ./traces/ -e Builtin.GoalSuccessRate --region us-east-1
```

### Design Decision
Existing project agents continue to use `cloud.resource_id` filtering — no behavior change. The `service.name` filter only activates when `--custom-service-name` is explicitly provided, matching the [AWS observability docs](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/observability-configure.html#observability-configure-3p) where `service.name` is the only reliable identifier for external agents.

## Test plan
- [x] 9 schema tests for custom log source validation (partial fields, neither source, both sources)
- [x] 2 primitive tests for external agent add/field omission
- [x] 7 on-demand eval tests (custom mode resolution, error cases, service.name query, output)
- [x] 8 local trace eval tests (file load, directory load, raw span format, error cases, output)
- [x] All 44 run-eval tests pass
- [x] Typecheck, lint, prettier all pass
- [ ] E2E: `agentcore add online-eval` with external agent path
- [ ] E2E: `agentcore run evals --custom-service-name <name> --custom-log-group-name <group> --region <region> -e Builtin.GoalSuccessRate`
- [ ] E2E: `agentcore run evals --input-path <trace-file> -e Builtin.GoalSuccessRate --region <region>`


src/cli/commands/run/command.tsx
```diff
@@ -58,11 +64,16 @@ export const registerRun = (program: Command) => {
...
+        customServiceName?: string;
+        customLogGroupName?: string;
+        inputPath?: string;
         output?: string;
         json?: boolean;
       }) => {
         const isArnMode = !!(cliOptions.agentArn && cliOptions.evaluatorArn);
-        if (!isArnMode) {
+        const isCustomMode = !!cliOptions.customServiceName;
+        const isInputMode = !!cliOptions.inputPath;
```
jariy17: **`isCustomMode` gating only checks `customServiceName`**

This skips `requireProject()` when only `--custom-service-name` is passed, even if `--custom-log-group-name` is missing. The deeper validation in `resolveFromCustom` catches this, but the user gets a confusing error about the missing log group name rather than being told they're outside a project.

Consider widening the gate so either custom flag triggers non-project mode:

```ts
const isCustomMode = !!(cliOptions.customServiceName || cliOptions.customLogGroupName);
```

That way `resolveFromCustom` gives the precise "both flags required" error in all cases, and users who pass only `--custom-log-group-name` by mistake don't get a misleading "not in an AgentCore project" error.

jariy17 (commented):
## Missing `filter ispresent(scope.name)` in custom-mode CloudWatch queries

Both `discoverSessions` and `fetchSessionSpans` omit the `| filter ispresent(scope.name)` filter when using custom service name mode. The standard mode includes this filter to ensure only actual OTel span documents are returned, but the custom mode branches skip it.

### `fetchSessionSpans` — custom mode is missing the filter:

```typescript
// Custom mode (MISSING the filter):
if (opts.customServiceName) {
    spanQuery = `fields @message, attributes.session.id as sessionId, traceId
     | filter resource.attributes.service.name = '${sanitizeQueryValue(opts.customServiceName)}'`;
    //  ⛔ NO `| filter ispresent(scope.name)` — non-OTel entries slip through
}

// Standard mode (HAS the filter):
else {
    spanQuery = `fields @message, attributes.session.id as sessionId, traceId
     | parse resource.attributes.cloud.resource_id "runtime/*/" as parsedAgentId
     | filter parsedAgentId = '${sanitizeQueryValue(runtimeId!)}'
     | filter ispresent(scope.name)`;
    //  ✅ Only OTel span entries pass through
}
```

### Same issue in `discoverSessions`:

```typescript
if (opts.customServiceName) {
    agentFilter = `filter resource.attributes.service.name = '...'`;
    //  ⛔ No scope.name filter here either
}
```

### Why this matters

External agents are the *most likely* to have noisy, shared log groups — that's the whole point of `--custom-log-group-name`. Without `ispresent(scope.name)`, any log entry matching the `service.name` filter (application logs, error messages, health checks) will be returned and passed downstream as spans. These won't crash — they'll parse as incomplete span documents and produce **silently wrong evaluation results**.

### Suggested fix

Add `| filter ispresent(scope.name)` to both custom-mode query branches:

```typescript
// In fetchSessionSpans:
if (opts.customServiceName) {
    spanQuery = `fields @message, attributes.session.id as sessionId, traceId
     | filter resource.attributes.service.name = '${sanitizeQueryValue(opts.customServiceName)}'
     | filter ispresent(scope.name)`;  // ← add this
}

// In discoverSessions:
if (opts.customServiceName) {
    agentFilter = `filter resource.attributes.service.name = '${sanitizeQueryValue(opts.customServiceName)}'
     | filter ispresent(scope.name)`;  // ← add this
}
```

jariy17 (commented):
## No mutual exclusion between `--input-path`, `--custom-service-name`, and `--agent-arn`

If a user passes multiple mode flags (e.g. `--input-path ./trace.json --agent-arn arn:aws:...`), the if/else chain silently drops all but the first match:

```typescript
if (options.inputPath) {
    resolution = resolveFromInput(options);        // wins
} else if (options.customServiceName) {
    resolution = resolveFromCustom(options);        // silently ignored if inputPath set
} else if (options.agentArn) {
    resolution = resolveFromArn(options);           // silently ignored if either above set
} else {
    // project mode
}
```

No warning, no error — the extra flags are just swallowed.

### Suggested fix

Add an early validation guard before the mode routing, ideally using a Zod schema with `.refine()` on `RunEvalOptions` to enforce mutual exclusivity at the type/validation level:

```typescript
const RunEvalModeSchema = z.object({
  inputPath: z.string().optional(),
  customServiceName: z.string().optional(),
  agentArn: z.string().optional(),
}).refine(
  data => {
    const modes = [data.inputPath, data.customServiceName, data.agentArn].filter(Boolean);
    return modes.length <= 1;
  },
  {
    message: 'Only one of --input-path, --custom-service-name, or --agent-arn may be specified',
  }
);
```

Alternatively, a simpler imperative check at the top of `handleRunEval` works too:

```typescript
const modeFlags = [options.inputPath && '--input-path', options.customServiceName && '--custom-service-name', options.agentArn && '--agent-arn'].filter(Boolean);
if (modeFlags.length > 1) {
  return { success: false, error: `Only one mode flag may be specified, but got: ${modeFlags.join(', ')}` };
}
```

Either way, the user should get a clear error instead of silent flag dropping.

jariy17 (commented):
## No mutual exclusion between `--input-path`, `--custom-service-name`, and `--agent-arn`

If a user passes multiple mode flags (e.g. `--input-path ./trace.json --agent-arn arn:aws:...`), the if/else chain silently drops all but the first match:

```typescript
if (options.inputPath) {
    resolution = resolveFromInput(options);        // wins
} else if (options.customServiceName) {
    resolution = resolveFromCustom(options);        // silently ignored if inputPath set
} else if (options.agentArn) {
    resolution = resolveFromArn(options);           // silently ignored if either above set
} else {
    // project mode
}
```

No warning, no error — the extra flags are just swallowed.

### Suggested fix

Add an early guard at the top of `handleRunEval`:

```typescript
const modeFlags = [
  options.inputPath && '--input-path',
  options.customServiceName && '--custom-service-name',
  options.agentArn && '--agent-arn',
].filter(Boolean);

if (modeFlags.length > 1) {
  return { success: false, error: `Only one mode flag may be specified, but got: ${modeFlags.join(', ')}` };
}
```

The user should get a clear error instead of silent flag dropping.

src/schema/schemas/primitives/online-eval-config.ts
```diff
@@ -13,19 +13,61 @@ export const OnlineEvalConfigNameSchema = z
...
+    /** Evaluator names (custom), Builtin.* IDs, or evaluator ARNs */
+    evaluators: z.array(z.string().min(1)).min(1, 'At least one evaluator is required'),
+    /** Sampling rate as a percentage (0.01 to 100) */
+    samplingRate: z.number().min(0.01).max(100),
+    /** Optional description for the online eval config */
+    description: z.string().max(200).optional(),
+    /** Whether to enable execution on create (default: true) */
+    enableOnCreate: z.boolean().optional(),
+    /** Custom CloudWatch log group name for evaluating agents not hosted on AgentCore Runtime */
+    customLogGroupName: LogGroupNameSchema.optional(),
```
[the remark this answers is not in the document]
notgitika: fixed

src/cli/commands/run/command.tsx
notgitika: fixed

notgitika:
Addressed all three review items:

1. **`ispresent(scope.name)` filter** — Added to both `discoverSessions` and `fetchSessionSpans` custom-mode query branches. External agents with noisy shared log groups will no longer pick up non-OTel entries.

2. **Mutual exclusion between mode flags** — Added an early guard in `handleRunEval` that rejects conflicting flags (`--input-path`, `--custom-service-name`, `--agent-arn`) with a clear error message instead of silently dropping them.

3. **`isCustomMode` gate** — Widened to trigger on either custom flag so the user always gets the precise "both flags required" error from `resolveFromCustom`.

All 94 eval tests pass, tsc clean.

tejaskash:
Closing this out, seems to be stale

PR #687, opened 2026-03-26
fix(e2e): prevent API key credential provider quota exhaustion

jesseturner21:
## Summary

- E2E tests for non-Bedrock model providers (OpenAI, Gemini, Anthropic) were failing because the test account hit its API key credential provider quota limit
- Stale providers accumulated across CI runs because teardown deleted providers **after** CFN stack teardown, which frequently fails when the original deploy never succeeded
- This has been broken on `main` for all recent runs — not caused by any specific PR

### `e2e-tests/e2e-helper.ts`
- **Move credential provider deletion before CFN teardown** in `teardownE2EProject()` — ensures cleanup runs even if `remove all` + `deploy` fails
- **Add `cleanupStaleCredentialProviders()`** called in `beforeAll` — purges any leaked `E2e*` providers from prior runs before tests start
- **Add error logging** instead of silently swallowing deletion failures

### `.github/workflows/e2e-tests.yml`
- **Add "Clean up stale credential providers" step** before running tests — belt-and-suspenders defense using AWS CLI directly

## Root Cause

1. E2E creates API key credential providers during `agentcore deploy` (pre-deploy identity step)
2. Teardown runs: `remove all` → `deploy --yes` (to destroy CFN stack) → `DeleteApiKeyCredentialProviderCommand`
3. When the CFN teardown fails (common when deploy itself never succeeded), the credential provider deletion at the end still runs but is **best-effort with a silent catch**
4. Over many CI runs, providers accumulate → account quota hit → all non-Bedrock deploys fail → more providers leak (snowball)

## Test plan

- [x] TypeScript typecheck passes
- [x] ESLint passes
- [x] Prettier passes
- [x] Secretlint passes
- [x] `ListApiKeyCredentialProvidersCommand` import verified in SDK
- [ ] E2E tests pass after manual cleanup of 50 stale providers (re-run triggered on PR #620)

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -98,6 +98,18 @@ jobs:
       - run: npm run build
       - name: Install CLI globally
         run: npm install -g "$(npm pack | tail -1)"
+      - name: Clean up stale credential providers
+        run: |
+          # Delete leaked E2e* API key credential providers from previous CI runs
+          # to prevent account quota exhaustion
+          for name in $(aws bedrock-agentcore-control list-api-key-credential-providers \
+            --region ${{ inputs.aws_region || 'us-east-1' }} \
+            --query 'credentialProviders[?starts_with(name, `E2e`)].name' \
+            --output text); do
+              echo "Deleting stale provider: $name"
+              aws bedrock-agentcore-control delete-api-key-credential-provider \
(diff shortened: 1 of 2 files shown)


Hweinstock (commented):
had a similar change https://github.com/aws/agentcore-cli/pull/660, I'm good to go with either approach if we can avoid the race condition.

PR #794, opened 2026-04-08
feat(templates): add conversation history persistence to HTTP agent templates

### LINKED ISSUE #808: OpenAI Agents, Google ADK, and LangGraph HTTP templates are stateless — no conversation history across invocations
### Description

The OpenAI Agents, Google ADK, and LangGraph HTTP agent templates have no conversation history persistence. Every invocation starts with zero context — the agent cannot recall anything from prior turns within the same session.

For example, if a user says "my name is Aidan" and then in a follow-up invocation asks "what's my name?", the agent has no idea. Each invoke is treated as a completely independent interaction.

### Steps to Reproduce

1. `agentcore create --name test --defaults --framework OpenAIAgents` (or GoogleADK, or LangChain_LangGraph)
2. `agentcore dev`
3. Invoke with `{"prompt": "my name is Aidan"}`
4. Invoke again with `{"prompt": "what is my name?"}`

### Expected Behavior

The agent should recall "Aidan" from the previous turn within the same session, since both invocations share the same session context.

### Actual Behavior

The agent responds with something like "I don't know your name" because it has no memory of the prior invocation.

### Additional Context

This affects three of the four HTTP templates:
- **OpenAI Agents**: No session persistence mechanism — each invocation creates a fresh `Runner.run()` call with no prior context.
- **Google ADK**: `InMemorySessionService` creates a new session every invocation via `create_session()`, discarding prior turns.
- **LangGraph**: No checkpointer configured — the graph has no mechanism to persist state between invocations.

The Strands template is NOT affected by this issue — it reuses the `Agent` instance which accumulates messages internally (though it has a separate conversation bleed bug).

### LINKED ISSUE #809: Strands HTTP template leaks conversation history across sessions and users
### Description

The Strands HTTP agent template uses a single global `_agent` instance shared across all sessions and users. Since the Strands `Agent` accumulates messages internally in `self.messages`, conversation history from one user/session bleeds into another.

If user A has a conversation, user B's next invocation will see user A's messages in the agent's context.

This violates the AgentCore Runtime's session model. The runtime provides a `session_id` on every invocation via `context.session_id` — the expectation is that agents use this to isolate conversation state per session. The Strands template ignores it entirely and routes all invocations through a single shared agent instance.

### Steps to Reproduce

1. `agentcore create --name test --defaults --framework Strands`
2. `agentcore dev`
3. Invoke as session A: `{"prompt": "my name is Alice, this is a secret"}`
4. Invoke as session B (different session ID): `{"prompt": "what secrets do you know?"}`

### Expected Behavior

Session B should have no knowledge of session A's conversation. Each session should be isolated — this is how AgentCore Runtime is designed to work. The runtime assigns distinct `session_id` values to each session, and agents are expected to maintain separate conversation state per session.

### Actual Behavior

Session B can see session A's conversation history because both sessions share the same global `_agent` instance. The agent may reveal information from session A to session B.

### Additional Context

The root cause is in the template's agent initialization pattern:

```python
_agent = None

def get_or_create_agent():
    global _agent
    if _agent is None:
        _agent = Agent(...)
    return _agent
```

A single `Agent` instance is reused for every invocation regardless of session ID or user ID. Since Strands agents accumulate messages internally, all conversations share the same message history.

The fix is to key agent instances by `context.session_id` so each session gets its own isolated agent with its own conversation history, matching the session isolation model that AgentCore Runtime expects.

### LINKED ISSUE #810: OpenAI Agents and LangGraph HTTP templates have no system prompt
### Description

The OpenAI Agents and LangGraph HTTP agent templates ship with no system prompt. The agent has zero baseline instructions — it doesn't know it's supposed to be helpful, use tools, or behave in any particular way.

The Strands and Google ADK templates both include a system prompt (`"You are a helpful assistant. Use tools when appropriate."`), but OpenAI Agents and LangGraph were missed.

### Steps to Reproduce

1. `agentcore create --name test --defaults --framework OpenAIAgents` (or `LangChain_LangGraph`)
2. Inspect `app/test/main.py`
3. Observe there is no `instructions=` on the OpenAI `Agent()` constructor, or no `prompt=` on `create_react_agent()`

### Expected Behavior

All templates should include a baseline system prompt consistent with the other frameworks, so agents have a default personality and know to use available tools.

### Actual Behavior

- **OpenAI Agents**: `Agent()` constructor has no `instructions=` parameter — the agent runs with no system prompt.
- **LangGraph**: `create_react_agent()` has no `prompt=` parameter — same issue.

### Additional Context

This is inconsistent with the Strands and Google ADK templates which both include:
```
"You are a helpful assistant. Use tools when appropriate."
```

Without a system prompt, the agent's behavior is unpredictable and it may not reliably use the tools registered to it.

aidandaly24:
## Description

All four HTTP agent templates (Strands, OpenAI Agents, Google ADK, LangGraph) were stateless per-invocation — each call started with zero conversation context, so multi-turn recall within a session didn't work. Strands additionally had a session-bleed risk because all invocations sharing a Python process used a single global `_agent` whose `Agent.messages` accumulated across calls.

### Note on Strands and Runtime session isolation

AgentCore Runtime isolates each `session_id` to its own dedicated microVM (terminated and memory-sanitized at session end), so the Strands cross-session bleed cannot manifest in a deployed AgentCore Runtime agent. The bleed is real, however, in:
- `agentcore dev` (one local process serving multiple sessions — the exact reproduction in #809)
- Any deployment of the template outside AgentCore Runtime (Lambda, ECS, FastAPI, etc.) where one process may serve multiple sessions

Keying the agent by `session_id` makes the template correct independent of the runtime's isolation guarantee, costs the same memory profile as the previous global, and aligns Strands with the per-session pattern the other three templates use.

### Per-template fixes

- **Strands**: replace global `_agent` with `@lru_cache(maxsize=128) get_or_create_agent(session_id)`. Strands `Agent` accumulates messages internally across `stream_async()` calls, so reusing the per-session instance gives natural multi-turn memory and isolates sessions. Bound at 128 entries to cap process memory; evicted sessions silently start fresh on their next invoke (documented in a code comment, suggesting durable session stores for production). Applied to both the no-memory/no-config-bundle branch and the no-memory/config-bundle branch — `hasMemory` already keys per `(session_id, user_id)`. The `ConfigBundleHook` callbacks re-read bundle state at invocation time, so caching the agent is safe and consistent with how `hasMemory` already attaches the hook.
- **OpenAI Agents**: `@lru_cache(maxsize=128) get_session(session_id)` returning `SQLiteSession`, threaded via `session=` on `Runner.run()`. Auto-loads/saves prior history in-process. Same 128-session cap and documentation comment as Strands.
- **Google ADK**: module-level `_session_service` + `_runner` with a `get_or_create_session()` helper that calls `get_session()` then `create_session()` if missing — avoids `AlreadyExistsError` on repeat calls. ADK's `InMemorySessionService` is itself a single instance retained across invocations, so no per-session cache cap is needed.
- **LangGraph**: module-level `_checkpointer = InMemorySaver()` passed to `create_react_agent()`. `thread_id` mapped to `context.session_id` via `configurable` on `graph.ainvoke()`. `InMemorySaver` retains all checkpoints in process memory.

All conversation state is in-memory (best-effort) — persists across invocations within the runtime process and resets on cold starts. This is the correct default for starter templates; users can swap to durable backends (Strands `FileSessionManager`, OpenAI `SQLiteSession` with file path, ADK `DatabaseSessionService`, LangGraph persistent checkpointers) when they need durability.

Note: #810 (missing system prompts on OpenAI Agents and LangGraph) was already addressed by other PRs that landed on `main` after this PR was opened — both templates now ship `INSTRUCTIONS` / `DEFAULT_SYSTEM_PROMPT` constants. This PR's scope is therefore #808 + #809.

## Related Issue

Closes #808
Closes #809

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test:unit` (275 test files passing)
- [x] `npm run test:update-snapshots`

**End-to-end deploy + invoke testing (from earlier iteration of this branch):**
- Strands (Bedrock): Deployed, invoked 3x — recall ✅, session isolation ✅
- OpenAI Agents (OpenAI): Deployed, invoked 3x — recall ✅, session isolation ✅
- LangGraph (Bedrock): Deployed, invoked 3x — recall ✅, session isolation ✅
- Google ADK (Gemini): Deployed, code structurally verified — Gemini API quota prevented live invocation testing

## Checklist

---

### DIFF
--- a/src/assets/python/http/openaiagents/base/main.py
+++ b/src/assets/python/http/openaiagents/base/main.py
@@ -99,6 +99,9 @@ You have persistent storage at {{sessionStorageMountPath}}. Use file tools to re
 """
 
 
+# Caches up to 128 active sessions; LRU eviction silently resets history for
+# the oldest session. For production use, replace with a durable session store
+# (e.g. SQLiteSession with a file path).
 @lru_cache(maxsize=128)
 def get_session(session_id):
     return SQLiteSession(session_id)
--- a/src/assets/python/http/openaiagents/base/main.py
+++ b/src/assets/python/http/openaiagents/base/main.py
@@ -1,5 +1,6 @@
 import os
-from agents import Agent, Runner, function_tool
+from functools import lru_cache
+from agents import Agent, Runner, SQLiteSession, function_tool
 from bedrock_agentcore.runtime import BedrockAgentCoreApp
 from model.load import load_model
 {{#if hasGateway}}
@@ -97,8 +98,14 @@ You have persistent storage at {{sessionStorageMountPath}}. Use file tools to re
 {{/if}}
 """
 
+
+@lru_cache(maxsize=128)
+def get_session(session_id):
+    return SQLiteSession(session_id)
+
+
 # Define the agent execution
-async def main(query):
+async def main(query, session):
     ensure_credentials_loaded()
     try:
         {{#if hasGateway}}
@@ -110,7 +117,7 @@ async def main(query):
                 mcp_servers=mcp_servers,
                 tools=tools
             )
-            result = await Runner.run(agent, query)
+            result = await Runner.run(agent, query, session=session)
             return result
         else:
             agent = Agent(
@@ -120,7 +127,7 @@ async def main(query):
                 mcp_servers=[],
                 tools=tools
             )
-            result = await Runner.run(agent, query)
+            result = await Runner.run(agent, query, session=session)
             return result
         {{else}}
         if mcp_servers:
@@ -133,7 +140,7 @@ async def main(query):
                     mcp_servers=active_servers,
                     tools=tools
                 )
-                result = await Runner.run(agent, query)
+                result = await Runner.run(agent, query, session=session)
                 return result
         else:
             agent = Agent(
@@ -143,7 +150,7 @@ async def main(query):
                 mcp_servers=[],
                 tools=tools
             )
-            result = await Runner.run(agent, query)
+            result = await Runner.run(agent, query, session=session)
             return result
         {{/if}}
     except Exception as e:
@@ -157,9 +164,11 @@ async def invoke(payload, context):
 
     # Process the user prompt
     prompt = payload.get("prompt", "What can you help me with?")
+    session_id = getattr(context, "session_id", "default-session")
+    session = get_session(session_id)
 
-    # Run the agent
-    result = await main(prompt)
+    # Run the agent (session automatically loads/saves conversation history)
+    result = await main(prompt, session)
 
     # Return result
     return {"result": result.final_output}
--- a/src/assets/python/http/strands/base/main.py
+++ b/src/assets/python/http/strands/base/main.py
@@ -165,8 +165,12 @@ def agent_factory():
     return get_or_create_agent
 get_or_create_agent = agent_factory()
 {{else}}
+# Caches up to 128 active sessions; LRU eviction silently resets history for
+# the oldest session. For production use, replace with a durable session store
+# (e.g. Strands FileSessionManager).
 {{#if hasConfigBundle}}
-def create_agent():
+@lru_cache(maxsize=128)
+def get_or_create_agent(session_id):
     return Agent(
         model=load_model(),
         system_prompt=DEFAULT_SYSTEM_PROMPT,
@@ -194,12 +198,8 @@ async def invoke(payload, context):
     user_id = getattr(context, 'user_id', 'default-user')
     agent = get_or_create_agent(session_id, user_id)
 {{else}}
-{{#if hasConfigBundle}}
-    agent = create_agent()
-{{else}}
     session_id = getattr(context, 'session_id', 'default-session')
     agent = get_or_create_agent(session_id)
-{{/if}}
 {{/if}}
 
     # Execute and format response
--- a/src/assets/python/http/strands/base/main.py
+++ b/src/assets/python/http/strands/base/main.py
@@ -1,3 +1,4 @@
+from functools import lru_cache
 from typing import Any
 
 from strands import Agent, tool
@@ -173,17 +174,13 @@ def create_agent():
         hooks=[ConfigBundleHook()],
     )
 {{else}}
-_agent = None
-
-def get_or_create_agent():
-    global _agent
-    if _agent is None:
-        _agent = Agent(
-            model=load_model(),
-            system_prompt=DEFAULT_SYSTEM_PROMPT,
-            tools=tools
-        )
-    return _agent
+@lru_cache(maxsize=128)
+def get_or_create_agent(session_id):
+    return Agent(
+        model=load_model(),
+        system_prompt=DEFAULT_SYSTEM_PROMPT,
+        tools=tools
+    )
 {{/if}}
 {{/if}}
 
@@ -200,7 +197,8 @@ async def invoke(payload, context):
 {{#if hasConfigBundle}}
     agent = create_agent()
 {{else}}
-    agent = get_or_create_agent()
+    session_id = getattr(context, 'session_id', 'default-session')
+    agent = get_or_create_agent(session_id)
 {{/if}}
 {{/if}}
 
--- a/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
+++ b/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
@@ -3417,21 +3417,39 @@ agent = Agent(
 )
 
 
-# Session and Runner
-async def setup_session_and_runner(user_id, session_id):
-    ensure_credentials_loaded()
-    session_service = InMemorySessionService()
-    session = await session_service.create_session(
+# Module-level session service and runner (preserves history across invocations)
+_session_service = InMemorySessionService()
+_runner = None
+
+
+def get_or_create_runner():
+    global _runner
+    if _runner is None:
+        ensure_credentials_loaded()
+        _runner = Runner(
+            agent=agent,
+            app_name=APP_NAME,
+            session_service=_session_service,
+        )
+    return _runner
+
+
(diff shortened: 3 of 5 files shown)


src/assets/python/http/strands/base/main.py
```diff
@@ -55,19 +55,18 @@ def get_or_create_agent(session_id, user_id):
     return get_or_create_agent
 get_or_create_agent = agent_factory()
 {{else}}
-_agent = None
+_agents = {}
```
notgitika: this dict grows unbounded and becomes heavy, every new session_id will add an Agent instance (with its full message history) that's never evicted. same applies to `_sessions` in the OpenAI template.

ADK and LangGraph don't have this issue since they use a single service/checkpointer instance.

I'm fine with this for a starter template but should we at least add a comment about this? We can also try to improve this by adding an LRU cache evictor? Is that overkill? 😓  Your call

src/assets/python/http/openaiagents/base/main.py
```diff
@@ -68,16 +78,17 @@ async def main(query):
                     mcp_servers=active_servers,
                     tools=[add_numbers]
                 )
```
notgitika, line 145: you added instructions everywhere else but seems like you missed it here.

notgitika (changes requested):
2 comments but LGTM otherwise

src/assets/python/http/strands/base/main.py
aidandaly24: > try to improve this by adding an LRU cache evictor? Is that overkill?

I had the same thought I'm not sure. Will discuss in standup.

aidandaly24: Added LRU cache

src/assets/python/http/openaiagents/base/main.py
aidandaly24: Fixed

aidandaly24:
@notgitika Pushed a fresh rebase onto main with both of your review comments addressed:
- Strands: `@lru_cache(maxsize=128)` on `get_or_create_agent` (caps unbounded growth)
- OpenAI Agents: `instructions=INSTRUCTIONS` on all four `Agent()` constructors (now using the existing `INSTRUCTIONS` constant rather than inlining the string)

Also dropped #810 from scope since both LangGraph and OpenAI templates already have system prompts on main from other PRs that landed in the meantime. Scope is now #808 + #809.

All checks green; ready for re-review when you have a chance.

src/assets/python/http/strands/base/main.py
```diff
@@ -173,17 +174,13 @@ def create_agent():
...
-def get_or_create_agent():
-    global _agent
-    if _agent is None:
-        _agent = Agent(
-            model=load_model(),
-            system_prompt=DEFAULT_SYSTEM_PROMPT,
-            tools=tools
-        )
-    return _agent
+@lru_cache(maxsize=128)
```
notgitika, line 181: what happens for turn 129? they would lose the whole history? 

if so, can we add some logging so the user is aware?

src/assets/python/http/strands/base/main.py
```diff
@@ -173,17 +174,13 @@ def create_agent():
         hooks=[ConfigBundleHook()],
     )
```
padmak30, line 179: The config bundle path should also get conversation persistence across sessions

src/assets/python/http/strands/base/main.py
padmak30: When the 129th session arrives, the session #1's agent will be evicted and its message history lost. The next call to session #1, will get a new Agent with no prior memory without any error/warning. lru_cache doesn't support eviction callbacks, so this cannot be logged. Maybe add a comment in the template, "Caches up to 128 active sessions; LRU eviction silently resets history for the oldest session. For production use, replace with a durable session store."

aidandaly24: Yep — turn 129 evicts session #1, and the next call to session #1 starts fresh with no warning. `functools.lru_cache` doesn't expose an eviction callback so we can't log on it directly. Took @padmak30's suggestion below and added a documentation comment above the cache in both Strands and OpenAI templates explaining the cap and pointing to durable session stores (Strands `FileSessionManager`, `SQLiteSession` with file path) for production use.

src/assets/python/http/strands/base/main.py
aidandaly24: Good catch — fixed in the latest commit. The `hasConfigBundle` branch now uses the same `@lru_cache(128) get_or_create_agent(session_id)` pattern as the non-bundle branch, with `hooks=[ConfigBundleHook()]` attached at construction. Verified safe to cache: `ConfigBundleHook` reads bundle state inside its callbacks (`_inject_system_prompt`, `_override_tool_desc`) at invocation time, not at hook registration, so reusing the same `Agent` across invocations still picks up bundle changes. The `hasMemory` branch already attaches the hook the same way to a cached agent, so this matches existing precedent in the template.

src/assets/python/http/strands/base/main.py
aidandaly24: Used your suggested wording (lightly adapted). Added the comment above the `@lru_cache` in both Strands and OpenAI Agents templates pointing users to `FileSessionManager` / `SQLiteSession` with a file path for production use. Thanks for the clean phrasing.

aidandaly24:
Superseded by #1639, which carries the complete per-session-isolation fix across all five HTTP templates (strands, openaiagents, googleadk, langchain_langgraph, autogen) with bounded LRU eviction, validated by deploy+invoke. Closing in favor of #1639.

PR #828, opened 2026-04-13
feat: add e2e tests for import command

jesseturner21:
## Description

Add end-to-end tests for `import runtime`, `import memory`, and `import evaluator` commands against real AWS resources.

Python fixture scripts create resources via the Bedrock AgentCore control plane API in `beforeAll`, then the tests import them into a CLI project and verify status + invocation. A cleanup script in `afterAll` tears down the CFN stack and deletes any leftover resources.

Key design decisions:
- Setup scripts run sequentially because `save_resource()` does a read-modify-write on a shared `bugbash-resources.json` file
- The default evaluator model ID is passed from the CLI TypeScript source to the Python setup script via env var to stay in sync automatically
- `cleanup_resources.py` preserves the resource file on partial failure so leaked resources can be retried

## Related Issue

Closes #

## Documentation PR

N/A — no user-facing changes.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I ran `npm run lint`

Additional testing:
- All 5 e2e import tests pass locally against real AWS (us-east-1)
- Full e2e suite passes in CI via `e2e-tests-full.yml` workflow dispatch

## Checklist

---

terms of your choice.

### DIFF
--- a/e2e-tests/fixtures/import/cleanup_resources.py
+++ b/e2e-tests/fixtures/import/cleanup_resources.py
@@ -3,13 +3,37 @@
 
 Called from afterAll in import e2e tests as a fallback cleanup
 for resources that were not successfully imported into CloudFormation.
+
+Note: The IAM role (bugbash-agentcore-role) is intentionally left in place —
+it is shared across test runs via ensure_role() in common.py.
 """
 import json
 import os
 import sys
 
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
-from common import RESOURCES_FILE, get_control_client
+from common import REGION, RESOURCES_FILE, get_control_client, get_account_id
+
+import boto3
+
+
+def cleanup_s3_code_objects():
+    """Delete uploaded code.zip objects from the bugbash S3 bucket."""
+    account_id = get_account_id()
+    bucket_name = f"bugbash-agentcore-code-{account_id}-{REGION}"
+    s3 = boto3.client("s3", region_name=REGION)
+    try:
+        resp = s3.list_objects_v2(Bucket=bucket_name)
+        objects = resp.get("Contents", [])
+        if not objects:
+            return
+        s3.delete_objects(
+            Bucket=bucket_name,
+            Delete={"Objects": [{"Key": o["Key"]} for o in objects]},
+        )
+        print(f"Deleted {len(objects)} object(s) from s3://{bucket_name}")
+    except Exception as e:
+        print(f"Could not clean up S3 objects: {e}")
 
 
 def main():
@@ -47,6 +71,8 @@ def main():
     else:
         os.remove(RESOURCES_FILE)
         print("Cleaned up bugbash-resources.json")
+
+    cleanup_s3_code_objects()
 
 
 if __name__ == "__main__":
--- a/e2e-tests/fixtures/import/cleanup_resources.py
+++ b/e2e-tests/fixtures/import/cleanup_resources.py
@@ -9,7 +9,7 @@ import os
 import sys
 
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
-from common import REGION, RESOURCES_FILE, get_control_client
+from common import RESOURCES_FILE, get_control_client
 
 
 def main():
--- a/e2e-tests/fixtures/import/cleanup_resources.py
+++ b/e2e-tests/fixtures/import/cleanup_resources.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python3
+"""Delete all resources tracked in bugbash-resources.json.
+
+Called from afterAll in import e2e tests as a fallback cleanup
+for resources that were not successfully imported into CloudFormation.
+"""
+import json
+import os
+import sys
+
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+from common import REGION, RESOURCES_FILE, get_control_client
+
+
+def main():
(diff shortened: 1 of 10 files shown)


e2e-tests/fixtures/import/setup_runtime_basic.py
```diff
@@ -0,0 +1,68 @@
...
+        protocolConfiguration={"serverProtocol": "HTTP"},
+    )
+
+    runtime_id = resp["agentRuntimeId"]
+    runtime_arn = resp["agentRuntimeArn"]
+    print(f"Runtime ID: {runtime_id}")
+    print(f"Runtime ARN: {runtime_arn}")
+
+    save_resource("runtime-basic", runtime_arn, runtime_id)
+    wait_for_runtime(client, runtime_id)
```
jariy17: `wait_for_runtime()` returns `False` when the resource fails to reach READY or times out, but the return value is discarded. The script exits 0 regardless, so `beforeAll` proceeds with a broken resource ARN, causing confusing downstream test failures that don't point to the root cause.

Same pattern in `setup_memory_full.py:63` and `setup_evaluator.py:68`.

```suggestion
    if not wait_for_runtime(client, runtime_id):
        sys.exit(1)
```

e2e-tests/fixtures/import/common.py
```diff
@@ -0,0 +1,253 @@
+"""Shared helpers for bugbash setup scripts."""
+import json
+import os
+import time
+import zipfile
+import tempfile
+
+import boto3
+
+REGION = os.environ.get("AWS_REGION") or os.environ.get("AWS_DEFAULT_REGION") or "us-west-2"
```
jariy17: The TS test file and `e2e-helper.ts` both default to `us-east-1`. While `AWS_REGION` is passed explicitly from the test, anyone running these Python scripts standalone for debugging will create resources in `us-west-2` that don't match the CLI's expectations.

```suggestion
REGION = os.environ.get("AWS_REGION") or os.environ.get("AWS_DEFAULT_REGION") or "us-east-1"
```

e2e-tests/fixtures/import/cleanup_resources.py
```diff
@@ -0,0 +1,53 @@
...
+        with open(RESOURCES_FILE, "w") as f:
+            json.dump(remaining, f, indent=2)
+        print(f"WARNING: {len(failed)} resources could not be deleted, kept in {RESOURCES_FILE}")
+    else:
+        os.remove(RESOURCES_FILE)
+        print("Cleaned up bugbash-resources.json")
+
+
+if __name__ == "__main__":
+    main()
```
jariy17, line 79: The cleanup script only deletes runtimes/memories/evaluators. The S3 bucket (`bugbash-agentcore-code-{account}-{region}`) and its uploaded `code.zip` objects accumulate on every run and are never cleaned up. The IAM role (`bugbash-agentcore-role`) with `AmazonBedrockFullAccess` persists indefinitely.

For the IAM role: if it's intentionally persistent/shared across runs, a comment documenting that decision would be sufficient. For S3: at minimum, the uploaded code objects should be deleted in cleanup.

e2e-tests/import-resources.test.ts
```diff
@@ -0,0 +1,212 @@
...
+
+  afterAll(async () => {
+    // 1. Tear down CFN stack created by import (this deletes all imported resources)
+    if (projectPath && hasAws) {
+      await runAgentCoreCLI(['remove', 'all', '--json'], projectPath);
+      const deployResult = await runAgentCoreCLI(['deploy', '--yes', '--json'], projectPath);
+      if (deployResult.exitCode !== 0) {
+        console.warn('Teardown deploy failed:', deployResult.stderr);
+      }
+    }
```
jariy17, line 99: Consider reusing `teardownE2EProject()` from `e2e-helper.ts` instead of reimplementing the `remove all` + `deploy` pattern inline. The shared helper also logs stdout on failure (not just stderr). You can pass `'Bedrock'` as `modelProvider` to skip credential provider cleanup:

```ts
import { teardownE2EProject } from './e2e-helper.js';
// ...
await teardownE2EProject(projectPath, agentName, 'Bedrock');
```

jariy17 (commented):
A few items to address — see inline comments.

e2e-tests/fixtures/import/setup_runtime_basic.py
jesseturner21: Good catch, fixed.

e2e-tests/fixtures/import/common.py
jesseturner21: Fixed.

e2e-tests/fixtures/import/cleanup_resources.py
jesseturner21: Added S3 code object cleanup to `cleanup_resources.py`. The IAM role is intentionally shared/reused across runs, added a comment to explain that decision.

e2e-tests/import-resources.test.ts
jesseturner21: Keeping the inline teardown here, the import test needs extra cleanup that `teardownE2EProject()` doesn't handle: the Python fallback script deletes resources that weren't successfully imported into CloudFormation, and now also cleans up S3 code objects. Added a comment above `afterAll` explaining this.

jariy17 (approved):

PR #829, opened 2026-04-13
fix: remove docker info check from container runtime detection

jesseturner21:
## Problem

During `agentcore deploy`, the "Check dependencies" preflight calls `detectContainerRuntime()` which runs `docker info`. This command communicates with the Docker daemon through the Unix socket (`/var/run/docker.sock`). If the user isn't in the `docker` group, the OS prompts for a password — a confusing experience since the container runtime check is non-blocking (deploy falls back to CodeBuild for container builds anyway).

The same issue affects `agentcore dev` with Container agents and users of Colima (which uses the standard `docker` CLI binary).

## Options considered

1. **Remove `docker info` entirely** — detect runtimes with `which` + `--version` only. If the daemon isn't running, the user finds out when they actually use it (`docker build` fails with a clear "Cannot connect to the Docker daemon" error).

2. **Keep `docker info` but suppress the password prompt** — not possible. The OS-level polkit/sudo prompt fires before our code can react to the result.

3. **Keep `docker info` but catch the failure silently** — same problem: the prompt appears *before* the command returns a result. We can't prevent it.

## Solution

Option 1. Remove the `docker info` call from `detectContainerRuntime()` and rely on `which` + `--version` for binary detection. This matches the approach already used by `detectContainerRuntimeSync()` in `src/lib/packaging/container.ts`.

The tradeoff is losing the proactive "start Docker" hint when the daemon is down. But the `docker build` error message is clear enough, and avoiding a surprise password prompt is worth that tradeoff.

## Changes

- Remove `docker info` probe from `detectContainerRuntime()`
- Remove `notReadyRuntimes` tracking (only populated by the `docker info` failure path)
- Remove `getStartHint()` helper and `START_HINTS` constant (only used to format `notReadyRuntimes` messages)
- Simplify `requireContainerRuntime()` to just throw "not found" if no runtime passes `which` + `--version`
- Simplify `container-dev-server.ts` prepare() to just check if `runtime` is null

## Test plan

- [x] Unit tests: 35/35 pass across detect.test.ts and container-dev-server.test.ts
- [x] Full test suite: 3295/3295 pass
- [x] TypeScript compilation: 0 errors
- [x] Pre-commit hooks: eslint, prettier, secretlint, typecheck all pass
- [x] Manual: `detectContainerRuntime()` returns Docker info via `which` + `--version` only (no password prompt)
- [x] Manual: `agentcore dev -l` with Container agent — full flow completes (build, start, run) with no password prompt
- [ ] Manual: `agentcore deploy` with Container agent on a machine without docker group membership — verify no password prompt

### DIFF
--- a/src/cli/external-requirements/__tests__/detect.test.ts
+++ b/src/cli/external-requirements/__tests__/detect.test.ts
@@ -1,4 +1,4 @@
-import { detectContainerRuntime, getStartHint, requireContainerRuntime } from '../detect.js';
+import { detectContainerRuntime, requireContainerRuntime } from '../detect.js';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 
 const { mockCheckSubprocess, mockRunSubprocessCapture } = vi.hoisted(() => ({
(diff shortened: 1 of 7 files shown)


aidandaly24 (approved):
this lgtm approved

PR #844, opened 2026-04-14
fix: only exclude root-level agentcore/ directory from packaging artifacts

### LINKED ISSUE #843: agentcore directory incorrectly excluded from CodeZip artifact when used as a Python package name
### Description

The packaging helpers (copySourceTree, createZipFromDir, and their sync counterparts) unconditionally excluded any directory named agentcore from the zip artifact. This caused a false exclusion when a third-party Python dependency — such as langgraph-checkpoint-aws — ships a package with a sub-module also named agentcore (e.g., langgraph_checkpoint_aws/agentcore/). Those nested packages were silently dropped from the deployment artifact, causing ImportError at runtime.

### Steps to Reproduce

1. Create an AgentCore project that uses langgraph-checkpoint-aws (or any package that contains a sub-directory named agentcore).
2. Run agentcore deploy (CodeZip build type) or agentcore package.
3. Inspect the generated .zip artifact.

### Expected Behavior

Only the top-level agentcore/ directory (the CLI's own config/infrastructure directory) is excluded from the artifact. Any agentcore sub-directory that lives inside a dependency (e.g., langgraph_checkpoint_aws/agentcore/) is included normally.

### Actual Behavior

All directories named agentcore at any depth in the source tree were excluded. The agentcore entry was in a flat EXCLUDED_ENTRIES set that was checked by name alone, without regard to its path relative to the project root. As a result, langgraph_checkpoint_aws/agentcore/__init__.py (and any sibling files) were silently omitted from the zip, breaking agents that depend on those packages at runtime.

### CLI Version

0.8.0

### Operating System

macOS

### Additional Context

**Root cause:** 'agentcore' was a member of EXCLUDED_ENTRIES (src/lib/packaging/helpers.ts:55), which was applied to every directory entry by name during recursive traversal — no path-depth check was performed.

**Fix**: Remove 'agentcore' from EXCLUDED_ENTRIES and instead pass the original rootDir through all recursive calls.  Before descending into or copying any entry, an explicit check gates exclusion on both the entry name matching CONFIG_DIR and the full resolved path equalling join(rootDir, CONFIG_DIR). This means only the project-root-level agentcore/ is skipped; nested occurrences are unaffected.

aidandaly24:
## Description

The packaging helpers (`copySourceTree`, `createZipFromDir`, and their sync counterparts) unconditionally excluded any directory named `agentcore` from the zip artifact. This caused a false exclusion when a third-party Python dependency — such as `langgraph-checkpoint-aws` — ships a package with a sub-module also named `agentcore` (e.g., `langgraph_checkpoint_aws/agentcore/`). Those nested packages were silently dropped from the deployment artifact, causing `ImportError` at runtime.

**Changes:**

1. **Remove `'agentcore'` from `EXCLUDED_ENTRIES`** — it no longer participates in the flat name-based exclusion check.
2. **Thread `rootDir` through all recursive traversal functions** (`copyEntry`, `collectFiles`, `copyEntrySync`, `collectFilesSync`) — the `agentcore` directory is now only excluded when its resolved path equals `join(rootDir, CONFIG_DIR)`, i.e., it sits at the project root.
3. **Remove the hand-written fflate type shim** (`src/lib/packaging/types/fflate.d.ts`) — this `declare module 'fflate'` file only declared `zipSync`, which completely shadowed the package's own type declarations, making all other fflate exports (including `unzipSync`) invisible to TypeScript. The real fflate v0.8.2 ships complete types that resolve correctly under `moduleResolution: "bundler"`.

## Related Issue

Closes #843

## Documentation PR

N/A

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I ran `npm run lint`

Added 4 new tests covering async/sync variants of both `copySourceTree` and `createZipFromDir`:
- Top-level `agentcore/` directory is excluded (project config dir)
- Nested `agentcore/` directories inside dependencies are preserved
- Zip contents verified via `fflate.unzipSync` (now importable after removing the type shim)

## Checklist

---

terms of your choice.


jesseturner21 (approved):
ltgm

PR #856, opened 2026-04-15
fix: defer policy engine write and harden policy flow UX

jesseturner21:
## Problem

Two issues in the policy engine TUI flow:

1. **Escape skipped to success instead of going back** — When adding a policy engine through the interactive TUI (`agentcore` → add → policy engine), pressing Escape on the "Attach to gateways" screen would skip forward to the success screen instead of going back. The engine was written to `agentcore.json` immediately at the name step, so the user couldn't back out without leaving a half-configured engine on disk.

2. **Policy generation failures crashed the TUI** — Unhandled rejections from the `generate()` call hit the global `process.exit(1)` handler in `index.ts`, causing the TUI to exit immediately. The error screen also had no active `useInput` handler, so Escape did nothing despite the hint text.

### Commit 1: Defer engine write until flow completes
- **Defer the disk write** — `policyEnginePrimitive.add()` is no longer called at the name step. A new `commitEngine` helper writes the engine (and optionally attaches gateways) only when the user reaches the final confirmation.
- **Escape goes back** — on the gateway selection screen, Escape now returns to the name step instead of skipping to success. Since nothing is persisted yet, going back is clean.
- **Three terminal write paths** — the engine is written to disk only when:
  1. No unprotected gateways exist (flow completes at name step)
  2. User presses Enter with no gateways selected
  3. User selects gateways and picks an enforcement mode

### Commit 2: Preserve engine name on back navigation
- When navigating back from the gateway screen, the previously entered engine name is restored via `pendingEngineName` state, instead of being replaced with a generated default.

### Commit 3: Prevent TUI crash on policy generation failure
- Add `.catch()` safety net on `generate()` to prevent unhandled rejections from reaching the global `process.exit(1)` handler
- Add `useInput` handler for Escape on the error screen so users can go back and retry
- Add `ExecLogger` to write structured logs to `.cli/logs/policy-generation/`
- Show log file path on the error screen

## Test plan

- [x] Escape on gateway screen → goes back to name step, `agentcore.json` unchanged
- [x] Enter with no gateways selected → engine written, success screen shown
- [x] Select gateways + pick mode → engine + gateway config written, success screen shown
- [x] Non-interactive CLI (`agentcore add policy-engine --name X`) still works
- [x] Back navigation preserves previously entered engine name
- [x] Policy generation failure shows error screen (no TUI crash)
- [x] Escape on generation error screen goes back to previous step
- [x] All 3304 unit tests pass


jesseturner21 (commented):
# ISSUE BELOW FIXED
### Used Option B

---

### Issue: Escape → back to engine-wizard loses the user's previously entered name

**Severity: Medium — UX regression**

When the user presses Escape on the gateway selection screen, the flow transitions to `{ name: 'engine-wizard' }`, which remounts `AddPolicyEngineScreen`. That component generates a fresh `initialValue` via `generateUniqueName('MyPolicyEngine', existingEngineNames)` — the name the user previously typed and submitted is discarded.

**Scenario:**
1. User types engine name `ProductionGuard` → submits
2. Gateway selection screen appears (unprotected gateways exist)
3. User presses Escape to go back
4. Engine name screen reappears with `MyPolicyEngine` (or similar generated name) instead of `ProductionGuard`

**Options to address:**

**Option A — Carry the name in flow state:** Add an optional `engineName` field to the `engine-wizard` flow state variant. When navigating back from `attach-gateways`, pass `flow.engineName` into it. Then thread that through to `AddPolicyEngineScreen` as an `initialValue` override. This is the cleanest fix — it's one new field on the state union and a prop on the child component.

```tsx
// Flow state addition:
| { name: 'engine-wizard'; prefillName?: string }

// Back handler:
onBack={() => setFlow({ name: 'engine-wizard', prefillName: flow.engineName })}

// In AddPolicyEngineScreen rendering:
initialValue={flow.prefillName ?? generateUniqueName('MyPolicyEngine', existingEngineNames)}
```

**Option B — Store the pending name in component state:** Add a `pendingEngineName` state variable alongside the existing `engineNames` state. Set it when `handleEngineComplete` fires, clear it on success. Use it as the initial value when rendering `engine-wizard`. Slightly more state to manage but doesn't require changing the flow union type.

Option A is simpler and keeps the state local to the navigation path.

avi-alpert (approved):

PR #858, opened 2026-04-15
feat: add AG-UI (AGUI) as fourth first-class protocol mode

### LINKED ISSUE #796: New cli does not support agui
### Description

I was in the process of migrating from the old cli to the new one, but it turned out that the new cli does not support agui

### Steps to Reproduce

Import a project using ag-ui protocol

### Expected Behavior

The import should succeed

### Actual Behavior

I get an error complaining that only http, mcp and a2a are supported

### CLI Version

0.7.1

### Operating System

Linux

### Additional Context

_No response_

aidandaly24:
## Description

Adds AG-UI (AGUI) as a fourth first-class protocol mode alongside HTTP, MCP, and A2A. AG-UI is an SSE-based agent-to-UI protocol that enables rich streaming interactions (text, tool calls, reasoning, state management) between agents and frontends.

### What's included

**Schema & types:**
- `AGUI` added to `ProtocolModeSchema`, `PROTOCOL_FRAMEWORK_MATRIX`, `RESERVED_PROJECT_NAMES` in both CLI and CDK
- `AGUI` added to `agentcore.schema.v1.json` protocol enum
- Complete `AguiEvent` typed union (27 event types) in `src/cli/aws/agui-types.ts`
- `AguiRunInput` request type with `buildAguiRunInput` helper

**Python agent templates (3 frameworks):**
- **Strands** — `ag-ui-strands >= 0.1.7` with `session_manager_provider` for per-thread memory isolation
- **LangChain+LangGraph** — StateGraph with ToolNode + ReAct loop, MemorySaver checkpointer, LangchainInstrumentor
- **GoogleADK** — `ag-ui-adk` with `ADKAgent`, `AGUIToolset`, `create_adk_app`. Uses plain `opentelemetry-distro` (not `aws-opentelemetry-distro`) due to google-adk version conflict

**Unified SSE parser (`src/cli/aws/agui-parser.ts`):**
- Single parser shared between dev (`invoke-agui.ts`) and deployed (`agentcore.ts`) paths
- Bounded cursor-based pruning with `singleConsumer` mode for dev path
- CRLF-safe line splitting (`/\r?\n/`)
- Malformed JSON skipped (never yielded as content)

**Dev mode:**
- Persistent `threadId` per dev session for multi-turn conversations (reset on clearConversation)
- Tool call tracking with `toolCallId`-based matching
- Retry logic with exponential backoff (no retry after streaming starts)

**Container support:**
- Dockerfile `{{#if enableOtel}}` conditional respects `instrumentation.enableOtel` schema field
- `enableOtel` threaded through `AgentRenderConfig` → `BaseRenderer` → Handlebars template
- MCP+Container latent bug fixed (enableOtel: false was ignored by Dockerfile)

**Import path hardening:**
- `ProtocolModeSchema.safeParse` replaces unsafe `as` cast in `yaml-parser.ts`
- MCP `enableOtel` clamped to `false` regardless of YAML value in import actions

**DX improvements:**
- `formatZodIssue` falls back to `issue.code` instead of printing literal `"undefined"`
- New `dockerfile-render.test.ts` snapshot test covering both enableOtel branches
- Dynamic imports converted to static where non-intentional (AGENTS.md compliance)

### Deployed & tested

| Config | Create | Template | Deploy | Invoke |
|---|---|---|---|---|
| Strands+Bedrock CodeZip | ✅ | ✅ | ✅ | ✅ |
| Strands+Bedrock Container | ✅ | ✅ | ✅ | ✅ "2+3=5" |
| LangGraph+Bedrock CodeZip | ✅ | ✅ | ✅ | ✅ "325435+435634563456456=435634563781891" |
| ADK+Gemini CodeZip | ✅ | ✅ | ✅ | ✅ |
| ADK+Gemini Container | ✅ | ✅ container builds | ✅ | ✅ |
| Dev mode multi-turn | ✅ | — | — | ✅ "favorite color is blue" |
| Dev mode tool call | ✅ | — | — | ✅ TOOL_CALL_START→ARGS→END→RESULT |

## Related Issue

Closes #796 

## Type of Change

- [x] New feature
- [x] Bug fix

## Testing

- [x] I ran `npm run lint`
- [x] End-to-end deployment testing across all 3 frameworks × 2 build types
- [x] Dev mode testing with curl against local uvicorn server
- [x] 4-lane parallel code review (Planner/Architect/Critic consensus + OMC code-reviewer)

## Checklist


src/assets/python/agui/langchain_langgraph/base/main.py
```diff
@@ -0,0 +1,68 @@
+import os
+
+# Enable LangGraph FastAPI mode (must be set before langgraph imports)
+os.environ["LANGGRAPH_FAST_API"] = "true"
+
```
avi-alpert, line 4: im thinking we will also want to add instrumentation like we do for the other langchain templates

https://github.com/aws/agentcore-cli/blob/main/src/assets/python/http/langchain_langgraph/base/main.py#L14

src/assets/python/agui/langchain_langgraph/base/main.py
```diff
@@ -0,0 +1,68 @@
...
+add_langgraph_fastapi_endpoint(app=app, agent=agent, path="/invocations")
+
+
+@app.get("/ping")
+async def ping():
+    return {"status": "healthy"}
+
+
+if __name__ == "__main__":
+    uvicorn.run(app, host="0.0.0.0", port=8080)
```
avi-alpert:
```suggestion
    uvicorn.run(app, host="0.0.0.0", port=int(os.environ.get("PORT", "8080")))
```
for local testing it would be nice if we could optionally specify a different port through an env var. (If you end up making this change please apply it to all the agui templates)

src/assets/python/agui/langchain_langgraph/base/main.py
aidandaly24: updated

src/assets/python/a2a/googleadk/base/main.py
```diff
@@ -6,14 +6,16 @@
...
+
 
 def add_numbers(a: int, b: int) -> int:
     """Return the sum of two numbers."""
     return a + b
 
 
 agent = Agent(
-    model=load_model(),
+    model="gemini-2.5-flash",
```
avi-alpert, line 78: why is the AG-UI PR hardcoding a model in the google a2a agent template?

src/cli/tui/screens/add/AddScreen.tsx
```diff
@@ -2,7 +2,7 @@ import type { SelectableItem } from '../../components';
 import { SelectScreen } from '../../components';
 
 const ADD_RESOURCES = [
-  { id: 'agent', title: 'Agent', description: 'Deploy an HTTP, MCP, or A2A agent' },
+  { id: 'agent', title: 'Agent', description: 'Deploy an HTTP, MCP, A2A, or AGUI agent' },
```
avi-alpert: nit: in strings we are displaying to users, do we use "AG-UI" or "AGUI"?

jesseturner21:
## Missing JSON Schema Update

The PR description states that `AGUI` was added to `agentcore.schema.v1.json` protocol enum, but this file is not included in the diff. The current schema at `schemas/agentcore.schema.v1.json` still has:

```json
"protocol": {
  "type": "string",
  "enum": ["HTTP", "MCP", "A2A"]
}
```

This needs to be updated to include `"AGUI"` so that users who validate their config against the JSON schema can use the new protocol.

**Fix:** Add `"AGUI"` to the protocol enum in `schemas/agentcore.schema.v1.json`.

jesseturner21:
This pr has a corresponding change in the l3 constructs package here https://github.com/aws/agentcore-l3-cdk-constructs/pull/151

src/assets/python/a2a/googleadk/base/main.py
aidandaly24: This isn't a new hardcode — it's a fix for a pre-existing bug where the A2A GoogleADK template was passing `model=load_model()` to the ADK `Agent` constructor. For GoogleADK, `load_model()` doesn't return a model object — it sets the `GOOGLE_API_KEY` env var as a side effect and returns `None`. So the agent was getting `model=None` and crashing.

The ADK `Agent` takes a model **name string** (e.g. `"gemini-2.5-flash"`), not a model object. The API key comes from the environment, separately. So the correct pattern is:

```python
load_model()  # sets GOOGLE_API_KEY env var
agent = Agent(model="gemini-2.5-flash", ...)  # model is a string
```

This matches how the HTTP GoogleADK template works (`MODEL_ID = "gemini-2.5-flash"` at line 20). The function is just misnamed — it should be `setup_credentials()`, not `load_model()`. Filed as a follow-up to rename it across all GoogleADK templates.

src/cli/operations/dev/invoke.ts
```diff
@@ -224,6 +224,11 @@ export async function* invokeForProtocol(
     case 'A2A':
       yield* invokeA2AStreaming(options);
       break;
+    case 'AGUI': {
+      const { invokeAguiStreaming } = await import('./invoke-agui');
```
avi-alpert: is this dynamic import intentional?

avi-alpert (approved):

PR #904, opened 2026-04-21
fix(invoke): show full session ID and print resume command on exit

aidandaly24:
## Description

The invoke TUI truncated the session ID to 8 characters (`550e8400`), making it impossible to copy the full UUID needed for `agentcore invoke --session-id <id>` to resume a session. Additionally, there was no guidance on how to resume a session after exiting.

This PR:
- Displays the full session ID in the TUI header instead of truncating to 8 chars
- Prints a colored resume command (`agentcore invoke --session-id <full-uuid>`) after TUI exit
- Uses Ink's `unmount()` instead of `process.exit(0)` for clean shutdown, which also fixes the update notifier not appearing on Esc exit

Screenshot 2026-04-21 at 3 34 25 PM

## Related Issue

Closes #

## Documentation PR

N/A

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Manually tested all exit paths:
- **Ctrl+C**: Resume message and update notifier both display ✅
- **Double Esc**: Resume message and update notifier both display ✅
- **`agentcore invoke` (direct)**: Works correctly ✅
- **`agentcore` → invoke (TUI)**: Works correctly ✅
- **Session resumption**: Copied session ID from output, used `--session-id` flag to resume successfully ✅

- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.

### DIFF
--- a/src/cli/commands/invoke/command.tsx
+++ b/src/cli/commands/invoke/command.tsx
@@ -168,10 +168,10 @@ export const registerInvoke = (program: Command) => {
             });
           } else {
             // No CLI options - interactive TUI mode (headers still passed if provided)
-            const { waitUntilExit } = render(
+            const { waitUntilExit, unmount } = render(
               <InvokeScreen
                 isInteractive={true}
-                onExit={() => process.exit(0)}
+                onExit={() => unmount()}
                 initialSessionId={cliOptions.sessionId}
                 initialUserId={cliOptions.userId}
                 initialHeaders={headers}
--- a/src/cli/cli.ts
+++ b/src/cli/cli.ts
@@ -224,4 +224,10 @@ export const main = async (argv: string[]) => {
 
   // Telemetry notice already printed above; only run update check here.
   await printPostCommandNotices(false, updateCheck);
+
+  const exitMessage = getExitMessage();
+  if (exitMessage) {
+    console.log(`\n${exitMessage}`);
+    clearExitMessage();
+  }
 };
--- a/src/cli/tui/screens/invoke/InvokeScreen.tsx
+++ b/src/cli/tui/screens/invoke/InvokeScreen.tsx
@@ -147,12 +147,12 @@ export function InvokeScreen({
   const mcpFetchTriggeredRef = useRef(false);
 
   useEffect(() => {
-    if (sessionId) {
+    if (sessionId && messages.length > 0) {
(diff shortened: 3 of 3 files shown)


jesseturner21:
The `useEffect` that sets the exit message will cause issues when `InvokeScreen` is used from the main TUI (`agentcore` → invoke → Esc → help → exit). The message is set as soon as a session is created, but it's never cleared when the user navigates away from the invoke screen. This means users will see "To resume this session..." even if they left invoke and did other things before exiting.

**Options to fix:**

1. **Add a cleanup function to clear the message on unmount:**
```tsx
useEffect(() => {
  if (sessionId) {
    const cyan = '\\x1b[36m';
    const reset = '\\x1b[0m';
    setExitMessage(`To resume this session, run: ${cyan}agentcore invoke --session-id ${sessionId}${reset}`);
  }
  return () => clearExitMessage();
}, [sessionId]);
```

2. **Only set the message when actually exiting** (similar to how `CreateScreen` does it) - wrap the `onExit` prop to set the message before calling it:
```tsx
const handleExit = useCallback(() => {
  if (sessionId) {
    const cyan = '\\x1b[36m';
    const reset = '\\x1b[0m';
    setExitMessage(`To resume this session, run: ${cyan}agentcore invoke --session-id ${sessionId}${reset}`);
  }
  onExit();
}, [sessionId, onExit]);
```
Then use `handleExit` instead of `onExit` throughout the component.

Option 1 is simpler but means the message won't show if the user navigates away and then exits. Option 2 ensures the message only shows when exiting directly from invoke, which matches the `CreateScreen` pattern.

src/cli/commands/invoke/command.tsx
```diff
@@ -168,10 +168,10 @@ export const registerInvoke = (program: Command) => {
             });
           } else {
             // No CLI options - interactive TUI mode (headers still passed if provided)
-            const { waitUntilExit } = render(
+            const { waitUntilExit, unmount } = render(
               <InvokeScreen
                 isInteractive={true}
-                onExit={() => process.exit(0)}
+                onExit={() => unmount()}
```
aidandaly24, line 174: This is what fixes the issue with upgrade only coming out on ctrl + c

notgitika (approved):
LGTM

PR #847, opened 2026-04-14
chore: remove preview bump type from release workflow

jesseturner21:
## Summary
- Removed `preview` and `preview-major` options from the release workflow dispatch inputs
- Removed all preview versioning logic from `scripts/bump-version.ts` (type, parsing, formatting, bump cases, help text, validation)
- Preview versioning is no longer used; remaining bump types: `patch`, `minor`, `major`, `prerelease`

## Test plan
- [x] Verified `npx tsx scripts/bump-version.ts patch --dry-run` works correctly
- [x] Pre-commit hooks pass (prettier, eslint, typecheck, secretlint)
- [ ] CI checks pass

PR #905, opened 2026-04-21
feat: support preview releases from feature branches

tejaskash:
## Summary
- Release workflow now supports publishing preview versions from non-main branches using the `preview` npm dist-tag
- Stable bump types (patch/minor/major) are blocked on non-main branches to prevent accidental overwrites of the `latest` tag
- Release PRs target the source branch instead of hardcoded `main`
- Publish job no longer gated to `main`-only — it uses the dynamically determined base branch and dist-tag

## Test plan
- [ ] Run release workflow from `main` with `patch` — should behave identically to before (dist-tag: `latest`, PR targets `main`)
- [ ] Run release workflow from a feature branch with `preview` — should set dist-tag to `preview` and PR targets the feature branch
- [ ] Run release workflow from a feature branch with `patch` — should fail at the metadata step

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,13 +33,28 @@ jobs:
     outputs:
       version: ${{ steps.bump.outputs.version }}
       branch: ${{ steps.bump.outputs.branch }}
+      dist_tag: ${{ steps.release-meta.outputs.dist_tag }}
+      base_branch: ${{ steps.release-meta.outputs.base_branch }}
 
     steps:
-      - name: Validate running from main
+      - name: Determine release metadata
+        id: release-meta
         run: |
-          if [[ "${{ github.ref }}" != "refs/heads/main" ]]; then
-            echo "⚠️  WARNING: Running from ${{ github.ref }}"
-            echo "⚠️  Production releases should only run from main branch"
+          BRANCH_NAME="${{ github.ref_name }}"
+          VERSION_BUMP="${{ github.event.inputs.bump_type }}"
+
+          if [[ "$BRANCH_NAME" == "main" ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -35,13 +35,28 @@ jobs:
...
-            echo "⚠️  WARNING: Running from ${{ github.ref }}"
-            echo "⚠️  Production releases should only run from main branch"
+          BRANCH_NAME="${{ github.ref_name }}"
+          VERSION_BUMP="${{ github.event.inputs.bump_type }}"
+
+          if [[ "$BRANCH_NAME" == "main" ]]; then
+            echo "dist_tag=latest" >> $GITHUB_OUTPUT
+            echo "base_branch=main" >> $GITHUB_OUTPUT
+          else
+            if [[ "$VERSION_BUMP" != "preview" && "$VERSION_BUMP" != "preview-major" && "$VERSION_BUMP" != "prerelease" ]]; then
```
notgitika: **P2 — bump_type input may not list the new valid values**

The validation here accepts preview, preview-major, and prerelease, but the workflow_dispatch input options list (above this diff) likely only has patch, minor, major. If those new values are not added to the input dropdown, users have to type them manually which is error-prone. Make sure the input options include all valid bump types.

tejaskash: All three accepted non-main types (`preview`, `preview-major`, `prerelease`) are already in the `workflow_dispatch` input options dropdown (lines 11-16). No change needed here.

tejaskash:
Rebased on latest main (which merged #847 removing `preview`/`preview-major` bump types). The guard now only accepts `prerelease` from non-main branches, matching the current dropdown options.

notgitika (approved):
LGTM

PR #921, opened 2026-04-22
fix: update asset snapshots and fix useDevDeploy skip test

notgitika:
## Summary

- Update 3 asset snapshots that drifted from the harness-implementation branch content:
  - File listing (new harness files added)
  - `cdk/bin/cdk.ts`
  - `cdk/lib/cdk-stack.ts`
- Fix `useDevDeploy` test: when `skip=true`, `isComplete` is correctly `true` (skipped = done), not `false` as the test asserted

## Root cause

Asset snapshots were not regenerated after harness-related changes to CDK templates and file structure. The `useDevDeploy` test had a stale assertion — the hook returns `isComplete = skip || deployDone`, so skip=true means immediately complete.

## Test plan

- [x] All 95 tests pass (90 snapshot + 5 useDevDeploy)
- [x] Pre-commit hooks pass (eslint, prettier, secretlint, typecheck)

### DIFF
--- a/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
+++ b/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
@@ -110,6 +110,8 @@ async function main() {
     memoryName?: string;
     containerUri?: string;
     hasDockerfile?: boolean;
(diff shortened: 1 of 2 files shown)


tejaskash (approved):
GLTM

PR #925, opened 2026-04-22
fix(deploy): honor aws-targets.json region for all SDK and CDK calls

### LINKED ISSUE #924: CLI ignores region from aws-targets.json for API calls unless AWS_DEFAULT_REGION is set
### Description

The agentcore CLI does not use the region field from aws-targets.json when making direct API calls (e.g., creating runtimes, evaluators, online eval configs). It only
  respects the AWS SDK default region resolution chain (AWS_DEFAULT_REGION → AWS_REGION → ~/.aws/config).                                                               
                                                                                                                                                                        
  Changing the region in aws-targets.json has no effect on where resources are actually created unless AWS_DEFAULT_REGION is also set. 

### Steps to Reproduce

 Configure aws-targets.json with a non-default region:                                                                                                              
                                                                                                                                                                        
     [                                                                                                                                                                  
       {                                                                                                                                                                
         "name": "default",                                                                                                                                             
         "account": "123456789012",                                                                                                                                     
         "region": "ap-southeast-2"                                                                                                                                     
       }                                                                                                                                                                
     ]                                                                                                                                                                  
                                                                                                                                                                        
  2. Do not set AWS_DEFAULT_REGION                                                                                                                                      
  3. Run agentcore deploy  

### Expected Behavior

The CLI should read the region from aws-targets.json and use it for all API calls related to that deployment target, including resource creation (runtimes,           
  evaluators, etc.).   

### Actual Behavior

 The CDK stack is synthesized with the correct region from aws-targets.json, but the CLI's own API calls (outside of CDK) use the AWS SDK's default region resolution  
  chain. Resources end up being created in the wrong region.                               

### CLI Version

_No response_

### Operating System

macOS

### Additional Context

_No response_

aidandaly24:
## Description

The agentcore CLI did not use the `region` field from `aws-targets.json` when making API calls for resources like runtimes, evaluators, and online eval configs. Changing `aws-targets.json` had no effect on where resources were created unless `AWS_DEFAULT_REGION` was also set.

### Root cause

The CLI passes `target.region` explicitly to its own SDK clients, but `@aws-cdk/toolkit-lib` constructs internal SDK clients (for CloudFormation, S3 asset upload, asset publishing, etc.) that do not receive an explicit region option and fall back to the SDK default region resolution chain: `AWS_REGION` → `AWS_DEFAULT_REGION` → shared config profile. When `aws-targets.json` specified a non-default region but those env vars were unset, CDK's internal clients resolved the wrong region and resources were created in the SDK default region.

### Fix

New helper `src/cli/aws/target-region.ts` exports:
- `applyTargetRegionToEnv(region)` — sets `AWS_REGION` and `AWS_DEFAULT_REGION` to the target region, returns a restore function.
- `withTargetRegion(region, fn)` — try/finally wrapper.

Applied at:
- `src/cli/commands/deploy/actions.ts` — `handleDeploy()` calls `applyTargetRegionToEnv(target.region)` after target resolution; restore runs in the existing `finally` block.
- `src/cli/operations/deploy/teardown.ts` — `destroyTarget()` wraps CDK `initialize()`/`destroy()` in `withTargetRegion(...)`.
- `src/cli/tui/hooks/useCdkPreflight.ts` — TUI path applies the override after `validateProject()`; restored on unmount, `cancelTeardown`, preflight retry, SIGINT/SIGTERM.

Invoke/status/eval/logs commands already pass `target.region` explicitly to every SDK client — no changes needed there.

## Related Issue

Closes #924

## Documentation PR

N/A — behavior-only fix.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I ran \`npm run test:unit\` and \`npm run test:integ\`
- [x] I ran \`npm run typecheck\`
- [x] I ran \`npm run lint\`
- [ ] If I modified \`src/assets/\`, I ran \`npm run test:update-snapshots\` and committed the updated snapshots

### Unit tests

7 new unit tests in \`src/cli/aws/__tests__/target-region.test.ts\` cover:
- sets both env vars to the provided region
- restore clears env vars when they were previously unset
- restore preserves previous values when they were set
- restore handles each env var independently
- applies inside callback and restores afterwards
- restores even when the callback throws
- returns the callback result

All 7 pass. Relevant targeted suites (aws, operations/deploy, config-io, tui/hooks): 354/354 pass. tsc: 0 errors. eslint: 0 errors on modified files.

### End-to-end testing against live AWS

Tested full lifecycle (create → add agent → deploy → invoke → teardown) against AWS account 603141041947 across 5 scenarios:

| # | Scenario | Region | Result |
|---|---|---|---|
| 1 | CLI non-interactive, no env vars | \`ap-southeast-2\` | ✅ Runtime in apse2, none in us-east-1 |
| 2 | Env isolation — success + error paths | \`ap-southeast-2\` | ✅ Env vars remain unset after both paths |
| 3 | CLI non-interactive, no env vars | \`eu-west-1\` | ✅ Runtime in euw1, invoke returned \"Hello!\" |
| 4 | Interactive TUI (tui-harness) | \`ap-southeast-2\` | ✅ Full interactive deploy + teardown |
| 5 | **Key test: target beats env var** | \`eu-west-2\` + \`AWS_DEFAULT_REGION=us-east-1\` | ✅ Runtime in euw2, none in us-east-1 |

All 5 verified via \`aws bedrock-agentcore-control list-agent-runtimes\` and teardown confirmed cleanup.

## Checklist

---

### DIFF
--- a/src/cli/aws/__tests__/target-region.test.ts
+++ b/src/cli/aws/__tests__/target-region.test.ts
@@ -1,0 +1,99 @@
+import { applyTargetRegionToEnv, withTargetRegion } from '../target-region.js';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+
+describe('target-region', () => {
+  let savedRegion: string | undefined;
+  let savedDefaultRegion: string | undefined;
+
+  beforeEach(() => {
+    savedRegion = process.env.AWS_REGION;
+    savedDefaultRegion = process.env.AWS_DEFAULT_REGION;
+    delete process.env.AWS_REGION;
+    delete process.env.AWS_DEFAULT_REGION;
+  });
+
+  afterEach(() => {
+    if (savedRegion !== undefined) process.env.AWS_REGION = savedRegion;
+    else delete process.env.AWS_REGION;
+    if (savedDefaultRegion !== undefined) process.env.AWS_DEFAULT_REGION = savedDefaultRegion;
+    else delete process.env.AWS_DEFAULT_REGION;
+  });
+
+  describe('applyTargetRegionToEnv', () => {
+    it('sets AWS_REGION and AWS_DEFAULT_REGION to the provided region', () => {
+      applyTargetRegionToEnv('ap-southeast-2');
+      expect(process.env.AWS_REGION).toBe('ap-southeast-2');
+      expect(process.env.AWS_DEFAULT_REGION).toBe('ap-southeast-2');
+    });
+
+    it('returns a restore function that clears env vars when they were previously unset', () => {
+      const restore = applyTargetRegionToEnv('eu-west-1');
+      restore();
+      expect(process.env.AWS_REGION).toBeUndefined();
+      expect(process.env.AWS_DEFAULT_REGION).toBeUndefined();
+    });
+
+    it('returns a restore function that restores previous env var values', () => {
+      process.env.AWS_REGION = 'us-east-1';
+      process.env.AWS_DEFAULT_REGION = 'us-east-1';
+
+      const restore = applyTargetRegionToEnv('ap-south-1');
+      expect(process.env.AWS_REGION).toBe('ap-south-1');
+      expect(process.env.AWS_DEFAULT_REGION).toBe('ap-south-1');
+
+      restore();
+      expect(process.env.AWS_REGION).toBe('us-east-1');
+      expect(process.env.AWS_DEFAULT_REGION).toBe('us-east-1');
+    });
+
+    it('restores each env var independently (only one was previously set)', () => {
+      process.env.AWS_REGION = 'us-west-2';
+      // AWS_DEFAULT_REGION intentionally left unset
+
+      const restore = applyTargetRegionToEnv('eu-central-1');
+      expect(process.env.AWS_REGION).toBe('eu-central-1');
+      expect(process.env.AWS_DEFAULT_REGION).toBe('eu-central-1');
+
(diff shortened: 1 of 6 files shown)


notgitika (commented):
Nice fix — the approach of promoting the target region onto env vars is the right call since CDK toolkit-lib's internal clients don't take an explicit region.

Two small things:

1. **TUI error states don't restore the env override** — when `run()` errors out after the region override is applied (e.g. build/synth/stack-status failures), we call `setPhase('error')` but never call `restoreRegionEnv()`. It stays overridden until the user retries or unmounts. Probably fine in practice since the TUI is short-lived, but a small `useEffect` guard would close this gap cleanly:
   ```ts
   useEffect(() => {
     if (phase === 'error') restoreRegionEnv();
   }, [phase, restoreRegionEnv]);
   ```

2. **Minor export inconsistency** — `withTargetRegion` is exported from the barrel (`aws/index.ts`) but `applyTargetRegionToEnv` is imported directly from `target-region.ts` in both `actions.ts` and `useCdkPreflight.ts`. Not a big deal, just noticed the mismatch — either export both from the barrel or import both directly.

Everything else looks solid — tests are thorough, the teardown double-wrapping is a nice defensive touch, and the env var save/restore logic correctly handles the `undefined` vs set distinction.

notgitika (approved):
LGTM

PR #926, opened 2026-04-22
fix(harness): complete credential flow for non-Bedrock model providers

aidandaly24:
## Description

The harness API key credential flow had multiple gaps blocking end-to-end deploy+invoke for OpenAI and Gemini harnesses. This PR fixes the credential lifecycle end-to-end: TUI/CLI collection, schema validation, deploy-time resolution, and the vended CDK's IAM policy wiring.

### What was broken

- **Invokes 401'd with AccessDeniedException** — the vended CDK passed `apiKeyArn` to `AgentCoreHarnessRole`, but our new flow stores the credential by name (`apiKeyCredential`). The role never got the `bedrock-agentcore:GetResourceApiKey` policy.
- **Cross-provider credential contamination** — two harnesses (OpenAI + Gemini) with coincidentally matching keys would reuse the same credential, wiring the wrong provider's ARN into the other provider's model config.
- **Silent failures** — skip-on-API-key for non-Bedrock providers produced a broken harness; dangling credential references deferred their failure to a confusing deploy-time error; missing `.env.local` entries silently fell through to CDK synth; Secrets Manager ARNs in `apiKeyArn` were rejected only at Zod time with no case/whitespace tolerance.
- **Multi-write without rollback** — primitive wrote project spec, harness spec, and `.env.local` in an order that could leave a harness referencing a non-existent credential on partial failure.
- **CLI flag confusion** — agent path silently accepted and ignored `--api-key-arn`; both `--api-key` and `--api-key-arn` could be set at once with no error.

### What changed

- **Vended CDK** (`src/assets/cdk/bin/cdk.ts`) resolves `model.apiKeyCredential` to the token-vault provider ARN from deployed state per-target before passing to `AgentCoreStack`.
- **`CredentialPrimitive.resolveCredentialStrategy`** filters reuse candidates by `authorizerType === 'ApiKeyCredentialProvider'` AND `cred.name.endsWith(modelProvider)`, preventing cross-provider contamination.
- **New preflight validator** (`preflight.ts#validateHarnessCredentialReferences`) rejects dangling `apiKeyCredential` references and OAuth credentials used as API key credentials, before `pre-deploy-identity` runs.
- **Pre-deploy identity** escalates skipped credentials that harnesses reference (clear error naming the missing env var), and tracks newly-created token-vault providers to roll them back when the subsequent CDK deploy fails.
- **`HarnessPrimitive.add()`** rejects non-Bedrock provider without a credential source, rejects both `--api-key` and `--api-key-arn` set together, and writes in rollback-safe order: `.env.local` → `harness.json` → `agentcore.json` (the commit point).
- **Schema** trims `apiKeyArn`, enforces `.min(1)` on `apiKeyCredential`, rejects Secrets Manager ARNs case-insensitively, and fails non-Bedrock providers that ship with neither credential field.
- **TUI** removes the skip option from `ApiKeySecretInput` when rendered in the harness wizard; CLI guards surface `--api-key-arn` on the agent path and both-flags combinations at `validate` time.
- **Refactor** — extracted `computeCredentialName` from `schema-mapper.ts` into `credential-utils.ts` so the agent and harness flows share the same naming.

## Related Issue

Closes #

## Type of Change

- [x] Bug fix

## Testing

- [x] I ran `npm run lint`

**Automated coverage**
- 85 schema tests — empty-string mutex bypass, whitespace ARN, mixed-case Secrets Manager rejection, missing credential on non-Bedrock provider.
- 6 preflight validator tests — no harnesses, no apiKeyCredential, dangling reference, wrong authorizerType, multi-harness error collection, valid reference.
- 4 cross-provider isolation tests in resolve-credential-strategy.test.ts.
- 5 HarnessPrimitive tests covering the primary apiKey path, reuse path, BYO apiKeyCredentialArn, and dual-flag rejection.
- 6 end-to-end integration tests using real ConfigIO against tmpdir.
- Source-level wiring guard (useCreateFlow.wiring.test.ts) preventing regression of field drops between TUI and primitive layers.

**End-to-end AWS verification** (account 603141041947)
- OpenAI harness (us-east-1): create → deploy → invoke "Say hello in one word" → returned "Hello". CDK IAM fix verified.
- Gemini harness: confirmed end-to-end in real deploys.
- M4 (dangling credential): fails at preflight with exact remediation text.
- M5 (missing env var): fails before CDK synth with exact error text.
- M3 (Secrets Manager ARN) and M7 (both-flags mutex): agentcore create produces exact error.

## Checklist

---

### DIFF
--- a/src/assets/cdk/bin/cdk.ts
+++ b/src/assets/cdk/bin/cdk.ts
@@ -59,7 +59,7 @@ async function main() {
   // so we read them dynamically via specAny (same pattern as gateways above).
   // Harness paths in agentcore.json are relative to the project root (parent of agentcore/).
   const projectRoot = path.resolve(configRoot, '..');
-  const harnessConfigs: {
+  interface HarnessRawConfig {
     name: string;
     executionRoleArn?: string;
     memoryName?: string;
@@ -67,12 +67,14 @@ async function main() {
     hasDockerfile?: boolean;
     tools?: { type: string; name: string }[];
     apiKeyArn?: string;
-  }[] = [];
+    apiKeyCredential?: string;
+  }
+  const harnessRawConfigs: HarnessRawConfig[] = [];
   for (const entry of specAny.harnesses ?? []) {
     const harnessPath = path.resolve(projectRoot, entry.path, 'harness.json');
     try {
       const harnessSpec = JSON.parse(fs.readFileSync(harnessPath, 'utf-8'));
-      harnessConfigs.push({
+      harnessRawConfigs.push({
         name: entry.name,
         executionRoleArn: harnessSpec.executionRoleArn,
         memoryName: harnessSpec.memory?.name,
@@ -80,6 +82,7 @@ async function main() {
         hasDockerfile: !!harnessSpec.dockerfile,
         tools: harnessSpec.tools,
         apiKeyArn: harnessSpec.model?.apiKeyArn,
+        apiKeyCredential: harnessSpec.model?.apiKeyCredential,
       });
     } catch (err) {
       throw new Error(
@@ -102,6 +105,20 @@ async function main() {
     const credentials = targetResources?.credentials as
       | Record<string, { credentialProviderArn: string; clientSecretArn?: string }>
       | undefined;
+
+    // Resolve apiKeyCredential (credential name in agentcore.json) to the token-vault
+    // provider ARN from deployed state. This is what the harness execution role needs
+    // to grant bedrock-agentcore:GetResourceApiKey on.
+    const harnessConfigs = harnessRawConfigs.map(raw => {
+      if (raw.apiKeyArn) {
+        return raw;
+      }
+      if (raw.apiKeyCredential) {
+        const resolved = credentials?.[raw.apiKeyCredential]?.credentialProviderArn;
+        return { ...raw, apiKeyArn: resolved };
+      }
+      return raw;
+    });
 
     new AgentCoreStack(app, stackName, {
       spec,
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,4 @@ ProtocolTesting/
 
 # Auto-cloned CDK constructs (from scripts/bundle.mjs)
 .cdk-constructs-clone/
+.omc/
--- a/src/cli/commands/create/__tests__/harness-action.test.ts
+++ b/src/cli/commands/create/__tests__/harness-action.test.ts
@@ -49,7 +49,8 @@ describe('createProjectWithHarness', () => {
       cwd: testDir,
       modelProvider: 'open_ai',
       modelId: 'gpt-4',
-      apiKeyArn: 'arn:aws:secretsmanager:us-east-1:123456789012:secret:my-key',
+      apiKeyCredentialArn:
+        'arn:aws:bedrock-agentcore:us-east-1:123456789012:token-vault/default/apikeycredentialprovider/my-key',
       skipMemory: true,
       maxIterations: 10,
       maxTokens: 2000,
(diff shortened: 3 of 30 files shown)


src/assets/cdk/bin/cdk.ts
```diff
@@ -59,27 +59,30 @@ async function main() {
   // so we read them dynamically via specAny (same pattern as gateways above).
   // Harness paths in agentcore.json are relative to the project root (parent of agentcore/).
   const projectRoot = path.resolve(configRoot, '..');
-  const harnessConfigs: {
+  interface HarnessRawConfig {
```
aidandaly24, line 62: ### Why per-target credential resolution in `cdk.ts`

`AgentCoreHarnessRole` scopes its `bedrock-agentcore:GetResourceApiKey` policy to the specific ARN passed in via `harness.apiKeyArn`. Since our flow stores the credential by name (`apiKeyCredential`), we have to resolve that name → ARN from deployed state before CDK synth so the IAM statement is concrete.

That's why the harness config is built in two steps:
- `harnessRawConfigs` — read once from each `harness.json` on disk (pre-loop)
- `harnessConfigs` — resolved per-target inside the `for (const target of targets)` loop, using that target's `credentials` map from deployed state

### Alternative we could have taken

`AgentCoreRuntime` skips per-credential resolution entirely and grants a wildcard policy via `grantCredentialAccess()`:

```ts
actions: ['bedrock-agentcore:GetResourceApiKey', ...],
resources: [
  `arn:aws:bedrock-agentcore:*:${account}:token-vault/*`,
  `arn:aws:bedrock-agentcore:*:${account}:apikeycredentialprovider/*`,
],
```

We could change `AgentCoreHarnessRole` to do the same. That would eliminate the resolution block in `cdk.ts` entirely — but it relaxes the harness role's IAM scoping from "this one credential" to "any credential in this account." Current approach preserves the tighter scoping the construct already had.

jesseturner21:
In `src/assets/cdk/bin/cdk.ts` (around lines 112-121), when `raw.apiKeyCredential` is set but cannot be resolved from `credentials?.[raw.apiKeyCredential]?.credentialProviderArn` (e.g. `credentials` is `undefined` for this target, or the credential name isn't present in deployed state), `resolved` silently becomes `undefined` and is spread into the harness config as `apiKeyArn: undefined`. Downstream in `AgentCoreHarnessRole`, the `if (harness.apiKeyArn)` branch then skips the `GetResourceApiKey` policy entirely — producing a stack that synths and deploys successfully but whose harness invocations will 401 at runtime with `AccessDeniedException`. This is exactly the class of silent failure the rest of the PR is working to eliminate (via preflight + `escalateSkippedCredentialsReferencedByHarnesses`), so it's worth closing this defense-in-depth gap too.

In the current flow those upstream validators do cover the known paths, so in practice this shouldn't fire. But if any future refactor weakens an invariant (a target's credentials map isn't written to deployed state, a new code path bypasses preflight, etc.), the failure mode is a working-looking deploy that then 401s — much harder to diagnose than a loud synth-time error.

A couple of options:

1. Throw in `cdk.ts` when resolution fails, e.g.:
```ts
if (raw.apiKeyCredential) {
  const resolved = credentials?.[raw.apiKeyCredential]?.credentialProviderArn;
  if (!resolved) {
    throw new Error(
      `Harness "${raw.name}" references credential "${raw.apiKeyCredential}" but no credentialProviderArn is present in deployed state for target "${target.name}". This indicates an internal bug — the credential should have been created by pre-deploy identity setup before synth.`
    );
  }
  return { ...raw, apiKeyArn: resolved };
}
```

2. Or, change `AgentCoreHarnessRole` to fall back to the wildcard `grantCredentialAccess()`-style policy when `apiKeyArn` is unset but the harness is non-Bedrock (as you note in your existing review comment, this relaxes IAM scoping — probably not what you want here).

Option 1 seems cleaner given the PR's overall posture on silent-failure elimination.

aidandaly24:

PR #958, opened 2026-04-24
fix: use pull_request_target for fork PR support

jesseturner21:
## Summary
- Switches the AI review workflow trigger from `pull_request` to `pull_request_target`
- `pull_request` gives fork PRs a read-only `GITHUB_TOKEN`, which blocks labels, secrets, and the review entirely
- `pull_request_target` runs in the base repo context with full write permissions and secret access
- This is safe because we never check out or execute fork code — the Checkout step gets the base branch (where the review scripts live), and the harness fetches the PR diff via the GitHub API

## Test plan
- [ ] Open/reopen a fork PR from an `agentcore-cli-devs` member and verify the label is added, the review runs, and the label is removed
- [ ] Verify `workflow_dispatch` still works

### DIFF
--- a/.github/workflows/pr-ai-review.yml
+++ b/.github/workflows/pr-ai-review.yml
@@ -87,6 +87,7 @@ jobs:
           echo "number=$PR_NUM" >> "$GITHUB_OUTPUT"
 
       - name: Add agentcore-harness-reviewing label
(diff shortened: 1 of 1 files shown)


Hweinstock (approved):
noice

PR #966, opened 2026-04-24
chore: disable automatic PR review trigger

jesseturner21:
## Summary
- Comments out the `pull_request` trigger on the AI review workflow so it no longer runs automatically on PR open/reopen
- The workflow can still be triggered manually via `workflow_dispatch`

## Test plan
- [ ] Verify the workflow no longer triggers on PR open/reopen
- [ ] Verify manual workflow_dispatch still works

### DIFF
--- a/.github/workflows/pr-ai-review.yml
+++ b/.github/workflows/pr-ai-review.yml
@@ -1,8 +1,8 @@
 name: AgentCore Harness Reviewing
 
 on:
-  pull_request:
-    types: [opened, reopened]
+  # pull_request:
+  #   types: [opened, reopened]
   workflow_dispatch:
     inputs:
       pr_url:


jesseturner21:
This PR doesn't actually achieve its stated goal — it's effectively a no-op after merging.

The branch was created from a commit before #958 merged to `main`. PR #958 changed the workflow trigger from `pull_request` to `pull_request_target` (for fork PR support). This PR's diff comments out `pull_request:`, but on current `main` that line doesn't exist — it's `pull_request_target:` instead.

I verified this by merging the branch locally: `git diff origin/main` against the merge result returns empty, so merging this PR would leave `pull_request_target:` still active and the workflow would keep auto-triggering on PR open/reopen.

To actually disable the automatic trigger, you'll need to rebase onto current `main` and comment out (or remove) the `pull_request_target:` block instead:

```yaml
on:
  # pull_request_target:
  #   types: [opened, reopened]
  workflow_dispatch:
    ...
```

PR #969, opened 2026-04-25
feat: add project-name option to create

tejaskash:
## Summary

Adds an optional --project-name flag to agentcore create so the project/folder name can be set independently from the resource name supplied by --name.

On main, --name continues to identify the default runtime resource while --project-name controls the generated project name and directory. Existing calls without --project-name keep the prior behavior.

## Validation

- npm run build
- Created a real project with: node dist/cli/index.mjs create --project-name CliProjMainFinal --name CliAgentResourceFinal123456 --defaults --skip-git --skip-install --json --output-dir /tmp/agentcore-cli-create-check-main-final
- Verified agentcore.json has project CliProjMainFinal, runtime CliAgentResourceFinal123456, and the runtime app directory exists.

### DIFF
--- a/src/cli/commands/create/command.tsx
+++ b/src/cli/commands/create/command.tsx
@@ -76,6 +76,7 @@ function printCreateSummary(
 /** Handle CLI mode with progress output */
 async function handleCreateCLI(options: CreateOptions): Promise<void> {
   const cwd = options.outputDir ?? getWorkingDirectory();
+  const projectName = options.projectName ?? options.name!;
 
   const validation = validateCreateOptions(options, cwd);
   if (!validation.valid) {
@@ -89,7 +90,7 @@ async function handleCreateCLI(options: CreateOptions): Promise<void> {
 
   // Handle dry-run mode
   if (options.dryRun) {
-    const result = getDryRunInfo({ name: options.name!, cwd, language: options.language });
+    const result = getDryRunInfo({ name: options.name!, projectName, cwd, language: options.language });
     if (options.json) {
       console.log(JSON.stringify(result));
     } else {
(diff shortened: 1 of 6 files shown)


src/cli/commands/create/command.tsx
```diff
@@ -173,7 +175,11 @@ export const registerCreate = (program: Command) => {
   program
     .command('create')
     .description(COMMAND_DESCRIPTIONS.create)
-    .option('--name <name>', 'Project name (start with letter, alphanumeric only, max 23 chars) [non-interactive]')
+    .option('--name <name>', 'Resource name [non-interactive]')
```
notgitika: why remove the description? it is needed

src/cli/commands/create/command.tsx
```diff
@@ -76,6 +76,7 @@ function printCreateSummary(
 /** Handle CLI mode with progress output */
 async function handleCreateCLI(options: CreateOptions): Promise<void> {
   const cwd = options.outputDir ?? getWorkingDirectory();
+  const projectName = options.projectName ?? options.name!;
```
notgitika: The `name!` assertion can technically be undefined when someone passes `--no-agent --project-name Foo` without `--name`. 

it works right now because the .action() handler patches name before calling this function, but if handleCreateCLI ever gets called from somewhere else it'll break silently. Can we move that backfill logic into `handleCreateCLI` methoid directly?

notgitika (changes requested):

src/cli/commands/create/command.tsx
tejaskash: Updated the description to `'Resource name (agent or harness) [non-interactive]'` — the old project-name constraints now live on `--project-name`.

src/cli/commands/create/command.tsx
tejaskash: Moved the backfill into the handler function itself — it now derives `name` and `projectName` with cross-fallback at the top, and the `.action()` caller no longer patches `opts.name`.

notgitika (approved):
LGTM

PR #970, opened 2026-04-25
feat: add project-name option to preview create

tejaskash:
## Summary

Adds an optional --project-name flag to agentcore create on preview so the project/folder name can be set independently from the resource name supplied by --name.

On preview, default create produces a harness. With this change, --name identifies the harness resource while --project-name controls the generated project name and directory. Existing calls without --project-name keep the prior behavior.

## Validation

- npm run build
- Created a real preview harness project with: node dist/cli/index.mjs create --project-name CliProjPrevFinal --name CliHarnessResourceFinal123456 --skip-git --skip-install --json --output-dir /tmp/agentcore-cli-create-check-preview-final
- Verified agentcore.json has project CliProjPrevFinal, harness CliHarnessResourceFinal123456, path app/CliHarnessResourceFinal123456, and the harness config exists.

### DIFF
--- a/src/cli/commands/create/harness-validate.ts
+++ b/src/cli/commands/create/harness-validate.ts
@@ -1,8 +1,9 @@
-import { HarnessNameSchema } from '../../../schema';
+import { HarnessNameSchema, ProjectNameSchema } from '../../../schema';
 import { validateFolderNotExists } from './validate';
 
 export interface CreateHarnessCliOptions {
   name?: string;
+  projectName?: string;
   modelProvider?: string;
   modelId?: string;
   apiKeyArn?: string;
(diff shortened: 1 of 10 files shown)


src/cli/commands/create/harness-validate.ts
```diff
@@ -50,12 +51,18 @@ export function validateCreateHarnessOptions(options: CreateHarnessCliOptions, c
     return { valid: false, error: '--name is required' };
   }
 
+  const projectName = options.projectName ?? options.name;
+  const projectNameResult = ProjectNameSchema.safeParse(projectName);
```
aidandaly24, line 55: Do we ever check the length?

tejaskash: Yes — `ProjectNameSchema` enforces a 23-char max via `.max(23)`. The `HarnessNameSchema` enforces its own 48-char max separately (via `.max(48)` and the regex). Both are validated in `validateCreateHarnessOptions`.

notgitika (approved):
LGTM

PR #979, opened 2026-04-27
feat: runtime endpoint support in AgentCore CLI

jariy17:
## Summary
- Add `endpoints` dictionary field to `AgentEnvSpec` schema for named version aliases
- Add `RuntimeEndpointPrimitive` with full add/remove/preview lifecycle
- TUI: multi-field add wizard (tab between name/version/description) + remove flow
- Status: endpoints nested under agents with `[Deployed]`/`[Local only]` badges
- Deploy pipeline: `parseRuntimeEndpointOutputs` + `buildDeployedState` tracks endpoint IDs/ARNs in `deployed-state.json`

## CLI Usage
```bash
agentcore add runtime-endpoint --runtime MyAgent --endpoint prod --version 3 --description "Production"
agentcore remove runtime-endpoint --name prod
agentcore status  # shows endpoints nested under agents
```

## Known Issue
First-time deploy of runtime + endpoints in same stack creation may fail due to service eventual consistency. Workaround: deploy runtime first, then add endpoints and redeploy.

## Depends on
- CDK: aws/agentcore-l3-cdk-constructs#180

## Test plan
- [ ] Schema tests pass (endpoints field validation)
- [ ] Primitive tests pass (add/remove/preview)
- [ ] CLI commands: add (6 cases), remove (2 cases), status display
- [ ] TUI flows verified via TUI harness with screenshots
- [ ] Deploy verified end-to-end (two-phase) with CFN resources confirmed in AWS
- [ ] Edge cases: same name on different runtimes, remove runtime with endpoints, remove all endpoints


tejaskash (commented):
Good feature implementation that follows the established primitive patterns closely. Schema design, deploy pipeline integration, and TUI wizard are well done.

**Main issues to address before merge:**
1. `remove()` is ambiguous when same-name endpoints exist across runtimes (your own test case 7a). Needs composite key like policies use.
2. Status `--type runtime-endpoint` agent matching parses `detail` strings — fragile if agent names share prefixes.

See inline comments for details and suggestions.

src/cli/primitives/RuntimeEndpointPrimitive.ts
```diff
@@ -0,0 +1,344 @@
...
+      return { success: false, error: getErrorMessage(err) };
+    }
+  }
+
+  async remove(name: string): Promise<RemovalResult> {
+    try {
+      const project = await this.readProjectSpec();
+
+      // Find the runtime that contains this endpoint
+      let found = false;
```
tejaskash: **Bug: `remove()` is ambiguous across runtimes with same-name endpoints.**

This searches all runtimes and deletes from the first match. If two runtimes both have a `prod` endpoint, this always removes from whichever runtime appears first in the array. The PR description (test case 7a) explicitly says "same endpoint name on different runtimes" should work.

`previewRemove()`, `getRemovable()`, and the TUI remove flow all share this issue — `name` alone isn't unique.

**Suggestion:** Use `runtimeName/endpointName` as the composite key throughout the remove path, matching the pattern `PolicyPrimitive` uses with `engineName/policyName`. `getRemovable()` should set `name` to the composite form.

src/cli/commands/status/command.tsx
```diff
@@ -153,15 +155,41 @@ export const registerStatus = (program: Command) => {
...
+                  const agentEndpoints = runtimeEndpoints.filter(ep => ep.detail?.startsWith(`${entry.name} `));
+                  return (
+                    <Box key={`${entry.resourceType}-${entry.name}`} flexDirection="column">
+                      <ResourceEntry entry={entry} showRuntime />
+                      {entry.invocationUrl && (
+                        <Text dimColor>
+                          {'  '}URL: {entry.invocationUrl}
+                        </Text>
+                      )}
+                      {agentEndpoints.map(ep => (
```
tejaskash, line 169: **Fragile: matching endpoints to agents via `detail` string prefix.**

```tsx
const agentEndpoints = runtimeEndpoints.filter(ep => ep.detail?.startsWith(`${entry.name} `));
```

Parsing display strings for data routing is brittle. If `getLocalDetail` formatting changes, this silently breaks. And if an agent name contains a space or is a prefix of another name followed by a space, this will misattribute endpoints.

**Suggestion:** Add a structured field (e.g., `parentName`) to `ResourceStatusEntry` and filter on that instead of parsing `detail`.

src/cli/primitives/RuntimeEndpointPrimitive.ts
```diff
@@ -0,0 +1,344 @@
...
+  addScreen(): AddScreenComponent {
+    return null;
+  }
+
+  /**
+   * Stub for future cross-reference validation.
+   * Checks if any gateway targets reference a given runtime endpoint.
+   */
+  private findGatewayTargetReferences(
+    _project: AgentCoreProjectSpec,
```
tejaskash: **Dead code: `findGatewayTargetReferences` is never called.**

The implementation plan intended this to be called from `previewRemove()` for cross-reference validation (blocking removal of endpoints referenced by gateway targets). The shipped `previewRemove()` doesn't call it.

Either wire it into `previewRemove` as the plan intended, or remove it. Dead stubs with `_`-prefixed params add noise.

src/cli/primitives/__tests__/RuntimeEndpointPrimitive.test.ts
```diff
@@ -0,0 +1,14 @@
...
+  const primitive = new RuntimeEndpointPrimitive();
+
+  it('has kind "runtime-endpoint"', () => {
+    expect(primitive.kind).toBe('runtime-endpoint');
+  });
+
+  it('has label "Runtime Endpoint"', () => {
+    expect(primitive.label).toBe('Runtime Endpoint');
+  });
+});
```
tejaskash, line 354: **Thin coverage — only tests two string constants.**

The `add()`, `remove()`, `previewRemove()`, and `getRemovable()` methods have zero unit tests. These contain the core business logic: runtime lookup, duplicate detection, version validation against deployed state, empty-dict cleanup, etc.

The plan outlined tests with mocked ConfigIO for these methods. Worth adding before merge, especially given the remove-ambiguity bug.

src/cli/primitives/RuntimeEndpointPrimitive.ts
```diff
@@ -0,0 +1,344 @@
...
+
+      // Validate version is a positive integer
+      const version = options.version ?? 1;
+      if (!Number.isInteger(version) || version < 1) {
+        return { success: false, error: `Version must be a positive integer (got ${version}).` };
+      }
+
+      // Check version against latest deployed version
+      try {
+        const configIO = new ConfigIO();
```
tejaskash: **Separation of concerns: `add()` constructs its own `ConfigIO`.**

`BasePrimitive` already manages ConfigIO internally via `readProjectSpec()`/`writeProjectSpec()`. Creating a second `ConfigIO()` here bypasses whatever setup or state BasePrimitive maintains.

**Suggestion:** Expose deployed-state reading through BasePrimitive or accept it as a parameter, rather than constructing ConfigIO ad-hoc.

src/cli/tui/screens/runtime-endpoint/AddRuntimeEndpointScreen.tsx
```diff
@@ -0,0 +1,269 @@
...
+            if (isDeployed && num > maxVersion) {
+              setError(`Version must be between 1 and ${maxVersion} (latest deployed version)`);
+              return;
+            }
+          }
+          setActiveField(ENDPOINT_FIELDS[idx + 1]!);
+          setError(null);
+          return;
+        }
+        // Last field — validate and submit
```
tejaskash, line 138: **`parseInt` silently truncates floats in TUI.**

The Commander parser was updated (commit c1ad7ea) to use `Number` instead of `parseInt` so that `3.5` is caught. But this TUI path still uses `parseInt(endpointVersion, 10)`, which silently truncates `3.5` to `3`. The downstream `add()` does an `isInteger` check so it won't cause data corruption, but the user gets no TUI-level feedback about the truncation.

Consider using `Number(endpointVersion)` here too, then checking `Number.isInteger()` to show an inline error.

src/cli/commands/status/command.tsx
```diff
@@ -153,15 +155,41 @@ export const registerStatus = (program: Command) => {
...
+              </Box>
+            )}
+
+            {agents.length === 0 && runtimeEndpoints.length > 0 && (
+              <Box flexDirection="column" marginTop={1}>
+                <Text bold>Runtime Endpoints</Text>
+                {runtimeEndpoints.map(ep => (
+                  <Text key={`ep-${ep.detail}-${ep.name}`}>
+                    {'  '}◉ {ep.name} <Text dimColor>{ep.detail}</Text>{' '}
+                    <Text color={DEPLOYMENT_STATE_COLORS[ep.deploymentState] ?? 'gray'}>
```
tejaskash, line 189: Minor: Using `ep.detail` in a React key is fragile since `detail` is a display string that could change format. Prefer a structured identifier like `agentName/ep.name` if you add the `parentName` field suggested above.

src/cli/primitives/RuntimeEndpointPrimitive.ts
tejaskash: **Bug: `remove()` is ambiguous across runtimes with same-name endpoints.**

This searches all runtimes and deletes from the first match. If two runtimes both have a `prod` endpoint, this always removes from whichever runtime appears first in the array. The PR description (test case 7a) explicitly says "same endpoint name on different runtimes" should work.

`previewRemove()`, `getRemovable()`, and the TUI remove flow all share this issue — `name` alone isn't unique.

**Suggestion:** Use `runtimeName/endpointName` as the composite key throughout the remove path, matching the pattern `PolicyPrimitive` uses with `engineName/policyName`. `getRemovable()` should set `name` to the composite form.

src/cli/commands/status/command.tsx
tejaskash, line 169: **Fragile: matching endpoints to agents via `detail` string prefix.**

```tsx
const agentEndpoints = runtimeEndpoints.filter(ep => ep.detail?.startsWith(`${entry.name} `));
```

Parsing display strings for data routing is brittle. If `getLocalDetail` formatting changes, this silently breaks. And if an agent name contains a space or is a prefix of another name followed by a space, this will misattribute endpoints.

**Suggestion:** Add a structured field (e.g., `parentName`) to `ResourceStatusEntry` and filter on that instead of parsing `detail`.

src/cli/primitives/RuntimeEndpointPrimitive.ts
tejaskash: **Dead code: `findGatewayTargetReferences` is never called.**

The implementation plan intended this to be called from `previewRemove()` for cross-reference validation (blocking removal of endpoints referenced by gateway targets). The shipped `previewRemove()` doesn't call it.

Either wire it into `previewRemove` as the plan intended, or remove it. Dead stubs with `_`-prefixed params add noise.

src/cli/primitives/__tests__/RuntimeEndpointPrimitive.test.ts
tejaskash, line 354: **Thin coverage — only tests two string constants.**

The `add()`, `remove()`, `previewRemove()`, and `getRemovable()` methods have zero unit tests. These contain the core business logic: runtime lookup, duplicate detection, version validation against deployed state, empty-dict cleanup, etc.

The plan outlined tests with mocked ConfigIO for these methods. Worth adding before merge, especially given the remove-ambiguity bug.

src/cli/primitives/RuntimeEndpointPrimitive.ts
tejaskash: **Separation of concerns: `add()` constructs its own `ConfigIO`.**

`BasePrimitive` already manages ConfigIO internally via `readProjectSpec()`/`writeProjectSpec()`. Creating a second `ConfigIO()` here bypasses whatever setup or state BasePrimitive maintains.

**Suggestion:** Expose deployed-state reading through BasePrimitive or accept it as a parameter, rather than constructing ConfigIO ad-hoc.

src/cli/tui/screens/runtime-endpoint/AddRuntimeEndpointScreen.tsx
tejaskash, line 138: **`parseInt` silently truncates floats in TUI.**

The Commander parser was updated (commit c1ad7ea) to use `Number` instead of `parseInt` so that `3.5` is caught. But this TUI path still uses `parseInt(endpointVersion, 10)`, which silently truncates `3.5` to `3`. The downstream `add()` does an `isInteger` check so it won't cause data corruption, but the user gets no TUI-level feedback about the truncation.

Consider using `Number(endpointVersion)` here too, then checking `Number.isInteger()` to show an inline error.

src/cli/commands/status/command.tsx
tejaskash, line 189: Minor: Using `ep.detail` in a React key is fragile since `detail` is a display string that could change format. Prefer a structured identifier like `agentName/ep.name` if you add the `parentName` field suggested above.

tejaskash (commented):
Good feature implementation that follows the established primitive patterns closely. Schema design, deploy pipeline integration, and TUI wizard are well done.

**Main issues to address before merge:**
1. `remove()` is ambiguous when same-name endpoints exist across runtimes (your own test case 7a). Needs composite key like policies use.
2. Status `--type runtime-endpoint` agent matching parses `detail` strings — fragile if agent names share prefixes.

See inline comments for details and suggestions.

src/cli/primitives/RuntimeEndpointPrimitive.ts
jariy17: Fixed in 11a7a86. `remove()`, `previewRemove()`, and `getRemovable()` now use composite `runtimeName/endpointName` keys throughout, matching the PolicyPrimitive pattern. Legacy bare name fallback kept for backwards compatibility.

src/cli/commands/status/command.tsx
jariy17: Fixed in 11a7a86. Added `parentName` field to `ResourceStatusEntry`. Status command now filters with `ep.parentName === entry.name` instead of parsing detail strings. React keys use `parentName/name`. Detail field no longer includes agent name prefix.

src/cli/commands/status/command.tsx
jariy17: Also fixed in 11a7a86. React keys now use `${ep.parentName}/${ep.name}` in both the nested and standalone sections.

src/cli/primitives/__tests__/RuntimeEndpointPrimitive.test.ts
jariy17: Fixed in d6389c8. Added 23 unit tests with mocked ConfigIO covering:
- `add()`: runtime lookup, duplicate detection, version validation (int, float, deployed range), response shape
- `remove()`: composite key targeting, legacy fallback, empty-dict cleanup, same-name-across-runtimes
- `previewRemove()`: summary content, schema changes, not-found error
- `getRemovable()`: composite keys, multi-runtime, empty states

src/cli/primitives/RuntimeEndpointPrimitive.ts
jariy17: Removed in b9f158f. Gateway targets don't reference endpoint qualifiers yet (out of scope for this proposal), so the stub was dead code. Can be added back when that feature ships.

src/cli/primitives/RuntimeEndpointPrimitive.ts
jariy17: Fixed in 0ff1db8. Now uses `this.configIO` from BasePrimitive instead of constructing a new ConfigIO. Removed the unused import.

src/cli/tui/screens/runtime-endpoint/AddRuntimeEndpointScreen.tsx
jariy17: Fixed in be47f35. Both TUI validation spots now use `Number()` + `Number.isInteger()` instead of `parseInt`. Floats like 3.5 now show an inline error instead of being silently truncated.

jariy17:
## Final Verification Report

All test flows verified with bundle `aws-agentcore-0.11.0-1777326506`:

- **5/5 CLI commands** — add, duplicate rejection, not-found, float version, composite key remove
- **6/6 TUI flows** — add menu, runtime picker, multi-field form, confirm, success, remove
- **2/2 deploys** — phase 1 (runtime only), phase 2 (runtime + endpoints) to account 725476964917
- **AWS verified** — `AWS::BedrockAgentCore::RuntimeEndpoint` resources `CREATE_COMPLETE`, ARNs in deployed-state.json
- **14/14 review comments** resolved

Full HTML report with inline screenshots: https://gist.github.com/jariy17/f73e36a15be05e18cc92ac09ef6ec67d

jariy17:
**View the full verification report (rendered HTML with inline screenshots):** https://htmlpreview.github.io/?https://gist.githubusercontent.com/jariy17/f73e36a15be05e18cc92ac09ef6ec67d/raw/test-report-final.html

tejaskash (approved):
All feedback addressed. Composite keys, parentName field, comprehensive tests, and TUI fixes all look good. One minor nit: dangling JSDoc comment before the closing brace of RuntimeEndpointPrimitive (from the removed findGatewayTargetReferences stub) — can clean up in a follow-up.

PR #751, opened 2026-04-01
feat: add full TypeScript agent support

arepageek:
# Add full TypeScript agent support

 ##  Description
  Adds TypeScript as a first-class language for AgentCore agents. TypeScript agents use the @strands-agents/sdk with Express and deploy via Container
  build, since AgentCore Runtime's CodeZip mode only accepts Python runtimes (PYTHON_3_10 through PYTHON_3_14).

 ##  Key design decisions
  - Container build is automatic: When --language TypeScript is selected, the CLI defaults to Container build. Users don't need to specify --build
  Container.
  - Dev server runs locally without Docker: agentcore dev for TypeScript agents uses tsx watch directly (same as CodeZip dev mode), giving fast hot-reload
  without container overhead. Docker/Podman/Finch is only needed for agentcore deploy and agentcore package.
  - CodeZip + TypeScript is explicitly blocked: If a user passes --build CodeZip --language TypeScript, they get a clear error explaining the AWS API
  limitation rather than a cryptic deploy failure.

  ## What's included

  Template assets:
  - src/assets/typescript/http/strands/base/ — Strands TS agent template (Express + @strands-agents/sdk, Handlebars model loader for
  Bedrock/Anthropic/OpenAI/Gemini)
  - src/assets/typescript/mcp/standalone/base/ — MCP server template with @modelcontextprotocol/sdk
  - src/assets/container/typescript/ — Dockerfile (Node 22 slim) + dockerignore

  CLI changes:
  - create/command.tsx — Removed hardcoded CodeZip default so language-aware default flows through
  - create/action.ts — buildType defaults to Container for TypeScript, CodeZip for Python
  - create/validate.ts, add/validate.ts — TypeScript enabled; TS + CodeZip rejected with clear message
  - schema-mapper.ts — Language-aware agent spec: main.ts + NODE_22 + Container for TS
  - defaults.ts — Added DEFAULT_NODE_VERSION and DEFAULT_NODE_ENTRYPOINT
  - types.ts — Enabled TypeScript in LANGUAGE_OPTIONS

  Dev server:
  - config.ts — isNodeAgent() helper; .ts/.js entrypoints now dev-supported
  - codezip-dev-server.ts — ensureNodeDeps() runs npm install; Node spawn uses tsx watch
  - server.ts — TypeScript Container agents use CodeZipDevServer (no Docker for dev)
  - useDevServer.ts — Added 'listening on port' to Express ready detection patterns

  Docs:
  - README.md — Prerequisites, project structure for TS
[...]

PR #981, opened 2026-04-27
fix(typescript): typescript agent template and streaming fixes

jesseturner21:
## Description

Fixes several issues with TypeScript agent support in the CLI:

1. **Stream event types**: The TypeScript Strands template was using incorrect stream event field names, causing streaming responses to not render.
2. **HTTP invoke Accept header**: The invoke command was missing `text/event-stream` in the Accept header, breaking SSE streaming for HTTP agents.
3. **Memory not supported in TS**: The generate wizard offered memory configuration for TypeScript agents, but the Strands TS SDK doesn't support the memory session manager yet. This disables the memory step and removes the memory-related Handlebars blocks and `session.ts` capability file from TypeScript templates.
4. **Protocol options UX**: Instead of showing disabled protocol options for TypeScript, the wizard now filters to only show HTTP.

## Related Issue

Closes #

## Documentation PR

N/A

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.


docs/TYPESCRIPT_SUPPORT_TEST_PLAN.md
```diff
@@ -0,0 +1,433 @@
...
+Hands-on verification checklist for TypeScript agent support. Work top-to-bottom. Each step has explicit commands, an
+expected outcome, and a status box. Record results inline (`[x]` pass, `[!]` fail + note, `[~]` partial / skipped).
+
+**Scope:** TypeScript + Strands HTTP agents. Other frameworks (LangChain, CrewAI, GoogleADK, OpenAIAgents) in TS are
+explicitly out of scope and should reject.
+
+**AWS test account:** `325335451438` via `AWS_PROFILE=deploy`. Refresh with:
+
+```bash
+ada credentials update --account 325335451438 --provider isengard --role Admin --profile deploy --once
```
avi-alpert: this should not be included

avi-alpert:
Local traces isnt working - need to auto instrument like we do with python agents

avi-alpert:
Note: just saw your PR to fix SDK to not use hardcoded 8080 port. Once that's merged, I will make the corresponding change here in the web ui start.ts

jesseturner21:
merged in the 8080 update in the typescript sdk - https://github.com/aws/bedrock-agentcore-sdk-typescript/commit/25a5ef748968ac0b9af1f823160bb518483f1099

docs/TYPESCRIPT_SUPPORT_TEST_PLAN.md
jesseturner21: fixed my bad

docs/BUGBASH-TS-STATUS.md
```diff
@@ -0,0 +1,154 @@
+# TypeScript Agent Support Bugbash — Status & Handoff
```
jesseturner21: remove this

docs/commands.md
```diff
@@ -51,6 +51,13 @@ agentcore create \
 # Skip agent creation
 agentcore create --name MyProject --no-agent
 
+# TypeScript (Strands-only)
```
jesseturner21: fix this it can be vercel or strands

docs/frameworks.md
```diff
@@ -3,6 +3,17 @@
 AgentCore CLI supports multiple agent frameworks for template-based agent creation, plus a BYO (Bring Your Own) option
 for existing code.
 
+## Supported Languages
+
+| Language   | Supported Frameworks | Runtime      | Notes                                                                              |
+| ---------- | -------------------- | ------------ | ---------------------------------------------------------------------------------- |
+| Python     | All frameworks       | Python 3.12+ | Default language. Uses `uv` for dependency management.                             |
+| TypeScript | Strands only         | Node 22      | Uses `npm` + `tsx` for the dev loop. Other frameworks are not yet available in TS. |
```
jesseturner21: fix this this can be vercel

docs/TYPESCRIPT_SUPPORT_HANDOFF.md
```diff
@@ -0,0 +1,176 @@
+# TypeScript (Strands) support — work-in-progress handoff
```
jesseturner21: remove this file

src/cli/commands/create/validate.ts
```diff
@@ -184,9 +192,12 @@ export function validateCreateOptions(options: CreateOptions, cwd?: string): Val
       return { valid: false, error: `Invalid model provider: ${options.modelProvider}` };
     }
 
-    // Validate language is supported
-    if (options.language === 'TypeScript') {
-      return { valid: false, error: 'TypeScript is not yet supported. Currently supported: Python' };
+    // TypeScript is Strands-only for now
```
jesseturner21: can we fix this comment

src/assets/typescript/http/strands/base/main.ts
```diff
@@ -0,0 +1,71 @@
...
+  description: 'Return the sum of two numbers',
+  inputSchema: {
+    type: 'object',
+    properties: {
+      a: { type: 'number' },
+      b: { type: 'number' },
+    },
+    required: ['a', 'b'],
+  },
+  handler: async ({ a, b }: { a: number; b: number }) => a + b,
```
aidandaly24: **[CRITICAL — verified against `@strands-agents/sdk@1.0.0-rc.4`]** This passes `handler:` but the SDK's `FunctionToolConfig` (`dist/src/tools/function-tool.d.ts:45-54`) and `ZodToolConfig` (`dist/src/tools/zod-tool.d.ts:16-34`) require **`callback:`**. Neither config has a `handler` field — the user-supplied function is silently never invoked. The `addNumbers` example tool ships broken in every Strands TS scaffold.

**Fix:** rename `handler` → `callback`. Add a unit test that mocks `tool()` and asserts the example registers a callable tool.

src/assets/typescript/http/strands/base/model/load.ts
```diff
@@ -0,0 +1,102 @@
...
+
+async function getApiKey(): Promise<string> {
+  if (process.env.LOCAL_DEV === '1') {
+    const apiKey = process.env[IDENTITY_ENV_VAR] ?? process.env.ANTHROPIC_API_KEY;
+    if (!apiKey) {
+      throw new Error(`${IDENTITY_ENV_VAR} or ANTHROPIC_API_KEY not found. Add your key to agentcore/.env.local`);
+    }
+    return apiKey;
+  }
+  return withApiKey({ providerName: IDENTITY_PROVIDER_NAME })(async (apiKey: string) => apiKey)();
```
aidandaly24, line 23: **[CRITICAL — verified against `bedrock-agentcore@0.2.4`]** `withApiKey` is a HOF wrapper that **injects** the apiKey as the last parameter of a wrapped function — verified against `dist/src/identity/wrappers.d.ts:52`. Per `dist/src/identity/types.d.ts:80-89`, `ApiKeyWrapperConfig.workloadIdentityToken` is optional with the doc string: *"if not provided, automatically falls back to `context.workloadAccessToken` when called within a request handler. **If neither is available, an error is thrown.**"*

`loadModel()` is invoked from `getOrCreateAgent()` at module-init / agent-cache time — **before** any request context exists. The SDK throws `"neither workloadIdentityToken nor context.workloadAccessToken is available"` on the first invocation. Bedrock paths don't go through `withApiKey`, which is why bug-bash didn't catch this.

Same pattern at lines `:55`, `:87`, and `src/assets/typescript/http/vercelai/base/model/load.ts:38, 71, 104`.

**Fix:** call `loadModel()` from inside `process(payload, context)` and pass `workloadIdentityToken: context.workloadAccessToken`. The current cache-at-boot pattern is incompatible with the SDK's request-scoped credential design.

src/schema/constants.ts
```diff
@@ -152,10 +155,25 @@ export const DEFAULT_PYTHON_VERSION: PythonRuntime = 'PYTHON_3_14';
...
+export const DEFAULT_NODE_VERSION: NodeRuntime = 'NODE_22';
+
 /** Combined runtime version schema supporting both Python and Node/TypeScript runtimes */
 export const RuntimeVersionSchema = z.union([PythonRuntimeSchema, NodeRuntimeSchema]);
 export type RuntimeVersion = z.infer<typeof RuntimeVersionSchema>;
 
+/** Default entrypoint filename for each target language (create path). */
+export const DEFAULT_ENTRYPOINT_BY_LANGUAGE: Record<'Python' | 'TypeScript', string> = {
+  Python: 'main.py',
+  TypeScript: 'main.js',
```
aidandaly24, line 168: **[HIGH — entrypoint convention incoherent across 5 sites]**

1. This line records `'main.js'`
2. `src/cli/operations/agent/generate/schema-mapper.ts:124-125` writes `'main.js'` into `agentcore-project.json`
3. `src/lib/packaging/node.ts:95` and `:149` hardcode `join(srcDir, 'main.ts')` — ignoring `spec.entrypoint` entirely
4. `src/assets/container/typescript/Dockerfile:25` hardcodes `CMD ["npx", "tsx", "main.ts"]` — also ignoring spec
5. CDK `agentcore-l3-cdk-constructs/.../AgentCoreRuntime.ts:177-189` writes `agent.entrypoint` straight into CFN

Default scaffolds work because `main.js` ends up in the zip and CFN reads `main.js`. **But BYO TS users cannot customize `entrypoint`** — the packager and Dockerfile both ignore it. If a user changes `entrypoint` to `server.ts`, nothing honors it.

**Fix:** pick one convention end-to-end. Recommended: record `'main.ts'` (the source file the user edits), have the packager rewrite `entryPoint` to `main.js` for Node. Or follow `spec.entrypoint` everywhere. Add a unit test asserting the file named in `agent.entrypoint` exists in the produced zip.

src/assets/typescript/http/strands/base/package.json
```diff
@@ -0,0 +1,39 @@
...
+    {{#if (eq modelProvider "Anthropic")}}
+    "@anthropic-ai/sdk": "^0.71.2",
+    {{/if}}
+    {{#if (eq modelProvider "OpenAI")}}
+    "openai": "^6.7.0",
+    {{/if}}
+    {{#if (eq modelProvider "Gemini")}}
+    "@google/genai": "^1.40.0",
+    {{/if}}
+    "@aws-sdk/client-s3": "^3.943.0",
```
aidandaly24: **[HIGH — bloat + version skew + misleading docs]**

- `@aws-sdk/client-s3` (line 22) — never imported by `main.ts`/`load.ts`/`client.ts`
- `express` (line 26) — never imported (Strands SDK ships its own server)
- `zod` `^4.1.12` (line 28) — never imported, AND VercelAI ships `zod@^3.24` → peer-dep hell when sharing utilities between agents

Compounded by README at `src/assets/typescript/http/strands/base/README.md:11-12` which claims a "Fastify-based HTTP app" — but the package ships `express`, not `fastify`. `main.ts` doesn't import either.

**Fix:** drop unused deps. Pin `zod` to one major across both TS templates. Fix README — either remove the framework name (server is an SDK detail) or correct it.

src/assets/typescript/http/strands/base/main.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+const app = new BedrockAgentCoreApp({
+  invocationHandler: {
+    async *process(payload: any, context: any) {
+      const agent = await getOrCreateAgent();
+
+      for await (const event of agent.stream(payload.prompt ?? '')) {
+        if (
+          event.type === 'modelStreamUpdateEvent' &&
```
aidandaly24, line 60: **[HIGH — silent-failure on next SDK rev]** This three-level nested filter (`modelStreamUpdateEvent` → `modelContentBlockDeltaEvent` → `textDelta`) is exactly the kind of string matching that drifts between Strands TS SDK RCs and produces empty agent output without any error.

The snapshot test at `assets.snapshot.test.ts.snap` locks bytes. When commit `32d006a9` updated the field names, the snapshot was updated in lockstep — meaning the next field-name regression will silently re-pass against new bytes. **A snapshot-update commit hides the very regression the test should catch.**

The e2e test asserts `json.success`, not stream contents — an empty stream still passes.

**Fix:** add a unit test that mocks `agent.stream(...)` to return a fixture event matching v1.0.0-rc.4 shape; assert response stream is non-empty. Have e2e assert non-empty content. Add an inline comment linking the SDK source line / version that defines these field names.

src/assets/typescript/http/strands/base/README.md
```diff
@@ -0,0 +1,37 @@
...
+| Variable | Required | Description |
+| --- | --- | --- |
+{{#if hasIdentity}}| `{{identityProviders.[0].envVarName}}` | Yes | {{modelProvider}} API key (local) or Identity provider name (deployed) |
+{{/if}}| `LOCAL_DEV` | No | Set to `1` to use `.env.local` instead of AgentCore Identity |
+
+# Developing locally
+
+If installation was successful, `node_modules/` is already populated with dependencies.
+
+`agentcore dev` will start a local server on 0.0.0.0:8080 using `npx tsx watch main.ts` for hot reload.
```
aidandaly24: **[HIGH — stale + LAN-exposure risk]**

1. **Port:** commit `485cbdab` ("fix(dev): use dynamic port assignment for TS HTTP agents in web UI") and `web-ui/handlers/start.ts:94-115` make the port dynamic. README claims fixed `8080`. Same staleness in `vercelai/base/README.md:27`.
2. **Bind address:** `main.ts:71` only specifies `port`, not `host`. Python equivalent explicitly binds `127.0.0.1` (`codezip-dev-server.ts:190`). The actual host depends on `BedrockAgentCoreApp.run()`'s default. If it's `0.0.0.0`, a developer on coffee-shop Wi-Fi exposes their local agent (with AWS creds in scope) to the network.

**Fix:**
1. Replace with: "`agentcore dev` will start a local server on a port logged to the terminal (default `8080` in single-agent mode, dynamic in web-UI mode)."
2. Have templates explicitly pass `host: '127.0.0.1'` to `app.run()` to match the Python path's localhost-only binding.

src/assets/container/typescript/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+WORKDIR /app
+
+ENV NODE_ENV=production \
+    DOCKER_CONTAINER=1
+
+RUN userdel -r node 2>/dev/null || true
+RUN useradd -m -u 1000 bedrock_agentcore
+
+COPY package.json package-lock.json* ./
+RUN npm ci --omit=dev || npm install --omit=dev
```
aidandaly24, line 12: **[HIGH — supply-chain + prod-bloat]**

1. **`npm ci --omit=dev || npm install --omit=dev`** silently regenerates `package-lock.json` if `npm ci` fails. Any drift between lockfile and `package.json` causes the build to switch from a pinned install to a floating one — exactly the attack surface lockfiles exist to close.
2. **Lifecycle scripts run by default** — combined with `^` ranges in template `package.json`, the *first build* of any scaffolded project runs whatever postinstall hooks the latest matching versions ship.
3. **Line 25:** `CMD ["npx", "tsx", "main.ts"]` runs through `tsx` on every container start, shipping a full TS toolchain into the customer's prod image. Cold-start penalty + supply-chain surface.

**Fix:**
- `npm ci --ignore-scripts`; fail the build if the lockfile is missing
- Compile to JS at build time, run `node dist/main.js`
- Make CMD honor `spec.entrypoint` via build arg (cross-references the entrypoint convention finding above)

src/cli/operations/dev/codezip-dev-server.ts
```diff
@@ -67,6 +67,35 @@ function ensurePythonVenv(
...
+ * install if `node_modules` is missing. Detects pnpm/yarn via lockfile, else npm.
+ */
+function ensureNodeDeps(cwd: string, onLog: (level: LogLevel, message: string) => void): boolean {
+  if (existsSync(join(cwd, 'node_modules'))) {
+    return true;
+  }
+
+  let cmd = 'npm';
+  let args = ['install'];
+  if (existsSync(join(cwd, 'pnpm-lock.yaml'))) {
```
aidandaly24, line 81: **[HIGH — supply-chain + UX]**

Three problems in `ensureNodeDeps`:
1. **No `--ignore-scripts` / `--frozen-lockfile`** — every `agentcore dev` runs full lifecycle scripts of every dep at full privilege, with AWS creds in the same shell.
2. **pnpm/yarn detection vs `setup.ts`** — `setupNodeProject` always uses npm; this file detects lockfiles and picks pnpm/yarn. Templates ship without lockfiles, so this branch only fires if a user dropped a foreign lockfile from another project. When pnpm/yarn isn't installed, `spawnSync` returns `status: null` and the check at line 91 (`result.status !== 0`) treats it as a generic install failure — confusing error for the user.
3. **Synchronous spawnSync blocks the Ink renderer** — minutes on a slow connection. Force-killing during install leaves a half-broken `node_modules/` and the next run trusts `existsSync(node_modules)` (line 75) and skips install.

**Fix:**
- Default to `--ignore-scripts`; require explicit `--allow-scripts` flag (or interactive prompt) before lifecycle execution
- Drop pnpm/yarn detection; use `npm install` consistently
- Replace `spawnSync` with async `spawn` + line-buffered streaming through `onLog`

src/cli/operations/dev/codezip-dev-server.ts
```diff
@@ -118,9 +147,11 @@ export class CodeZipDevServer extends DevServer {
     }
 
     if (!isPython) {
+      // TS entrypoint is already a file path like "main.ts" — pass it straight to tsx.
+      const entryFile = module.split(':')[0] ?? module;
       return {
         cmd: 'npx',
-        args: ['tsx', 'watch', (module.split(':')[0] ?? module).replace(/\./g, '/') + '.ts'],
+        args: ['tsx', 'watch', entryFile],
```
aidandaly24, line 154: **[HIGH — process orphan on Ctrl-C]** `npx tsx watch` spawns a 3-level process tree: `npx` → `tsx` → `node`. `DevServer.kill()` (`dev-server.ts:81`) only sends SIGTERM to the top-level `npx`. On Linux, `npx` does not propagate SIGTERM to descendants by default, and `tsx` doesn't always tear down the child Node process when its parent dies.

Result: bound socket on port 8080 + the user's compiled agent stay alive after the dev server is "stopped". Subsequent `agentcore dev` runs hit "Port 8080 in use, using 8081" and so on. Python path doesn't have this layering (uvicorn spawned directly).

**Repro:** `agentcore dev` against a TS project → "Server ready" → Ctrl-C twice → `lsof -iTCP:8080 -sTCP:LISTEN` shows the node process still bound.

**Fix:** use `detached: true` + `process.kill(-child.pid, 'SIGTERM')` so SIGTERM hits the whole process group. Or invoke `tsx` directly via `./node_modules/.bin/tsx watch ...` to remove the npm shim.

aidandaly24 (changes requested):
## Synthesized review from a 15-reviewer team pass

Thanks for the big push on TS support. I ran a multi-reviewer review (security, determinism, CDK design, cross-PR contract, tests, perf, template UX, silent failures, schema, dev UX, type design, observability, docs, SDK contract, maintainability) and verified the two most damaging findings against the actual published SDKs (`@strands-agents/sdk@1.0.0-rc.4` and `bedrock-agentcore@0.2.4` tarballs) before stamping severity.

### 2 CRITICAL findings (verified against published SDK type defs) — see inline

- **Strands template's `tool({ ..., handler: ... })`** — SDK requires `callback`, not `handler`. The example `addNumbers` tool ships **broken** in every Strands TS scaffold (verified against `dist/src/tools/{function-tool,zod-tool}.d.ts`).
- **`withApiKey({ providerName })(async (apiKey) => apiKey)()`** is misuse — `withApiKey` is a HOF that injects apiKey, requires `workloadIdentityToken` or a request-scoped `context.workloadAccessToken`, and `loadModel()` is called at module/agent-cache time before any context exists. Every non-Bedrock TS deploy will throw on first invocation (verified against `dist/src/identity/wrappers.d.ts` + `types.d.ts`).

### HIGH findings — see inline

- Entrypoint convention is incoherent across CLI persists / packager reads / Dockerfile uses / CDK writes — default scaffolds happen to work, BYO entrypoint customization is broken
- Strands `package.json` ships unused deps and README claims wrong framework
- Strands stream-event filter has zero test (snapshot updates lockstep with template)
- README claims `0.0.0.0:8080` but dev server uses dynamic ports; templates don't bind `127.0.0.1`
- Dockerfile `npm ci || npm install` regenerates lockfile; runs lifecycle scripts; ships `tsx` toolchain into prod
- `agentcore dev` runs `npm install` with full lifecycle scripts; pnpm/yarn detection produces misleading errors
- `tsx watch` orphan on Ctrl-C → ports stay bound across runs

### HIGH findings on lines this PR doesn't directly touch (action requested but anchored to an unchanged file, so they live in the body)

**`src/cli/aws/__tests__/agentcore.test.ts:200` — stale Accept-header assertion.** Production code (`src/cli/aws/agentcore.ts:321, 417`) was fixed to `'application/json, text/event-stream'` in commit `a21dbfda`. The test still asserts `expect(headers.Accept).toBe('application/json')` and passes only because it constructs the helper with a different second arg than production does. **No test verifies the SigV4 paths actually request both content types.** Update assertion + add a regression test that mocks the SDK client and asserts `accept` includes `text/event-stream`.

**`src/cli/telemetry/schemas/common-shapes.ts:70` — `Framework` enum missing `vercelai`.** This PR adds VercelAI as a first-class TS framework but the telemetry enum still only knows `['strands', 'langchain_langgraph', 'googleadk', 'openaiagents']`. `resilientParse` replaces unknown values with `'unknown'`, so every `agentcore create --framework VercelAI` reports `framework=unknown`. Cannot measure VercelAI adoption or regression rate. Compounded by `src/cli/commands/deploy/utils.ts:5-31` (`computeDeployAttrs`) having zero language attribution. Add `'vercelai'` to `Framework` and extend `computeDeployAttrs` with `has_typescript_runtime` / `frameworks: string[]`.

**`src/lib/packaging/helpers.ts:214,415` — CLI zip writer has no `mtime` pin.** Both `collectFiles` and `collectFilesSync` pass `[buffer, { level: 6 }]` to fflate — no `mtime`. The CDK side at `agentcore-l3-cdk-constructs/.../helpers.ts:368` pins `mtime: new Date('1980-01-02T00:00:00Z')` precisely to avoid this. Every CLI-driven CodeZip publish creates a new AgentCore Runtime version even with zero source changes. Mirror the CDK constant.

### Top 3 things to look at yourself

1. **Reconcile the entrypoint contract end-to-end** — picks a convention and unblocks customization for BYO TS users.
2. **Decide who owns `npm install`** — CLI runs it on `dev`/`create`, CDK construct doesn't, the bundler silently no-ops if deps missing.
3. **`'agentcore'` regression in `EXCLUDED_ENTRIES`** — paired CDK PR re-introduces a fix that was reverted in `4883e1c`.

Paired CDK review: aws/agentcore-l3-cdk-constructs#181.

src/assets/typescript/http/strands/base/main.ts
jesseturner21: Fixed in 402af8e8 — renamed `handler` to `callback` to match `FunctionToolConfig` in `@strands-agents/sdk`.

src/assets/typescript/http/strands/base/model/load.ts
jesseturner21: Investigated — this is a false positive. `withApiKey` is called as a self-contained expression: `withApiKey({ providerName })(async (apiKey) => apiKey)()`. It doesn't need request context because the AgentCore Runtime provides ambient workload identity credentials via the environment. We verified non-Bedrock providers (OpenAI, Anthropic) work end-to-end when deployed.

src/schema/constants.ts
jesseturner21: Acknowledged — the entrypoint convention is intentionally `main.js` (the compiled output) for the CFN/runtime layer, while the source is `main.ts`. The packager handles the ts→js compilation. BYO TS customization is not yet a supported path. Will track as a follow-up when BYO TS is added.

src/assets/typescript/http/strands/base/package.json
jesseturner21: Fixed in 904cd703 — removed `@aws-sdk/client-s3`, `express`, and `zod` from the template package.json. None were imported.

src/assets/typescript/http/strands/base/main.ts
jesseturner21: Acknowledged. The stream event shape is coupled to the SDK version which is pinned (`1.0.0-rc.4`). When we bump the SDK, we'll update the filter in lockstep. Adding a unit test that mocks the stream shape is a good idea — will track as a follow-up for test coverage improvements.

src/assets/typescript/http/strands/base/README.md
jesseturner21: Fixed in 18e4f89c — removed hardcoded `0.0.0.0:8080` from both Strands and VercelAI READMEs, and corrected the "Fastify-based" description. Re: bind address — `BedrockAgentCoreApp.run()` binds to `127.0.0.1` by default when `LOCAL_DEV=1`, so localhost-only is already enforced in dev mode.

src/assets/container/typescript/Dockerfile
jesseturner21: Acknowledged. The `npm ci || npm install` fallback is intentional — scaffolded projects won't have a lockfile on first build, so `npm ci` alone would always fail. The `tsx` runtime in prod is a known tradeoff for DX (no build step needed). Compiling to JS at build time is planned for a future optimization pass. Will track `--ignore-scripts` as a hardening follow-up.

src/cli/operations/dev/codezip-dev-server.ts
jesseturner21: Acknowledged. The `--ignore-scripts` concern applies to npm globally, not specific to us. The sync spawn and pnpm/yarn detection are areas for improvement. Will track as follow-up hardening work.

src/cli/operations/dev/codezip-dev-server.ts
jesseturner21: Good catch. Will fix — switching to process group kill (`detached: true` + `process.kill(-pid)`) or invoking `tsx` directly via `./node_modules/.bin/tsx` to avoid the npx shim layer. Tracking as a follow-up fix.

aidandaly24 (approved):
Thanks for making the updates, tested some of these false positives out and they were indeed false positives. LGTM!

notgitika:
Thank you to @arepageek for contributing with an initial prototype of typescript runtime support in the CLI earlier.

https://github.com/aws/agentcore-cli/pull/751 🚀

PR #1001, opened 2026-04-28
fix(e2e): add debug logging for gateway import CI failures

jesseturner21:
## Summary
- Add import log file output and CloudFormation stack events to the gateway import test when it fails
- This will help diagnose the `IMPORT_ROLLBACK_IN_PROGRESS. Reason: Unknown` error seen in CI

## Context
The gateway import test passes locally but fails in CI with an opaque CloudFormation error. This PR adds debug output so we can see the actual failure reason on the next CI run.

## Test plan
- [ ] CI runs and either passes or produces useful debug output

### DIFF
--- a/e2e-tests/e2e-helper.ts
+++ b/e2e-tests/e2e-helper.ts
@@ -14,7 +14,7 @@ import {
 } from '@aws-sdk/client-bedrock-agentcore-control';
 import { execSync } from 'node:child_process';
 import { randomUUID } from 'node:crypto';
-import { mkdir, rm, writeFile } from 'node:fs/promises';
+import { mkdir, readFile, rm, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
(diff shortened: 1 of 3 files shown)


e2e-tests/e2e-helper.ts
```diff
@@ -377,3 +377,44 @@ export async function teardownE2EProject(projectPath: string, agentName: string,
...
+  console.log(`Import ${label} stdout:`, result.stdout);
+  console.log(`Import ${label} stderr:`, result.stderr);
+
+  const logMatch = /Log: (.+)/.exec(result.stderr);
+  if (logMatch) {
+    try {
+      const logContents = await readFile(join(projectPath, logMatch[1]!), 'utf-8');
+      console.log(`Import ${label} log:\n`, logContents);
+    } catch {
+      /* log file may not exist */
```
aidandaly24, line 397: Nit: maybe we should throw a warning here?

aidandaly24 (approved):

PR #989, opened 2026-04-27
test: speed up CI and fix mock cleanup gaps

tejaskash:
## Summary

Day 1 of the test suite improvement plan (see `docs/test-improvement-plan.md` in workspace root).

- **Node 20 only on PRs**: Full matrix (20/22/24) still runs on push to main, but PRs only run Node 20. Cuts CI jobs from 3x to 1x for PRs.
- **3-way vitest sharding**: Unit tests split across 3 parallel shard jobs using blob reporter. A merge-reports job recombines results and generates coverage. Wall-clock for unit tests should drop from ~3 min to ~1-1.5 min.
- **deps.optimizer**: Pre-bundles `@aws-sdk/*`, `@smithy/*`, `zod`, and `commander` via vitest's SSR optimizer to reduce per-worker import overhead.
- **Exclude tui-harness from unit tests**: These are proof-of-concept tests for the TUI harness, not production code. Excluded from the unit test project.
- **Mock cleanup fixes**: Added `afterEach(vi.restoreAllMocks)` to 3 test files that had `vi.mock` without cleanup. Moved inline `consoleSpy.mockRestore()` calls to `afterEach` in logs-eval tests.
- **PTY test skip**: Added `describe.skipIf(!ptyAvailable)` around PTY-dependent tests that fail when `node-pty` spawn is unavailable.

Note: The CDK repo (`agentcore-l3-cdk-constructs`) Node 20 matrix change will be a separate PR.

## Test plan

- [ ] CI runs 3 shard jobs (Node 20 only) on this PR instead of 3 Node version jobs
- [ ] merge-reports job successfully combines blob reports and produces coverage
- [ ] Coverage report comment appears on this PR
- [ ] No new test failures compared to main
- [ ] PTY tests show as skipped, not failed

PR #1016, opened 2026-04-29
ci: cut full e2e time in half via vitest sharding

Hweinstock:
## Description

The full E2E suite takes 19-25 min because all 17 test files run on a single runner per CDK source. Each test deploys its own CloudFormation stack, and measured timings show parallel deploys have zero degradation (77s stays 77s with concurrent stacks).

This adds 6-way vitest sharding via the GitHub Actions matrix, going from 2 to 12 parallel runners. Also isolates import test resources per run to prevent conflicts when `npm` and `main` matrix entries hit the same AWS account concurrently.

**Measured results:**

| Shards | Wall-clock | Slowest shard |
|--------|-----------|---------------|
| 1 (baseline) | 19-25 min | — |
| 4 | ~15 min | 890s |
| 6 | **~11.5 min** | 691s |
| 8 | ~12.0 min | 717s |

6 shards appears to be the sweet spot. Adding more shards creates diminishing returns since the main bottleneck is the deploy (274s) and container build (204s). This is roughly what the e2e tests take on PRs when running minimal files so we're likely close to how fast we can get it. 

Inspired by https://github.com/aws/agentcore-cli/pull/989.

Docs: https://vitest.dev/guide/improving-performance.html#sharding

### Import test resource isolation

The `import-resources.test.ts` setup creates AWS resources (runtime, memory, evaluator) via Python scripts, then saves their ARNs to `bugbash-resources.json`. The cleanup script deletes those resources and also calls `cleanup_s3_code_objects()` which previously deleted **all** objects from the shared S3 bucket `bugbash-agentcore-code-{account}-{region}`.

The `npm` and `main` matrix entries are separate runners hitting the same AWS account simultaneously. Both run `import-resources.test.ts`. When one job's cleanup finishes first, it nuked all S3 objects — including the `code.zip` the other job uploaded and was still using, causing downstream resource creation failures.

The fix: each run now gets a unique `RESOURCE_SUFFIX` (the test's timestamp-based suffix). This scopes:
- The resources file: `bugbash-resources-{suffix}.json` instead of `bugbash-resources.json`
- The S3 prefix: `bugbash-{suffix}/code.zip` instead of `bugbash/code.zip`
- The S3 cleanup: only deletes objects under its own prefix

## Related Issue

N/A — CI performance improvement.

## Documentation PR

N/A

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Other (please describe): CI performance — shard E2E tests for faster wall-clock time

## Testing

Tried different shard sizes, runs are here: https://github.com/aws/agentcore-cli/actions/runs/25165632571, https://github.com/aws/agentcore-cli/actions/runs/25166733599
- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.

### DIFF
--- a/.github/workflows/e2e-tests-full.yml
+++ b/.github/workflows/e2e-tests-full.yml
@@ -27,6 +27,7 @@ jobs:
       fail-fast: false
       matrix:
         cdk-source: [npm, main]
+        shard: ['1/6', '2/6', '3/6', '4/6', '5/6', '6/6']
     steps:
       - uses: actions/checkout@v6
         with:
@@ -70,7 +71,7 @@ jobs:
           CDK_REPO: ${{ secrets.CDK_REPO_NAME }}
       - name: Install CLI globally
         run: npm install -g "$(npm pack | tail -1)"
-      - name: Run E2E tests (${{ matrix.cdk-source }})
+      - name: Run E2E tests (${{ matrix.cdk-source }}, shard ${{ matrix.shard }})
         env:
           AWS_ACCOUNT_ID: ${{ steps.aws.outputs.account_id }}
           AWS_REGION: ${{ inputs.aws_region || 'us-east-1' }}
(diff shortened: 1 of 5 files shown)


tejaskash (dismissed):

Hweinstock:
Addressed comments and re ran https://github.com/aws/agentcore-cli/actions/runs/25175265881.

tejaskash (approved):

PR #1018, opened 2026-04-29
sync-preview: real merge main into preview

aidandaly24:
## Summary
- Real merge commit (two parents) to restore ancestry link between `main` and `preview`
- Resolves merge conflicts in CHANGELOG.md, package.json, package-lock.json, and browser-mode.ts (preview values kept)
- Previous sync PR #1017 was squash-merged, which broke the `merge-base` check in the release workflow

## Context
The "Release Both" workflow preflight checks `git merge-base HEAD origin/preview == MAIN_SHA`. A squash merge destroys the two-parent ancestry, so the check fails even though the code is present. This PR restores the link with a real merge commit.


notgitika (approved):
LGTM

PR #1044, opened 2026-04-29
fix: preserve caret range for @aws/agentcore-cdk dependency

aidandaly24:
## Summary
- Fix release workflow to write `^$VERSION` instead of exact `$VERSION` for @aws/agentcore-cdk
- Strip caret when comparing versions so it doesn't rewrite unnecessarily
- Restore `^` in asset file and snapshot (removed by #1042)

## Test plan
- [ ] Snapshot test passes locally (verified)
- [ ] Next release should produce `^x.y.z` in the vended CDK template

### DIFF
--- a/.github/workflows/release-main-and-preview.yml
+++ b/.github/workflows/release-main-and-preview.yml
@@ -109,23 +109,6 @@ jobs:
 
       - run: npm ci
 
-      - name: Sync @aws/agentcore-cdk to latest npm version
-        run: |
-          LATEST_CDK=$(npm view @aws/agentcore-cdk version 2>/dev/null || echo "")
-          if [ -n "$LATEST_CDK" ]; then
-            TEMPLATE_PKG="src/assets/cdk/package.json"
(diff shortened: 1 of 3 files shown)


.github/workflows/release-main-and-preview.yml
```diff
@@ -114,12 +114,12 @@ jobs:
           LATEST_CDK=$(npm view @aws/agentcore-cdk version 2>/dev/null || echo "")
           if [ -n "$LATEST_CDK" ]; then
             TEMPLATE_PKG="src/assets/cdk/package.json"
-            CURRENT_CDK=$(node -p "require('./$TEMPLATE_PKG').dependencies['@aws/agentcore-cdk']")
+            CURRENT_CDK=$(node -p "require('./$TEMPLATE_PKG').dependencies['@aws/agentcore-cdk'].replace(/^\^/, '')")
```
notgitika: we will have to remove the whole CDK flow here

notgitika (approved):

notgitika (approved):
LGTM

PR #1014, opened 2026-04-29
feat: add telemetry audit mode with FileSystemSink

Hweinstock:
## Description

Adds telemetry audit mode: when enabled, CLI commands write telemetry data to `~/.agentcore/telemetry/` as JSONL files for local inspection.

- **FilesystemSink** — metric-agnostic `MetricSink` that appends `{ value, attrs }` as JSONL. Each `record()` appends a line to the file. 
- **TelemetryClientAccessor** — static class managing the client lifecycle. `init()` creates eagerly, `get()` creates lazily if needed, `shutdown()` proxies to the client. Called from `main()` at startup/exit.
- **Audit activation** — `AGENTCORE_TELEMETRY_AUDIT=1` env var or `telemetry.audit: true` in config. The `agentcore telemetry audit` command is intentionally not introduced since we are not introducing this functionality to customers yet. 
- **Instrumented commands** — `help` and `help.modes` wrapped with `withCommandRun()` as proof of concept. Same/similar logic will follow for rest of commands. 

## Related Issue

Closes #

## Documentation PR

N/A — hidden flag, no user-facing docs yet.

## Type of Change

- [x] New feature

## Testing

In addition to tests added, also tested e2e with packaged tarball. 

- [x] I ran `npm run test:unit` and `npm run test:integ`
- [x] I ran `npm run typecheck`
- [x] I ran `npm run lint`
- [ ] If I modified `src/assets/`, I ran `npm run test:update-snapshots` and committed the updated snapshots

## Checklist

- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the
terms of your choice.

PR #1050, opened 2026-04-29
feat: wire telemetry into all add.* commands

Hweinstock:
## Description

Wire telemetry `withCommandRun()` into all `add.*` CLI command handlers and TUI add flows. CLI paths use a `cliCommandRun` helper that wraps the action handler with telemetry recording, error formatting, and `process.exit`. TUI paths use a `withAddTelemetry` helper that wraps each `.add()` call site in the TUI hooks/screens.

Each command emits a metric with command-specific attributes (e.g., `credential_type`, `strategy_count`, `evaluator_type`) on success, and `exit_reason: failure` with `error_name` on failure.

Supporting changes:
- `standardize()` helper for Zod enum narrowing (lowercases CLI values, uses `safeParse` to never throw)
- `TelemetryHelper` test utility with `assertMetricEmitted` for telemetry verification
- `runCLI` refactored to accept `{ env }` option for passing audit env vars in tests

## Related Issue

Closes #1014

## Type of Change

- [x] New feature

## Testing

- Telemetry assertions added to existing integ tests (`add-remove-resources.test.ts`, `help.test.ts`) verifying audit JSONL entries are written with correct command names, exit reasons, and attrs
- Manual e2e verification: `npm pack` → install tarball → run add commands with `AGENTCORE_TELEMETRY_AUDIT=1` → verified success and failure entries in audit files

## Checklist

---

terms of your choice.


Hweinstock:
1. standardize() type safety — Intentional tradeoff. The as cast only affects telemetry attrs, which are validated at runtime by COMMAND_SCHEMAS[command].parse(attrs) in recordCommandRun's try/catch. Invalid values silently drop the metric — correct behavior for telemetry.
Throwing here would fail the command after the operation already succeeded. Added a doc comment explaining this.

2. Dependency status — PR #1014 is merged. This branch is rebased onto main.

3. Test coverage — added some more basic coverage here. 

4. Unreachable code — No unreachable code exists. cliCommandRun is inside if (cliMode) { ... } else { TUI } — the else branch is reachable. Returning
exit codes instead of process.exit() would be a larger refactor; the existing primitives already used process.exit() before this PR.

5. TUI error handling — Intentionally different from CLI. CLI needs JSON/text error formatting + process.exit. TUI renders Ink components with their own
lifecycle (onExit, unmount). A shared helper wouldn't fit both patterns.

6. Gateway target type mapping — Addressed. Moved targetTypeMap into common-shapes.ts as GATEWAY_TARGET_TYPE_MAP next to the GatewayTargetType enum so
they stay in sync.

7. withAddTelemetry fallback — Already addressed in earlier commit. Changed from double-execution bug to returning a failure sentinel.

8. String coercion in test helper — Necessary because the JSONL audit sink serializes all values as strings. The coercion matches the actual data format.

Follow-up: Plan to update the telemetry schemas to allow 'unknown' as a valid value for enum attrs, so a single unrecognized attribute doesn't drop the
entire metric. However, this is OOS for this change.

tejaskash (commented):
Good PR overall — the  /  split is clean, the telemetry-never-blocks-CLI design is solid, and the test infrastructure with `TelemetryHelper` is well done. A few comments inline.

src/cli/telemetry/cli-command-run.ts
```diff
@@ -0,0 +1,68 @@
...
+      result = await fn();
+      if (!result.success) throw new Error(result.error);
+      return attrs;
+    });
+  } catch {
+    // withCommandRun re-throws after recording failure telemetry.
+    // result is set if fn() ran; if not, something unexpected happened.
+  }
+  return result ?? ({ success: false, error: 'Telemetry wrapper failed unexpectedly' } as AddResult<T>);
+}
```
tejaskash, line 71: If `fn()` throws an unexpected error (not a `result.success === false` path), the original error message gets discarded and replaced with the generic `"Telemetry wrapper failed unexpectedly"`. TUI hooks that call this will surface that string to the user, hiding the real cause.

Consider preserving the original error:

```ts
} catch (err) {
  if (!result) {
    return { success: false, error: getErrorMessage(err) } as AddResult<T>;
  }
}
return result;
```

This way `result` is still returned if `fn()` succeeded but `withCommandRun` re-threw, and the actual error propagates if `fn()` itself threw.

src/cli/primitives/GatewayTargetPrimitive.ts
```diff
@@ -412,7 +422,7 @@ export class GatewayTargetPrimitive extends BasePrimitive<AddGatewayTargetOption
             } else {
```
tejaskash, line 427: The same attrs object `{ target_type: telemetryTargetType, host: telemetryHost, outbound_auth: telemetryOutboundAuth }` is returned from 4 different code paths in this function. Extracting it once at the top would reduce repetition:

```ts
const telemetryAttrs = { target_type: telemetryTargetType, host: telemetryHost, outbound_auth: telemetryOutboundAuth };
```

Also note this line hardcodes `host: 'lambda'` instead of using `telemetryHost` like the other 3 paths. Is that intentional (lambdaFunctionArn targets always run on Lambda regardless of the `--host` flag), or should it use `telemetryHost` for consistency?

src/cli/primitives/AgentPrimitive.tsx
```diff
@@ -264,92 +277,106 @@ export class AgentPrimitive extends BasePrimitive<AddAgentOptions, RemovableReso
...
+                  process.exit(0);
+                },
+              })
+            );
+          } catch (error) {
+            console.error(getErrorMessage(error));
+            process.exit(1);
+          }
         }
       });
```
tejaskash, line 382: Question: The TUI fallback for agent doesn't use `withAddTelemetry` (unlike the other TUI hooks like `useCreateMemory`, `useCreateEvaluator`, etc.). Is agent TUI telemetry planned for a follow-up, or is it intentionally omitted because the `AddFlow` component has its own telemetry wiring internally?

src/cli/telemetry/cli-command-run.ts
```diff
@@ -0,0 +1,68 @@
...
+    // withCommandRun records success/failure telemetry, then re-throws on failure
+    await client.withCommandRun(command, fn);
+    process.exit(0);
+  } catch (error) {
+    if (json) {
+      console.log(JSON.stringify({ success: false, error: getErrorMessage(error) }));
+    } else {
+      console.error(getErrorMessage(error));
+    }
+    process.exit(1);
```
tejaskash, line 36: Question: The old code would return the full result object from `.add()` on failure (which might include extra fields beyond `success` and `error`). The new code constructs a fresh `{ success: false, error: getErrorMessage(error) }`. Are there any downstream consumers of the `--json` failure output that depend on additional fields?

src/cli/telemetry/cli-command-run.ts
Hweinstock: Good catch, we should preserve the original error here.

src/cli/primitives/GatewayTargetPrimitive.ts
Hweinstock: Agree we can hoist this up a level. I assume if `lambdaFunctionArn` is passed the host must be a lambda? However, the outcome should be the same and we can avoid making an assumption by swapping back to telemetryHost.

src/cli/primitives/AgentPrimitive.tsx
Hweinstock: I think this is a miss. Good catch!

src/cli/telemetry/cli-command-run.ts
Hweinstock: I don't think its possible to include fields beyond success and error based on the type: https://github.com/aws/agentcore-cli/blob/0214f863af65fbaaa09d4da89d7c1620b5927b58/src/cli/primitives/types.ts#L9-L11.

tejaskash (approved):
LGTM

PR #1056, opened 2026-04-30
fix(harness): add error handling for invoke_harness API call

jesseturner21:
## Summary
- Wraps the `invoke_harness_streaming` call in a try/except block so boto3 errors (bad credentials, network issues, invalid ARN) produce a clean error message instead of a raw Python traceback in GitHub Actions logs.

## Test plan
- [x] Verified Python syntax is valid
- [ ] Trigger harness review workflow with an invalid ARN to confirm clean error output


tejaskash (approved):
LGTM

PR #1062, opened 2026-04-30
fix: default invalid telemetry attributes to 'unknown' instead of dropping metric

Hweinstock:
## Description

When a single telemetry attribute fails validation (invalid enum, missing field, wrong type), the entire metric is silently dropped. This change makes attribute parsing resilient: invalid fields default to `'unknown'` while the metric is still published.

Adds `resilientParse()` to `common-shapes.ts` which validates each schema field individually via `safeParse()`. Updates `recordCommandRun()` in `client.ts` to use it instead of strict `.parse()`.

## Related Issue

Closes #

## Documentation PR

N/A

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.

### DIFF
--- a/src/cli/telemetry/schemas/common-shapes.ts
+++ b/src/cli/telemetry/schemas/common-shapes.ts
@@ -8,6 +8,24 @@ export function safeSchema<T extends Record<string, SafeField>>(shape: T) {
   return z.object(shape);
 }
 
+/**
+ * Validate each field in a schema individually, defaulting to 'unknown' on failure.
+ * This ensures a single invalid attribute never blocks the entire metric from being published.
+ * Keys in attrs not present in the schema are omitted from the result.
+ */
+export function resilientParse(
+  schema: z.ZodObject<z.ZodRawShape>,
+  attrs: Record<string, unknown>
+): Record<string, unknown> {
+  const result: Record<string, unknown> = {};
+  for (const key of Object.keys(schema.shape)) {
+    const field = schema.shape[key] as z.ZodType;
+    const parsed = field.safeParse(attrs[key]);
+    result[key] = parsed.success ? parsed.data : 'unknown';
+  }
+  return result;
+}
+
 // Primitive types
 export const Count = z.number().int().nonnegative();
 
--- a/src/cli/telemetry/__tests__/client.test.ts
+++ b/src/cli/telemetry/__tests__/client.test.ts
@@ -116,18 +116,49 @@ describe('TelemetryClient', () => {
       expect(sink.metrics[0]!.attrs.check_only).toBe('true');
     });
 
-    it('silently drops invalid success payloads', async () => {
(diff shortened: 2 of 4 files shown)


avi-alpert (approved):

Hweinstock:
Response to agent: 
1. not relevant. 
2. the alternative is that downstream never recieves the metric. Now we can adjust downstream to accept unknown here. 
3. this is acceptable as long as otel finds it acceptable. The OTEL TS Client should handle this for us by specifying the type as "string" on the wire format, making this ok as long as the backend accepts it. The alternative is that they are completely dropped before being sent. 

I think its missing context on the fact that this "shouldn't" happen. The resiliency with `unknown` is that it can alert us to bugs in our implementation that allow unknown values to pass through.

src/cli/telemetry/schemas/common-shapes.ts
```diff
@@ -8,6 +8,24 @@ export function safeSchema<T extends Record<string, SafeField>>(shape: T) {
...
+ * Validate each field in a schema individually, defaulting to 'unknown' on failure.
+ * This ensures a single invalid attribute never blocks the entire metric from being published.
+ * Keys in attrs not present in the schema are omitted from the result.
+ */
+export function resilientParse(
+  schema: z.ZodObject<z.ZodRawShape>,
+  attrs: Record<string, unknown>
+): Record<string, unknown> {
+  const result: Record<string, unknown> = {};
+  for (const key of Object.keys(schema.shape)) {
```
tejaskash, line 21: `'unknown'` works well for enum fields, but some schemas have `z.boolean()` and `z.number()` fields (e.g. `has_agent`, `runtime_count`). If those are missing or invalid, the metric would carry the string `"unknown"` where downstream consumers expect `"true"`/`"false"` or a number.

If that's intentional (OTel attrs are all strings anyway so `"unknown"` is a valid sentinel for any field type), a short comment here would save a future reader from wondering. Something like:

```ts
// 'unknown' is acceptable for all field types — OTel attrs are serialized as strings,
// and downstream dashboards filter on this sentinel regardless of the original type.
result[key] = parsed.success ? parsed.data : 'unknown';
```

Otherwise, a type-aware default (`false` for booleans, `0` for numbers) would keep the data clean.

tejaskash (commented):
Looks good — one nit inline.

src/cli/telemetry/schemas/common-shapes.ts
Hweinstock: Whoops sorry, missed this before merging. The strands agent called out the same thing, and my understanding is that the OTEL client will handle this for us as long as we accept it in the backend. Agree a comment could be helpful here because its confusing that we're using a different type.

Issue #1066, opened 2026-04-30
E2E test shard 2/6 failing on main


Hweinstock:
## Summary

E2E test shard 2/6 is failing on `main`.

## Details

- **Workflow:** [E2E Tests (Full Suite) #266](https://github.com/aws/agentcore-cli/actions/runs/25186056841/job/73844827394)
- **Failed job:** `e2e (main, 2/6)` — exited with code 1 after 4m 38s
- **Trigger:** merge commit `e9dfc16` from PR #1050
- **Failed step:** "Run E2E tests (main, shard 2/6)"

PR #1067, opened 2026-04-30
fix: resolve e2e import test concurrency races

Hweinstock:
## Description

Two issues in running the e2e tests in parallel: 

### naming conflict
```
botocore.errorfactory.ConflictException: An error occurred (ConflictException) when calling the CreateAgentRuntime operation: An agent with the specified name already exists. Use a different name and try again.
```

Resource names are using time (in seconds) to suffix, but this creates an issue when they create at the same time. Switch to UUID or millisecond precision in resource name suffix. 

### list then get race condition
```
[error] Online evaluation configuration not found
```

The code performs a check that the given evaluator isn't already in an eval config, because if it is, the import will fail. 
When iterating through all online eval configs in the account, we call list then get, but its possible an eval config is deleted between the list and get. This will cause the promise to reject with a ResourceNotFoundError. Therefore we can ignore these errors. 

## Related Issue

Closes #https://github.com/aws/agentcore-cli/issues/1066

## Documentation PR

N/A

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.

### DIFF
--- a/e2e-tests/fixtures/import/common.py
+++ b/e2e-tests/fixtures/import/common.py
@@ -2,6 +2,7 @@
 import json
 import os
 import time
+import uuid
 import zipfile
 import tempfile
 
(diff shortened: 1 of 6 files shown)


Hweinstock:
removed unused time import. Good call review agent!

tejaskash (approved):
LGTM

jesseturner21 (approved):
lgtm

PR #1068, opened 2026-04-30
feat: evo preview features — config bundles, batch evaluation, recommendations, AB testing

notgitika:
## Summary

Adds preview support for the Evo feature set: config bundles, batch evaluation, recommendations, and AB testing.

### Config Bundles [preview]
- `add config-bundle` — add versioned runtime configuration bundles
- `cb versions` — list version history for a bundle
- `cb diff` — diff two versions of a bundle
- `cb create-branch` — create a new branch on an existing bundle
- `--with-config-bundle` flag on agent creation auto-wires config bundle support
- Config bundle baggage passed on invoke for runtime config injection

### Batch Evaluation [preview]
- `run batch-evaluation` — run evaluators across all agent sessions in CloudWatch
- `stop batch-evaluation` — stop a running batch evaluation
- Ground truth support (assertions, expected trajectory, turns)
- Name validation against API pattern `[a-zA-Z][a-zA-Z0-9_]{0,47}`

### Recommendations [preview]
- `run recommendation` — optimize system prompts or tool descriptions using agent traces
- Supports inline, file, and config bundle input sources
- Config bundle integration: reads current prompt, writes optimized version back
- JSONPath resolution from `--runtime` flag for multi-component bundles

### AB Testing [preview]
- Target-based AB test routing
- AB test detail screen with p-value significance display

### Other
- TUI routing fixes for `agentcore add config-bundle` and `agentcore add ab-test`
- Documentation for all preview features (`docs/config-bundles.md`, `docs/batch-evaluation.md`, `docs/recommendations.md`)
- README updated with preview commands and doc links

## Companion PR
- CDK constructs: aws/agentcore-l3-cdk-constructs (separate PR)

## Test plan
- [x] Unit tests passing
- [x] Manual CLI testing: batch eval (valid name, hyphens rejected, fake evaluator error, ground truth, multiple evaluators, lookback days, stop)
- [x] Manual CLI testing: recommendations (inline, file, config bundle, tool descriptions, nonexistent agent)
- [x] Manual CLI testing: config bundles (versions, diff, create-branch, add)
- [x] Manual CLI testing: status shows config bundles
- [x] Validate command passes on current schema


jariy17 (changes requested):
Reviewed for regressions against the current private repo baseline (v0.12.2). Found 5 issues that need addressing before merge.

---

## 1. Breaking flag rename: `--days` → `--lookback` on `run eval`

**File:** `src/cli/commands/run/command.tsx`

The `--days` flag was renamed to `--lookback` with no alias or deprecation shim. Any user script or CI job using `agentcore run eval --days 7` will silently get no results (or an error) without explanation. Either keep `--days` as a deprecated alias, or add a clear error message when the old flag is detected.

---

## 2. CloudWatch traces removed from `agentcore dev` browser mode

**File:** `src/cli/commands/dev/browser-mode.ts`

Both `onListCloudWatchTraces` and `onGetCloudWatchTrace` handlers were deleted entirely. These wired the browser dev UI's Traces tab to live CloudWatch data. After this PR, `agentcore dev` no longer serves CloudWatch trace data to the web UI at all. The companion browser test was simultaneously simplified to stop asserting span tree, log panel, and CloudWatch data — which masks this regression in CI. If this is intentional (e.g., moved to a different mechanism), please document it. If not, the handlers need to be restored.

---

## 4. E2E test race condition — `RESOURCE_SUFFIX` removed

**File:** `e2e-tests/import-resources.test.ts`

The per-run `RESOURCE_SUFFIX` that namespaced the fixture file was removed. All runs now read/write a shared `bugbash-resources.json`. Two concurrent E2E runs (e.g., two open PRs) will stomp on each other's fixture data and cause flaky failures. The suffix should be kept, or the fixture path should use a run-specific directory.

---

## 5. `PRIVATE_DEV_DISTRO` package rename has no migration path

**File:** `src/cli/constants.ts`

`PRIVATE_DEV_DISTRO.packageName` changed from `@aws/agentcore` → `@aws/agentcore-dev`. Users currently installed from `npm.pkg.github.com` as `@aws/agentcore` will not get update notifications and will need to manually reinstall under the new name. `checkForUpdates: false` means they won't even get a hint. If this rename is intentional, a one-time migration notice or install script should accompany it.

jariy17 (changes requested):
Updated review after re-pulling the branch (HEAD c292bd1bf1ff). Regressions #3 (version downgrade) and #5 (PRIVATE_DEV_DISTRO rename) are resolved — thank you. Three issues remain, all of which contradict the current state of public `main`:

---

## 1. Breaking flag rename: `--days` → `--lookback` on `run eval` (still present)

**File:** `src/cli/commands/run/command.tsx`

Public `main` still has `--days`. This PR renames it to `--lookback` with no alias or deprecation path. Any user or CI script using `agentcore run eval --days 7` will silently get the default value instead (commander ignores unknown options). Either keep `--days` as a deprecated alias pointing to the same internal field, or add an explicit error when the old flag is detected.

---

## 2. CloudWatch traces deleted from `agentcore dev` (still present)

**File:** `src/cli/commands/dev/browser-mode.ts`

Public `main` has both `onListCloudWatchTraces` and `onGetCloudWatchTrace` handlers. This PR deletes them entirely and simultaneously weakens the browser test to stop asserting CloudWatch-sourced span/log data, masking the removal in CI. After merging, `agentcore dev` will serve no CloudWatch trace data in the web UI. If this is intentional (replaced by a new mechanism), please document what replaces it. Otherwise these handlers need to be restored.

---

## 4. E2E concurrency race reintroduced (still present)

**File:** `e2e-tests/import-resources.test.ts`

Public `main` commit bd6f841 (`fix: resolve e2e import test concurrency races #1067`) specifically restored `RESOURCE_SUFFIX` to namespace the fixture file per run. This PR's merge conflict resolution removed it again — the resource path is hardcoded back to `bugbash-resources.json` and `RESOURCE_SUFFIX` is not passed to either setup or cleanup scripts. Concurrent E2E runs will stomp on each other's fixture data. The fix from #1067 needs to be carried forward into this branch.

jariy17 (approved):
Re-reviewed at HEAD 6e085f429767. All previously flagged regressions are resolved:

- ✅ `--days` flag restored on `run eval`
- ✅ `onListCloudWatchTraces` / `onGetCloudWatchTrace` restored in `agentcore dev`
- ✅ `RESOURCE_SUFFIX` isolation restored in e2e import tests
- ✅ Version and `agent-inspector` dep back to 0.12.2 / 0.3.0
- ✅ `PRIVATE_DEV_DISTRO` config reverted

No regressions against the private repo baseline. The 4 issues flagged by `agentcore-cli-automation` (hardcoded `amazonaws.com` in recommendation/config-bundle wrappers, stale JSON schema, silent `agentcore.json` mutation on deploy, config bundle/AB test teardown leak) are separate functional issues worth addressing but not regressions from this PR's changes.

jariy17 (dismissed):
Updating review — the 4 issues flagged by agentcore-cli-automation are blocking and need to be addressed before merge.

---

## 1. Hardcoded `amazonaws.com` breaks non-commercial partitions

**Files:**
- `src/cli/aws/agentcore-recommendation.ts:228`
- `src/cli/aws/agentcore-config-bundles.ts:181`

Both hardcode `https://bedrock-agentcore..amazonaws.com` / `https://bedrock-agentcore-control..amazonaws.com`. The sibling wrappers in this same PR (`agentcore-ab-tests.ts`, `agentcore-batch-evaluation.ts`, `agentcore-http-gateways.ts`) correctly use `dnsSuffix(region)` from `./partition`. Recommendations and config bundles will silently fail in GovCloud and China partitions.

Fix: import `dnsSuffix` from `./partition` and replace the hardcoded literal in both files.

---

## 2. `schemas/agentcore.schema.v1.json` is stale

The Zod schemas now include `configBundles`, `abTests`, and `httpGateways` as top-level fields on `AgentCoreProjectSpecSchema`, but the committed JSON schema was not regenerated. Users whose editors validate `agentcore.json` against the published schema (VS Code, etc.) will see false "property not allowed" errors on every new preview field.

Fix: run `npm run build:lib && npm run build:schema` and commit the regenerated `schemas/agentcore.schema.v1.json`.

---

## 3. `validateProject()` silently rewrites `agentcore.json` on every deploy

**File:** `src/cli/operations/deploy/preflight.ts`

The deploy preflight injects `type: "ConfigurationBundle"` into config bundle entries and writes the file back with `JSON.stringify(rawJson, null, 2)`. This runs on every `agentcore deploy`, producing surprise git diffs for users and clobbering their file's original formatting (tabs, trailing newlines, key order). The Zod `ConfigBundleSchema` already applies this default in-memory, so the write-back is unnecessary.

Fix options:
1. Fix the CDK side to consume the Zod-parsed spec with defaults applied, and drop this patching code.
2. Inject `type` only into the in-memory object the CDK reads at synth time, without persisting to disk.
3. If persisting is truly required: only write when `patched === true` (already done), preserve trailing newline, and print a warning to the user that their file was modified.

---

## 4. Config bundles and AB tests are leaked on stack teardown

`performStackTeardown` explicitly calls `deleteHttpGatewayWithTargets` for HTTP gateways before destroying the CFN stack, but there is no equivalent cleanup for config bundles or AB tests. When a user runs `agentcore remove all` + deploy, the CFN stack is destroyed but any config bundles and AB tests in deployed-state remain orphaned in AWS — silently accumulating charges with no CLI surface to clean them up.

Fix: extend `performStackTeardown` to iterate `deployedState.targets[target].resources.configBundles` and `.abTests` and delete them, mirroring what it already does for `httpGateways`.

notgitika:
Follow up PR for addressing:

1. validateProject() silently rewrites agentcore.json on every deploy


2. Config bundles and AB tests are leaked on stack teardown

padmak30 (approved):

PR #1071, opened 2026-04-30
fix: stop rewriting agentcore.json on deploy and clean up evo resources on teardown

### LINKED ISSUE #1070: fix: deploy preflight silently rewrites agentcore.json and config bundles/AB tests leak on teardown
## Follow-up from #1068

Two issues identified during review that need addressing:

### 1. `validateProject()` silently rewrites agentcore.json on every deploy

**File:** `src/cli/operations/deploy/preflight.ts`

The deploy preflight injects `type: "ConfigurationBundle"` into config bundle entries and writes the file back with `JSON.stringify(rawJson, null, 2)`. This runs on every `agentcore deploy`, producing surprise git diffs for users and clobbering their file's original formatting (tabs, trailing newlines, key order). The Zod `ConfigBundleSchema` already applies this default in-memory, so the write-back is unnecessary.

**Options:**
- Fix the CDK side to consume the Zod-parsed spec with defaults applied, and drop this patching code
- Inject `type` only into the in-memory object the CDK reads at synth time, without persisting to disk
- If persisting is truly required: only write when `patched === true` (already done), preserve trailing newline, and print a warning to the user that their file was modified

### 2. Config bundles and AB tests are leaked on stack teardown

`performStackTeardown` explicitly calls `deleteHttpGatewayWithTargets` for HTTP gateways before destroying the CFN stack, but there is no equivalent cleanup for config bundles or AB tests. When a user runs `agentcore remove all` + `deploy`, the CFN stack is destroyed but any config bundles and AB tests in deployed-state remain orphaned in AWS — silently accumulating charges with no CLI surface to clean them up.

**Fix:** Extend `performStackTeardown` to iterate `deployedState.targets[target].resources.configBundles` and `.abTests` and delete them, mirroring what it already does for `httpGateways`.

notgitika:
## Summary

Two fixes for issues identified during #1068 review:

### 1. Deploy preflight no longer rewrites agentcore.json
`validateProject()` was injecting `type: "ConfigurationBundle"` into config bundle entries and writing the file back to disk on every `agentcore deploy`. This caused surprise git diffs and clobbered file formatting. The type discriminator is now applied in-memory only — the CDK reads the patched object at synth time without touching the user's file.

### 2. Config bundles and AB tests cleaned up on stack teardown
`performStackTeardown` already deleted HTTP gateways before destroying the CFN stack, but config bundles and AB tests were left orphaned in AWS. Now iterates `deployedState.targets[target].resources.configBundles` and `.abTests` and deletes them, matching the existing HTTP gateway pattern.

Closes #1070

## Test plan
- [x] `npm run typecheck` passes
- [x] Pre-commit hooks pass (eslint, prettier, secretlint)
- [ ] Manual: `agentcore deploy` with config bundles does not modify agentcore.json on disk
- [ ] Manual: `agentcore remove all` + `agentcore deploy` cleans up config bundles and AB tests


notgitika:
Recreating with clean history — branched from wrong base.

PR #1078, opened 2026-05-01
fix: sync-preview pushes directly on clean merge, PRs only on conflict

notgitika:
## Summary

Reworks the sync-preview workflow to push directly to `preview` when the merge is clean (using the `agentcore-devx-automation` GitHub App to bypass branch protection), and only creates a PR when there are real conflicts.

### How it works

1. **Clean merge** → merges main, restores preview's package version, pushes directly to `preview`. No PR.
2. **Only version conflicts** (package.json/package-lock.json) → accepts main's content, restores preview's version field, pushes directly. No PR.
3. **Real conflicts in other files** → opens a PR with conflict markers for manual resolution.

### Key changes

- Uses `actions/create-github-app-token` with the `agentcore-devx-automation` app to get a token that bypasses branch protection
- Instead of ignoring package.json/lock entirely (which discards new deps, scripts, etc. from main), accepts main's full content and surgically restores only the `version` field
- CHANGELOG.md and schema merge normally from main — no special handling
- Dedup check prevents multiple conflict PRs

## Test plan
- [ ] Workflow YAML is valid (verified locally)
- [ ] Push to main triggers sync-preview workflow
- [ ] Clean merge pushes directly to preview with version restored
- [ ] Version-only conflicts auto-resolve and push directly
- [ ] Real conflicts in other files produce a PR with conflict markers
- [ ] Existing open PR prevents duplicate creation
- [ ] App token has push access to preview branch

### DIFF
--- a/.github/workflows/sync-preview.yml
+++ b/.github/workflows/sync-preview.yml
@@ -17,11 +17,11 @@ jobs:
     name: Merge main into preview
     runs-on: ubuntu-latest
     steps:
-      - name: Generate App token
+      - name: Generate GitHub App Token
         id: app-token
         uses: actions/create-github-app-token@v1
         with:
-          app-slug: agentcore-devx-automation
+          app-id: ${{ vars.APP_ID }}
           private-key: ${{ secrets.APP_PRIVATE_KEY }}
 
       - name: Checkout preview
--- a/.github/workflows/sync-preview.yml
+++ b/.github/workflows/sync-preview.yml
@@ -65,8 +65,9 @@ jobs:
           if git merge origin/main --no-edit -m "chore: merge main into preview"; then
             echo "status=clean" >> $GITHUB_OUTPUT
           else
-            # If package.json or package-lock.json conflict, accept main's content
-            # (we'll fix the version back afterwards)
+            # preview carries a higher version string than main (e.g. 1.0.0-preview.X vs 0.13.X).
+            # This means package.json/package-lock.json almost always conflict on the version field.
+            # Accept main's content here; the version is restored in the next step.
             for f in package.json package-lock.json; do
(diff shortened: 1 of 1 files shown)


notgitika:
1/ is a real issue. unfortunately we need a PAT token or github app token to resolve that
2/ acceptable trade-off imo
3/ `Automatically delete head branches` is enabled in settings so this is not an issue

Hweinstock (dismissed):

avi-alpert (dismissed):

.github/workflows/sync-preview.yml
```diff
@@ -17,93 +17,140 @@ jobs:
...
 
           if git merge origin/main --no-edit -m "chore: merge main into preview"; then
-            git push origin preview
-            echo "✅ main merged into preview and pushed"
-            echo "status=merged" >> $GITHUB_OUTPUT
+            echo "status=clean" >> $GITHUB_OUTPUT
           else
-            git merge --abort
-            echo "status=conflict" >> $GITHUB_OUTPUT
+            # If package.json or package-lock.json conflict, accept main's content
```
Hweinstock: sorry if I forgot the answer here, but why we do need to special case package.json/package-lock.json? Does preview have new dependencies we want to manage separately?

Hweinstock (dismissed):

.github/workflows/sync-preview.yml
notgitika: the reason is that the package version itself is different in main and preview
example:
main: `0.13.1`
preview: `1.0.0-preview.5`

Everything else should ideally be identical

notgitika: added inline comment

Hweinstock (dismissed):
LGTM once CI passes.

Hweinstock (approved):

PR #1087, opened 2026-05-01
test: collapse schema enumeration tests and remove duplicates

tejaskash:
## Summary
- Collapse `it.each` per-value enumeration patterns into consolidated tests across 9 schema test files (constants, agent-env, agentcore-project, aws-targets, deployed-state, mcp-defs, mcp, evaluator, memory)
- Remove deploy/dev `--help` unit tests already covered by `integ-tests/help.test.ts`
- Remove 5 `it.todo` placeholders from credential-ops.test.ts

Schema tests: ~506 → ~415. All cross-field validation (`superRefine`), boundary, and discriminated union tests preserved.

## Test plan
- [ ] `npm test` passes — schema subset verified locally (415 tests), full suite should pass in CI
- [ ] No coverage regression for `src/schema/` (tests collapsed, not deleted — same code paths covered)

### DIFF
--- a/src/cli/commands/deploy/__tests__/deploy.test.ts
+++ b/src/cli/commands/deploy/__tests__/deploy.test.ts
@@ -5,23 +5,6 @@ import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { afterAll, beforeAll, describe, expect, it } from 'vitest';
 
-describe('deploy --help', () => {
-  it('shows verbose option', async () => {
-    const result = await runCLI(['deploy', '--help'], process.cwd());
-    expect(result.exitCode).toBe(0);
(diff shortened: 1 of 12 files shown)


tejaskash:
Good catch — restored the deploy/dev help flag assertions and the `--invoke` negative regression guard. The integ smoke test is indeed only exit-code + "Usage:".

tejaskash:
Addressed: replaced the representative-sample tests with `.options` assertions for both `AgentCoreRegionSchema` and `GatewayTargetTypeSchema`. This catches accidental removal of any enum value without per-value `it.each` tests.

jesseturner21 (approved):
LGTM

PR #1105, opened 2026-05-04
fix: prefix HTTP gateway names with project name to prevent cross-project collisions

jariy17:
## Root Cause

HTTP gateways were being deployed without a project-scoped prefix, causing cross-project name collisions when multiple projects shared the same AWS account or region. Gateway names were globally unique identifiers but were derived only from the gateway's local config, not the project context.

## What Changed

6 files updated to prefix gateway names with the project name:

- Gateway creation logic now prepends `<project-name>-` to all HTTP gateway names
- Gateway lookup/describe paths updated to match the new naming scheme
- Delete/update flows updated to resolve the prefixed name correctly
- Validation updated to account for the prefix in name length checks

## Customer Impact

**Breaking change**: Gateway names are now limited to 24 characters (to accommodate the project name prefix within AWS name length limits). Customers with existing gateways deployed under the old naming scheme will need to redeploy — old gateways will not be automatically renamed.

## Test Results

- Unit tests: 13/13 passing
- Integration tests: 184/184 passing


jariy17:
Addressed in commit `5c178e6`.

The root cause was that `ABTestPrimitive.resolveVariants()` was constructing target names without the project prefix, while `post-deploy-http-gateways.ts` creates targets on AWS with name `${projectName}-${tgt.name}`. The fix threads `projectName` as a required parameter down through `ABTestPrimitive` → `resolveVariants()`, so the variant config now resolves to `${projectName}-${runtime}-${endpoint}` — matching what was actually created on AWS.

You raised a valid design point about whether target names need prefixing at all since they're scoped within a gateway. The current approach (Option 2 from your list) was chosen to keep naming consistent: gateway names are prefixed, so target names follow the same `${projectName}-` convention. If there's a strong preference to revert target-name prefixing and rely solely on gateway scoping, happy to revisit — but wanted to call out the rationale explicitly.

jariy17:
Addressed in commit `92b1fe5`.

The lookup mismatch was exactly as you described — `existingTargetsByName` was keyed by the AWS-returned name (prefixed) but the lookup used the unprefixed spec name, so it never matched. Fixed at line 104 of `post-deploy-http-gateways.ts` to use `existingTargetsByName.get(\`${projectName}-${tgt.name}\`)`, so the qualifier-change detection and idempotency logic now work correctly on re-deploy.

You're right that there was no test covering the `existingGateway` + `gwSpec.targets` re-deploy path — that's a gap worth noting for follow-up test coverage.

jariy17:
Addressed in commits `232a284` (implementation) and `1807d98` (unit test).

The fix adds a two-step fallback in the state-loss recovery path:

1. Look up the prefixed name (`${projectName}-${gwSpec.name}`) first — the normal post-PR case.
2. If not found, fall back to the unprefixed legacy name. If a match is found there, a `console.warn()` migration notice is emitted so the user gets a clear signal that they're running against a pre-PR gateway and should redeploy to get the correctly-named one.

This avoids silently creating a duplicate gateway. The orphaned old gateway concern is still valid for users who *do* have state — they'll need a manual `agentcore teardown` + redeploy to clean up — but the state-loss case (your specific scenario) is now handled gracefully. A unit test covering both the prefixed-found and legacy-fallback branches was added in `1807d98`.

jariy17:
This comment is now stale — the `NODE_OPTIONS=--experimental-require-module` changes to `package.json` were removed from the PR in commit `8452ac4` (`chore: remove NODE_OPTIONS test script changes from gateway fix`). Those changes are no longer part of this PR.

jariy17:
## Verification summary

Smoke-tested the branch end-to-end before requesting review:

**Build**
- `npm run bundle` completed successfully → `aws-agentcore-0.13.0-20260504181229.tgz` (1.7 MB, 335 files)
- Installed globally; `agentcore --version` → `0.13.0-20260504181229`

**CLI commands**
- `agentcore add ab-test --help` — all flags render correctly (Config-Bundle and Target-Based mode flags present)
- `agentcore add ab-test` (non-interactive) → `{"success":true,"abTestName":"mytest"}` — `agentcore.json` updated with full AB test entry (80/20 C/T weights, gateway ref)
- `agentcore remove ab-test --name mytest --yes --json` → `{"success":true,...}` — `agentcore.json` confirmed clean afterward

**TUI flow**
- 13-step Ink wizard traced end-to-end from source: mode → name → description → gateway → variant config → eval config → weights → enable → confirm

**PR fix verification (bundle inspection)**
- `${projectName}-${gatewayName}` prefix present in the installed bundle's gateway creation paths
- `${projectName}-${targetName}` prefix present in the AB test target resolution path
- Both the gateway prefix fix and the AB test prefix fix are active in the built artifact

**Unit tests**
- New unit tests for the pre-migration gateway name fallback (commit `1807d98`) pass

src/schema/schemas/primitives/http-gateway.ts
```diff
@@ -7,10 +7,13 @@ import { z } from 'zod';
...
   .string()
   .min(1, 'Name is required')
-  .max(48)
+  .max(
+    24,
+    'Gateway name must be 1–24 alphanumeric/hyphen characters (combined with project name must fit 48-char AWS limit)'
+  )
   .regex(
-    /^[a-zA-Z][a-zA-Z0-9-]{0,47}$/,
-    'Must begin with a letter and contain only alphanumeric characters and hyphens (max 48 chars)'
```
notgitika, line 13: project names vary from 1 to ~48 chars right? A 1-char project name wastes 22 chars of headroom with this validation. 

The preflight validation in validateHttpGatewayNames() already does the correct combined check against 48 chars. I dont think this is needed?

src/cli/operations/deploy/post-deploy-ab-tests.ts
```diff
@@ -409,7 +409,8 @@ async function findABTestByName(
 /**
  * Resolve variant config bundle references.
  * If bundleArn is a name (not an ARN), look it up in deployed config bundles.
- * Target-based variants are passed through as-is.
+ * Target-based variants have their target name prefixed with projectName to match
+ * what post-deploy-http-gateways.ts creates on AWS (e.g. `${projectName}-${tgt.name}`).
```
notgitika, line 413: target names are prefixed without a combined-length preflight check for targets. If a target name + project prefix exceeds the AWS limit, the user gets a raw API error. can we add validation here as well?

src/cli/operations/deploy/post-deploy-ab-tests.ts
```diff
@@ -142,7 +142,7 @@ export async function setupABTests(options: SetupABTestsOptions): Promise<SetupA
       const existingTest = existingABTests?.[testSpec.name];
 
       // Resolve ARN references from deployed state
-      const resolvedVariants = resolveVariants(testSpec.variants, deployedResources);
+      const resolvedVariants = resolveVariants(testSpec.variants, projectSpec.name, deployedResources);
```
notgitika, line 145: if someone has an existing AB test pointing at a legacy (unprefixed) gateway target, the resolved name won't match the actual target name on AWS, right?

notgitika (changes requested):

notgitika (approved):
LGTM

PR #1114, opened 2026-05-04
chore: update namespace design for data plane

nborges-aws:
## Description

Adds support for the new `namespacePath` data-plane field introduced by the memory namespace redesign. `namespacePath` is a hierarchical path-prefix filter that sits alongside the existing namespace field on ListMemoryRecords and RetrieveMemoryRecords. The existing namespace field will migrate to function as exact-match.

Replaced required namespace with unions for the option types. Forcing callers to supply exactly one of namespace / namespacePath at compile-time. Defensive runtime validation is kept for paths that bypass the type system (JSON body from web UI requests).

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Other (please describe): New field added to DP methods, existing method logic changed

## Testing

- [x] I ran `npm run lint`

## Checklist

- [ ] Any dependent changes have been merged and published -- SDK DP PR must be merged first

---

terms of your choice.


nborges-aws:
> **Blocker: generated Python agent code will fail at runtime — `bedrock-agentcore` Python SDK does not accept `namespace_path=`.**
> 
> In `src/cli/operations/agent/import/langgraph-translator.ts` and `src/cli/operations/agent/import/strands-translator.ts`, the emitted code calls:
> 
> ```python
> memory_client.retrieve_memories(memory_id=memory_id, namespace_path=f'/users/{user_id}/facts', ...)
> ```
> 
> But the `memory_client.retrieve_memories` method in the high‑level Python `bedrock-agentcore` package (which is what's pinned via `bedrock-agentcore[memory] >= 1.0.3` in `pyproject-generator.ts`) only accepts a `namespace` keyword — I checked every published release up to and including the current latest (1.8.0) and none define a `namespace_path` parameter:
> 
> ```python
> # bedrock_agentcore/memory/client.py @ 1.8.0
> def retrieve_memories(
>     self, memory_id: str, namespace: str, query: str, actor_id: Optional[str] = None, top_k: int = 3
> ):
> ```
> 
> So every agent generated with `memoryOption: 'longAndShortTerm'` will raise `TypeError: retrieve_memories() got an unexpected keyword argument 'namespace_path'` on the first retrieval call. The two new translator tests pass because they only assert the string contents of the generated file — they never exercise the Python.
> 
> This is distinct from the raw data‑plane SDK (`@aws-sdk/client-bedrock-agentcore`), which does take `namespacePath` at the wire level — that part of the PR looks fine. The issue is specifically the high‑level Python helper the generated agents import.
> 
> Options:
> 
> 1. Hold this PR's translator changes until the Python `bedrock-agentcore` SDK ships `namespace_path` support on `retrieve_memories`, then bump the minimum pinned version in `pyproject-generator.ts` (`MEMORY_DEPS` / `BASE_DEPS`) to that release.
> 2. Split the PR: land the data‑plane (`listMemoryRecords` / `retrieveMemoryRecords` / web‑ui handler) changes now, and revert the translator changes in `langgraph-translator.ts` + `strands-translator.ts` until the Python SDK supports it.
> 3. If the intent is for generated agents to call the low-level boto3 `bedrock-agentcore` client directly (which does support `namespacePath`), rewrite the emitted retrieval code to use boto3 instead of `memory_client.retrieve_memories`.

This is expected. The following PR adds the missing field in the SDK: https://github.com/aws/bedrock-agentcore-sdk-python/pull/449

jesseturner21 (approved):
LGTM

PR #1115, opened 2026-05-04
chore: update CP semantics to expect redesigned namespaces field

nborges-aws:
## Description

Adds support for the new `namespaceTemplates` control-plane field. `namespaceTemplates` replaces `namespaces` on strategy configurations with clearer naming, as strategy namespace values have always been templates resolved at ingestion time, not literal namespaces. The old namespaces field is deprecated but remains accepted for backward compatibility.

- **agentcore.json schema** accepts namespaceTemplates and reflectionNamespaceTemplates on strategy definitions. Old namespaces / reflectionNamespaces still validate. Specifying both the deprecated and preferred form for the same concept is rejected.
- **Strategy creation flows** (`agentcore create --memory longAndShortTerm`, `agentcore add memory`) now write namespaceTemplates / reflectionNamespaceTemplates in the generated config.
- **Response parsing** (getMemoryDetail) reads either field from the service and surfaces the new name to downstream code.
- **Import flow** (`agentcore import memory`) writes the new field names to local config regardless of which form the service returned.
- **Dev web UI resources view** exposes `namespaceTemplates` in the API, with fallback to `namespaces` for legacy data.
- **JSON schema** (schemas/agentcore.schema.v1.json) regenerated from Zod to include the new fields.
- **Docs** (docs/memory.md, docs/configuration.md) updated to show `namespaceTemplates` as primary with a deprecation note for the old name.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ x ] Other (please describe): Deprecation of field and addition of replacement in CP methods

## Testing

- [ x ] I ran `npm run test:unit` and `npm run test:integ`
- [ x ] I ran `npm run typecheck`
- [ x ] I ran `npm run lint`

## Checklist

- [ x ] I have read the CONTRIBUTING document
- [ x ] I have added any necessary tests that prove my fix is effective or my feature works
- [ x ] I have updated the documentation accordingly
- [ x ] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [ x ] My changes generate no new warnings

---

terms of your choice.

### DIFF
--- a/src/cli/commands/import/import-memory.ts
+++ b/src/cli/commands/import/import-memory.ts
@@ -42,14 +42,16 @@ function filterInternalNamespaces(namespaces: string[]): string[] {
 function toMemorySpec(memory: MemoryDetail, localName: string): Memory {
   const strategies: Memory['strategies'] = memory.strategies.map(s => {
     const mappedType = mapStrategyType(s.type);
-    const filteredNamespaces = s.namespaces ? filterInternalNamespaces(s.namespaces) : [];
+    const filteredTemplates = s.namespaceTemplates ? filterInternalNamespaces(s.namespaceTemplates) : [];
     return {
       type: mappedType as Memory['strategies'][number]['type'],
       ...(s.name && { name: s.name }),
       ...(s.description && { description: s.description }),
-      ...(filteredNamespaces.length > 0 && { namespaces: filteredNamespaces }),
-      ...(s.reflectionNamespaces &&
-        s.reflectionNamespaces.length > 0 && { reflectionNamespaces: s.reflectionNamespaces }),
+      ...(filteredTemplates.length > 0 && { namespaceTemplates: filteredTemplates }),
+      ...(s.reflectionNamespaceTemplates &&
+        s.reflectionNamespaceTemplates.length > 0 && {
+          reflectionNamespaceTemplates: s.reflectionNamespaceTemplates,
+        }),
     };
   });
 
--- a/src/schema/llm-compacted/agentcore.ts
+++ b/src/schema/llm-compacted/agentcore.ts
@@ -103,8 +103,12 @@ interface MemoryStrategy {
   type: MemoryStrategyType;
   name?: string; // @regex ^[a-zA-Z][a-zA-Z0-9_]{0,47}$ @max 48
   description?: string;
+  namespaceTemplates?: string[];
+  reflectionNamespaceTemplates?: string[]; // EPISODIC only: templates for cross-episode reflections
+  /** @deprecated Use namespaceTemplates instead. */
   namespaces?: string[];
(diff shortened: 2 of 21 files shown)


src/cli/commands/import/import-memory.ts
```diff
@@ -42,14 +42,16 @@ function filterInternalNamespaces(namespaces: string[]): string[] {
...
-    const filteredNamespaces = s.namespaces ? filterInternalNamespaces(s.namespaces) : [];
+    const filteredTemplates = s.namespaceTemplates ? filterInternalNamespaces(s.namespaceTemplates) : [];
     return {
       type: mappedType as Memory['strategies'][number]['type'],
       ...(s.name && { name: s.name }),
       ...(s.description && { description: s.description }),
-      ...(filteredNamespaces.length > 0 && { namespaces: filteredNamespaces }),
-      ...(s.reflectionNamespaces &&
-        s.reflectionNamespaces.length > 0 && { reflectionNamespaces: s.reflectionNamespaces }),
+      ...(filteredTemplates.length > 0 && { namespaceTemplates: filteredTemplates }),
```
Hweinstock, line 50: q: if an existing memory has the old field on it does import still work?

src/schema/llm-compacted/agentcore.ts
```diff
@@ -103,8 +103,12 @@ interface MemoryStrategy {
   type: MemoryStrategyType;
   name?: string; // @regex ^[a-zA-Z][a-zA-Z0-9_]{0,47}$ @max 48
   description?: string;
+  namespaceTemplates?: string[];
+  reflectionNamespaceTemplates?: string[]; // EPISODIC only: templates for cross-episode reflections
+  /** @deprecated Use namespaceTemplates instead. */
```
Hweinstock, line 108: when the backend deprecates this field, how will we know we need to deprecate it as well? Also, how do we communicate this to users on old versions?

Hweinstock (commented):
Looks like there are some merge conflicts, but seems reasonable.

src/cli/commands/import/import-memory.ts
nborges-aws: Yep, and this was one of the flows tested in bug bash. It will resolve to the new fields in local config, regardless of old/new fields on the existing memory.

src/schema/llm-compacted/agentcore.ts
nborges-aws: Any users calling the methods and passing the old fields will receive a deprecation warning logged out. The service team will give us some notice of when the grace-period will be ending on the deprecated field, at which point we'll have to do a breaking change release to remove the old fields

Hweinstock (approved):

PR #1121, opened 2026-05-05
feat: add archive command for batch evaluations and recommendations 

padmak30:
Cherrypick 7586092e3cd9c9a967944a36af689c8bf000ad37

* feat: add archive command for batch evaluations and recommendations. Introduces a new top-level archive command with two subcommands: archive batch-evaluation and archive recommendation. Each calls the corresponding service delete API and removes the matching local .cli/ history file.

### DIFF
--- a/docs/policies/iam-policy-user.json
+++ b/docs/policies/iam-policy-user.json
@@ -159,6 +159,169 @@
         "s3:GetObject"
       ],
       "Resource": "*"
+    },
+    {
+      "Sid": "AgentCoreResourceManagement",
+      "Effect": "Allow",
+      "Action": [
+        "bedrock-agentcore:CreateAgentRuntime",
+        "bedrock-agentcore:UpdateAgentRuntime",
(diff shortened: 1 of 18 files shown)


padmak30:
For the 2 comments from Harness review
1. The additional IAM changes were added to fix a merge conflict, since both branches were out of sync for this file. The existing policy on main has `ACCOUNT_ID`, so keeping it for now. This policy is not used anywhere, only used for docs.
2. All other existing commands (recommendations history, evals history) call requireProject() outside the try/catch — same pattern as archive. None of them handle the --json contract on a project-not-found failure; they all fall through to requireProject's Ink render + process.exit(1). So keeping the behavior consistent with other patterns and main.

docs/policies/iam-policy-user.json
notgitika, line 1: is this file just an artifact for main preview sync?

notgitika (approved):

docs/policies/iam-policy-user.json
padmak30: yes

PR #1122, opened 2026-05-05
test: trim scheduled e2e matrix and harden dev-server process cleanup

tejaskash:
## Summary
- Scheduled Monday e2e run now targets 6 priority suites (CodeZip, Container, custom JWT auth, evals, import, local dev). Manual dispatch still runs the full e2e directory to exercise every framework/model-provider combo — halves the weekly AWS test burn
- Replace fire-and-forget SIGTERM in \`integ-tests/dev-server.test.ts\` with a process-group terminator: signal via \`-pid\`, wait on exit event, SIGKILL after 5s timeout. Spawn uses \`detached:true\` so the group signal reaches subprocesses

## Paired PR
The CDK construct test isolation fix (\`setupMocks()\` mock pollution in \`evaluator.test.ts\`) lives in the \`agentcore-l3-cdk-constructs\` repo as a separate PR, since it's a different package.

## Test plan
- [x] Integ tests still pass — \`npm run test:integ\`
- [x] \`e2e-tests-full.yml\` YAML is valid (validated locally with Python yaml)
- [x] Manual workflow_dispatch of \`e2e-tests-full.yml\` still runs the full suite (sanity-check before relying on the schedule split)

### DIFF
--- a/.github/workflows/e2e-tests-full.yml
+++ b/.github/workflows/e2e-tests-full.yml
@@ -28,6 +28,17 @@ jobs:
       matrix:
         cdk-source: [npm, main]
         shard: ['1/6', '2/6', '3/6', '4/6', '5/6', '6/6']
+    env:
+      # Scheduled runs cover 6 priority suites spanning CodeZip, Container, Auth,
+      # Evals, Import, and local Dev paths. Manual workflow_dispatch runs the
+      # full e2e directory to exercise every framework/model-provider combo.
+      SCHEDULED_TESTS: >-
+        e2e-tests/strands-bedrock.test.ts
+        e2e-tests/container-strands-bedrock.test.ts
+        e2e-tests/byo-custom-jwt.test.ts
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-tests-full.yml
```diff
@@ -28,6 +28,13 @@ jobs:
       matrix:
         cdk-source: [npm, main]
         shard: ['1/6', '2/6', '3/6', '4/6', '5/6', '6/6']
+    env:
+      # Scheduled runs cover 6 priority suites spanning CodeZip, Container, Auth,
+      # Evals, Import, and local Dev paths. Manual workflow_dispatch runs the
+      # full e2e directory to exercise every framework/model-provider combo.
+      SCHEDULED_TESTS: >-
```
Hweinstock, line 35: whats the advantage of scheduled tests if we run them on every commit?

Hweinstock (approved):

PR #1123, opened 2026-05-05
test: per-directory coverage + E2E cleanup hardening (Days 5-6)

tejaskash:
## Summary

Two days of test infrastructure work bundled together since they're orthogonal but small.

### Day 5 — Per-directory coverage thresholds

- New `scripts/check-coverage.mjs` enforces per-directory line coverage thresholds (ratchet values ~5% below current coverage; target values documented inline)
- PR-changed-line coverage check (≥50%) with per-file breakdown
- Switched `davelosert/vitest-coverage-report-action` to `file-coverage-mode: changes` so PR comments highlight uncovered new lines
- Ratchet thresholds: schema 78%, operations 52%, lib 82%, aws 40%, tui/hooks 14%, tui/components 68%, commands 46%

### Day 6 — E2E cleanup hardening + sweeper spec

- `teardownE2EProject()` retries deploy up to 3× (15s backoff) and **throws on failure** instead of swallowing silently. Stops silent AWS resource leaks.
- CDK stacks now tagged with `Environment=e2e-test` / `CreatedAt` / `CreatedBy` when `AGENTCORE_E2E_TEST=1` is set (opt-in, user projects unaffected)
- `cleanupStaleCredentialProviders` window shrunk from 30min → 5min — catches orphans faster
- Cognito cleanup in `byo-custom-jwt.test.ts` retries 3× with 5s delay instead of swallowing errors
- New `docs/test-resource-sweeper-spec.md` for the infra team to implement a scheduled sweeper

## Test plan
- [x] `node scripts/check-coverage.mjs` passes against current coverage locally
- [x] Coverage job in CI runs and posts per-directory table on this PR
- [x] Intentionally lower a threshold → CI fails (verify the gate works)
- [x] Typecheck passes on all packages


tejaskash:
Addressed — reverted the default to 30 min and added an opt-in `E2E_STALE_CRED_MAX_AGE_MS` env var so the scheduled sweeper can still use a tight window without risking normal CI.

tejaskash:
Addressed — wrapped the `teardownE2EProject` call in `try/finally` across `createE2ESuite` + all 5 standalone callers (`ab-test-target-based`, `ab-test-config-bundle`, `archive-lifecycle`, `config-bundle-eval-rec`, `evals-lifecycle`) so `testDir` cleanup always runs.

tejaskash:
Addressed — extended `vitest.config.ts` coverage include to `src/**/*.tsx` (and excluded `*.test.tsx`), plus hardened `check-coverage.mjs` to treat a changed production file missing from coverage as 0% covered instead of skipping. Adjusted the `src/cli/commands/` ratchet from 46% → 34% to reflect the new denominator (3424 → 4553 lines with .tsx included).

e2e-tests/archive-lifecycle.test.ts
```diff
@@ -67,10 +67,13 @@ describe.sequential('e2e: archive command lifecycle', () => {
   }, 300000);
 
   afterAll(async () => {
-    if (projectPath && hasAws) {
-      await teardownE2EProject(projectPath, agentName, 'Bedrock');
+    try {
```
Hweinstock, line 70: nit: does it make sense to consolidate this cleanup into a common util?

e2e-tests/byo-custom-jwt.test.ts
```diff
@@ -48,7 +48,36 @@ const region = process.env.AWS_REGION ?? 'us-east-1';
...
+  return runCLI(args, cwd, {
+    skipInstall: false,
+    env: {
+      AGENTCORE_E2E_TEST: '1',
+      AGENTCORE_E2E_CREATOR: process.env.AGENTCORE_E2E_CREATOR ?? 'github-actions',
+    },
+  });
+}
+
+async function deleteCognitoResourceWithRetry(
```
Hweinstock, line 60: It looks like this function is called `deleteCognitoResourceWithRetry`, but is actually a generalized retry with cognito specific logs. 

Does it make sense to fully generalize it?

Hweinstock:
Is this functionality supported by CodeCov: https://about.codecov.io/ ? I know we already leverage this in Python SDK and wondering if we could leverage here as well.

PR #1125, opened 2026-05-05
refactor: unify result types with discriminated Result<T, E> union

Hweinstock:
### Problem 
The codebase uses an inconsistent result-like type that changes depending on the context. The implementations vary slightly, and are not explicit about (usually) sharing a common shape. This common shape also strips error types, and distills their information down to a string, which leads to brittle behavior when catching errors. 

As an example, systems that span modules (ex. telemetry) must handle each result type individually, resulting in excessive branching and complexity. Additionally, loss of error information, means loss of this information in telemetry, making it difficult to impossible to tell the type of error, and if it was user/application caused. 

### Solution
- unify result types behind a common flexible interface loosely based on Rust results. The main difference is that data is hoisted to the top level to be consistent with existing patterns. 
- store full error information on the result, allowing for `instanceof` checking and error name matching for stronger exception handling and telemetry. 

### Future Work
There are other result types that could be generalized with further work to support spread operators on the error as well. I left this OOS since the PR is already very large. Some other small patterns that this PR continues, but I think should change in the future:
- excessive mocking (Ex. some tests mock entire modules, rather than the underlying dependencies. We mock our own code in some places). 
- testing implementation rather than behavior (error messages are matches verbatim in many places, likely because thats all we had before. However, we should change these tests the verify error "shape" rather than exact content. 
- higher level utilities for working with results. There are common patterns that could be useful to abstract. As an example, a `mapResult` function that allows data transformations with clear error propagation. This could avoid some nesting of try/catches in a few places. 
- add more error types as we see fit. 
- avoid using console.log directly for more flexibility. 
## Related Issue

Closes #

## Documentation PR

N/A — internal refactor, no user-facing changes.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Other (please describe): Internal refactor — unify result types

## Testing

- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.

### DIFF
--- a/src/cli/commands/import/__tests__/import-runtime-handler.test.ts
+++ b/src/cli/commands/import/__tests__/import-runtime-handler.test.ts
@@ -10,6 +10,7 @@
  * - Fails when runtime name already exists in project
  */
 import { handleImportRuntime } from '../import-runtime';
+import assert from 'node:assert';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 
 // ── Mock dependencies ────────────────────────────────────────────────────────
@@ -25,7 +26,7 @@ const mockParseAndValidateArn = vi.fn();
 const mockFindResourceInDeployedState = vi.fn();
 const mockFailResult = vi.fn((...args: unknown[]) => ({
   success: false,
-  error: args[1] as string,
+  error: new Error(args[1] as string),
   resourceType: args[2] as string,
   resourceName: args[3] as string,
   logPath: 'test.log',
@@ -205,8 +206,9 @@ describe('handleImportRuntime', () => {
       });
 
       expect(result.success).toBe(false);
-      expect(result.error).toContain('Could not determine entrypoint');
-      expect(result.error).toContain('--entrypoint');
+      assert(!result.success);
+      expect(result.error.message).toContain('Could not determine entrypoint');
+      expect(result.error.message).toContain('--entrypoint');
     });
 
     it('fails with clear error when entryPoint is undefined', async () => {
@@ -231,7 +233,8 @@ describe('handleImportRuntime', () => {
       });
 
       expect(result.success).toBe(false);
-      expect(result.error).toContain('Could not determine entrypoint');
+      assert(!result.success);
+      expect(result.error.message).toContain('Could not determine entrypoint');
     });
 
     it('fails with clear error when entryPoint is empty array', async () => {
@@ -256,7 +259,8 @@ describe('handleImportRuntime', () => {
       });
 
       expect(result.success).toBe(false);
-      expect(result.error).toContain('Could not determine entrypoint');
+      assert(!result.success);
+      expect(result.error.message).toContain('Could not determine entrypoint');
     });
 
     it('uses --entrypoint flag when provided, bypassing auto-detection', async () => {
@@ -369,7 +373,8 @@ describe('handleImportRuntime', () => {
       });
 
       expect(result.success).toBe(false);
-      expect(result.error).toContain('Multiple runtimes found');
+      assert(!result.success);
+      expect(result.error.message).toContain('Multiple runtimes found');
     });
 
     it('errors when no runtimes exist', async () => {
@@ -382,7 +387,8 @@ describe('handleImportRuntime', () => {
       });
 
       expect(result.success).toBe(false);
-      expect(result.error).toContain('No runtimes found');
+      assert(!result.success);
+      expect(result.error.message).toContain('No runtimes found');
     });
   });
 
@@ -546,7 +552,8 @@ describe('handleImportRuntime', () => {
       });
 
       expect(result.success).toBe(false);
-      expect(result.error).toContain('--code');
+      assert(!result.success);
+      expect(result.error.message).toContain('--code');
     });
 
     it('fails when source path does not exist', async () => {
@@ -574,7 +581,8 @@ describe('handleImportRuntime', () => {
       });
 
       expect(result.success).toBe(false);
-      expect(result.error).toContain('does not exist');
+      assert(!result.success);
+      expect(result.error.message).toContain('does not exist');
     });
 
     it('fails when runtime name already exists in project', async () => {
@@ -606,7 +614,8 @@ describe('handleImportRuntime', () => {
       });
 
       expect(result.success).toBe(false);
-      expect(result.error).toContain('already exists');
+      assert(!result.success);
+      expect(result.error.message).toContain('already exists');
     });
   });
 });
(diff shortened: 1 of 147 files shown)


padmak30:
## Bug Bash Report — refactor: unify result types with discriminated Result<T, E> union

**Branch:** `refactor/unified-result-type-v2` | **Version:** 0.13.0 | **Base:** `main`

### TL;DR
No in-scope blockers, highs, or mediums found. Refactor is behavior-preserving: every happy/error path exercised produces identical output to `main`. Two pre-existing quirks re-observed (documented below as pre-existing). 1 note for maintainers.

### In-scope bugs
None.

### Pre-existing bugs (reproduce identically on `main`)
**[medium] Commands exit 0 on error**
- **Reproducer:**
  ```bash
  cd /tmp/empty-dir
  node dist/cli/index.mjs validate; echo "exit=$?"   # prints "No agentcore project found…" exit=0
  node dist/cli/index.mjs status;   echo "exit=$?"   # same
  node dist/cli/index.mjs status --json; echo "exit=$?"
  # inside a project, with malformed agentcore.json
  echo '{invalid json' > agentcore/agentcore.json
  node dist/cli/index.mjs validate;  echo "exit=$?"  # exit=0
  node dist/cli/index.mjs status --json; echo "exit=$?"  # exit=0
  ```
- **Evidence it's pre-existing:** Checked out `origin/main` (9afd3f08), rebuilt, reproduced identical output + `exit=0` for every case. Not introduced by this PR.

**[low] `status --json` prints human-readable text on error, not JSON**
- **Reproducer:** With a malformed `agentcore.json`, `status --json` prints `Error: Failed to parse JSON in config file…` as plain text instead of a JSON error envelope.
- **Evidence it's pre-existing:** Same output on `origin/main`.

### Notes
- **`invoke`/`logs`/`traces` without a TTY print command help instead of an error.** When running `invoke --name X --prompt hello` in a non-TTY shell, the CLI renders the subcommand help and exits 0 instead of reporting that an interactive terminal is required (or attempting non-interactive execution). Reproduces identically on `main`, so out of scope for this refactor; flagging for future UX work. The sole "interactive terminal required" message only surfaces for `deploy`.

### What works ✓
- `agentcore create --project-name bbtest --name myagent --defaults --skip-git --skip-install` produces a valid project with correct `agentcore.json`.
- `validate` on a good project → `Valid`, exit=0.
- `status` / `status --json` on a good project → correct human + JSON output; `resources` array and `deploymentState: local-only` reported.
- Schema violations are surfaced with precise per-field errors:
  - `{}` → `name: expected "string"`, `version: expected "number"`.
  - `runtimes[0].build: "BogusBuild"` → `expected "CodeZip" | "Container"`.
- Malformed JSON → `Invalid JSON in agentcore.json: Expected property name…` (position included).
- `remove agent --name nonexistent -y` → structured JSON `{"success":false,"error":"Agent \"nonexistent\" not found."}`.
- `remove agent --name myagent -y` → structured success JSON; subsequent `status --json` shows `resources: []`.
- `remove all -y` → `{"success":true,"message":"All schemas reset to empty state"}`.
- Build clean, `--version` prints `0.13.0`, all CLI help rendered correctly.

### Test coverage
| Surface | Tested | Result |
|---------|--------|--------|
| Build | yes | pass |
| CLI happy path (create/validate/status) | yes | pass |
| Error paths (missing project, malformed JSON, bad schema) | yes | pass (behavior identical to base) |
| Schema validation | yes | pass |
| Removal paths (`remove agent`, `remove all`, nonexistent) | yes | pass |
| Partition correctness | n/a | PR does not touch ARN/endpoint code |

Refactor LGTM from a black-box behavioral standpoint — every tested surface produces identical stdout/stderr/exit code to `main`.

jariy17:
## PR Test Report
8 passed, 0 failed. Refactor is behavior-preserving end-to-end: JSON error output stays backwards-compatible (additive `errorType`, `errors`, `cause` fields). Minor note: `agentcore status --json` output changed from pretty-printed to compact JSON (still parseable).
Report: https://gistpreview.github.io/?e5dbf4120a54b23a120eeff6072a3fd4/pr-agentcore-cli-1125-test-report.html

Hweinstock:
Will fix security audit as a follow-up since its unrelated to changes here. 

https://github.com/aws/agentcore-cli/pull/1211

src/cli/commands/import/__tests__/import-runtime-handler.test.ts
```diff
@@ -205,8 +206,9 @@ describe('handleImportRuntime', () => {
       });
 
       expect(result.success).toBe(false);
```
avi-alpert, line 208: wouldn't the `expect`s in this test file be redundant? if you have assert(!result.success); right after

Hweinstock: Yep, I tried to swap to `assert` everywhere where we needed the narrowed type, but it looks like I missed removing the existing expect here. It should be harmless to have both, and can also follow-up with a clean-up to remove this.

src/cli/commands/validate/action.ts
```diff
@@ -39,22 +35,22 @@ export async function handleValidate(options: ValidateOptions): Promise<Validate
...
+    return { success: false, error: new Error(formatError(err, 'aws-targets.json'), { cause: err }) };
   }
 
   // Validate deployed state if it exists (.cli/state.json)
   if (configIO.configExists('state')) {
     try {
       await configIO.readDeployedState();
     } catch (err) {
-      return { success: false, error: formatError(err, '.cli/state.json') };
+      return { success: false, error: new Error(formatError(err, '.cli/state.json'), { cause: err }) };
```
avi-alpert, line 53: why arent the errors in this file ValidationErrors?

avi-alpert:
Can you update the Agents.MD file i see a lot of stuff in there referencing the old result shapes

src/cli/commands/validate/action.ts
Hweinstock: good question, we technically don't know if its an error from the schema being invalidated, or from issues reading from the filesystem. See implementation: https://github.com/aws/agentcore-cli/blob/6dc4b7fd5ad5c53beee52e51ac0dd10386f2986b/src/lib/schemas/io/config-io.ts#L195. 

We could have some logic that checks the error thrown and determines if it should be validation or something else, but I left that OOS.

avi-alpert (dismissed):

Hweinstock:
> Can you update the Agents.MD file i see a lot of stuff in there referencing the old result shapes

Good call, let me update this before merging.

avi-alpert (approved):

PR #1127, opened 2026-05-06
fix: honor aws-targets.json region for AWS SDK calls when AWS_DEFAULT_REGION is unset

### LINKED ISSUE #924: CLI ignores region from aws-targets.json for API calls unless AWS_DEFAULT_REGION is set

aidandaly24:
## Description

Make the resolved deployment target's region from `aws-targets.json` authoritative on the process environment for every CLI command that may construct AWS SDK clients (or invoke CDK toolkit-lib internals) without an explicit `region` option.

### Background

Previously, when a user had a non-default region in `aws-targets.json` but no `AWS_DEFAULT_REGION` set in their shell, the CLI would silently fall back to the SDK's default region (typically `us-east-1`) when making API calls — creating runtimes, evaluators, online eval configs, etc. in the wrong region. This was inconsistent with CDK synthesis, which correctly reads the region from `aws-targets.json`.

A partial fix already existed for `deploy`, `teardown`, and the TUI preflight hook via `applyTargetRegionToEnv()` in `src/cli/aws/target-region.ts`. This PR closes the remaining gaps and adds a structural guarantee plus tests so the behaviour cannot silently regress.

### Changes

- **`src/cli/aws/target-region.ts`** — adds a `runWithTargetRegion(getTarget, fn)` convenience wrapper for handlers where target resolution and the AWS SDK work live in the same lexical scope. Adds a top-of-file policy comment documenting that any new CLI entry point that may construct SDK clients without an explicit `region` option **must** apply the override after resolving its target.
- **`src/cli/aws/index.ts`** — re-exports `runWithTargetRegion`.
- **`src/cli/commands/import/actions.ts`** — replaces the bare `process.env.AWS_REGION = …` / `process.env.AWS_DEFAULT_REGION = …` mutation (which leaked env state to subsequent in-process work and was not undone on early returns) with `applyTargetRegionToEnv()`. Re-applies the override with the resolved target region (which overrides the YAML-based initial region) and restores in a top-level `finally` block on both success and error paths.
- **`src/cli/commands/abtest/command.ts`** — `getRegion()` now also promotes the resolved region onto `AWS_REGION` / `AWS_DEFAULT_REGION` so any helper constructing an SDK client without an explicit region behaves consistently. (The previous fallback could silently land on `us-east-1` even when `aws-targets.json` was present.)
- **Tests**:
  - New `src/cli/commands/deploy/__tests__/region-override.test.ts` — verifies `handleDeploy` propagates `target.region` to `process.env` during execution and restores after, on both happy and error paths, and that env is not mutated when no target can be resolved.
  - New `src/cli/commands/import/__tests__/region-override.test.ts` — analogous coverage for `handleImport`, including the no-physical-IDs (light) path.
  - Extended `src/cli/aws/__tests__/target-region.test.ts` with cases for the new `runWithTargetRegion` wrapper (success, no-target skip, throw-from-callback, throw-from-resolver).

### Out of scope

The CDK construct library (`agentcore-l3-cdk-constructs`) is not affected — CDK synthesis already reads the region directly from `aws-targets.json`. This bug was exclusively in the CLI's runtime AWS SDK calls.

## Related Issue

Closes #924

## Documentation PR

N/A — internal behaviour fix, no user-facing documentation changes.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [ ] I ran `npm run lint`

Targeted test runs that pass on this branch:

- `src/cli/aws/__tests__/target-region.test.ts` — 11/11 pass (includes new `runWithTargetRegion` cases).
- `src/cli/commands/deploy/__tests__/region-override.test.ts` — 3/3 pass.
- `src/cli/commands/import/__tests__/region-override.test.ts` — 3/3 pass.
- Regression checks on touched `import/actions.ts`: `import-no-deploy.test.ts` + `idempotency.test.ts` — 39/39 pass.
- `npm run typecheck` is clean.

## Checklist

---

terms of your choice.

### DIFF
--- a/src/cli/aws/__tests__/target-region.test.ts
+++ b/src/cli/aws/__tests__/target-region.test.ts
@@ -1,4 +1,5 @@
-import { applyTargetRegionToEnv, withTargetRegion } from '../target-region.js';
+import type { AwsDeploymentTarget } from '../../../schema';
+import { applyTargetRegionToEnv, runWithTargetRegion, withTargetRegion } from '../target-region.js';
 import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 
 describe('target-region', () => {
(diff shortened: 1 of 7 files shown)


aidandaly24:
Closing: created by a test run with broken extract phase (0-file diff → insufficient review)

PR #1148, opened 2026-05-06
fix(deploy): pass stack selection to diff for --target filtering

### LINKED ISSUE #980: bug(deploy): --diff flag diffs all stacks instead of filtering to the specified --target
### Description

`agentcore deploy --target <name> --diff` shows CDK diff for all stacks in the synthesized app, not just the stack corresponding to the specified --target. When multiple targets are defined in aws-targets.json, the --diff flag ignores the --target filter.

### Steps to Reproduce

1. Configure two or more targets in `aws-targets.json`:
```
[
  { "name": "preprod-east2", "account": "...", "region": "us-east-2" },
  { "name": "default", "account": "...", "region": "us-west-2" }
]
```
2. Run `agentcore deploy --target default --diff`
3. Observe the output shows diff for both `Stack AgentCore-<project>-preprod-east2` and `Stack AgentCore-<project>-default`

### Expected Behavior

Only the stack for the specified target (`AgentCore-<project>-default`) should be diffed. The `--target` flag should filter the diff to the selected target's stack, consistent with how `--target `scopes other deploy operations.

### Actual Behavior

Both stacks are diffed. The output shows `Number of stacks with differences: 2` and includes full diff output for both targets, even though only `default` was specified.

### CLI Version

0.11.0

### Operating System

macOS

### Additional Context

Root cause is in `actions.ts`. The diff code path calls `toolkitWrapper.diff()` without passing a stack selection filter:

```
// line ~197 in actions.ts
await toolkitWrapper.diff();
```

The CdkToolkitWrapper.diff() method in wrapper.ts accepts a DiffOptions with optional stacks selection, but it's never provided. Since 
cdk.ts synthesizes a stack per target, CDK diffs all of them.

**Suggested fix - pass the target-specific stack name:**
```
import { StackSelectionStrategy } from '@aws-cdk/toolkit-lib';

await toolkitWrapper.diff({
  stacks: {
    strategy: StackSelectionStrategy.PATTERN_MUST_MATCH,
    patterns: [stackName],
  },
});
```

Note: the same issue possibly also affects toolkitWrapper.deploy() which is also called without stack selection.

Introduced in PR #485 (feat: add --diff flag to deploy command).

Hweinstock:
## Description

When `--diff` or `--deploy` is used with `--target`, the CLI now passes a stack selection pattern (`*${targetName}*`) to the CDK toolkit so only the selected target's stack is operated on, not all stacks in the synthesized app.

Extracted `runDiff` and `runDeploy` functions for testability via dependency injection.

## Related Issue

Closes #980

## Type of Change

- [x] Bug fix

## Testing

- [x] I ran `npm run lint`

## Checklist

### DIFF
--- a/src/cli/commands/deploy/__tests__/deploy.test.ts
+++ b/src/cli/commands/deploy/__tests__/deploy.test.ts
@@ -1,4 +1,6 @@
 import { runCLI } from '../../../../test-utils/index.js';
+import { runDeploy, runDiff } from '../actions.js';
+import { StackSelectionStrategy } from '@aws-cdk/toolkit-lib';
 import { randomUUID } from 'node:crypto';
 import { mkdir, rm, writeFile } from 'node:fs/promises';
 import { tmpdir } from 'node:os';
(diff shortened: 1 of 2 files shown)


Hweinstock:
Addressed by no longer relying on arbitrary first stack name behavior.

Hweinstock:
Also fixes https://github.com/aws/agentcore-cli/issues/1150

jesseturner21 (approved):
good fix!

PR #1158, opened 2026-05-07
fix(dev): honor explicit --port literally and log silent runtime-index offsets (#1079)

### LINKED ISSUE #1079: agentcore dev -p <port> is silently offset by runtime index when running multiple runtimes in parallel
### Description

When running multiple `agentcore dev` instances in parallel, `getAgentPort()` in
  [`src/cli/operations/dev/config.ts`](https://github.com/aws/agentcore-cli/blob/main/src/cli/operations/dev/config.ts) silently offsets the port passed via
  `-p` by the runtime's index in `agentcore.json` (`basePort + runtimeIndex`). The shift is not logged and the flag's help text (`"Port for development
  server"`) gives no hint that `-p` is a *base* port.

## Why this is a problem

  The actual bound port depends on **which other runtimes are configured in `agentcore.json` and which ones happen to be running**, not just on the `-p` value.
   This breaks local development and integration testing in practical ways:

  - Downstream services (backend servers, integration tests, scripts) that target `http://localhost:<-p value>/invocations` silently talk to the wrong port —
  or get connection errors — depending on which runtimes are up.
  - Which runtime owns which port changes if someone reorders, adds, or removes entries in `agentcore.json`. A config-only change silently breaks every
  developer's local setup.
  - To hit a specific port, a user has to compute `-p = desiredPort - runtimeIndex` and keep that in sync with `agentcore.json`. The `-p` value in
  `package.json` / README / CI scripts no longer means what it says.
  - A second free-port search (`findFreePort` in `src/cli/operations/dev/utils.ts`) can shift the port *again* if `basePort + index` is occupied, and **only**
  that shift is announced. The first, index-based shift is invisible.

## Suggested Fixes (any one resolves the surprise)

1. Treat `-p` as the literal port; fail fast on conflict.
2. Add a per-runtime `devPort` field in `agentcore.json` so the mapping is explicit and stable across reorderings.
3. Log the offset whenever `getAgentPort` shifts the port, mirroring the existing `Port ${ee} in use, using ${re}` message.
4. At minimum, document the current behavior in `--help` and the README.

### Steps to Reproduce

1. `agentcore.json` with two runtimes:
     ```json
     { "runtimes": [{ "name": "AgentA", ... }, { "name": "AgentB", ... }] }
     ```
2. Run both in parallel with the ports you actually want:
     ```bash
     agentcore dev -r AgentA -p 8787 &
     agentcore dev -r AgentB -p 8788 &
     ```
3. `lsof -iTCP -sTCP:LISTEN -n -P` — `AgentB` is on `8789`, not `8788`.

### Expected Behavior

`AgentB` binds to `8788` (the value passed via `-p`).

### Actual Behavior

`AgentB` binds to `8789` (`8788 + index(AgentB)` = `8788 + 1`). No log line explains why.

### CLI Version

`@aws/agentcore@0.12.0`; logic still on `main` / v0.13.0.

### Operating System

macOS

### Additional Context

_No response_

aidandaly24:
## Description

Fix for issue #1079: `agentcore dev --port <port>` silently offsets the user's port by the runtime's index in `agentcore.json`, so `agentcore dev -r AgentB -p 8788` actually binds to 8789 (`8788 + index(AgentB)` = `8788 + 1`) without any log line. This is the bug the issue describes; it breaks downstream services that target `http://localhost:<-p value>/invocations`, makes `-p` values in `package.json` / README / CI scripts misleading, and means a config-only reorder of runtimes silently changes which port each agent binds.

This PR resolves the surprise by combining suggestions #1 and #3 from the issue:

- **Explicit `--port` is honored literally.** Any non-default Commander value source (`'cli'`, `'env'`, `'config'`) disables the runtime-index offset and triggers fail-fast on conflict, so the user gets exactly the port they asked for or a clear error.
- **Implicit `--port` (no flag) keeps the auto-offset, but the offset is now logged.** The historical `basePort + runtimeIndex` allocation is preserved for users who rely on it, but the previously-silent shift now emits `Runtime "AgentB" is at index 1; using port 8081 (pass --port 8081 explicitly to override).` so it's visible.
- **`--port` help text is rewritten** to document the multi-runtime semantics, so users learn about the offset from `--help` instead of by debugging.

The fix touches three independent port-allocation pathways that all had the same bug:

1. `agentcore dev --logs` (non-interactive) — `src/cli/commands/dev/command.tsx`
2. The web-UI start handler — `src/cli/operations/dev/web-ui/handlers/start.ts`
3. The TUI dev-server hook — `src/cli/tui/hooks/useDevServer.ts` (this one was actually missing the offset entirely; it now applies the same offset + log + fail-fast as the other two)

The branching logic is extracted into pure, unit-tested helpers in `src/cli/commands/dev/port-resolution.ts` (`isPortExplicit`, `computeTargetPort`, `resolveDevPort`) so the explicit-vs-implicit semantics have a single source of truth.

### Backwards compatibility

- **Behavior change**: passing `-p 8080` (matching the registered default) is now treated as explicit, which disables the runtime-index offset and fails fast on conflict. CI scripts that previously did `-p = desiredPort - runtimeIndex` arithmetic will now bind to the wrong port — they should drop the arithmetic and pass the port they actually want. Documented in `CHANGELOG.md`.
- `getAgentPort()` is preserved as a deprecated wrapper around the new `resolveAgentPort()` (returning `{ port, offset }`) so external callers don't break.
- `agentcore-l3-cdk-constructs` was checked — no callers of `WebUIOptions` / `WebUIServer`, so no schema or CDK changes required.

## Related Issue

Closes #1079

## Documentation PR

N/A — `--port` help text and `CHANGELOG.md` are updated in this PR; no agent-docs changes needed.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Breaking change 
- [x] Documentation update 

## Testing

- [x] I ran `npm run lint` 

### Test coverage added

- `src/cli/commands/dev/__tests__/port-resolution.test.ts` (new file, 21 tests):
  - `isPortExplicit` — all source values (`cli`/`env`/`config`/`default`) plus the defensive `console.warn` path for missing `getOptionValueSource` API
  - `computeTargetPort` — A2A/MCP fixed ports, HTTP explicit-honored, HTTP implicit-offset, offset-log emission
  - `resolveDevPort` — explicit-conflict error vs. implicit silent-shift, A2A/MCP-specific conflict messages
- `src/cli/operations/dev/__tests__/config.test.ts` (+8 tests for `resolveAgentPort`):
  - Explicit vs. implicit, project-null, agent-not-found, empty-string agent name, zero-runtimes project, first-runtime offset 0, `getAgentPort` backwards-compat
- `src/cli/operations/dev/web-ui/__tests__/start.test.ts` (new file, 9 tests):
  - Offset log when `basePortIsExplicit=false` and index > 0
  - Honors basePort literally when explicit
  - Falls back to `uiPort + 1 + index` when `basePort` is undefined
  - Multi-runtime explicit-conflict hint
  - Programmatic-misuse guard (`basePortIsExplicit` ignored when `basePort` undefined)
  - A2A and MCP protocol gates verifying no spurious "index N" log fires and the protocol-specific conflict message is returned

### Manual verification

Reproducing the issue's repro steps:

```bash
agentcore dev -r AgentA -p 8787 --logs &
agentcore dev -r AgentB -p 8788 --logs &
lsof -iTCP -sTCP:LISTEN -n -P | grep 87
```

Confirmed `AgentB` binds to `8788` (not `8789`). Without `-p`, the offset still applies but emits a clear `Runtime "AgentB" is at index 1; using port 8081 (pass --port 8081 explicitly to override).` log.

## Checklist

### Review history

This PR went through 4 rounds of review:

- **Round 1 (16 findings):** Mocked `waitForServerReady` in tests (cut runtime from ~3min → 1.25s); guarded `basePortIsExplicit` to require `basePort`; broadened explicit detection to any non-default Commander source; added programmatic-misuse guards; documented the `-p 8080` behavior change in `CHANGELOG`; added improved multi-runtime conflict hint; tightened `useDevServer` restart fallback; documented `WebUIOptions.basePort`/`basePortIsExplicit` pairing for external callers.
- **Round 2 (5 findings):** Extracted port-resolution branching from `command.tsx` into a pure, unit-testable `port-resolution.ts` module; added unit tests for the defensive `console.warn` path; added missing-conflict-fallback case for the misuse guard; added A2A/MCP gate tests for the start handler; locked the `index <= 0` guard with empty-name and zero-runtimes cases.
- **Round 3 (4 findings):** Fixed a regression where browser-mode would assign HTTP agents to 8081 (the web UI's own port); split `resolveDevPort` into `computeTargetPort` + `resolveDevPort` so the CLI flow has a single source of truth for the target port; symmetric `findAvailablePort` probe in the TUI explicit-port branch; simplified restart logic so the resolved port no longer depends on a stale `portFree` flag.
- **Round 4: 0 findings, all approved.**

---

terms of your choice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+### Changed
+- **Breaking:** `agentcore dev --port <port>` now honors the value literally
+  instead of silently offsetting it by the runtime's index in `agentcore.json`
+  (issue #1079). When `--port` is passed (including `-p 8080`, which matches
+  the registered default), per-runtime offsetting is disabled and a port
+  conflict fails fast with a clear error message. When `--port` is omitted,
+  the previous `basePort + runtimeIndex` auto-allocation continues to apply
+  but now logs the offset whenever it shifts. Migration: drop the manual
+  `-p = desiredPort - runtimeIndex` arithmetic from your scripts and pass the
+  port you actually want.
+
 ## [0.13.1] - 2026-05-06
 
 ### Added
--- a/src/cli/commands/dev/__tests__/dev.test.ts
+++ b/src/cli/commands/dev/__tests__/dev.test.ts
@@ -13,6 +13,10 @@ describe('dev command', () => {
       expect(result.stdout.includes('--stream'), 'Should show --stream option').toBeTruthy();
       expect(result.stdout.includes('--logs'), 'Should show --logs option').toBeTruthy();
       expect(result.stdout.includes('8080'), 'Should show default port').toBeTruthy();
+      expect(
+        /base.*port|runtime\s+index/i.test(result.stdout),
+        'Should document multi-runtime --port semantics (issue #1079)'
+      ).toBeTruthy();
     });
 
     it('does not show --invoke flag', async () => {
--- a/src/cli/commands/dev/__tests__/dev.test.ts
+++ b/src/cli/commands/dev/__tests__/dev.test.ts
@@ -13,8 +13,10 @@ describe('dev command', () => {
       expect(result.stdout.includes('--stream'), 'Should show --stream option').toBeTruthy();
       expect(result.stdout.includes('--logs'), 'Should show --logs option').toBeTruthy();
       expect(result.stdout.includes('8080'), 'Should show default port').toBeTruthy();
+      // Stable contract: the --port description must mention "runtime index"
+      // so users learn the multi-runtime semantics from --help (issue #1079).
       expect(
-        /base.*port|runtime\s+index/i.test(result.stdout),
+        result.stdout.includes('runtime index'),
         'Should document multi-runtime --port semantics (issue #1079)'
       ).toBeTruthy();
     });
--- a/src/cli/commands/dev/__tests__/port-resolution.test.ts
+++ b/src/cli/commands/dev/__tests__/port-resolution.test.ts
@@ -1,0 +1,220 @@
+import type { AgentCoreProjectSpec, DirectoryPath, FilePath } from '../../../../schema';
+import { isPortExplicit, resolveDevPort } from '../port-resolution';
+import { describe, expect, it, vi } from 'vitest';
+
+// Helper to cast strings to branded path types for testing
+const filePath = (s: string) => s as FilePath;
+const dirPath = (s: string) => s as DirectoryPath;
+
+const makeProject = (...names: string[]): AgentCoreProjectSpec => ({
+  name: 'TestProject',
+  version: 1,
+  managedBy: 'CDK' as const,
+  runtimes: names.map(name => ({
+    name,
+    build: 'CodeZip' as const,
+    runtimeVersion: 'PYTHON_3_12' as const,
+    entrypoint: filePath('main.py'),
+    codeLocation: dirPath(`./agents/${name}`),
+    protocol: 'HTTP' as const,
+  })),
+  memories: [],
+  credentials: [],
+  evaluators: [],
+  onlineEvalConfigs: [],
+  agentCoreGateways: [],
+  policyEngines: [],
+  configBundles: [],
+  abTests: [],
+  httpGateways: [],
+});
+
+describe('isPortExplicit', () => {
+  it('returns true when source is "cli"', () => {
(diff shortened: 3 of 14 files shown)


aidandaly24:
Closing: test run from batch agent

PR #1186, opened 2026-05-08
feat: multi-environment-deploy

aidandaly24:
## Summary

Autonomous implementation of feature: **multi-environment-deploy**

## Implementation Progress

# Progress — feature/multi-environment-deploy

## Tasks
- [x] T1: (commit: d4d9fcca) Add environment schemas to aws-targets schema
- [x] T2: (commit: dc58d3c7) Add cross-validation: env target refs must exist
- [x] T3: (commit: 8f432cff) Unit tests for environments schema
- [x] T4: (commit: 8e3e87de) Implement environment resolution and override merge
- [x] T5: (commit: 215c1979) Implement sequential multi-target deploy orchestrator
- [x] T6: (commit: 96d197d4) Register --env option and DeployOptions field
- [x] T7: (commit: d374cfb1) Wire --env resolution into deploy action with mutual-exclusion
- [x] T8: (commit: 0f92c8c1) Register --parallel and --continue-on-error options
- [x] T9: (commit: cfca7be4) Add parallel + continue-on-error paths to orchestrator
- [x] T10: (commit: cc149f1d) Add --env to status command with table output
- [x] T11: (commit: c91c5784) Expose environments via useAwsTargetConfig hook
- [x] T12: (commit: 19a2d84b) Add environment picker to deploy TUI screen
- [x] T13: (commit: a3006096) Add environment setup step to create TUI
- [x] T14: (commit: a57223d3) Target-to-environment assignment + write environments on create

## Review Summary

Reviewers: 5, Approved: False, Findings: 5

## Agentic Test Results

- Build: PASS
- Pack: PASS
- Install: PASS
- Version check: PASS
- Feature test: FAIL
- Overall: FAIL

Parallel Testers:
  - Schema acceptance: valid environments file parses; invalid configs are rejected with clear errors: FAIL (0 tests, 1 bugs)
      Bug: None confirmed in the feature itself — the feature could not be exercised because nothing is installed.
  - Backward compatibility: legacy array-shaped aws-targets.json still works end-to-end: FAIL (0 tests, 1 bugs)
      Bug: None reportable against the feature itself, because no feature artifact is present to test.
  - Unknown-environment error path returns the available list (resolveEnvironment + status integration): FAIL (0 tests, 1 bugs)
      Bug: None attributable to the `multi-environment-deploy` feature, because the feature could not be exercised. However, the test harness/environment itself 
  - Mutual-exclusion + flag-dependency validation (CLI surface, no AWS calls): FAIL (6 tests, 6 bugs)
      Bug: BLOCKER / ENVIRONMENT BUG: The `agentcore` CLI is not installed anywhere on PATH ($PATH = `/opt/amazon/bin:/opt/amazon/python3.10/bin:/usr/local/sbin:
      Bug: `command -v agentcore` → not found
      Bug: Exhaustive search: `find / -iname "*agentcore*" 2>/dev/null` only returns `/opt/amazon/bin/bedrock-agentcore`, which is an unrelated Python script and
      Bug: No `node_modules`, no `npm`, no source tree, no `package.json` for an `agentcore` project anywhere on disk.
      Bug: Repro (any of these reproduces it):
      Bug: Because the binary doesn'\''t exist, the validateDeployOptions code path added by T7/T8 cannot be exercised at all from the CLI surface. None of the mutu

Errors:
  - None confirmed in the feature itself — the feature could not be exercised because nothing is installed.
  - None reportable against the feature itself, because no feature artifact is present to test.
  - None attributable to the `multi-environment-deploy` feature, because the feature could not be exercised. However, the test harness/environment itself is broken in three independent ways:
  - BLOCKER / ENVIRONMENT BUG: The `agentcore` CLI is not installed anywhere on PATH ($PATH = `/opt/amazon/bin:/opt/amazon/python3.10/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`).
  - `command -v agentcore` → not found
  - Exhaustive search: `find / -iname "*agentcore*" 2>/dev/null` only returns `/opt/amazon/bin/bedrock-agentcore`, which is an unrelated Python script and is itself broken (`ModuleNotFoundError: No module
  - No `node_modules`, no `npm`, no source tree, no `package.json` for an `agentcore` project anywhere on disk.
  - Repro (any of these reproduces it):
  - Because the binary doesn'\''t exist, the validateDeployOptions code path added by T7/T8 cannot be exercised at all from the CLI surface. None of the mutual-exclusion / flag-dependency rules can be valida

---
*This PR was created by the autonomous feature builder pipeline.*

### DIFF
--- a/src/cli/commands/deploy/actions.ts
+++ b/src/cli/commands/deploy/actions.ts
@@ -6,7 +6,6 @@ import type {
   EnvironmentOverrides,
   Environments,
 } from '../../../schema';
-import { AwsTargetsSchema } from '../../../schema';
 import { applyTargetRegionToEnv } from '../../aws';
 import { validateAwsCredentials } from '../../aws/account';
 import { CdkToolkitWrapper, createSwitchableIoHost } from '../../cdk/toolkit-lib';
@@ -53,7 +52,6 @@ import { enableOnlineEvalConfigs } from '../../operations/deploy/post-deploy-onl
 import { toStackName } from '../import/import-utils';
 import type { DeployResult } from './types';
 import { StackSelectionStrategy } from '@aws-cdk/toolkit-lib';
-import { readFile } from 'node:fs/promises';
 
 export interface ValidatedDeployOptions {
   target: string;
@@ -735,6 +733,10 @@ export interface EnvDeployOptions {
   verbose?: boolean;
   plan?: boolean;
   diff?: boolean;
+  /** Run env targets concurrently via Promise.allSettled. */
+  parallel?: boolean;
+  /** Continue past per-target failures (sequential mode only). */
+  continueOnError?: boolean;
   onProgress?: (step: string, status: 'start' | 'success' | 'error') => void;
   onResourceEvent?: (message: string) => void;
   /** Sink for orchestrator progress + summary lines. Defaults to console.log. */
@@ -754,22 +756,21 @@ export interface EnvDeployResult {
  * `{ targets, environments }` object shape. Returns environments only when
  * the file uses the object shape; otherwise environments is undefined.
  */
+/**
+ * Read aws-targets.json with environments if present. Delegates to
+ * `ConfigIO.readAwsTargetsFull` and then applies region fallback so the same
+ * environment/profile precedence used by the rest of the CLI is honored.
+ */
 async function readAwsTargetsWithEnvironments(
   configIO: ConfigIO
 ): Promise<{ targets: AwsDeploymentTarget[]; environments?: Environments }> {
-  const filePath = configIO.getPathResolver().getAWSTargetsConfigPath();
-  const raw = await readFile(filePath, 'utf8');
-  const parsed: unknown = JSON.parse(raw);
-  if (Array.isArray(parsed)) {
-    // Legacy shape: hand off to ConfigIO so region fallback still applies.
-    const targets = await configIO.resolveAWSDeploymentTargets();
-    return { targets };
-  }
-  // Object shape: validate and return both halves.
-  const validated = AwsTargetsSchema.parse(parsed);
-  // Region fallback for object-shape entries (mirrors resolveAWSDeploymentTargets).
-  const targets = await configIO.resolveAWSDeploymentTargets();
-  return { targets, environments: validated.environments };
+  const full = await configIO.readAwsTargetsFull();
+  // Apply the standard region fallback (env vars / profile config) by going
+  // through resolveAWSDeploymentTargets — its result preserves saved regions
+  // and fills in only blanks, so we can safely substitute it for the targets
+  // we just read while keeping the environments map from the object shape.
+  const resolved = await configIO.resolveAWSDeploymentTargets();
+  return full.environments ? { targets: resolved, environments: full.environments } : { targets: resolved };
 }
 
 /**
@@ -796,7 +797,12 @@ export async function handleEnvDeploy(options: EnvDeployOptions): Promise<EnvDep
   const results: DeployResult[] = [];
   const aggregate = await deployToTargets(
     resolved.targets,
-    { environmentName: options.env, log: options.onLog },
+    {
+      environmentName: options.env,
+      parallel: options.parallel,
+      continueOnError: options.continueOnError,
+      log: options.onLog,
+    },
     async target => {
       const result = await handleDeploy({
         target: target.name,
--- a/src/cli/commands/deploy/actions.ts
+++ b/src/cli/commands/deploy/actions.ts
@@ -1,5 +1,12 @@
 import { ConfigIO, SecureCredentials } from '../../../lib';
-import type { AgentCoreMcpSpec, DeployedState } from '../../../schema';
+import type {
+  AgentCoreMcpSpec,
+  AwsDeploymentTarget,
+  DeployedState,
+  EnvironmentOverrides,
+  Environments,
+} from '../../../schema';
+import { AwsTargetsSchema } from '../../../schema';
 import { applyTargetRegionToEnv } from '../../aws';
 import { validateAwsCredentials } from '../../aws/account';
 import { CdkToolkitWrapper, createSwitchableIoHost } from '../../cdk/toolkit-lib';
@@ -33,7 +40,9 @@ import {
   synthesizeCdk,
   validateProject,
 } from '../../operations/deploy';
+import { mergeOverrides, resolveEnvironment } from '../../operations/deploy/environment';
 import { formatTargetStatus, getGatewayTargetStatuses } from '../../operations/deploy/gateway-status';
+import { deployToTargets } from '../../operations/deploy/multi-target';
 import { deleteOrphanedABTests, setupABTests } from '../../operations/deploy/post-deploy-ab-tests';
 import {
   resolveConfigBundleComponentKeys,
@@ -44,6 +53,7 @@ import { enableOnlineEvalConfigs } from '../../operations/deploy/post-deploy-onl
 import { toStackName } from '../import/import-utils';
 import type { DeployResult } from './types';
 import { StackSelectionStrategy } from '@aws-cdk/toolkit-lib';
+import { readFile } from 'node:fs/promises';
 
 export interface ValidatedDeployOptions {
   target: string;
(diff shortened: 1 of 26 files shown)


src/cli/operations/deploy/multi-target.ts
```diff
@@ -0,0 +1,117 @@
...
+  const successes: TargetDeployResult[] = [];
+  const failures: TargetDeployResult[] = [];
+
+  if (options.parallel) {
+    targets.forEach((target, i) => {
+      log(`[${i + 1}/${targets.length}] Deploying to ${target.name} (${target.region})...`);
+    });
+    const settled = await Promise.allSettled(targets.map((target, i) => deployFn(target, i)));
+    settled.forEach((result, i) => {
+      const target = targets[i]!;
```
notgitika, line 57: **BLOCKER: `--parallel` is unsafe due to shared process-level state in `handleDeploy`.**

Each `handleDeploy` call invokes `applyTargetRegionToEnv(target.region)` which mutates `process.env.AWS_REGION` and `AWS_DEFAULT_REGION` globally (restored in `finally`). When two targets in different regions run concurrently via this `Promise.allSettled`, they'll corrupt each other's environment.

Additionally, `validateProject()`, `synthesizeCdk()`, and the CDK toolkit wrapper likely share process-level state (e.g., CDK context, temp file paths).

Recommendation: either (a) remove `--parallel` for v1 and document it as a future enhancement requiring subprocess isolation, or (b) spawn each target deploy in a child process (e.g., `fork()`) so they get independent `process.env` and CDK state.

src/cli/commands/deploy/actions.ts
```diff
@@ -705,3 +726,106 @@ export async function handleDeploy(options: ValidatedDeployOptions): Promise<Dep
...
+async function readAwsTargetsWithEnvironments(
+  configIO: ConfigIO
+): Promise<{ targets: AwsDeploymentTarget[]; environments?: Environments }> {
+  const full = await configIO.readAwsTargetsFull();
+  // Apply the standard region fallback (env vars / profile config) by going
+  // through resolveAWSDeploymentTargets — its result preserves saved regions
+  // and fills in only blanks, so we can safely substitute it for the targets
+  // we just read while keeping the environments map from the object shape.
+  const resolved = await configIO.resolveAWSDeploymentTargets();
+  return full.environments ? { targets: resolved, environments: full.environments } : { targets: resolved };
```
notgitika, line 773: **HIGH: Dual-shape compatibility gap will break existing `--target` deploys.**

`readAwsTargetsWithEnvironments` calls `configIO.resolveAWSDeploymentTargets()` which internally uses `readAWSDeploymentTargets()` → validates against `AwsDeploymentTargetsSchema` (the array-only schema). Once a user writes the new `{ targets, environments }` object shape to disk, **every** existing code path that calls `resolveAWSDeploymentTargets()` will throw `ConfigValidationError`.

This means the moment a user adopts environments, `agentcore deploy --target prod` (without `--env`) will break.

You need to either:
1. Update `ConfigIO.readAWSDeploymentTargets()` to accept both shapes (extract `targets` array from object shape), or
2. Add a normalization layer that `resolveAWSDeploymentTargets` uses to handle both on-disk formats transparently.

src/cli/commands/deploy/command.tsx
```diff
@@ -157,12 +194,21 @@ export const registerDeploy = (program: Command) => {
       }) => {
         try {
           requireProject();
-          if (cliOptions.json || cliOptions.target || cliOptions.dryRun || cliOptions.yes || cliOptions.verbose) {
+          if (
```
notgitika, line 197: **MEDIUM: `target` is set to `undefined` when `--env` is present, but `DeployOptions.target` is non-optional downstream.**

```ts
target: cliOptions.env ? cliOptions.target : (cliOptions.target ?? 'default'),
```

When `--env` is used and `--target` is not provided, this evaluates to `undefined`. But `handleDeployCLI` checks `options.env` early and returns, so it works at runtime — however the type assertion `as DeployOptions` hides a type hole. A safer pattern:

```ts
target: cliOptions.target ?? (cliOptions.env ? undefined : 'default'),
```

Also: `validateDeployOptions` should run on the **raw** CLI options (before this defaulting logic) so the mutual-exclusion check (`--env` + `--target`) actually fires when both are explicitly passed.

src/cli/tui/screens/deploy/DeployScreen.tsx
```diff
@@ -192,6 +206,39 @@ export function DeployScreen({
...
+  useEffect(() => {
+    if (envChoice !== 'running' || !selectedEnvName) return;
+    let cancelled = false;
+    void handleEnvDeploy({
+      env: selectedEnvName,
+      // Honor the outer DeployScreen's autoConfirm prop so teardown deploys
+      // are NOT silently approved when the user opens the TUI without -y.
+      // Inside handleDeploy this gates the teardown-confirmation prompt.
+      autoConfirm: autoConfirm,
+      onLog: line => {
```
notgitika, line 219: **MEDIUM: This `useEffect` has a race condition and no error handling for `handleEnvDeploy` rejections.**

1. If the component unmounts mid-deploy (e.g., user presses Esc), `cancelled = true` prevents state updates — but `handleEnvDeploy` continues running the full CDK deploy in the background with no way to abort it. The cleanup function gives a false sense of cancellation.

2. There's no `.catch()` on the promise. If `handleEnvDeploy` throws (vs returning `{ success: false }`), the promise rejection is unhandled and the UI gets stuck in the 'running' state forever.

Suggestion:
```ts
}).then(result => { ... }).catch(err => {
  if (cancelled) return;
  setEnvDeployResult({ success: false, envName: selectedEnvName, results: [], error: err.message });
  setEnvChoice('done');
});
```

src/cli/tui/screens/create/CreateScreen.tsx
```diff
@@ -209,6 +213,65 @@ export function CreateScreen({ cwd, isInteractive, onExit, onNavigate }: CreateS
...
+  const [envWizardError, setEnvWizardError] = useState<string | null>(null);
+
+  // Kick the wizard off once scaffolding succeeds (interactive only).
+  useEffect(() => {
+    if (!allSuccess || envWizardStage !== 'idle' || !isInteractive) return;
+    let cancelled = false;
+    void (async () => {
+      try {
+        const configRoot = findConfigRoot(projectRoot);
+        if (!configRoot) {
```
notgitika, line 234: **HIGH: The environment wizard fires on `allSuccess` but there's no guard against it running when `CreateScreen` re-renders after the project scaffold completes AND the env wizard is mid-flight.**

More critically — where does this wizard actually *render*? The `envWizardStage` state machine is defined and transitions happen, but I don't see `<EnvironmentStep>` or `<AssignTargetsPanel>` rendered in the JSX return below. The state machine runs and calls `writeAwsTargetsFull`, but the user never sees the interactive wizard UI.

If these components ARE rendered further down the file (past what I can see in the diff), please confirm. Otherwise this is the dead-code issue flagged earlier — the logic exists but the UI is unreachable.

src/cli/commands/status/action.ts
```diff
@@ -553,3 +562,118 @@ export async function handleRuntimeLookup(
...
+  error?: string;
+}
+
+export type StackInfoFetcher = (region: string, stackName: string) => Promise<{ status?: string; lastUpdated?: Date }>;
+
+/**
+ * Default stack-info fetcher: DescribeStacks via SDK. Returns undefined fields
+ * when the stack does not exist or the call fails (caller decides how to display).
+ */
+export const defaultStackInfoFetcher: StackInfoFetcher = async (region, stackName) => {
```
notgitika, line 590: **MINOR: `defaultStackInfoFetcher` creates a new `CloudFormationClient` per target.**

For environments with multiple targets in the same region, this creates redundant clients. Consider caching by region:

```ts
const cfnClients = new Map<string, CloudFormationClient>();
function getCfnClient(region: string) {
  if (!cfnClients.has(region)) cfnClients.set(region, new CloudFormationClient({ region, credentials: getCredentialProvider() }));
  return cfnClients.get(region)!;
}
```

Also: the bare `catch {}` swallows all errors silently. Consider logging at debug level so users troubleshooting can discover auth/network failures via `--verbose`.

notgitika (changes requested):
Good architecture overall — the schema → resolution → orchestrator → command decomposition is clean, and the `deployToTargets` orchestrator (never throws, returns aggregate) is the right pattern. The environment.ts + mergeOverrides are well-factored with proper immutability. Tests use real temp dirs and DI rather than over-mocking which is great.

However, there are several blocking and high-severity issues that need addressing before this can merge. See inline comments.

ChrisTitos:
This is exactly something we need. It seems the work has stalled on this since the PR was opened. Any updates?

PR #1213, opened 2026-05-12
sync-preview: merge main into preview

notgitika:
## Summary

Syncs 54 commits from main into preview, resolving all conflicts.

### Resolution decisions

- **Telemetry**: Adopted main's refactor (`runCliCommand` + `withCommandRunTelemetry`), updated preview callers
- **Preview-specific features preserved**: harness system, tool commands (`registerAddTool`/`registerRemoveTool`), invoke harness flow, deploy `onDeployMessage` support
- **Version**: Kept `1.0.0-preview.8`
- **Schema/Changelog**: Kept preview's `schemas/agentcore.schema.v1.json` and `CHANGELOG.md`
- **Security**: Added `protobufjs@8.0.3` override (fixes vuln from `@opentelemetry/exporter-metrics-otlp-http@^0.217.0`), retained `fast-xml-parser` and `@aws-sdk/xml-builder` overrides

### Verification

- `npm run typecheck` — passes (0 errors)
- `npm audit --audit-level=high --omit=dev` — passes (0 high, 3 moderate)

## Test plan

- [ ] CI passes (build, typecheck, lint, security, unit tests)
- [ ] Harness invoke flow still works on preview
- [ ] Remove flow supports all resource types including harness

PR #1225, opened 2026-05-13
chore: record merge point with main

notgitika:
Records git merge ancestry so the sync-preview workflow knows main's commits are already incorporated (via squash-merged #1213). No file changes this shoold only updates the merge graph.


notgitika:
needs to be "create a merge commit" not squash commit

Hweinstock (approved):

PR #1269, opened 2026-05-15
fix(ci): persist-credentials: false on remaining workflows

jpr5:
## Summary
- **release.yml**: `persist-credentials: false` on `prepare-release` + `publish-npm` checkout steps; explicit `insteadOf` credential config before every `git push` so pushes still work without the persisted token
- **agent-restricted.yml**: `persist-credentials: false` on checkout (strands-action manages its own credentials via `pat_token` input)
- **strands-command.yml**: `persist-credentials: false` on checkout (same pattern as agent-restricted)

No changes to `test-and-build` job (already had `persist-credentials: false`) or `zizmor.yml` (no artipacked suppressions for these files).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -27,14 +27,35 @@ updates:
       prefix: 'chore'
       include: 'scope'
     open-pull-requests-limit: 10
+    cooldown:
+      default-days: 1
 
   - package-ecosystem: 'github-actions'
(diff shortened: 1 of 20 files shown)


jpr5:
Opened against the wrong repo by mistake. Recreating on CopilotKit/agentcore-cli.

PR #1284, opened 2026-05-18
feat(issue-1236): enhance docker build with additional build configuration

### LINKED ISSUE #1236: feat(config): support shared Dockerfile via buildContextPath and customDockerBuildArgs in agentcore.json
## Description

Currently, declaring multiple runtimes in `agentcore.json` effectively requires one `Dockerfile` per agent (via the `dockerfile` key). This forces users to maintain several nearly-identical Dockerfiles when the build logic is shared.

**Current config example:**
```json
{
  "name": "dummyagent",
  "build": "Container",
  "entrypoint": "dist/main.js",
  "codeLocation": ".",
  "dockerfile": "Dockerfile.dummy-agent",
  "networkMode": "PUBLIC",
  "protocol": "HTTP"
},
{
  "name": "generalchat",
  "build": "Container",
  "entrypoint": "dist/main.js",
  "codeLocation": ".",
  "dockerfile": "Dockerfile.general-chat",
  "networkMode": "PUBLIC",
  "protocol": "HTTP"
}
```

Two additions to the runtime config schema would enable sharing a single Dockerfile across all agents:

### 1. `buildContextPath` — explicit Docker build context

Add a `buildContextPath` key to specify the Docker build context directory independently from `codeLocation`. This mirrors the standard `docker build -f <dockerfile> <context>` syntax and gives users fine-grained control over what is sent to the daemon.

### 2. `customDockerBuildArgs` and/or automatic `AGENT_NAME` build arg

Either (or both):

- **Automatic injection**: always pass `AGENT_NAME=<runtime.name>` as a `--build-arg` so the Dockerfile can branch on it without any extra config.
- **Declarative args**: allow a `customDockerBuildArgs` map in the runtime config that is forwarded verbatim as `--build-arg` flags to the Docker build command.

**Proposed config with both additions:**
```json
{
  "name": "dummyagent",
  "build": "Container",
  "entrypoint": "dist/main.js",
  "codeLocation": ".",
  "dockerfile": "Dockerfile",
  "buildContextPath": ".",
  "customDockerBuildArgs": { "AGENT_NAME": "dummyagent" },
  "networkMode": "PUBLIC",
  "protocol": "HTTP"
}
```

Or, with automatic `AGENT_NAME` injection, the Dockerfile side would look like:
```dockerfile
ARG AGENT_NAME
# ... shared build steps ...
COPY dist/${AGENT_NAME} ./app
```

## Acceptance Criteria

- [ ] A `buildContextPath` key is recognised in a runtime entry and used as the Docker build context (equivalent to the positional argument in `docker build -f <file> <context>`)
- [ ] At least one of the following is supported:
  - `customDockerBuildArgs` map in the runtime config, forwarded as `--build-arg` flags
  - Automatic injection of `AGENT_NAME` (set to `runtime.name`) as a build arg
- [ ] Multiple runtimes can share a single `Dockerfile`, each producing a different image based on build args
- [ ] Existing configs without these new keys are unaffected (fully backwards-compatible)

## Additional Context

This is particularly useful in monorepo setups where several agents share the same base image and build process, differing only in their entry point or bundled code. Without this, every new agent forces a new Dockerfile, which creates duplication and maintenance overhead.

EdouardCourty:
## Description

Add support for two new options in the runtime configurations for Docker-based builds :
- change build context path with `buildContextPath`
- custom docker build arguments with `customDockerBuildArgs`

Both are optional and this change is backward-compatible.

## Related Issue

Closes #1236 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.


aidandaly24:
Hey @EdouardCourty after reviewing it seems the above comments are true regarding this feature not being honored by `agentcore deploy`. If you make these suggested updates we'd love to get this merged in. 

I can also go ahead and make the necessary updates if there is no activity on this PR by 06/05/2026 or with your explicit permission. Thank you!

EdouardCourty:
Hey @aidandaly24 if you can make the needed changes to `aws/agentcore-l3-cdk-constructs`, go ahead!

I could not find the repository myself as it does not seem to be public (the NPM package page has a GitHub link which shows a 404...).

nxf5025:
@aidandaly24 - I've been successfully running this branch as a fork of the CLI specifically for the `agentcore dev` functionality. Since the other projects are not public does it make sense to get this out for now and the `agentcore deploy` can come later? Just brainstorming here as it seems we are stuck.

tejaskash:
@nxf5025 We discussed this, and @aidandaly24 will make the changes to L3 CDK.

aidandaly24:
Hey @nxf5025 @EdouardCourty, going to get this reviewed and it will be released this Thursday. Sorry for the delay!

src/cli/operations/dev/container-dev-server.ts
```diff
@@ -65,9 +66,21 @@ export class ContainerDevServer extends DevServer {
     }
     this.runtimeBinary = runtime.binary;
 
-    // 2. Verify Dockerfile exists
+    // 2. Verify Dockerfile exists (resolved relative to the build context, matching deploy)
+    const buildContext = this.config.buildContextPath ?? this.config.directory;
+    // When the context is widened via buildContextPath, ensure a .dockerignore keeps secrets/junk out
+    // of the local image — the same file the deploy (CodeBuild) path honors, so both stay consistent.
+    if (this.config.buildContextPath) {
+      const created = ensureBuildContextDockerignore(buildContext);
```
tejaskash, line 82: **[Blocker — correctness] `dev` crashes instead of reporting a config error.**

`ensureBuildContextDockerignore(buildContext)` runs here inside `prepare()`, *before* the graceful missing-Dockerfile guard below, and calls `writeFileSync` with no try/catch. A `buildContextPath` that resolves to a missing or read-only directory throws ENOENT/EACCES, so `prepare()` never returns `false` — the throw propagates through `await this.prepare()`. In the TUI, `useDevServer` fires `void startServer()` with no catch around `await server.start()`, so this becomes an **unhandled rejection that crashes the dev process**; the web-ui path degrades to a generic 500 instead of the `{success:false}` handling the `if (!child)` block provides.

Shares a root cause with the preflight and packager comments: make `ensureBuildContextDockerignore` guard the directory's existence and run it *after* validation, via a shared notification callback.

src/cli/operations/deploy/preflight.ts
```diff
@@ -208,6 +217,20 @@ export function validateContainerAgents(projectSpec: AgentCoreProjectSpec, confi
       } else {
         warnDeprecatedBaseImage(dockerfilePath, agent.name);
       }
+      // buildContextPath widens the docker build context (e.g. the whole repo). Ensure a
+      // .dockerignore exists at that root so secrets/junk (.env, .git, agentcore/) are never baked
+      // into the image or uploaded to CodeBuild. Both local and deploy honor this same file; it is
+      // created only when absent (never overwrites the user's).
+      if (agent.buildContextPath) {
+        const created = ensureBuildContextDockerignore(buildContext);
```
tejaskash, line 224: **[Blocker — correctness] Preflight masks its own friendly error and writes into the repo on a failing deploy.**

`ensureBuildContextDockerignore` guards only the `.dockerignore` file's existence, never the directory. With `buildContextPath: './typo-dir'` (missing), the loop first pushes a clean `"Dockerfile not found"` into `errors[]`, but this call then `writeFileSync`s into the missing dir and throws raw ENOENT **before** the aggregated `throw` — so the user sees a cryptic fs error instead of the actionable one. When the dir exists but the Dockerfile is missing, it still leaves a stray `.dockerignore` in the repo even though the deploy aborts (dirtying CI/git checkouts).

Note the inconsistency with `ContainerPackager.pack()`, which does the Dockerfile `existsSync` check *after* this write; preflight does it before. Fix both by guarding the dir and deferring the write until validation passes.

src/lib/packaging/container.ts
```diff
@@ -35,7 +36,16 @@ export class ContainerPackager implements RuntimePackager {
...
     const configBaseDir = options.artifactDir ?? options.projectRoot ?? process.cwd();
     const codeLocation = resolveCodeLocation(spec.codeLocation, configBaseDir);
-    const dockerfilePath = getDockerfilePath(codeLocation, spec.dockerfile);
+    const buildContext = spec.buildContextPath
+      ? resolveCodeLocation(spec.buildContextPath, configBaseDir)
+      : codeLocation;
+    // When the context is widened via buildContextPath, ensure a .dockerignore keeps secrets/junk out
+    // of the local image — the same file the deploy (CodeBuild) path honors, so both stay consistent.
+    if (spec.buildContextPath) {
+      ensureBuildContextDockerignore(buildContext);
```
tejaskash: **[Blocker — correctness] `package` throws an untyped fs error and leaves a stray `.dockerignore`.**

`pack()` calls `ensureBuildContextDockerignore` here, *before* the `existsSync(dockerfilePath)` check that used to be its first guard. A bad `buildContextPath` throws a bare `Error` (ENOENT), bypassing the `PackagingError` path — telemetry's `classifyError` only maps `BaseError`, so it lands in `UnknownError` and the user sees a cryptic message. With a valid dir but a missing Dockerfile, the `.dockerignore` is written into the repo before `pack()` rejects.

Same root cause as the dev-server and preflight comments. Also: preflight validates the Dockerfile *before* this write, so the two callers are inconsistent — centralize the ordering.

src/lib/packaging/build-context-dockerignore.ts
```diff
@@ -0,0 +1,37 @@
...
+ * is set and none exists. Keeps secrets and build junk out of the image and out of the CodeBuild
+ * upload. Both local `agentcore dev`/`package` and `agentcore deploy` honor this same file, so what
+ * is excluded locally is exactly what is excluded on deploy.
+ */
+export const BUILD_CONTEXT_DOCKERIGNORE_TEMPLATE = `# Generated by agentcore because this agent sets buildContextPath — the whole of this directory is the
+# Docker build context, so secrets/junk here would otherwise be baked into the image and uploaded to
+# CodeBuild. Both 'agentcore dev'/'package' and 'agentcore deploy' honor this file. Edit freely; to
+# intentionally include a file listed here (e.g. a non-secret .env.production), remove its line.
+.env
+.env.*
```
tejaskash, line 18: **[Blocker — correctness] Generated `.dockerignore` doesn't exclude nested `node_modules`/secrets in the monorepo case.**

These patterns (`node_modules/`, `.venv/`, `__pycache__/`, `.env`, ...) have no `**/` prefix. Docker matches `.dockerignore` root-relative via Go `filepath.Match`, so `node_modules/` matches only the top-level directory. In the feature's headline case (`buildContextPath: '.'`), nested `app/agent-one/node_modules` and per-package `.env` files are **still uploaded to the Docker context and CodeBuild** — directly contradicting the module/inline comments that this keeps secrets and build junk out of the upload. A leaked nested `.env` is security-adjacent, not just a size regression.

Use `**/node_modules`, `**/.env`, `**/__pycache__`, etc.

src/lib/constants.ts
```diff
@@ -51,14 +51,20 @@ export type ContainerRuntime = 'docker' | 'podman' | 'finch';
...
+ * `dockerfile` may be a filename ('Dockerfile') or a forward-slash relative subpath
+ * ('docker/Dockerfile'); absolute paths, backslashes, and `..` traversal are rejected so it can
+ * never escape the build context.
  */
-export function getDockerfilePath(codeLocation: string, dockerfile?: string): string {
+export function getDockerfilePath(buildContext: string, dockerfile?: string): string {
   const name = dockerfile ?? DOCKERFILE_NAME;
-  if (name.includes('/') || name.includes('\\') || name.includes('..')) {
-    throw new Error(`Invalid dockerfile name: must be a filename without path separators or traversal`);
+  if (name.startsWith('/') || name.includes('\\') || name.split('/').includes('..')) {
```
tejaskash: **[Cleanup — reuse/altitude] Dockerfile-path validation is duplicated and already divergent.**

The same "relative subpath, no leading slash, no backslash, no `..`" rule is now maintained in two places this PR edits in lockstep: this runtime guard in `getDockerfilePath` and the Zod `DockerfilePathSchema` regex+refine in `agent-env.ts`. They **already disagree** — the schema's `^[a-zA-Z0-9]` anchor rejects `.docker/Dockerfile`, but this guard accepts it (no leading `/`, no `\`, no `..` segment). Extract a single `isValidDockerfilePath` predicate used by both. Non-blocking, but worth doing while this code is open since the divergence is live.

src/lib/packaging/container.ts
```diff
@@ -35,7 +36,16 @@ export class ContainerPackager implements RuntimePackager {
     const agentName = options.agentName ?? spec.name;
     const configBaseDir = options.artifactDir ?? options.projectRoot ?? process.cwd();
     const codeLocation = resolveCodeLocation(spec.codeLocation, configBaseDir);
-    const dockerfilePath = getDockerfilePath(codeLocation, spec.dockerfile);
+    const buildContext = spec.buildContextPath
```
tejaskash: **[Cleanup — altitude] Build-context resolution re-derived at 4 sites with two different resolvers.**

`buildContextPath ?? codeLocation` (then resolve) is re-implemented here, in `preflight.ts` (~line 210), and in `container-dev-server.ts` (~line 70) — while `dev/config.ts` resolves via a *different* private function (`resolveCodeDirectory`, which strips trailing slashes and uses `join`) than deploy/package (`resolveCodeLocation`, which uses `resolve`). The PR's own comments insist all paths must resolve context identically, so this is a latent local-vs-deploy divergence: a future path-normalization tweak applied to one resolver silently makes `agentcore dev` build from a different context than deploy. Collapse into one exported `resolveBuildContext(spec, baseDir)`. Non-blocking.

src/schema/schemas/agent-env.ts
```diff
@@ -94,6 +94,81 @@ export const EntrypointSchema = z
...
+
+/**
+ * Build-arg name. Must be a valid identifier because on deploy each arg is forwarded to CodeBuild as
+ * an environment variable (values are inherited from the environment, never interpolated onto the
+ * shell command line) and consumed by Dockerfile `ARG` directives, which are themselves identifiers.
+ */
+const BuildArgKeySchema = z
+  .string()
+  .regex(
+    /^[A-Za-z_][A-Za-z0-9_]*$/,
```
tejaskash: **[Cleanup — reuse] `BuildArgKeySchema` re-declares `EnvVarNameSchema`'s regex.**

This regex `/^[A-Za-z_][A-Za-z0-9_]*$/` is byte-identical to the exported `EnvVarNameSchema` above. Build args *become* CodeBuild env-var names, so reuse `EnvVarNameSchema` (with a custom message) — nothing is lost, since the `RESERVED_BUILD_ARG_KEYS` check lives separately in the superRefine. Non-blocking.

src/schema/schemas/agent-env.ts
```diff
@@ -94,6 +94,81 @@ export const EntrypointSchema = z
...
+ * (`buildContextPath` when set, otherwise `codeLocation`). Accepts a filename ('Dockerfile',
+ * 'Dockerfile.gpu') or a forward-slash relative subpath ('docker/Dockerfile'). Absolute paths
+ * and '..' traversal are rejected so the Dockerfile can never escape the build context.
+ */
+const DockerfilePathSchema = z
+  .string()
+  .min(1)
+  .max(255)
+  .regex(
+    /^[a-zA-Z0-9][a-zA-Z0-9._/-]*$/,
```
tejaskash: **[Cleanup — low severity] Broadened dockerfile regex accepts a directory-shaped value.**

Widening the regex to allow `/` for subpaths also lets a trailing-slash directory value like `docker/` through (`docker/Dockerfile/` and `a//b` are caught cleanly by `existsSync`/`join` normalization, but a value naming a real directory is not). That defers failure to `docker build -f <dir>` with a confusing error instead of rejecting at config-validation time. Low severity; a `.refine` rejecting a trailing slash closes it.

tejaskash (commented):
Automated review of the container-build enhancements. 8 findings below — 4 are correctness blockers (all but one share a single root cause), the rest are cleanup.

tejaskash (changes requested):
Requesting changes on the 4 correctness blockers flagged in the inline comments above (dev-server crash, preflight error-masking + repo write, packager untyped error + stray file, and the nested `.dockerignore` secret/junk leak). Findings 1–3 share one root cause: guard the directory in `ensureBuildContextDockerignore` and run it after validation via a shared notification callback; finding 4 is a `**/`-prefix fix on the template. The 4 cleanup items are non-blocking. (Refuted candidates dropped; add-agent wizard subpath conflict is pre-existing → follow-up ticket.)

tejaskash (approved):

PR #1285, opened 2026-05-18
ci: add Claude Code /security-review workflow on PRs

tejaskash:
## Summary
- Adds `.github/workflows/pr-security-review.yml` that runs `anthropics/claude-code-action@v1` with the built-in `/security-review` slash command on every PR (`pull_request_target`: opened/reopened/synchronize) and via `workflow_dispatch`.
- Authenticates to Bedrock 

Mirrors the auth gate from `pr-ai-review.yml`: PRs from non-collaborators are skipped to prevent prompt-injection-driven token abuse.

Model: `us.anthropic.claude-opus-4-7`, max-turns 20, region `us-west-2`.

## Test plan
- [x] Merge this PR, then open a follow-up PR — `pull_request_target` reads workflows from `main`, so the workflow only takes effect once merged.
- [x] Verify the `claude-security-reviewing` label appears, the workflow run completes, and a security review comment is posted by the bot.
- [x] Confirm OIDC role assumption succeeds in the workflow logs.

### DIFF
--- a/.github/workflows/pr-security-review.yml
+++ b/.github/workflows/pr-security-review.yml
@@ -2,7 +2,9 @@ name: Claude Security Review
 
 on:
   pull_request_target:
-    types: [opened, reopened, labeled]
+    types: [opened, reopened]
+  pull_request_review:
+    types: [submitted]
   workflow_dispatch:
     inputs:
       pr_number:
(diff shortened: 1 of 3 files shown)


tejaskash (commented):
Addressed in 0b32025b66d7e6eea7a695a748e1ea3353c28264:

1. **synchronize trigger** — dropped from `pull_request_target` types so the workflow now only fires on opened/reopened/labeled, matching pr-ai-review.yml.
2. **Script injection on `pr_number`** — moved `inputs.pr_number` into `env:` and read it via `process.env.PR_NUMBER_INPUT` with explicit base 10. Also applied the same env-var pattern to `steps.pr.outputs.number` in the two label-management github-script steps as defense in depth.
3. **timeout-minutes** — added `timeout-minutes: 30` to the security-review job.

Also, per follow-up request: updated the prompt to instruct Claude to post inline line-level review comments instead of a single summary comment (with a fallback top-level comment if there are no findings).

.github/workflows/pr-security-review.yml
notgitika, line 1: Can we have this workflow run for non-maintainers after someone approves the PR and before the merge? 
Instead of manually adding the label?

tejaskash: adressed

notgitika (dismissed):

notgitika (approved):

PR #1287, opened 2026-05-18
fix(invoke): preserve model maxTokens when overriding model at harness invoke

### LINKED ISSUE #1290: Harness invoke with --model-id drops model inference parameters (temperature, topP, maxTokens)
## Problem

When invoking a harness agent with `--model-id` to override the model at invoke time, all model inference parameters (temperature, topP, topK, maxTokens) configured in `harness.json` are dropped from the request. The server treats the invoke-time model override as a full replacement of the deployed model config, so omitting these fields results in unbounded/default inference behavior.

**Example:** A harness configured with `maxTokens: 500` will ignore that limit when invoked with `--model-id anthropic.claude-v3`, producing responses far exceeding 500 tokens.

## Root Cause

`buildHarnessBaseOpts` in the invoke command constructed the model override object with only the `modelId` (and `apiKeyArn` for non-Bedrock providers), without forwarding the inference parameters from the harness spec.

## Impact

- Affects all provider types (Bedrock, OpenAI, Gemini)
- Only triggered when using `--model-id` at invoke time
- Deploy-time configuration is unaffected (the parameters are correctly stored on the server)

## Fix

PR #1287 — forwards inference parameters (temperature, topP, topK, maxTokens) from the harness spec when constructing the model override, and uses CLI-provided values (e.g. `--api-key-arn`) as precedence overrides.

## Affected Versions

Preview branch only (harness feature has not shipped to mainline).

jesseturner21:
## Summary

- Fix `buildHarnessBaseOpts` to forward model inference parameters (temperature, topP, topK, maxTokens) from the harness spec when constructing invoke-time model overrides
- Import `HarnessModel` type from schema instead of maintaining a local duplicate interface
- Add unit tests for `buildHarnessBaseOpts` covering all provider branches, param omission, and CLI precedence

Fixes #1290

## Test plan

- [x] Unit tests added for all 3 providers (bedrock, open_ai, gemini) with inference params
- [x] Unit tests for omission of undefined params
- [x] Unit test for CLI apiKeyArn precedence over harness spec
- [x] Unit test for harness-level execution limits (maxTokens, maxIterations, timeoutSeconds)
- [x] Manual end-to-end: deployed harness with maxTokens:500, invoked with --model-id, confirmed `stopReason: "max_tokens"` and `outputTokens: 500`

### DIFF
--- a/src/cli/aws/agentcore-harness.ts
+++ b/src/cli/aws/agentcore-harness.ts
@@ -17,10 +17,34 @@ import { randomUUID } from 'node:crypto';
 
 export type HarnessStatus = 'CREATING' | 'READY' | 'UPDATING' | 'DELETING' | 'DELETED' | 'FAILED';
 
+export interface BedrockModelConfig {
+  modelId: string;
+  temperature?: number;
+  topP?: number;
+  maxTokens?: number;
+}
+
+export interface OpenAiModelConfig {
+  modelId: string;
+  apiKeyArn?: string;
+  temperature?: number;
+  topP?: number;
+  maxTokens?: number;
+}
+
+export interface GeminiModelConfig {
+  modelId: string;
+  apiKeyArn?: string;
+  temperature?: number;
+  topP?: number;
+  topK?: number;
+  maxTokens?: number;
+}
+
 export interface HarnessModelConfiguration {
-  bedrockModelConfig?: { modelId: string };
-  openAiModelConfig?: { modelId: string; apiKeyArn?: string };
-  geminiModelConfig?: { modelId: string; apiKeyArn?: string };
+  bedrockModelConfig?: BedrockModelConfig;
+  openAiModelConfig?: OpenAiModelConfig;
+  geminiModelConfig?: GeminiModelConfig;
 }
 
 export type HarnessSystemPrompt = { text: string }[];
--- a/src/cli/commands/invoke/action.ts
+++ b/src/cli/commands/invoke/action.ts
@@ -516,6 +516,10 @@ interface HarnessModel {
   provider?: string;
   modelId?: string;
   apiKeyArn?: string;
+  temperature?: number;
+  topP?: number;
+  topK?: number;
+  maxTokens?: number;
 }
 
 function buildHarnessBaseOpts(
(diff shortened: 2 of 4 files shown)


src/cli/commands/invoke/action.ts
```diff
@@ -527,15 +531,43 @@ function buildHarnessBaseOpts(
...
+        baseOpts.model = {
+          bedrockModelConfig: {
+            modelId,
+            ...(temperature !== undefined && { temperature }),
+            ...(topP !== undefined && { topP }),
+            ...(modelMaxTokens !== undefined && { maxTokens: modelMaxTokens }),
+          },
+        };
         break;
     }
```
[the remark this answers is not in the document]
jesseturner21: added test

src/cli/commands/invoke/action.ts
```diff
@@ -516,6 +516,10 @@ interface HarnessModel {
   provider?: string;
   modelId?: string;
   apiKeyArn?: string;
+  temperature?: number;
+  topP?: number;
+  topK?: number;
+  maxTokens?: number;
 }
```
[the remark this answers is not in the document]
jesseturner21: added

src/cli/aws/agentcore-harness.ts
```diff
@@ -17,10 +17,34 @@ import { randomUUID } from 'node:crypto';
 
 export type HarnessStatus = 'CREATING' | 'READY' | 'UPDATING' | 'DELETING' | 'DELETED' | 'FAILED';
 
+export interface BedrockModelConfig {
```
Hweinstock, line 20: was going to ask why these types don't come for free by using the AWS SDK, but noticed we aren't using it. 

Since harness is GA should we migrate off of https://github.com/aws/agentcore-cli/blob/preview/src/cli/aws/api-client.ts, and use a typed http handler?

src/cli/aws/agentcore-harness.ts
```diff
@@ -17,10 +17,34 @@ import { randomUUID } from 'node:crypto';
...
+
+export interface GeminiModelConfig {
+  modelId: string;
+  apiKeyArn?: string;
+  temperature?: number;
+  topP?: number;
+  topK?: number;
+  maxTokens?: number;
+}
+
```
tejaskash, line 43: The three new interfaces here (`BedrockModelConfig`, `OpenAiModelConfig`, `GeminiModelConfig`) duplicate fields already defined on `HarnessModel` in `src/schema/schemas/primitives/harness.ts`. The bug this PR fixes was caused by exactly this kind of drift — the local `HarnessModel` interface in `action.ts` was missing `temperature`/`topP`/`topK`/`maxTokens`, so model overrides silently dropped them.

Consider deriving these wire types from the schema, e.g.:

```ts
type BedrockModelConfig = Pick<HarnessModel, 'modelId' | 'temperature' | 'topP' | 'maxTokens'>;
```

That way the next inference param added to the schema (e.g. `stopSequences`, `presencePenalty`) won't need to be added in two places and won't recreate this bug.

src/cli/aws/agentcore-harness.ts
jesseturner21: Good call, but migrating `agentcore-harness.ts` off the hand-rolled `api-client.ts` HTTP calls to use the SDK commands directly would eliminate these local type definitions entirely and keep us in sync with the API automatically. It's a bigger refactor so i filed a follow-up issue for it rather than bundling it here.

 https://github.com/aws/agentcore-cli/issues/1295

src/cli/aws/agentcore-harness.ts
jesseturner21: Agree but better fix here is Harrison's suggestion above which are generated from the service model and stay in sync automatically and that eliminates both the local interfaces. I'll file a follow-up for that migration.

Hweinstock (approved):

PR #1288, opened 2026-05-18
fix: align harness memory config with Smithy API model

### LINKED ISSUE #1291: Harness memory not working when memory ARN is configured
## Problem

When deploying a harness with memory configured (either via `memory.name` referencing a deployed memory resource, or via `memory.arn` with a direct ARN), the memory does not function at invocation time. The model reports "I don't have the ability to permanently save or remember information" and no memory tools are available.

## Root Cause

The CLI was sending the wrong JSON structure for the `memory` field in the CreateHarness/UpdateHarness API calls.

**What the CLI was sending:**
```json
{ "memory": { "memoryArn": "arn:aws:bedrock-agentcore:..." } }
```

**What the API expects (per Smithy model):**
```json
{ "memory": { "agentCoreMemoryConfiguration": { "arn": "arn:aws:bedrock-agentcore:..." } } }
```

The API silently accepted the incorrect structure and stored `"memory": {}`, resulting in deployed harnesses having no memory capabilities despite the configuration appearing correct in `deployed-state.json`.

## Impact

All harness deployments with memory configured on the `preview` branch are affected. Memory appears deployed (shows in `agentcore status`) but does not function at runtime.

## Workaround

None — requires the CLI fix.

## Fix

PR #1288 — aligns the CLI's `HarnessMemoryConfiguration` type and `mapMemory()` function with the Smithy API model.

## How to verify after fix

1. Deploy a harness with memory: `agentcore create --memory longAndShortTerm --defaults`
2. Call GetHarness and confirm the response contains:
   ```json
   "memory": { "agentCoreMemoryConfiguration": { "arn": "arn:..." } }
   ```
   (instead of `"memory": {}`)
3. Invoke the harness and confirm the model has memory tool access

jesseturner21:
## Summary

- Fixes harness memory not working when a memory ARN is configured in `harness.json`
- The Harness API expects memory as a Smithy union type `{ agentCoreMemoryConfiguration: { arn: "..." } }` but the CLI was sending `{ memoryArn: "..." }`
- The API silently accepted the wrong structure and stored `"memory": {}`, resulting in deployed harnesses having no memory capabilities

Closes #1291

## Changes

- `src/cli/aws/agentcore-harness.ts`: Updated `HarnessMemoryConfiguration` interface to match Smithy model with `agentCoreMemoryConfiguration` union wrapper
- `src/cli/operations/deploy/imperative/deployers/harness-mapper.ts`: Updated `mapMemory()` to return correct nested structure
- `src/cli/operations/deploy/imperative/deployers/harness-deployer.ts`: Updated ARN extraction from new structure

## Test plan

- [x] `npm test -- "harness-mapper"` — 38 tests pass
- [x] `npm test -- "harness-deployer"` — 22 tests pass
- [x] `npm test -- "agentcore-harness"` — 16 tests pass
- [x] `npx tsc --noEmit` — no type errors
- [x] E2E: deployed harness with memory, verified `GetHarness` returns `{ agentCoreMemoryConfiguration: { arn: "..." } }` instead of `{}`

### DIFF
--- a/src/cli/aws/agentcore-harness.ts
+++ b/src/cli/aws/agentcore-harness.ts
@@ -37,8 +37,15 @@ export interface HarnessSkill {
   path: string;
 }
 
+export interface HarnessAgentCoreMemoryConfiguration {
+  arn: string;
+  actorId?: string;
+  messagesCount?: number;
+  retrievalConfig?: Record<string, { topK?: number; relevanceScore?: number }>;
+}
+
 export interface HarnessMemoryConfiguration {
-  memoryArn?: string;
+  agentCoreMemoryConfiguration: HarnessAgentCoreMemoryConfiguration;
 }
 
 export interface HarnessTruncationConfiguration {
--- a/src/cli/aws/agentcore-harness.ts
+++ b/src/cli/aws/agentcore-harness.ts
@@ -41,7 +41,7 @@ export interface HarnessAgentCoreMemoryConfiguration {
   arn: string;
   actorId?: string;
   messagesCount?: number;
-  retrievalConfig?: Record<string, { topK?: number; relevanceScore?: number }>;
+  retrievalConfig?: Record<string, { topK?: number; relevanceScore?: number; strategyId?: string }>;
 }
 
 export interface HarnessMemoryConfiguration {
--- a/src/cli/operations/deploy/imperative/deployers/__tests__/harness-mapper.test.ts
+++ b/src/cli/operations/deploy/imperative/deployers/__tests__/harness-mapper.test.ts
@@ -380,7 +380,9 @@ describe('mapHarnessSpecToCreateOptions', () => {
       });
 
       expect(result.memory).toEqual({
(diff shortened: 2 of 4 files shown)


src/cli/aws/agentcore-harness.ts
```diff
@@ -37,8 +37,15 @@ export interface HarnessSkill {
   path: string;
 }
 
+export interface HarnessAgentCoreMemoryConfiguration {
+  arn: string;
+  actorId?: string;
+  messagesCount?: number;
+  retrievalConfig?: Record<string, { topK?: number; relevanceScore?: number; strategyId?: string }>;
```
notgitika, line 44: these are in the interface but not populated. would be dropped if user goes in to add these fields right?

jesseturner21: Good catch. `messagesCount` and `retrievalConfig` are not yet in the harness.json schema (`HarnessMemoryRefSchema`), so users can't set them today — they won't be silently dropped because there's no input path for them. The interface matches the full Smithy model so these fields are ready when we add schema support for them.

notgitika (approved):
LGTM thank you for fixing this!

PR #1296, opened 2026-05-18
docs: sync CLI documentation with actual command surface

arepageek:
## Description

This PR brings the CLI documentation in line with the actual CLI surface. It does three things:

1. **Documents 5 stable commands that were not in `docs/commands.md` at all** — `import` (with all 5 subcommands), `add policy-engine`, `add policy`, `add runtime-endpoint`, and `telemetry` (with `enable`/`disable`/`status`).
2. **Adds missing flags to commands that were already documented but out of date** — `create`, `add agent`, `add evaluator`, `add online-eval`, `dev`, and `run eval`.
3. **Fixes incorrect command names in `docs/PERMISSIONS.md`** — `run batch-eval` and `run recommend` were used where the actual CLI commands are `run batch-evaluation` and `run recommendation`.

After these changes, **100% of stable CLI commands are documented or mentioned** in `docs/` or `README.md` (preview commands are intentionally not yet documented).

### Files changed

- `docs/commands.md` — most of the work (new sections + flag updates)
- `docs/evals.md` — kept in sync with `commands.md` for the `add evaluator`, `run eval`, and `add online-eval` tables
- `docs/PERMISSIONS.md` — command name corrections

### `docs/commands.md` — new sections

- `### import` — top-level flags + 5 subcommands (`runtime`, `memory`, `evaluator`, `online-eval`, `gateway`) with examples and per-subcommand flag tables
- `### add policy-engine` — Cedar policy engine creation, attach modes
- `### add policy` — Cedar policy authoring (file, inline statement, `--generate` from natural language)
- `### add runtime-endpoint` — named endpoints (version aliases) for runtimes
- `### telemetry` — anonymous usage analytics: `enable`, `disable`, `status`

### `docs/commands.md` — flag updates to existing sections

- `create`: clarified `--name` (it's the agent name, also used as project dir if `--project-name` is omitted), added `--project-name`, `--session-storage-mount-path`, `--with-config-bundle`, added `AGUI` to `--protocol`
- `add agent`: added `AGUI` to `--protocol`, `--request-header-allowlist`, `--session-storage-mount-path`, `--with-config-bundle`
- `add evaluator`: refactored to support both `llm-as-a-judge` (default) and `code-based` evaluator types via `--type`, `--lambda-arn`, `--timeout`, `--kms-key-arn`
- `add online-eval`: added `--endpoint <name>`
- `dev`: added `-b, --no-browser` and `--no-traces`
[...]

PR #1297, opened 2026-05-18
ci(security-review): add safe-to-review label as alternate trigger

tejaskash:
## Summary

Re-add the `labeled` event to the security-review workflow so maintainers can kick the review on a community PR by applying the `safe-to-review` label, in addition to the existing approving-review trigger.

Two paths for community PRs now:
- Submit an approving review → workflow runs (gates on the approver's write access).
- Apply `safe-to-review` label → workflow runs (gates on the labeler's write access).

Maintainer-authored PRs still auto-run on `opened`/`reopened`/`synchronize` as before.

The job-level `if:` filters labeled events to only `safe-to-review` so other label churn (e.g. `size/m` from the size-labeler) doesn't spawn API calls.

## Why

PR #1296 surfaced this gap — a maintainer applied a label expecting the review to fire, and it didn't. Approval-only is too coarse a gate when you want a security review *before* committing to an approval. The label flow gives maintainers a lighter-weight kick without endorsing the change.

## Test plan

- [ ] Once merged, on PR #1296 (or any open community PR) apply `safe-to-review` and confirm the workflow fires under the labeler's identity.
- [ ] Apply an unrelated label (e.g. `size/m`) and confirm the workflow does NOT fire.
- [ ] Confirm the existing approving-review path still works.

### DIFF
--- a/.github/workflows/pr-security-review.yml
+++ b/.github/workflows/pr-security-review.yml
@@ -2,7 +2,7 @@ name: Claude Security Review
 
 on:
   pull_request_target:
-    types: [opened, reopened, synchronize]
+    types: [opened, reopened, synchronize, labeled]
   pull_request_review:
     types: [submitted]
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


jesseturner21 (approved):
LGTM

PR #1310, opened 2026-05-19
ci(security-review): drop pull_request_review trigger (broken on fork PRs)

tejaskash:
## Why

The `pull_request_review` trigger was added to give maintainers a way to authorize the security review on community (fork) PRs by approving the review. It works on same-repo PRs but **fails silently on fork PRs**: GitHub does not inject repo/org secrets into `pull_request_review` runs when the PR head is a fork, regardless of the reviewer's access level. Today's run on PR #1299 (fork from `notgitika`) is the proof — the workflow fired, the auth gate passed, then `Generate GitHub App token` crashed with `Input required and not supplied: app-id` because `vars.APP_ID` and `secrets.APP_PRIVATE_KEY` both resolved to empty.

The `pull_request_target` event with the `labeled` action does receive secrets on fork PRs and is the only trigger that actually works end-to-end for community contributions.

## What changes

- Remove `pull_request_review` from the workflow triggers.
- Drop the related branches in the auth job (filter clause, `isApproval` path, `review.state == 'approved'` check).
- Update CONTRIBUTING.md to direct maintainers to the `safe-to-review` label and explain why approving review isn't the trigger.
- Same-repo maintainer-authored PRs continue to auto-run on `opened` / `reopened` / `synchronize` exactly as before.
- Community PRs require the `safe-to-review` label, gated on the labeler's write access.

## Test plan

- [x] Land this. On the next community fork PR, apply `safe-to-review` and confirm the workflow runs end-to-end (App token resolves, label adds, Bedrock review runs, inline comments + summary post).
- [x] On a same-repo PR, confirm `opened` continues to auto-run as before.
- [x] On a same-repo PR, confirm an approving review no longer triggers the workflow (since the trigger was removed).


Hweinstock (dismissed):
Thanks for fixing!

notgitika (dismissed):
LGTM thank you!

notgitika (approved):
Thank you!

PR #1318, opened 2026-05-20
test(security-review): verify /security-review path against fix-fork-secrets

tejaskash:
**Throwaway PR for testing — do not merge, do not review.**

Base = `fix-fork-secrets` (PR #1310's branch) so the security review workflow that runs on this PR is the *new* one with bundled-slash-command + honest summary + cancel-in-progress=false. Verifies the inline-comment posting path before #1310 merges.

The diff adds a single file with two deliberate findings the bundled `/security-review` skill should flag (hardcoded AWS creds + command injection via `exec`).

Will close once verified.


tejaskash:
Closing — base-branch testing strategy doesn't work for pull_request_target (workflow always reads from default branch). Will verify the fixes in #1310 on a real PR after that merges.

PR #1326, opened 2026-05-20
feat: add invoke functionality to web-ui

avi-alpert:
## Summary

- Add deployed invocation support to the web UI for all four agent protocols (HTTP, A2A, AGUI, MCP)
- Extract `resolveInvokeTarget` into a reusable module shared by CLI invoke and web-ui handlers
- Return `deploymentTargets` from the resources API so agent inspector can build target picker
- Add `?target=deployed` path to the MCP proxy (`/api/mcp`) for deployed MCP tool listing and calling
- Fix mid-stream errors in deployed invocation handlers — emit SSE error frame and always call `res.end()`

## Additions to agent inspector (web-ui) APIs

**Deployed invocation (`POST /invocations?target=deployed`)**
- Accepts `agentName`, `targetName`, `prompt`, `sessionId`, `userId` in the request body
- Routes to the correct handler based on agent protocol (HTTP/A2A/AGUI)
- Streams responses as SSE events matching the local invocation format

**Deployed MCP proxy (`POST /api/mcp?target=deployed`)**
- Supports `initialize`, `tools/list`, and `tools/call` JSON-RPC methods
- Reuses `mcpInitSession`, `mcpListTools`, `mcpCallTool` from the AWS SDK layer
- Returns JSON-RPC envelope matching the local proxy response shape

**Resources API (`GET /api/resources`)**
- Returns `deploymentTargets: [{ name, region, description? }]` from `aws-targets.json`

**Error handling**
- All three deployed streaming helpers (HTTP, A2A, AGUI) now wrap `for await` in try/catch/finally
- On mid-stream failure: writes an SSE error frame, then ends the response (previously left client hanging)

## Test plan

- [x] Unit tests for deployed invocation path (validation, protocol routing, mid-stream errors, config failures)
- [x] Existing MCP proxy tests updated and passing
- [x] Manual: `agentcore dev` → toggle to Deployed mode → invoke HTTP/A2A/AGUI agent
- [x] Manual: `agentcore dev` → MCP agent → toggle Deployed → verify tools/list and tools/call work
- [x] Manual: verify local invocation still works for all protocols (no regression)

### DIFF
--- a/src/cli/operations/dev/web-ui/handlers/invocations.ts
+++ b/src/cli/operations/dev/web-ui/handlers/invocations.ts
@@ -472,6 +472,7 @@ async function handleDeployedInvocation(
         sessionId,
         userId,
         bearerToken: resolved.bearerToken,
+        baggage: resolved.baggage,
       });
     }
   } catch (err) {
@@ -495,6 +496,7 @@ interface DeployedInvokeParams {
   sessionId?: string;
   userId?: string;
   bearerToken?: string;
+  baggage?: string;
 }
 
 async function handleDeployedHttpInvocation(
@@ -510,6 +512,7 @@ async function handleDeployedHttpInvocation(
     sessionId: params.sessionId,
     userId: params.userId,
     bearerToken: params.bearerToken,
+    baggage: params.baggage,
   });
 
   ctx.setCorsHeaders(res, origin);
--- a/src/cli/operations/dev/web-ui/handlers/invocations.ts
+++ b/src/cli/operations/dev/web-ui/handlers/invocations.ts
@@ -1,6 +1,10 @@
+import { ConfigIO } from '../../../../../lib';
+import { invokeA2ARuntime, invokeAgentRuntimeStreaming, invokeAguiRuntime } from '../../../../aws/agentcore';
+import { buildAguiRunInput } from '../../../../aws/agui-types';
+import { resolveInvokeTarget } from '../../../../commands/invoke/resolve';
 import { extractSSEEventText, extractTaskText, isStatusUpdateEvent } from '../../invoke-a2a';
 import { handleHarnessInvocation } from './harness-invocation';
-import type { RouteContext } from './route-context';
+import { type RouteContext, parseRequestUrl } from './route-context';
 import { randomUUID } from 'node:crypto';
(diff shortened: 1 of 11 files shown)


src/cli/operations/dev/web-ui/handlers/mcp-proxy.ts
```diff
@@ -80,3 +88,135 @@ export async function handleMcpProxy(
...
+        res.end(JSON.stringify({ success: false, error: 'tools/call requires params.name' }));
+        return;
+      }
+      const response = await mcpCallTool(mcpOpts, params.name, params.arguments ?? {});
+      ctx.setCorsHeaders(res, origin);
+      res.writeHead(200, { 'Content-Type': 'application/json' });
+      res.end(
+        JSON.stringify({
+          success: true,
+          result: { jsonrpc: '2.0', result: { content: [{ type: 'text', text: response }] } },
```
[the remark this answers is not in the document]
avi-alpert: The frontend only renders plain text for tool call results today — it doesn't display images, resources, or multi-part content. So the deployed path collapsing to text matches what the UI actually uses. We'll revisit this if/when the frontend adds rich content rendering.

src/cli/operations/dev/web-ui/handlers/invocations.ts
```diff
@@ -331,3 +341,265 @@ async function handleAguiInvocation(
...
+      await handleDeployedAguiInvocation(ctx, res, origin, {
+        region: resolved.region,
+        runtimeArn: resolved.runtimeArn,
+        prompt,
+        sessionId,
+        userId,
+        bearerToken: resolved.bearerToken,
+      });
+    } else {
+      await handleDeployedHttpInvocation(ctx, res, origin, {
```
padmak30, line 468: What about MCP invocations? They would fail for http payload.

avi-alpert: The frontend calls a different endpoint for mcp (`/api/mcp`) - see changes in src/cli/operations/dev/web-ui/handlers/mcp-proxy.ts

padmak30 (dismissed):

padmak30 (approved):

PR #1321, opened 2026-05-20
feat: add agentcore feedback command

tejaskash:
## Summary

Adds a new top-level `agentcore feedback` command:

- **One-shot:** `agentcore feedback "msg" [--screenshot path] [--json]`
- **Wizard:** `agentcore feedback` (no args) → multi-step Ink TUI

Both modes display the AWS Customer Agreement and require interactive y/N consent before submitting. Bare Enter defaults to **No**; non-TTY stdin is refused so the consent decision is always explicit.

The wizard walks Message → optional Screenshot → Consent → Submit, with Esc stepping back and message text preserved across declines/retries. Screenshots are validated locally (PNG/JPG/JPEG, ≤100MB) before any network call, then uploaded via presigned URL with SHA-256 checksum headers.

Submissions carry `cli-version`, `os`, `node-version`, and `cli-mode` metadata. The CLI does not attach AWS account IDs, credentials, project names, or telemetry IDs.

## What's in the diff

- `src/cli/commands/feedback/` — Commander registration, action, TTY consent prompt
- `src/cli/operations/feedback/` — submission orchestrator, payload builder, presigned-URL + S3 upload + form-ingestion client
- `src/cli/tui/screens/feedback/` — Ink wizard with `useFeedbackFlow` hook
- `src/cli/cli.ts`, `src/cli/tui/copy.ts` — wiring
- `README.md`, `docs/commands.md`, `docs/feedback.md` — customer-facing docs

29 unit tests across operations, command, consent-prompt, and TUI hook.

## Test plan

- [x] `npm test -- feedback` — 29/29 pass
- [x] `eslint` clean on all new files
- [x] `tsc --noEmit` clean on all new files
- [x] Manual prod submission verified end-to-end (message + screenshot, both submit and decline paths)
- [x] CLI consent prompt: y/yes/Y/YES accepts; n/no/bare-Enter declines; non-TTY refuses
- [x] TUI wizard: all phase transitions, Esc back-stepping, message preservation across consent decline
- [x] Plain mode and `--json` mode for success, declined, no-tty, validation, network error
- [x] Bug bash plan attached for testers (5 testers × 6 min)

PR #1327, opened 2026-05-20
ci(security-review): fix origin/HEAD missing so /security-review's git diff works

tejaskash:
## Why

Run [#26186045056](https://github.com/aws/agentcore-cli/actions/runs/26186045056) on PR #1321 failed with `Reached maximum number of turns (30)` because the very first thing the bundled `/security-review` skill does is `git diff origin/HEAD...`, and that resolves to:

```
fatal: ambiguous argument 'origin/HEAD...': unknown revision or path not in the working tree
```

`actions/checkout` clones the remote and sets `origin/<head-ref>` but does **not** set the remote's symbolic `HEAD` ref. So `origin/HEAD` is undefined, the skill's git command crashes, Claude tries 30 turns of recovery variants, then the action exits 1.

This affects every PR run after [#1310](https://github.com/aws/agentcore-cli/pull/1310) merged (we switched from inlining the prompt + custom git context to letting the bundled skill drive its own git commands).

## What changes

One step inserted right after `Checkout PR head`:

```yaml
- name: Set origin/HEAD for /security-review skill
  env:
    BASE_REF: ${{ steps.pr.outputs.base_ref }}
  run: |
    set -euo pipefail
    git remote set-head origin "$BASE_REF"
    git symbolic-ref refs/remotes/origin/HEAD
```

`git remote set-head origin <base-ref>` makes `origin/HEAD` resolve to the PR's base branch. The follow-up `git symbolic-ref` line is a sanity print that fails the step early if the head still isn't set.

## Test plan

- [ ] Merge.
- [ ] Open a small PR (or wait for the next community PR + `safe-to-review`); confirm the security review run no longer hits `error_max_turns`, that the skill's `git diff origin/HEAD...` resolves correctly, and that inline findings (or a "no findings" summary) post as expected.

### DIFF
--- a/.github/workflows/pr-security-review.yml
+++ b/.github/workflows/pr-security-review.yml
@@ -167,6 +167,19 @@ jobs:
           # the base branch locally too. fetch-depth: 0 grabs the full history.
           fetch-depth: 0
 
+      - name: Set origin/HEAD for /security-review skill
+        env:
+          BASE_REF: ${{ steps.pr.outputs.base_ref }}
+        run: |
+          set -euo pipefail
+          # actions/checkout doesn't set up the remote's symbolic HEAD ref, so
+          # `git diff origin/HEAD...` (the first command the bundled
+          # /security-review skill runs) fails with "ambiguous argument
+          # 'origin/HEAD...': unknown revision". Point origin/HEAD at the PR's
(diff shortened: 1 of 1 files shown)


Hweinstock (approved):
nice!

PR #1330, opened 2026-05-20
feat: add Lambda interceptors for AgentCore gateways

aidandaly24:
## Description

Adds end-to-end support for Lambda interceptors at gateway REQUEST and RESPONSE points: a new `InterceptorPrimitive`, three vended templates (`pass-through`, `jwt-scope-authorizer`, `tools-list-filter`) for both Python and Node.js, deploy-side preflight, and `logs interceptor` / `invoke interceptor` verbs.

### Primitive surface

- \`agentcore add interceptor --name --gateway --interception-points --template --runtime [--timeout] [--no-pass-request-headers] [--additional-policies]\` for managed mode
- External mode via direct \`agentcore.json\` edit referencing a customer-provided unqualified Lambda ARN
- \`agentcore remove interceptor --name\` reconciles deploy state on next \`agentcore deploy\`
- \`agentcore logs interceptor --name [--follow|--since|--until]\` tails CloudWatch for managed interceptors; external interceptors short-circuit to a remediation message
- \`agentcore invoke interceptor --name [--payload]\` invokes the underlying Lambda for managed interceptors

### Schema

- \`interceptors\` array on \`AgentCoreProjectSpec\` with cross-field \`superRefine\` for cardinality (max 1 per point per gateway, max 2 per gateway) and gateway-name reference checks
- Lambda ARN regex tightened to reject version/alias qualifiers
- \`passRequestHeaders\`, \`entrypoint\`, \`timeoutSeconds\`, \`runtime\` are optional with consumption-time defaults so user-edited \`agentcore.json\` stays sparse across CLI round-trips

### Deploy

- \`validateInterceptors()\` checks managed-mode codeLocation existence and emits a masked cross-account WARN for external mode (with full \`aws lambda add-permission\` remediation snippet); also wired into \`agentcore validate\` so the pre-check works without invoking deploy
- \`validateGatewayTargetLambdas()\` does a best-effort \`lambda:GetFunction\` per \`lambda-function-arn\` target and WARNs on any failure (NotFound, AccessDenied, throttle) so a typo'd ARN surfaces before CFN ROLLBACK
- Account IDs go through \`maskAccountId()\` for all user-visible warning output (PII masking)

### Templates

- \`pass-through\`, \`jwt-scope-authorizer\`, \`tools-list-filter\` for Python (\`pyproject.toml\` + hatchling) and Node.js (\`index.mjs\` + \`package.json\`)

### Bug fixes

- ANSI cursor-show escape (\`\\x1b[?25h\`) was being written to stdout on every CLI invocation by ink/cli-cursor on App unmount, corrupting machine-parseable output (broke \`--json\` JSON parsing). Fixed at the CLI entry by stripping cursor show/hide escapes from non-TTY stdout/stderr writes.

## Out of scope (P0)

The following were deliberately excluded from this PR. None are bugs — flagged here so reviewers don't ask "where is X":

- **PII-redaction template** — patterns are wildly customer-specific (a bank's PII isn't a hospital's). Better as a future config-driven template, or roll your own using \`pass-through\` as a skeleton.
- **Audit-logging template** (OpenSearch / S3) — non-trivial destination wiring (IAM, batching, SIEM plug-in points). Ship later when there's a clearer customer ask.
- **Provisioned concurrency** — adds CFN complexity and runs $$$ even when idle. Wait for actual latency-sensitive customer demand.
- **Multi-gateway shared interceptor pools** — schema is currently 1 interceptor → 1 gateway. Sharing across gateways needs schema redesign (\`gatewayNames: string[]\` or a separate attach relation) and harder cardinality validation.
- **Console UX parity** — separate team's mission. CLI guarantees programmatic parity, not click-for-click visual matching.
- **Streaming-aware first-invocation guard** — shipped as commented-in code in the pass-through Node template (\`if (invocationIndex > 0) { ... }\`). Activating by default would change non-streaming behavior; customers opt in when needed.

## Related Issue

Closes #

## Type of Change

- [x] New feature

## Testing

- [x] I ran \`npm run test:unit\` and \`npm run test:integ\`
- [x] I ran \`npm run typecheck\`
- [x] I ran \`npm run lint\`
- [x] If I modified \`src/assets/\`, I ran \`npm run test:update-snapshots\` and committed the updated snapshots

Additionally:
- End-to-end bug bash conducted across 60+ matrix cases against live AWS account, with 20+ parallel subagents covering: managed Python + Node templates × all 3 interception points, external same-account at REQUEST/RESPONSE/dual-point, cross-account preflight WARN, full deploy lifecycle (deploy → redeploy → remove → teardown), real MCP traffic with interceptor invocation proven via CloudWatch logs, JWT scope-authorizer allow/deny, RESPONSE-side \`statusCode\` mutation, tools-list filtering, additionalPolicies as file path AND managed ARN, IAM Sid hash uniqueness + size under 10240 bytes, schema rejects (qualifier ARN, unknown gateway, duplicate name, over-length name), --json output validity, TUI integration verified via tui-harness MCP

## Checklist

- [ ] Any dependent changes have been merged and published — depends on aws/agentcore-l3-cdk-constructs#225

---


src/cli/primitives/InterceptorPrimitive.ts
```diff
@@ -0,0 +1,428 @@
...
+        const targetDir = join(projectRoot, codeLocation);
+        try {
+          await renderInterceptorTemplate(options.name, runtime, template, targetDir);
+        } catch (renderErr) {
+          // Roll back the spec write so the project isn't left half-added.
+          // The user will see the original render error, not a recovery error.
+          try {
+            const project = await this.readProjectSpec();
+            project.interceptors = project.interceptors.filter(i => i.name !== options.name);
+            await this.writeProjectSpec(project);
```
tejaskash, line 95: Rollback only un-edits agentcore.json — if `renderInterceptorTemplate` throws after partial writes, the half-rendered files in `app/<name>/` stay on disk. A subsequent `agentcore add interceptor --name <same>` then fails at scaffolding because the directory exists. Either rm the partial tree on rollback, or document that the user has to clean it up before retrying.

tejaskash (commented):
Big PR but tight execution — the schema/CDK mirror, account masking, structured external-mode remediation, and cursor-escape fix are all well-handled. Four inline notes; none are blocking.

src/cli/commands/invoke/interceptor.ts
```diff
@@ -0,0 +1,123 @@
...
+      try {
+        payloadJson = await readFile(options.payloadFile, 'utf-8');
+      } catch (readErr) {
+        const msg = readErr instanceof Error ? readErr.message : String(readErr);
+        return {
+          success: false,
+          error: new ValidationError(`Cannot read --payload-file "${options.payloadFile}": ${msg}`),
+        };
+      }
+    } else if (options.payload) {
```
tejaskash, line 56: `--payload` and `--payload-file` aren't mutually exclusive — passing both silently prefers `--payload-file`. Most CLIs reject the combo with a clear error. A user passing both probably has a bug they'd want to know about.

src/cli/operations/deploy/preflight.ts
```diff
@@ -174,6 +188,153 @@ function validateHttpGatewayNames(projectSpec: AgentCoreProjectSpec): void {
...
+      if (!region) continue;
+
+      let client = clientsByRegion.get(region);
+      if (!client) {
+        client = new LambdaClient({ region, credentials });
+        clientsByRegion.set(region, client);
+      }
+
+      try {
+        await client.send(new GetFunctionCommand({ FunctionName: arn }));
```
tejaskash, line 383: Sequential `GetFunction` across all gateways × targets. For projects with many targets (and especially across regions, since each region gets its own client), the latency adds up on every deploy. `Promise.all` over a flattened `(gateway, target)` list would parallelize without changing the warn-only semantics — bonus: a single SDK throttle event won't serialize the rest.

src/cli/index.ts
```diff
@@ -2,6 +2,31 @@
...
+      return originalWrite(chunk.replace(CURSOR_ESCAPE_PATTERN, ''), ...(rest as [never]));
+    }
+    if (chunk instanceof Buffer) {
+      const filtered = chunk.toString('utf8').replace(CURSOR_ESCAPE_PATTERN, '');
+      return originalWrite(filtered, ...(rest as [never]));
+    }
+    return originalWrite(chunk as never, ...(rest as [never]));
+  }) as typeof stream.write;
+}
+patchStream(process.stdout);
```
tejaskash, line 30: Module-import side effect that monkey-patches `process.stdout.write` and `process.stderr.write` for the lifetime of the process. Scope is correct (non-TTY only, just `\x1b[?25[hl]`) and the motivation is real, but it's the kind of thing a future contributor will be very surprised by. Worth either moving to an explicit `installCursorFilter()` called from `main()` so it's greppable, or making the comment block more prominent so it doesn't read like a stray import-time bootstrap.

PR #1339, opened 2026-05-21
fix: avoid overwriting malformed global config

### LINKED ISSUE #1337: [bug]: invalid agentcore config is silently overwritten.
Actual Behavior: If the config at `~/.agentcore/config.json` is invalid json, the CLI silently overwrites it. 

Expected Behavior: Throw an error telling the customer to fix their config.json, or empty it to have it regenerated with defaults. 

Reproduction Steps: add invalid characters to `~/.agentcore/config.json`, run CLI commands that modify the json (ex. telemetry enable). Then verify the contents of the config were overridden.

kiwigitops:
## Summary
- keep malformed global config files from being replaced during config updates
- continue treating missing global config as an empty config
- add a regression test for malformed config preservation

Fixes #1337.

## Testing
- `npx prettier --check src/lib/schemas/io/global-config.ts src/cli/__tests__/global-config.test.ts`
- `npx vitest run --project unit src/cli/__tests__/global-config.test.ts`

### DIFF
--- a/src/cli/__tests__/global-config.test.ts
+++ b/src/cli/__tests__/global-config.test.ts
@@ -131,6 +131,8 @@ describe('global-config', () => {
     it('generates installationId on first run and returns created: true', async () => {
       const result = await getOrCreateInstallationId(tmp.configDir, tmp.configFile);
 
+      expect(result.success).toBe(true);
+      if (!result.success) throw result.error;
       expect(result.created).toBe(true);
       expect(result.id).toMatch(/^[0-9a-f-]{36}$/);
       const config = await readGlobalConfig(tmp.configFile);
@@ -142,7 +144,18 @@ describe('global-config', () => {
 
       const result = await getOrCreateInstallationId(tmp.configDir, tmp.configFile);
 
-      expect(result).toEqual({ id: 'existing-id', created: false });
+      expect(result).toEqual({ success: true, id: 'existing-id', created: false });
+    });
+
+    it('returns malformed config errors without generating an id', async () => {
+      await writeFile(tmp.configFile, '{ invalid json');
+
+      const result = await getOrCreateInstallationId(tmp.configDir, tmp.configFile);
+
+      expect(result.success).toBe(false);
+      if (result.success) throw new Error('expected failure');
+      expect(result.error.message).toMatch(/malformed/);
+      expect(await readFile(tmp.configFile, 'utf-8')).toBe('{ invalid json');
     });
   });
 });
--- a/src/cli/__tests__/global-config.test.ts
+++ b/src/cli/__tests__/global-config.test.ts
@@ -82,9 +82,9 @@ describe('global-config', () => {
     it('creates directory and writes config when none exists', async () => {
       const fresh = createTempConfig('gc-fresh');
 
-      const ok = await updateGlobalConfig({ telemetry: { enabled: false } }, fresh.configDir, fresh.configFile);
+      const result = await updateGlobalConfig({ telemetry: { enabled: false } }, fresh.configDir, fresh.configFile);
 
-      expect(ok).toBe(true);
+      expect(result.success).toBe(true);
       const written = JSON.parse(await readFile(fresh.configFile, 'utf-8'));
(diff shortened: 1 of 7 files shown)


src/lib/schemas/io/global-config.ts
Hweinstock: it looks like this catch will still silently swallow all errors from downstream. The result is that the user won't see a helpful error message of why the command failed when their config is invalid, but will see that the command failed. 

I think this might take a broader refactor to bubble up this error correctly, (and maybe unify these read config methods)

Hweinstock (commented):
Thanks for taking a look! Looking at this change, we might need a broader refactor here to get the right behavior we're looking for.

src/lib/schemas/io/global-config.ts
```diff
@@ -46,20 +47,55 @@ export function readGlobalConfigSync(configFile = GLOBAL_CONFIG_FILE): GlobalCon
...
+}
+
+type LoadConfigResult = { success: true; config: GlobalConfig } | { success: false; error: string };
+
+/**
+ * Reads the existing global config for an update. Distinguishes a missing file
+ * (treated as an empty config) from a malformed one (read/parse/schema failure),
+ * so the caller can avoid clobbering a config it could not understand.
+ */
+async function loadConfigForUpdate(configFile: string): Promise<LoadConfigResult> {
```
Hweinstock, line 83: this logic feels a bit convoluted to me. Do you think it makes sense to directly check if the file exists (via fs.exists or something), then use that instead of the error?

src/lib/schemas/io/global-config.ts
```diff
@@ -46,20 +47,55 @@ export function readGlobalConfigSync(configFile = GLOBAL_CONFIG_FILE): GlobalCon
...
+    }
+    return { success: false, error: `Could not read config at ${configFile}: ${toError(error).message}` };
+  }
+
+  try {
+    return { success: true, config: GlobalConfigSchema.parse(JSON.parse(data)) };
+  } catch (error) {
+    return {
+      success: false,
+      error: `Config at ${configFile} is malformed and was left unchanged: ${toError(error).message}`,
```
Hweinstock: nit: i don't think we need the "left unchanged"

src/lib/schemas/io/global-config.ts
```diff
@@ -46,20 +47,55 @@ export function readGlobalConfigSync(configFile = GLOBAL_CONFIG_FILE): GlobalCon
...
-    return true;
-  } catch {
-    return false;
+    return { success: true };
+  } catch (error) {
+    return { success: false, error: `Failed to write config to ${configFile}: ${toError(error).message}` };
+  }
+}
+
+type LoadConfigResult = { success: true; config: GlobalConfig } | { success: false; error: string };
```
Hweinstock: can we call the field errorMsg or make it store the full error object? 

What you did is in-line with an existing codebase pattern I'm trying to undo where we maintain errors as strings everywhere. This makes sense for displaying to the user, but creates issues when we want programmatic logic based on errors (ex. check if downstream threw an error of type X). Without the object, we lose error type information and are forced to rely on brittle string matching to determine the cause of the error.

src/cli/__tests__/global-config.test.ts
```diff
@@ -106,14 +106,24 @@ describe('global-config', () => {
       });
     });
 
-    it('returns false on write failures', async () => {
-      const ok = await updateGlobalConfig(
+    it('does not overwrite malformed config and reports why', async () => {
```
Hweinstock, line 109: nice simple test!

Hweinstock (commented):
Nice! The user experience makes sense to me. Few implementation notes, but I like the direction.

kiwigitops:
Updated, thanks. This revision addresses the implementation notes by:\n\n- checking file existence explicitly before reading, while still returning access/read failures instead of treating them as a missing config\n- returning an Error object from updateGlobalConfig so callers do not have to rely on string-only errors\n- trimming the malformed-config message\n\nFocused global-config tests and 
pm run typecheck pass.

PR #1365, opened 2026-05-22
fix: create global entrypoint for tui

### LINKED ISSUE #895: bug: duplicate header on entrypoint
### Description

```

  AgentCore Deploy


  AgentCore Deploy

  Project: {projectName}
  Target: us-west-2:{accountId}

  [done]    Validate project
  [done]    Check dependencies
  [done]    Build CDK project
  [done]    Synthesize CloudFormation
  [done]    Check stack status
  [done]    Publish assets
```

### Steps to Reproduce

`agentcore deploy`. 

### Expected Behavior

Only a single header. 

### Actual Behavior

two headers

### CLI Version

_No response_

### Operating System

macOS

### Additional Context

_No response_

Hweinstock:
## Description

Previously, commands like `agentcore add`, `deploy`, etc. rendered TUI screens inline via direct `render()` calls. This made it difficult to distinguish CLI vs TUI code paths, caused telemetry to be mislabeled as `mode="cli"` for interactive flows, and meant bare `agentcore` never initialized or shut down the telemetry client.

This PR creates a unified `renderTUI()` entrypoint that owns the telemetry lifecycle and replaces all inline `render()` calls. As part of this:

- Extracted rendering logic into `src/cli/tui/render.ts` and shared notices into `src/cli/notices.ts` to break circular imports
- Migrated `add`, `deploy`, `create`, `remove`, `invoke` to use `renderTUI()`
- Made `TelemetryClientAccessor.init()` handle re-initialization safely
- Preserved existing behavior (`diffMode`, `isInteractive`, `actionOnBack`)

This refactor fixes #895. 

## Type of Change

- [x] Bug fix
- [x] New feature

## Testing

- [x] Typecheck, lint, unit tests pass
- [x] Manual E2E testing of affected commands

---

https://github.com/user-attachments/assets/807f7c10-92ae-4ea1-b0d1-d18aae8b40fa

Manually tested flows:
- back on inline tui exists instead of going to help. 
- successful flow on tui exists instead of going back to help. 
- root level agentcore works as expected. 
- errors still propagate correctly in inline TUI. 
- telemetry is now emitted for both inline TUI and full screen. 


src/cli/tui/screens/invoke/useInvokeFlow.ts
```diff
@@ -114,80 +116,91 @@ export function useInvokeFlow(options: InvokeFlowOptions = {}): InvokeFlowState
...
-          setPhase('error');
-          return;
-        }
+      const result = await withCommandRunTelemetry(
+        'invoke',
+        {
+          has_stream: true,
+          has_session_id: !!initialSessionId,
+          auth_type: standardize(AuthType, initialBearerToken ? 'bearer_token' : 'sigv4'),
+          agent_protocol: standardize(AgentProtocol, 'unknown'),
```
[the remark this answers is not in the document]
Hweinstock: 2 is actually more correct in capturing the "command run" time. 1 looks like a real regression.

src/cli/commands/deploy/command.tsx
```diff
@@ -14,20 +14,14 @@ import React from 'react';
...
 const SPINNER_FRAMES = ['⠋', '⠙', '⠹', '⠸', '⠼', '⠴', '⠦', '⠧', '⠇', '⠏'];
 
-function handleDeployTUI(options: { autoConfirm?: boolean; diffMode?: boolean } = {}): void {
+function handleDeployTUI(options: { diffMode?: boolean } = {}): Promise<void> {
   requireProject();
-
-  const { unmount } = render(
-    <DeployScreen
-      isInteractive={false}
-      autoConfirm={options.autoConfirm}
```
Hweinstock, line 23: this field is dropped because it wasn't actually used: https://github.com/aws/agentcore-cli/blob/4adc9c08/src/cli/commands/deploy/command.tsx#L211-L214

src/cli/commands/invoke/command.tsx
```diff
@@ -241,33 +240,18 @@ export const registerInvoke = (program: Command) => {
               headers = parseHeaderFlags(cliOptions.header);
             }
 
-            const tuiResult = await withCommandRunTelemetry(
```
Hweinstock: telemetry is moved inside the invoke logic since the client is now initialized in renderTUI.

src/cli/commands/remove/command.tsx
```diff
@@ -84,15 +83,12 @@ export const registerRemove = (program: Command): Command => {
           });
         } else {
           requireTTY();
-          const { unmount } = render(
-            <RemoveAllScreen
-              isInteractive={false}
-              onExit={() => {
-                unmount();
-                process.exit(0);
```
Hweinstock, line 94: killing the process here can prevent clean up from running which includes telemetry client shutdown.

src/cli/notices.ts
```diff
@@ -0,0 +1,31 @@
+import { type UpdateCheckResult, printUpdateNotification } from './update-notifier';
+
+export function printTelemetryNotice(): void {
+  const yellow = '\x1b[33m';
```
Hweinstock, line 4: unrelated to work here, but I want to centralize these. Its very confusing as a reader of code to see `\x1b[33m` and know its yellow.

Hweinstock (commented):
few comments to help with review.

src/cli/commands/remove/command.tsx
```diff
@@ -84,15 +83,12 @@ export const registerRemove = (program: Command): Command => {
...
-              }}
-            />
-          );
+          await renderTUI({
+            initialRoute: { name: 'remove' },
```
avi-alpert: the remove all behavior gets lost here - running `agentcore remove all` now brings you to the remove picker instead of the "Reset all AgentCore schemas to empty state?" screen

src/cli/tui/screens/invoke/useInvokeFlow.ts
```diff
@@ -114,84 +116,104 @@ export function useInvokeFlow(options: InvokeFlowOptions = {}): InvokeFlowState
...
-        if (targetNames.length === 0) {
-          setError('No deployed targets found. Run `agentcore deploy` first.');
-          setPhase('error');
-          return;
-        }
+      const configIO = new ConfigIO();
+      const project = await configIO.readProjectSpec().catch(() => undefined);
+      const firstProtocol = project?.runtimes?.[0]?.protocol ?? 'unknown';
+
+      const result = await withCommandRunTelemetry(
```
avi-alpert, line 138: When i tested this PR locally, i noticed that the `value` in the emitted telemetry was always 1 (likely how long it took to load the config).

Before this change, the value was the actual duration of my entire `invoke` session.

src/cli/tui/render.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  await waitUntilExit();
+
+  if (inAltScreen) {
+    inAltScreen = false;
+    process.stdout.write(EXIT_ALT_SCREEN);
+    process.stdout.write(SHOW_CURSOR);
+  }
+
+  await TelemetryClientAccessor.shutdown();
```
avi-alpert: wouldnt shutting down TelemetryClientAccessor here make it so if telemetry is emitted during a post-exit action (i.e. launchBrowserDev) then it would stop working?

src/cli/commands/invoke/command.tsx
```diff
@@ -241,33 +240,18 @@ export const registerInvoke = (program: Command) => {
...
-                return { success: true };
-              }
-            );
-            if (!tuiResult.success) {
-              render(<Text color="red">Error: {getErrorMessage(tuiResult.error)}</Text>);
-              process.exit(1);
-            }
+              enterAltScreen: false,
+              actionOnBack: 'exit',
+              isInteractive: false,
```
avi-alpert: why was isInteractive switched to false here?

src/cli/commands/remove/command.tsx
Hweinstock: Good catch, fixed in follow-up.

src/cli/commands/invoke/command.tsx
Hweinstock: good question. I made this change thinking that `isInteractive=true` meant exit after first success (as it does for other commands), and I didn't want the tui to exit early. However, turns out its just dead code: https://github.com/aws/agentcore-cli/blob/69b4ce6dfee11ed035d54389cd8af6f77e28dec5/src/cli/tui/screens/invoke/InvokeScreen.tsx#L134. I think we can remove it as a followup.

src/cli/tui/screens/invoke/useInvokeFlow.ts
Hweinstock: Discussed in person, this is to be consistent with dev, and to avoid tying the duration to arbitrary user sessions.

src/cli/tui/render.ts
Hweinstock: Yes, good call, this is important for when we add telemetry for the browser actions. I've swapped it to flush before the blocking process, then the shutdown should be handled by the global one in the finally block, if it reaches there. Shutdown is "best-effort" so its fine if it doesn't.

avi-alpert (dismissed):

avi-alpert (approved):

PR #1341, opened 2026-05-21
feat: compile-time feature flag for preview/GA consolidation

jesseturner21:
## Summary

Consolidates the main and preview branches into a single branch using a compile-time feature flag. `npm run bundle` now produces two tarballs from the same source:

- **GA tarball** — public customers, all harness features stripped via dead code elimination
- **Preview tarball** — preview customers, full harness functionality enabled

This eliminates the merge-conflict-prone dual-branch workflow while ensuring GA customers never see unreleased features.

Also merges in main-branch work: datasets feature, feedback command, and version 0.15.0.

## How it works

1. `src/cli/feature-flags.ts` exports `isPreviewEnabled()` backed by a `__PREVIEW__` constant
2. esbuild's `define` block replaces `__PREVIEW__` with `true` or `false` at bundle time
3. When `false`, esbuild's minifier eliminates all dead code paths — GA bundles contain zero harness logic
4. All harness command registrations, TUI screens, deploy paths, and invoke options are gated behind the flag

### Feature flag infrastructure
- `src/cli/feature-flags.ts` — wrapper function for compile-time constant
- `esbuild.config.mjs` — `define: { __PREVIEW__: process.env.BUILD_PREVIEW === '1' ? 'true' : 'false' }`
- `vitest.config.ts` — matching define for test environments
- `scripts/bundle.mjs` — produces dual tarballs (GA build, then `BUILD_PREVIEW=1` rebuild)

### Command gating
- `invoke/command.tsx` — 13 harness options (`--harness`, `--harness-arn`, `--model-id`, etc.) conditionally registered
- `create/command.tsx` — 8 harness options (`--model-id`, `--api-key-arn`, `--truncation-strategy`, etc.) conditionally registered
- `cli.ts` — `add tool` / `remove tool` commands gated
- `primitives/registry.ts` — `harnessPrimitive` conditionally instantiated

### TUI gating
- `useCreateFlow.ts` — type selector (Harness/Agent/Skip) only shown in preview
- `useInvokeFlow.ts` — harness targets only displayed in preview
- `dev/command.tsx` — harness deploy + invoke flow gated

### Imperative deployment (harnesses)
- `src/cli/operations/deploy/imperative/` — plugin-based deployment manager that runs deployers by phase, separate from CDK pipeline
- `harness-deployer.ts` — creates/updates/deletes harness resources via API
- `harness-mapper.ts` — maps local harness config to API request shape
- `change-detection.ts` — hashes harness config to skip redundant deploys in dev mode

[...]

PR #1394, opened 2026-05-27
fix(deploy): add harness teardown to TUI deploy flow

jesseturner21:
## Summary

- The TUI deploy path (`useDeployFlow.ts`) was missing imperative harness deployment and teardown logic, causing "remove all and deploy" to skip harness deletion
- Added post-CDK harness deploy during normal deploys (persists harness state to `deployed-state.json`)
- Added imperative teardown before stack destroy during teardown deploys (deletes harnesses via API)
- Added E2E test validating the full create → deploy → remove all → deploy → verify deleted lifecycle

## Root Cause

When the feature flag PR (#1341) consolidated preview/main into a single branch, the TUI deploy hook was not updated to include the imperative harness manager calls. The CLI path (`actions.ts`) had them but `useDeployFlow.ts` did not.

## Test plan

- [x] Typecheck passes
- [x] Lint passes
- [x] Manual E2E: created harness, deployed, removed all, deployed (teardown), verified `ResourceNotFoundException` from `get-harness` API
- [ ] CI passes
- [ ] New E2E test (`harness-remove-all.test.ts`) passes with `BUILD_PREVIEW=1`


jesseturner21:
Consider folding the remove-all teardown verification into the existing `createHarnessE2ESuite` (in `harness-e2e-helper.ts`) instead of a separate test file. The existing suite already does create → deploy → invoke → status, so adding a final step that runs `remove all → deploy → verify deleted` would cover the same regression guard while saving an entire deploy cycle (~10+ min) in CI. Then `harness-remove-all.test.ts` can be removed.

.github/workflows/e2e-tests.yml
```diff
@@ -86,44 +86,78 @@ jobs:
...
         id: changed
         run: |
           BASE_SHA=${{ github.event.pull_request.base.sha || 'HEAD~1' }}
-          CHANGED=$(git diff --name-only "$BASE_SHA"..HEAD -- 'e2e-tests/*.test.ts' \
-            | grep -v '^e2e-tests/strands-bedrock\.test\.ts$' \
-            | tr '\n' ' ')
-          echo "extra_tests=$CHANGED" >> "$GITHUB_OUTPUT"
-          echo "Changed e2e tests: ${CHANGED:-none}"
+          # If any helper file changed, run all e2e tests
+          HELPERS_CHANGED=$(git diff --name-only "$BASE_SHA"..HEAD -- 'e2e-tests/*.ts' \
```
jesseturner21, line 119: This changes the e2e tests on pr so that if any helper files change it runs all the tests on the pr. This is important as when people are changing tests we want to see the changes for the tests.

e2e-tests/harness-e2e-helper.ts
```diff
@@ -30,10 +31,11 @@ export function createHarnessE2ESuite(cfg: HarnessE2EConfig) {
   const providerLabel =
     cfg.modelProvider === 'open_ai' ? 'OpenAI' : cfg.modelProvider === 'gemini' ? 'Gemini' : 'Bedrock';
 
-  describe.sequential(`e2e: harness/${providerLabel} — create → deploy → invoke`, () => {
+  describe.sequential(`e2e: harness/${providerLabel} — create → deploy → invoke → teardown`, () => {
```
jesseturner21, line 34: Add a check in these tests that remove all and deploy actually deletes the harness

src/cli/tui/screens/deploy/useDeployFlow.ts
```diff
@@ -319,6 +322,38 @@ export function useDeployFlow(options: DeployFlowOptions = {}): DeployFlowState
     setStackOutputs(outputs);
 
     const existingState = await configIO.readDeployedState().catch(() => undefined);
+
+    // Post-CDK: deploy imperative resources (harness) — preview mode only
+    let deployedHarnesses: Record<string, HarnessDeployedState> | undefined;
```
Hweinstock, line 327: Since this let is outside the feature flag it becomes dead code in the GA release. 

I think this specific pattern isn't very dangerous, but I'm worried this pattern may make it harder to cleanup since the codepaths are no longer isolated.

src/cli/tui/screens/deploy/useDeployFlow.ts
```diff
@@ -673,6 +709,37 @@ export function useDeployFlow(options: DeployFlowOptions = {}): DeployFlowState
...
         if (context?.isTeardownDeploy) {
+          // Teardown imperative resources (harnesses) before destroying the stack
+          if (isPreviewEnabled()) {
+            const teardownTarget = context.awsTargets[0];
+            if (teardownTarget) {
+              const imperativeManager = createDeploymentManager();
+              const teardownConfigIO = new ConfigIO();
+              const existingTeardownState = await teardownConfigIO
+                .readDeployedState()
+                .catch(() => ({ targets: {} }) as DeployedState);
```
Hweinstock, line 720: nit: why do we need to cast the type here? is there something we're assuming?

Hweinstock (approved):

PR #1397, opened 2026-05-27
fix: Patch CVE-2026-42010 GnuTLS auth bypass in Python Dockerfile

notgitika:
## Summary

Migrates the Python Dockerfile template from Debian Bookworm to Debian Trixie to address CVE-2026-42010 (GnuTLS RSA-PSK authentication bypass, CVSS 9.8).

## Changes

```diff
- FROM public.ecr.aws/docker/library/python:3.12-slim-bookworm
+ FROM public.ecr.aws/docker/library/python:3.12-slim-trixie
```

## Why

- Trixie (Debian 13) became stable May 16, 2026 and includes the GnuTLS fix (DLA-4595-1)
- Bookworm full support ends June 2026
- Managed runtime is not affected (runs AL2023 with OpenSSL, not GnuTLS)

## Testing

- `npm run build` ✅
- Dockerfile render snapshot tests updated and passing

## Related

- Ticket: https://t.corp.amazon.com/D448133494
- Samples PR: https://github.com/awslabs/agentcore-samples/pull/1575
- Previous fix: https://github.com/aws/agentcore-cli/pull/1250


Hweinstock (dismissed):
Is there anything we can do to notify existing customers they should modify their template?

notgitika:
good question @Hweinstock 

I am adding a follow up that would add a warning during deploy etc if user is still using bookworm. I think that might be the best method to inform users if they are still using an older vulnerable build.

notgitika:
Merge conflicts resolved (accepted main's workflow file changes — unrelated to this CVE fix).

Addressing review feedback:

**@agentcore-cli-automation (CHANGES_REQUESTED)** — The second code path in `src/cli/commands/import/actions.ts` already uses the shared `PYTHON_BASE_IMAGE` constant from `src/cli/constants.ts:75`, which is set to `python:3.12-slim-trixie`. It does not hardcode bookworm. The inline comment about drift is valid in principle, but both paths are already aligned:
- `src/assets/container/python/Dockerfile` — template file (updated to trixie)
- `src/cli/commands/import/actions.ts:387` — uses `PYTHON_BASE_IMAGE` constant (already trixie)
- `src/cli/commands/import/import-utils.ts:490` — reads from the template file

**@Hweinstock** — Existing customers are notified via a CLI deprecation warning added in `src/cli/operations/deploy/preflight.ts:213-229`. When a user deploys with a bookworm-based Dockerfile, the CLI emits:
> Warning: Agent "..." Dockerfile uses a base image containing "slim-bookworm" which is affected by CVE-2026-42010 (GnuTLS authentication bypass). Update the FROM line to use a Trixie-based variant.

This fires at deploy time, which is when customers would be impacted.

src/cli/operations/deploy/preflight.ts
```diff
@@ -206,6 +208,27 @@ export function validateContainerAgents(projectSpec: AgentCoreProjectSpec, confi
...
+function warnDeprecatedBaseImage(dockerfilePath: string, agentName: string): void {
+  try {
+    const content = readFileSync(dockerfilePath, 'utf-8');
+    for (const line of content.split('\n')) {
+      if (!/^\s*FROM\s+/i.test(line)) continue;
+      for (const image of DEPRECATED_BASE_IMAGES) {
+        if (line.includes(image)) {
+          console.warn(
+            `Warning: Agent "${agentName}" Dockerfile uses a base image containing "${image}" which is affected by ` +
+              `CVE-2026-42010 (GnuTLS authentication bypass). Update the FROM line to use a Trixie-based variant.`
```
Hweinstock: Wouldn't this display the same warning for all deprecated base images? I'm thinking if we add a new item to DEPRECATED_BASE_IMAGE, we would get the warning for bookworm which isn't clear from the function name or variable name. 

One idea could be to make DEPRECATE_BASE_IMAGE a map of image and versions to warning messages, then just logging that here.

notgitika: I like this idea! Raising a follow-up

Hweinstock (approved):

PR #1405, opened 2026-05-28
feat(telemetry): support agent_environment for runtime/harness distinction

Hweinstock:
### Problem 
Harness doesn't emit telemetry!

### Solution
re-use existing runtime metrics with an enum, and make runtime specific fields optional. This PR is essentially wiring that through everywhere. 

Alternatives:
- separate commands for harness (`invoke.runtime` and `invoke.harness`), but this causes cli.command_run `command` field to not represent the actual command run, and instead include the argument. 
- use discriminated union on attributes. This introduces stronger, but more complex schemas. The cost is that we must account for this in our parsing logic, and I don't think the complexity will scale well as we add fields. 

This PR also adds some tests for gaps in harness flows to add telemetry assertions. 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.


src/cli/commands/dev/command.tsx
```diff
@@ -353,6 +356,7 @@ export const registerDev = (program: Command) => {
             if (opts.logs) {
               // Preview: harness-only projects need deploy then print invoke instructions
               if (isPreviewEnabled() && supportedAgents.length === 0 && hasHarnesses) {
+                recorder.set({ agent_environment: 'harness' as const });
```
Hweinstock, line 359: harness vs runtime path is dynamic here so we overwrite default attribute when we hit the harness path.

src/cli/telemetry/schemas/common-shapes.ts
```diff
@@ -150,6 +152,7 @@ export type DeployMode = z.infer<typeof DeployModeSchema>;
   Keys are the field names as they appear in emitted metrics.
 */
 export const ATTRIBUTES = {
+  agent_environment: AgentEnvironment,
```
Hweinstock, line 158: callout: will need to sync backend schema.

integ-tests/dev-server.test.ts
```diff
@@ -162,9 +165,51 @@ describe('integration: dev server', () => {
...
+    const json = JSON.parse(createResult.stdout);
+    projectPath = json.projectPath;
+  });
+
+  afterAll(async () => {
+    telemetry.destroy();
+    if (projectPath) await rm(projectPath, { recursive: true, force: true });
+  });
+
+  // This test currently fails due to https://github.com/aws/agentcore-cli/issues/1406
```
Hweinstock, line 201: See issue for more info, but current behavior is we say "dev not supported on harness" but still exit code 0 which I think is incorrect.

src/cli/commands/create/command.tsx
```diff
@@ -139,78 +140,84 @@ async function handleCreateHarnessCLI(options: CreateOptions): Promise<void> {
...
-    maxLifetime: options.maxLifetime ? Number(options.maxLifetime) : undefined,
-    sessionStoragePath: options.sessionStorageMountPath,
-    skipGit: options.skipGit,
-    skipInstall: options.skipInstall,
-    onProgress,
-  });
+      const provider = (
+        options.modelProvider ? normalizeHarnessModelProvider(options.modelProvider) : 'bedrock'
+      ) as HarnessModelProvider;
+      const defaultModelIds: Record<string, string> = {
```
Hweinstock, line 178: we should move these constants somewhere else, but OOS here.

src/cli/tui/screens/create/useCreateFlow.ts
```diff
@@ -259,16 +260,22 @@ export function useCreateFlow(cwd: string): CreateFlowState {
...
-      agent_framework: standardize(AgentFramework, addAgentConfig?.framework),
-      model_provider: standardize(ModelProvider, addAgentConfig?.modelProvider),
-      memory_type: standardize(MemoryEnum, addAgentConfig?.memory ?? 'none'),
-      agent_protocol: standardize(AgentProtocol, addAgentConfig?.protocol ?? 'HTTP'),
-      build_type: standardize(BuildType, addAgentConfig?.buildType ?? 'CodeZip'),
-      agent_type: standardize(AgentType, addAgentConfig?.agentType ?? 'create'),
-      network_mode: standardize(NetworkMode, addAgentConfig?.networkMode ?? 'PUBLIC'),
-      has_agent: addAgentConfig !== null,
+      agent_environment: standardize(AgentEnvironment, isHarness ? 'harness' : 'runtime'),
+      has_agent: true,
```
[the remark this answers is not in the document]
Hweinstock: addressed.

src/cli/telemetry/schemas/common-shapes.ts
Hweinstock: OOS here, but I want to rename this to AgentSource. Would like do this now before it breaks more dashboards :)

Hweinstock: done and rebased

integ-tests/dev-server.test.ts
notgitika: thanks for cutting the issue! I will look into it

notgitika (approved):
LGTM, thank you!

PR #1424, opened 2026-05-29
fix(packaging): preserve top-level agentcore Python package in zip

### LINKED ISSUE #1408: Regression in #1068: CONFIG_DIR check re-introduced in zip stage, drops top-level `agentcore` Python packages from deployment artifact
## Summary

PR #1015 (#d1e52411) correctly removed the `CONFIG_DIR` exclusion check from the zip-stage functions (`collectFiles` / `collectFilesSync`) because the zip operates on the staging directory — not the project root — and the project config directory is already absent from staging by the time zipping begins.

PR #1068 (#9ccf802e, \"evo preview features\" merge) re-introduced that check due to a merge conflict resolution that took the wrong side. The check is currently present on \`main\`.

## Affected code

**File:** \`src/lib/packaging/helpers.ts\`

Lines 206 and 407 (both \`collectFiles\` and \`collectFilesSync\`) currently contain:

\`\`\`typescript
if (entry.name === CONFIG_DIR && resolve(directory) === resolve(rootDir)) continue;
\`\`\`

where \`rootDir\` equals \`stagingDir\` (passed from \`createZipFromDir(stagingDir, ...)\`).

## Impact

Any Python package that installs a top-level module named \`agentcore\` (i.e., places \`staging/agentcore/\` directly in the staging directory) will have that module silently excluded from the deployment zip.

This is distinct from the original issue #843 / #902, which affected *nested* \`agentcore\` subdirectories via the flat \`EXCLUDED_ENTRIES\` set. That bug was fixed by PR #844. The current regression only affects a top-level \`agentcore\` Python package — a less common case — but still incorrect behavior.

## Root cause

The zip stage should never exclude the project config directory because:
1. The config directory lives at \`<projectRoot>/agentcore/\` — NOT inside \`stagingDir\`
2. The copy stage (\`copySourceTree\`) already filters it out at the source before writing to staging
3. \`uv pip install --target stagingDir\` populates staging with third-party packages only

There is no \`agentcore/\` config directory in staging, so the check is both unnecessary and harmful.

## Fix

Restore the state from PR #1015 — remove the \`CONFIG_DIR\` check from \`collectFiles\` and \`collectFilesSync\`, and remove the now-unused \`rootDir\` parameter from those functions:

\`\`\`diff
-async function collectFiles(directory: string, rootDir: string, basePath = ''): Promise<Zippable> {
+async function collectFiles(directory: string, basePath = ''): Promise<Zippable> {
   const result: Zippable = {};
   const entries = await readdir(directory, { withFileTypes: true });
 
   for (const entry of entries) {
     if (EXCLUDED_ENTRIES.has(entry.name)) continue;
-    if (entry.name === CONFIG_DIR && resolve(directory) === resolve(rootDir)) continue;
 
     const fullPath = join(directory, entry.name);
     const zipPath = basePath ? \`\${basePath}/\${entry.name}\` : entry.name;
 
     if (entry.isDirectory()) {
-      Object.assign(result, await collectFiles(fullPath, rootDir, zipPath));
+      Object.assign(result, await collectFiles(fullPath, zipPath));
     } else if (entry.isFile()) {
       result[zipPath] = [await readFile(fullPath), { level: 6 }];
     }
   }
   return result;
 }
\`\`\`

Same change for \`collectFilesSync\`, and remove the extra argument in the callers \`createZipFromDir\` and \`createZipFromDirSync\`.

The existing test suite (\`helpers.test.ts\`, issue #843 describe block) already validates that nested \`agentcore/\` directories are preserved — those tests should continue to pass. The zip-stage tests should also be updated to reflect that \`createZipFromDir\` no longer excludes anything by \`CONFIG_DIR\` name (it already receives a clean staging directory).

## Commit history

| Commit | Description |
|--------|-------------|
| \`c3921ec0\` | PR #844 — removed \`'agentcore'\` from flat \`EXCLUDED_ENTRIES\`, added root-only \`CONFIG_DIR\` check in both copy and zip stages |
| \`d1e52411\` | PR #1015 — removed \`CONFIG_DIR\` check from zip stage (correct) |
| \`9ccf802e\` | PR #1068 — **re-introduced \`CONFIG_DIR\` check in zip stage** (regression) |

## Related

- Closes #902 (once fixed, the original report will also be addressed — though the flat \`EXCLUDED_ENTRIES\` bug was already fixed in PR #844)
- Fixes #843

NishchayMahor:
## Summary

Restores the state from #1015. The CONFIG_DIR check in the zip-stage `collectFiles` / `collectFilesSync` helpers was reintroduced by the #1068 merge and is currently on main. The zip stage runs against the staging directory, not the project root, and the project config dir is already filtered out at the copy stage, so the check is both unnecessary and harmful: any Python package whose top-level module is named `agentcore` gets silently dropped from the deployment artifact.

## What changed

- Drop the CONFIG_DIR check from `collectFiles` and `collectFilesSync` in `src/lib/packaging/helpers.ts`.
- Drop the now-unused `rootDir` parameter from both helpers and from the two internal call sites in `createZipFromDir` / `createZipFromDirSync`.
- Copy-stage exclusion (the `agentcore/` config dir check in `copyEntry`) is left as-is, since that one is correct.
- Update the existing zip-stage tests under "issue #843" so they no longer assert the broken behavior, and add a new test block under issue #1408 that builds a staging fixture with a top-level `agentcore/` Python package and asserts it is preserved in both the async and sync zip paths.

## Test plan

- [x] `npx vitest run src/lib/packaging/__tests__/helpers.test.ts` reports 39 passed (including the 2 new issue #1408 cases and the unchanged issue #843 copy-stage cases).
- [x] `npm run typecheck` clean.
- [x] `npx eslint src/lib/packaging/helpers.ts src/lib/packaging/__tests__/helpers.test.ts` clean.
- [x] `npx prettier --check` on both files clean.

Closes #1408

### DIFF
--- a/src/lib/packaging/__tests__/helpers.test.ts
+++ b/src/lib/packaging/__tests__/helpers.test.ts
@@ -475,44 +475,68 @@ describe('nested agentcore directory is preserved (issue #843)', () => {
     expect(existsSync(join(dest, 'lib', 'langgraph_checkpoint_aws', 'agentcore', '__init__.py'))).toBe(true);
     expect(existsSync(join(dest, 'lib', 'langgraph_checkpoint_aws', 'agentcore', 'core.py'))).toBe(true);
   });
+});
 
-  // ── createZipFromDir (async) ──
+// ── Issue #1408: zip stage must not drop top-level agentcore Python packages ──
 
-  it('zip: excludes top-level agentcore/ but includes nested agentcore/', async () => {
-    const src = buildFixture(join(root, 'zip-async'));
+describe('top-level agentcore Python package is preserved in zip (issue #1408)', () => {
+  let root: string;
+
+  beforeAll(() => {
+    root = mkdtempSync(join(tmpdir(), 'helpers-zip-agentcore-pkg-'));
+  });
+
+  afterAll(() => {
+    rmSync(root, { recursive: true, force: true });
+  });
+
+  /**
+   * Mimics a staging directory after `uv pip install --target staging`
+   * has installed a third-party package whose top-level module happens to
+   * be named `agentcore`. The zip stage runs against staging, where the
+   * project's own `agentcore/` config dir is already absent, so a
+   * top-level `agentcore/` here is a real Python package and must be
+   * included in the deployment artifact.
+   */
+  function buildStagingFixture(base: string): string {
+    const staging = join(base, 'staging');
+
+    mkdirSync(staging, { recursive: true });
+    writeFileSync(join(staging, 'main.py'), 'print("hello")');
+
+    // Top-level `agentcore` Python package (e.g. an installed dependency)
+    mkdirSync(join(staging, 'agentcore'), { recursive: true });
+    writeFileSync(join(staging, 'agentcore', '__init__.py'), '# package init');
+    writeFileSync(join(staging, 'agentcore', 'runtime.py'), 'def run(): pass');
+
+    return staging;
+  }
+
+  it('createZipFromDir includes a top-level agentcore Python package', async () => {
+    const staging = buildStagingFixture(join(root, 'zip-async'));
     const zipPath = join(root, 'zip-async.zip');
 
-    await createZipFromDir(src, zipPath);
+    await createZipFromDir(staging, zipPath);
 
     const zipBuffer = await readFile(zipPath);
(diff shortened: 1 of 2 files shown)


aidandaly24 (approved):
Thanks for your contribution.

Issue #699, opened 2026-03-27
Inconsistent deployment target handling across CLI commands


jesseturner21:
## Summary

Different CLI commands resolve the `--target` flag with different fallback strategies, making the CLI unintuitive and unpredictable for users managing multiple deployment targets.

## Current Behavior

| Command | Default when `--target` omitted | Multiple targets behavior | Error format when target not found |
|---------|-------------------------------|--------------------------|-----------------------------------|
| `deploy` | Hardcoded `'default'` | No error, silently uses `'default'` | Simple "not found" |
| `invoke` | Hardcoded `'default'` | Uses first deployed target | Shows available target list |
| `status` | First available target | Uses first available target | Shows available target list |
| `import` | `'default'` fallback | Errors, requires `--target` flag | Formatted multi-line list |

### Specific code locations

**Defaults to hardcoded `'default'` string:**
- `src/cli/commands/deploy/command.tsx:156` — `target: cliOptions.target ?? 'default'`
- `src/cli/commands/invoke/command.tsx:155` — `targetName: cliOptions.target ?? 'default'`
- `src/cli/commands/import/actions.ts:484` — `const targetName = target?.name ?? 'default'`

**Defaults to first available (different strategy):**
- `src/cli/commands/status/action.ts:225` — `options.targetName ?? targetNames[0]`
- `src/cli/commands/invoke/action.ts:45` — `options.targetName ?? targetNames[0]!`
- `src/cli/operations/resolve-agent.ts:70` — `targetNames[0]!`

### TUI vs CLI divergence

- **Deploy TUI** (`src/cli/tui/hooks/useAwsTargetConfig.ts`) allows multi-target selection; CLI only accepts one
- **Status TUI** (`src/cli/tui/screens/status/useStatusFlow.ts`) allows cycling through targets; CLI shows only one
- **Import** has no TUI mode at all

## Expected Behavior

All commands should use a single, consistent target resolution strategy:

1. If `--target` specified → validate it exists and use it
2. If exactly one target exists → use it automatically
3. If multiple targets exist and no `--target` specified → error with a list of available targets
4. Error messages should use the same format across all commands

## Proposed Solution

Create a shared `resolveTarget()` utility that all commands call, replacing the per-command ad-hoc logic. This would be a single source of truth for target resolution with consistent error handling.

## Impact

Users with multiple deployment targets (e.g., staging + prod) get different behavior depending on which command they run, which is confusing and can lead to accidentally operating on the wrong target.

[...]

PR #1478, opened 2026-06-08
fix(deploy): auto-populate default target on non-interactive deploy

jariy17:
## Problem

`agentcore deploy` on a freshly-created project fails in non-interactive mode with:

```
Target "default" not found in aws-targets.json
```

`agentcore create` (both interactive and non-interactive) writes an **empty** `aws-targets.json` by design — the target is meant to be populated at deploy time. The **interactive** deploy flow prompts the user for the target (region/account) via the TUI. But the **non-interactive** path (`deploy --yes` / `--json` / `--target`) routes through `handleDeploy`, which just looks up the `'default'` target and errors when the array is empty.

Notably, `agentcore dev` (`runCliDeploy`) already auto-populated the target from the detected AWS context — but plain non-interactive `deploy` did not, so the two diverged.

## Fix

Extract the auto-populate logic into a shared `ensureDefaultDeploymentTarget(configIO)` helper and call it in `handleDeploy` **before** the target lookup. On an empty `aws-targets.json`, it detects the account/region via the existing `detectAwsContext()` and writes a single `default` target. Best-effort: if the account can't be detected, nothing is written and the existing clear "target not found" error still surfaces.

`runCliDeploy` (used by `dev`) now reuses the same helper instead of its own inline copy, removing duplication.

## Testing

- **Unit** (`src/cli/operations/deploy/__tests__/ensure-target.test.ts`): empty → writes default; existing target → no-op; missing/unreadable file → populates; account undetectable → no write; detected region propagated.
- **Integ** (`integ-tests/deploy-autopopulate-target.test.ts`): `create` leaves `[]`; `deploy --json --dry-run` no longer errors "not found"; with AWS creds, `aws-targets.json` is populated with a `default` target.
- **Manual**: `npm run bundle` → installed the tarball globally → `agentcore create --name X --no-agent` (targets `[]`) → `agentcore deploy --json --dry-run` → `aws-targets.json` became `[{name:"default", account, region}]` and deploy advanced past target resolution (failed later only on the expected "No resources defined" for a `--no-agent` project).

Addresses #699


jariy17:
Thanks for the review. Addressed both points:

- **Integ test** — removed `deploy-autopopulate-target.test.ts` entirely. You're right that it would fail in the no-credentials integ environment (auto-populate is a no-op → the middle assertion breaks). The unit tests in `ensure-target.test.ts` already cover the empty/populated/missing-file/no-account branches without needing creds.
- **Consistency** — also updated `useDevDeploy` (the `dev` TUI path), which still had its own duplicated inline auto-populate block, to use the shared `ensureDefaultDeploymentTarget` helper. Now all deploy entry points (`handleDeploy`, `runCliDeploy`, and the dev TUI) go through the one function.

Both in `37d5a4b`.

integ-tests/deploy-autopopulate-target.test.ts
```diff
@@ -0,0 +1,64 @@
...
+  it('non-interactive deploy does not fail with "target not found"', async () => {
+    // --dry-run avoids real CDK deploy; the target lookup (and our auto-populate)
+    // still runs first. The key assertion: deploy must NOT bail with the
+    // pre-fix "Target default not found" error.
+    const result = await runCLI(['deploy', '--json', '--dry-run'], projectDir);
+
+    const combined = `${result.stdout}\n${result.stderr}`;
+    expect(combined.toLowerCase()).not.toContain('not found in aws-targets.json');
+    expect(combined.toLowerCase()).not.toContain('target "default" not found');
+  });
```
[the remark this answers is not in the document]
jariy17: remove integ test because unit tests verify this change

src/cli/operations/deploy/ensure-target.ts
```diff
@@ -0,0 +1,41 @@
...
+ * "target not found" error.
+ *
+ * @returns true if a default target was written, false otherwise.
+ */
+export async function ensureDefaultDeploymentTarget(configIO: ConfigIO): Promise<boolean> {
+  let targets;
+  try {
+    targets = await configIO.readAWSDeploymentTargets();
+  } catch {
+    // aws-targets.json missing or unreadable — treat as empty and try to populate.
```
Hweinstock: q: this behavior is likely reasonable if we don't think customers are manually editing it (which I think I agree with). However, still wondering if we might want to surface the read error instead of silently overwrite the file?

src/cli/operations/deploy/ensure-target.ts
```diff
@@ -0,0 +1,41 @@
...
+    targets = [];
```
Hweinstock, line 33: nit: can we just return false here?

src/cli/operations/deploy/index.ts
```diff
@@ -60,6 +60,9 @@ export {
   type OnlineEvalEnableResult,
 } from './post-deploy-online-evals';
 
+// Auto-populate a default deployment target for non-interactive deploys
```
Hweinstock: nit: unneeded comment.

Hweinstock (dismissed):
Great win! This always bugged me.

src/cli/operations/deploy/ensure-target.ts
jariy17: Good call — updated in 0a64ed8. It now only treats a genuinely-missing file (`ConfigNotFoundError`) as empty; any other read failure (corrupt JSON, validation, permissions) is rethrown so we don't silently overwrite a file that exists but couldn't be parsed. Added a unit test for the rethrow path.

src/cli/operations/deploy/ensure-target.ts
jariy17: Done in 0a64ed8 — the catch now rethrows non-`ConfigNotFoundError` errors and only falls through to `targets = []` for the missing-file case, so the structure is cleaner.

src/cli/operations/deploy/index.ts
jariy17: Removed in 0a64ed8.

Hweinstock (approved):

PR #1474, opened 2026-06-07
fix: resolve create harness Dockerfile paths from command cwd

kiwigitops:
## Description

When `agentcore create` adds a harness with a Dockerfile, the project directory is created before the harness is added. Relative Dockerfile paths were resolved from that new project root, so a path supplied from the directory where `agentcore create` was invoked could fail.

This threads the create command's original working directory into the harness add path and uses it as the base for relative Dockerfile paths. Existing harness add behavior still falls back to the project root.

## Related Issue

Closes #1128

## Documentation PR

N/A

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe):

## Testing

How have you tested the change?

- [ ] I ran `npm run test:unit` and `npm run test:integ`
- [x] I ran `npm run typecheck`
- [x] I ran `npm run lint`
- [ ] If I modified `src/assets/`, I ran `npm run test:update-snapshots` and committed the updated snapshots

Additional checks:

- `npx vitest run --project unit src/cli/primitives/__tests__/HarnessPrimitive.test.ts`
- Preview CLI smoke test with `agentcore create --container ./Dockerfile.custom`, verifying the Dockerfile is copied into the generated harness directory and recorded in `harness.json`

## Checklist

- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [x] I have updated the documentation accordingly
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the
terms of your choice.

PR #1482, opened 2026-06-08
fix(ci): make /security-review fail loudly when the model never runs

tejaskash:
## Summary
- Fix the silent failure mode that produced a false **"no high-confidence findings"** comment on #1474.
- Root cause: the bundled `/security-review` skill's SessionStart hook runs `git diff --name-only origin/HEAD...` as its first command. `actions/checkout@v6` with `ref: <fork-head-sha>` and `fetch-depth: 0` fetches the PR head's history but does not always fetch `origin/<base>`, so the diff resolves to `fatal: no merge base`. The hook errors, the SDK exits cleanly with `num_turns=0` and zero tokens, the inline-comment buffer is empty, and the workflow's summary step still posts "no findings."
- Verified in run [27159048605](https://github.com/aws/agentcore-cli/actions/runs/27159048605) — the SDK transcript shows `<local-command-stderr>fatal: origin/HEAD...HEAD: no merge base</local-command-stderr>` followed by `subtype: success, num_turns: 0`.

## Changes
1. **Make the merge base actually exist.** `Prepare base ref for /security-review skill` now explicitly `git fetch`es `origin/<base>` and runs `git merge-base origin/<base> HEAD` as a sanity check. Fails the step if the merge base can't be resolved, instead of letting the skill silently bail downstream.
2. **Catch silent skill bailouts in general.** New `Verify model actually ran` step reads the action's `execution_file` (a JSON array; final element is the SDK result envelope) and requires `num_turns > 0`, `output_tokens > 0`, and `is_error == false`. If the model never productively ran, the step fails and the summary comment posts *"the review did not actually analyze this PR (model took 0 turns — the skill likely failed during setup)"* instead of *"no high-confidence findings."*

Together: change 1 fixes the immediate bug; change 2 ensures any future cause of "SDK exits clean without the model doing real work" no longer reports a false-clean review.

## Test plan
- [ ] Re-run this workflow on a fork PR (e.g. #1474) and confirm the `Prepare base ref` step logs the merge base + file count.
- [ ] Confirm `Verify model actually ran` reports a sensible `num_turns` / `output_tokens` on a successful run.
- [ ] Manually break it (delete the `git fetch` line locally) and confirm the workflow now fails with the explicit error and the summary comment posts the failure message rather than "no findings."

### DIFF
--- a/.github/workflows/pr-security-review.yml
+++ b/.github/workflows/pr-security-review.yml
@@ -318,10 +318,12 @@ jobs:
             const runUrl = process.env.RUN_URL;
 
             let body;
-            if (modelRan === 'false') {
-              // Action exited 0 but the model never actually ran productively (e.g. the
-              // /security-review skill's SessionStart hook errored before the first turn).
-              // Reporting "no findings" here would be a lie — the diff was never analyzed.
+            if (modelRan !== 'true') {
+              // Two cases land here, both unsafe to report as "no findings":
+              //   - 'false': SDK exited 0 but transcript shows the model never ran productively
+              //     (e.g. /security-review's SessionStart hook errored before the first turn).
+              //   - 'unknown': claude-execution-output.json was missing, so we couldn't verify
+              //     the model ran at all. Treat as not-verified rather than silently green.
               body = `**Claude Security Review:** the review did not actually analyze this PR (model took ${numTurns} turn${numTurns === '1' ? '' : 's'} — the skill likely failed during setup). See the [run](${runUrl}) for details; a later push or re-run is needed for a real review.`;
             } else if (conclusion === 'success') {
               body =
--- a/.github/workflows/pr-security-review.yml
+++ b/.github/workflows/pr-security-review.yml
@@ -167,18 +167,38 @@ jobs:
           # the base branch locally too. fetch-depth: 0 grabs the full history.
           fetch-depth: 0
 
-      - name: Set origin/HEAD for /security-review skill
+      - name: Prepare base ref for /security-review skill
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-security-review.yml
```diff
@@ -240,17 +304,26 @@ jobs:
...
             const prNumber = parseInt(process.env.PR_NUMBER, 10);
             const count = parseInt(process.env.FINDING_COUNT || '0', 10);
             const conclusion = process.env.REVIEW_CONCLUSION || 'skipped';
+            const modelRan = process.env.MODEL_RAN || 'unknown';
+            const numTurns = process.env.NUM_TURNS || '0';
             const runUrl = process.env.RUN_URL;
 
             let body;
-            if (conclusion === 'success') {
+            if (modelRan === 'false') {
```
aidandaly24: will `unknown` hit on this?

aidandaly24: I think this should be `modelRan != 'true'` what do you think?

aidandaly24 (commented):
Totally willing to approve if there's a reason we want this set to 'false' bc otherwise lgtm

.github/workflows/pr-security-review.yml
tejaskash: Good catch — agreed. Switched to `modelRan !== 'true'` in c8b3be90 so the `'unknown'` case (execution-output JSON missing) also routes to the "did not actually analyze" message rather than falling through to "no findings". Both states mean we couldn't verify the model ran, so they should produce the same comment.

aidandaly24 (approved):

Issue #1485, opened 2026-06-08
feat(ci): add canary mechanism for detecting dependency failures


Hweinstock:
### Problems 
A breaking change or down time from our dependencies currently goes unnoticed (until manually reported, or spotted in telemetry). 

### Definition of Done
- run smoke tests on regular interval. On failure, create a GH issue with notification mechanisms to notify the team. 
- make the canary extendable to different test sets, intervals, or other behavior.

PR #1486, opened 2026-06-08
feat(ci): add hourly canary for smoke test

Hweinstock:
## Description
Add smoke tests once an hour that create an issue on failure. The issue will notify the team via the slack integration. 

- centralize issue creation logic into a script. 
- add script to tsbuild ignore and eslint ignore. 
- run across two dimensions: prerelease (latest main push) and released for both GA and preview. 

## Related Issue
https://github.com/aws/agentcore-cli/issues/1485

Closes #

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.

### DIFF
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -88,7 +88,7 @@ jobs:
           OPENAI_API_KEY: ${{ env.E2E_OPENAI_API_KEY }}
           GEMINI_API_KEY: ${{ env.E2E_GEMINI_API_KEY }}
         # Smoke test for now, only runs strands-bedrock.test.ts
-        run: npx vitest run --project e2e e2e-tests/strands-bedrock.test.ts
+        run: npx vitest run --project e2e e2e-tests/strands-bedrock.test.ts --retry 3
 
       - name: Generate GitHub App Token
         if: failure()
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -1,0 +1,116 @@
+name: Canary
+
+on:
+  # Run once an hour
+  schedule:
+    - cron: '0 * * * *'
+  workflow_dispatch:
+    inputs:
+      aws_region:
+        description: 'AWS region for deployment'
+        default: 'us-east-1'
+
+# Serialize canary runs so overlapping scheduled runs don't pile up. Matrix
+# jobs within a single run still execute in parallel.
+concurrency:
+  group: canary
+  cancel-in-progress: false
+
+permissions:
+  id-token: write
+  contents: read
+  issues: write
+
+env:
+  AGENTCORE_TELEMETRY_DISABLED: '1'
+  PRERELEASE_BASE_URL: https://github.com/aws/agentcore-cli/releases/download/prerelease
+
+jobs:
+  canary:
+    runs-on: ubuntu-latest
+    environment: e2e-testing
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        variant: [Released, Prerelease]
+        build: [GA, Preview]
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-node@v6
+        with:
+          node-version: '20.x'
+          cache: 'npm'
+      - name: Configure git
+        run: |
+          git config --global user.email "ci@amazon.com"
+          git config --global user.name "CI"
+      - uses: astral-sh/setup-uv@v7
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v6
+        with:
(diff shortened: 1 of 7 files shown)


.github/workflows/canary.yml
```diff
@@ -0,0 +1,116 @@
...
+
+      - name: Run canary smoke test
+        env:
+          AWS_ACCOUNT_ID: ${{ steps.aws.outputs.account_id }}
+          AWS_REGION: ${{ inputs.aws_region || 'us-east-1' }}
+          ANTHROPIC_API_KEY: ${{ env.E2E_ANTHROPIC_API_KEY }}
+          OPENAI_API_KEY: ${{ env.E2E_OPENAI_API_KEY }}
+          GEMINI_API_KEY: ${{ env.E2E_GEMINI_API_KEY }}
+        # Smoke test for now, only runs strands-bedrock.test.ts
+        run: npx vitest run --project e2e e2e-tests/strands-bedrock.test.ts
```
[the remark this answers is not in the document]
Hweinstock: Good call, added 3 retries via CLI.

.github/workflows/e2e-tests-full.yml
```diff
@@ -5,8 +5,6 @@ on:
       aws_region:
         description: 'AWS region for deployment'
         default: 'us-east-1'
-  schedule:
-    - cron: '0 14 * * 1' # Every Monday at 9 AM EST (14:00 UTC)
```
aidandaly24, line 9: This removes the Monday cron from `e2e-full-test` which in my opinion shrinks our coverage pretty significantly. The new canary replaces it with only `strands-bedrock.test.ts` (1 of 30 e2e suites). For catching CDK schema drift this is fine. But a regression specific to, say, LangGraph+Gemini or container builds is now only caught on push-to-main, not on a schedule. Is this intentional?

e2e-tests/e2e-helper.ts
```diff
@@ -76,7 +76,7 @@ export function createE2ESuite(cfg: E2EConfig) {
       testDir = join(tmpdir(), `agentcore-e2e-${randomUUID()}`);
       await mkdir(testDir, { recursive: true });
 
-      agentName = `E2e${cfg.framework.slice(0, 4)}${cfg.modelProvider.slice(0, 4)}${String(Date.now()).slice(-8)}`;
+      agentName = `E2e${cfg.framework.slice(0, 4)}${cfg.modelProvider.slice(0, 4)}${randomUUID().replace(/-/g, '').slice(0, 8)}`;
```
aidandaly24, line 79: `beforeAll` calls `cleanupStaleCredentialProviders()`, which sweeps orphaned `E2e*` credential providers from runs that died before `afterAll` teardown. But there's no equivalent janitor for the CloudFormation stack itself.

In the existing e2e flows this is low-risk — they run on push/PR, so a leaked stack is rare and gets noticed. The canary changes that calculus: at `4 matrix cells × hourly = ~96 deploy/teardown cycles/day`, a teardown that crashes between deploy and `afterAll` (timeout, runner eviction, AWS flake mid-tear-down) leaks an `AgentCore-E2e…-default` stack with nothing to reclaim it. A single persistent teardown bug could pile up ~96 stacks/day, and each carries real resources (runtime, IAM roles, KMS key from bootstrap) — so this compounds into cost and CFN/runtime quota exhaustion before anyone sees the Slack issue.

Suggest adding a `beforeAll` stack sweep that mirrors the `credential-provider` janitor — delete `AgentCore-E2e*` stacks older than the same cutoff. 

This can also be a follow up, should be non-blocking.

aidandaly24 (commented):
One comment that can be a follow up + a question.

aidandaly24 (approved):

.github/workflows/e2e-tests-full.yml
Hweinstock: Yes, intentional for a few reasons:
- the scheduled monday cron job didn't have a notification mechanism, so we didn't know when it failed. 
- running the entire test suite means we need to reduce the frequency we run it. This change proposed here is to lean towards frequent smoke tests instead of full e2e tests to justify a high frequency. 
- full e2e tests are already run on every commit to main, which IMO is where we're more likely to see regressions.

e2e-tests/e2e-helper.ts
Hweinstock: Good call, I think this is something we should do since I see 1000+ stacks siting in the CI account.

Hweinstock: added an issue for this: https://github.com/aws/agentcore-cli/issues/1493

PR #1518, opened 2026-06-10
fix(assets): prevent multi-gateway tool-name collisions across HTTP frameworks

### LINKED ISSUE #1519: bug(assets): multiple gateways crash/shadow agents due to duplicate tool names
### Description

When two (or more) AgentCore Gateways are attached to a single agent, every gateway exposes the built-in semantic search tool `x_amz_bedrock_agentcore_search` (and may also share target tool names). The vended agent templates wire **one MCP client/toolset/server per gateway** into a single agent, so the identical tool names collide. The symptom differs per framework:

| Framework | Behavior with 2 gateways |
|---|---|
| **Strands** | `ValueError: Tool name 'x_amz_bedrock_agentcore_search' already exists. Cannot register tools with exact same name.` at agent construction (hard crash) |
| **OpenAIAgents** | `UserError: Duplicate tool names found across MCP servers` |
| **GoogleADK** | last-wins dispatch shadow + duplicate `FunctionDeclaration` → Gemini 400 |
| **LangChain / LangGraph** | silent last-wins — one gateway's tool becomes unreachable |

Affects the four HTTP templates that iterate `gatewayProviders`: `strands`, `googleadk`, `langchain_langgraph`, `openaiagents`.

### Steps to Reproduce

1. `agentcore create --project-name demo --no-agent`
2. `agentcore add gateway --name gw1 --authorizer-type NONE`
3. `agentcore add gateway --name gw2 --authorizer-type NONE`
4. `agentcore add gateway-target --gateway gw1 --name a --type mcp-server --endpoint https://mcp.exa.ai/mcp`
5. `agentcore add gateway-target --gateway gw2 --name b --type mcp-server --endpoint https://mcp.deepwiki.com/mcp`
6. `agentcore add agent --name demo --framework Strands --model-provider Bedrock --memory none`
7. `agentcore deploy -y && agentcore invoke "hi"`

Both gateways have semantic search enabled (the default), so each exposes `x_amz_bedrock_agentcore_search`.

### Expected Behavior

The agent constructs successfully and can call tools from all attached gateways. Each gateway's tools are addressable and distinct.

### Actual Behavior

The agent fails to construct (Strands/OpenAIAgents raise on the duplicate tool name), or silently drops one gateway's duplicate-named tool (GoogleADK/LangChain), so the second gateway's search tool is unreachable.

### CLI Version

main (0.19.0)

### Operating System

Linux

### Additional Context

Each SDK ships a first-class namespacing primitive (Strands `MCPClient(prefix=)`, GoogleADK `MCPToolset(tool_name_prefix=)`, LangChain `MultiServerMCPClient(..., tool_name_prefix=True)`, OpenAIAgents `Agent(mcp_config={"include_server_in_tool_names": True})`); the templates simply weren't using them. Fixed in #1518.

### LINKED ISSUE #1520: bug(assets): NONE-auth gateways generate invalid Python (over-indented else) in googleadk/langchain/openaiagents templates
### Description

The `googleadk`, `langchain_langgraph`, and `openaiagents` gateway `client.py` templates generate **syntactically invalid Python** whenever a gateway uses the `NONE` authorizer type. The auth-type `{{#if}}/{{else}}/{{/if}}` Handlebars block is indented inside an `if url:` clause that is immediately followed by `else:`; the indented control tags leave stray whitespace that over-indents the Python `else:` (rendering `            else:` instead of `    else:`), producing an `IndentationError` at import time.

`NONE` is a valid, user-reachable gateway authorizer type (`GatewayAuthorizerTypeSchema = z.enum(['NONE', 'AWS_IAM', 'CUSTOM_JWT'])`), so any agent with a no-auth gateway in one of these frameworks fails to run. Strands is structurally immune (it uses an `if not url: return None` shape with no trailing Python `else:`).

### Steps to Reproduce

1. `agentcore create --project-name demo --no-agent`
2. `agentcore add gateway --name public --authorizer-type NONE`
3. `agentcore add gateway-target --gateway public --name a --type mcp-server --endpoint https://mcp.exa.ai/mcp`
4. `agentcore add agent --name demo --framework GoogleADK --model-provider Bedrock --memory none` (also reproduces with LangChain_LangGraph and OpenAIAgents)
5. Inspect `app/demo/mcp_client/client.py` — or run the agent

### Expected Behavior

The generated `client.py` is valid Python with the `else:` at the correct 4-space indent, for every authorizer type including `NONE`.

### Actual Behavior

The rendered `client.py` contains an over-indented `            else:`, raising `IndentationError: unexpected indent` when the module is imported.

### CLI Version

main (0.19.0)

### Operating System

Linux

### Additional Context

Root cause is indented standalone Handlebars control tags. Moving `{{#if}}/{{else}}/{{/if}}` to column 0 makes Handlebars strip them as standalone lines, keeping the Python `else:` correctly indented. Fixed in #1518.

aidandaly24:
## Description

When two AgentCore Gateways are attached to a single agent, both expose the built-in semantic search tool `x_amz_bedrock_agentcore_search` (and may share target tool names). Each HTTP framework template wired **one MCP client/toolset/server per gateway** into a single agent, so the identical tool names collided. The failure mode differs per framework:

| Framework | Behavior with 2 gateways (before this PR) |
|---|---|
| **Strands** | `ValueError: Tool name 'x_amz_bedrock_agentcore_search' already exists` at agent construction |
| **OpenAIAgents** | `UserError: Duplicate tool names found across MCP servers` |
| **GoogleADK** | last-wins dispatch shadow + duplicate `FunctionDeclaration` → Gemini 400 |
| **LangChain** | silent last-wins — one gateway's tool becomes unreachable |

Each SDK ships a first-class namespacing primitive; this PR uses the idiomatic one for each:

- **Strands** — `MCPClient(prefix="<snakeCase gateway name>")`
- **GoogleADK** — `MCPToolset(tool_name_prefix="<snakeCase gateway name>")`
- **LangChain** — `MultiServerMCPClient(servers, tool_name_prefix=True)` + snakeCase server-dict keys
- **OpenAIAgents** — `Agent(mcp_config={"include_server_in_tool_names": True})` plus an `AsyncExitStack` to `connect()` every gateway server (this also fixes a **pre-existing** bug where the gateway path never connected its MCP servers)

So two gateways now expose e.g. `ipinfo_x_amz_bedrock_agentcore_search` and `userdata_x_amz_bedrock_agentcore_search` — distinct, no collision.

**Version floors** are raised only when a gateway is present (non-gateway projects keep the lower floor), since the namespacing kwargs were introduced in later releases:

- `strands-agents` `1.13.0` → `1.15.0`
- `langchain-mcp-adapters` `0.1.11` → `0.2.0`
- `openai-agents` `0.4.2` → `0.16.0`

(GoogleADK needs no bump — `tool_name_prefix` is in the pinned `google-adk >= 1.17.0`.)

### Also fixed: NONE-auth invalid Python (pre-existing)

While verifying renders, I found that the `googleadk`, `langchain`, and `openaiagents` `client.py` templates generated **syntactically invalid Python** (an over-indented `else:`) whenever a gateway used `NONE` auth — caused by indented Handlebars `{{#if}}/{{else}}/{{/if}}` control tags. Moving those tags to column 0 lets Handlebars strip them as standalone lines, keeping the Python `else:` correctly indented. Strands was structurally immune and left unchanged.

## Related Issue

Closes #1519 (multi-gateway duplicate tool-name collision)
Closes #1520 (NONE-auth gateways generate invalid Python)

## Type of Change

- [x] Bug fix

## Testing

- [x] I ran `npm run test:unit` (full unit suite passes; the only failures are 5 pre-existing TUI tests in `useDevDeploy.test.tsx` / `LogsScreen.test.tsx`, which fail identically on `main`)
- [x] I ran `npm run lint` (0 errors)

Added a regression suite (`src/cli/templates/__tests__/multi-gateway-collision.test.ts`) that renders the **real** asset files and asserts: per-framework namespacing for two gateways, the conditional version floors, and the NONE-auth indentation fix across all affected frameworks.

### Verified end-to-end on real AWS — all four frameworks

For each affected framework I deployed a full project with **two live gateways** (Exa web-search + DeepWiki GitHub-docs, both `NONE` auth, so both expose `x_amz_bedrock_agentcore_search` — the collision trigger), then `agentcore invoke`d a prompt that exercises a tool from **each** gateway. I also ran a per-framework **counterfactual against the deployed gateways' real tool lists** in the agent's own venv (correct SDK versions): the unfixed template reproduces that framework's specific failure, the fixed template resolves it.

| Framework | Model | Deploy | Invoke (both gateways) | Counterfactual (old → fixed, live gateways) |
|---|---|---|---|---|
| **Strands** | Bedrock | ✅ | ✅ Exa + DeepWiki both executed | `ValueError: Tool name 'x_amz_bedrock_agentcore_search' already exists` → both register (`ipinfo_…` / `userdata_…`) |
| **LangChain / LangGraph** | Bedrock | ✅ | ✅ Exa + DeepWiki both executed | 2× identical `x_amz_…_search` (LangGraph `ToolNode` silently shadows one) → distinct `exagw_…` / `wikigw_…` |
| **GoogleADK** | Gemini | ✅ | ✅ Exa + DeepWiki both executed | dispatch `tools_dict` collapses 2→1 (last-wins shadow) → 2 distinct prefixed dispatch names |
| **OpenAIAgents** | OpenAI | ✅ | ✅ Exa executed; DeepWiki reached (upstream MCP server timed out — not a collision) | `UserError: Duplicate tool names found across MCP servers: x_amz_bedrock_agentcore_search` → `mcp_exagw__…` / `mcp_wikigw__…` |

For Strands and OpenAIAgents the unfixed template **hard-crashes** at agent construction; for GoogleADK and LangChain it **silently drops** one gateway's tool. In every case the fix makes both gateways' tools coexist and remain callable. OpenTelemetry runtime traces confirmed the prefixed tool names (e.g. `mcp_wikigw__deepwiki___ask_question`) were registered and invoked server-side. The OpenAIAgents fix additionally connects all gateway servers via `AsyncExitStack`, resolving a pre-existing "server not initialized" bug in that template's gateway path. All test stacks were torn down after verification.

## Checklist

---

### DIFF
--- a/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
+++ b/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
@@ -4106,7 +4106,7 @@ requires-python = ">=3.10"
 dependencies = [
     "opentelemetry-distro",
     "opentelemetry-exporter-otlp",
-    "google-adk >= 1.17.0, < 2.0.0",
+    "google-adk >= 1.35.0, < 2.0.0",
     "google-genai >= 1.0.0, < 2.0.0",
     "bedrock-agentcore >= 1.0.3",
     "botocore[crt] >= 1.35.0",
--- a/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
+++ b/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
@@ -4631,9 +4631,8 @@ dependencies = [
     "opentelemetry-instrumentation-langchain >= 0.59.0",
     "langgraph >= 1.0.2",
     "mcp >= 1.19.0",
-    {{#if hasGateway}}"langchain-mcp-adapters >= 0.2.0",
-    {{else}}"langchain-mcp-adapters >= 0.1.11",
-    {{/if}}"langchain >= 1.0.3",
+    "langchain-mcp-adapters >= 0.2.0",
+    "langchain >= 1.0.3",
     "bedrock-agentcore >= 1.8.0",
     "botocore[crt] >= 1.35.0",
 {{#if (eq modelProvider "Bedrock")}}
@@ -5079,9 +5078,7 @@ readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
     "aws-opentelemetry-distro",
-    {{#if hasGateway}}"openai-agents >= 0.16.0",
-    {{else}}"openai-agents >= 0.4.2",
-    {{/if}}
+    "openai-agents >= 0.16.0",
     "bedrock-agentcore >= 1.8.0",
     "botocore[crt] >= 1.35.0",
     {{#if hasGateway}}{{#if (includes gatewayAuthTypes "AWS_IAM")}}"mcp-proxy-for-aws >= 1.1.0",
@@ -5677,9 +5674,7 @@ dependencies = [
     {{#if (eq modelProvider "Gemini")}}"google-genai >= 1.0.0",
     {{/if}}"mcp >= 1.19.0",
     {{#if (eq modelProvider "OpenAI")}}"openai >= 1.0.0",
-    {{/if}}{{#if hasGateway}}"strands-agents >= 1.15.0",
-    {{else}}"strands-agents >= 1.13.0",
-    {{/if}}
+    {{/if}}"strands-agents >= 1.15.0",
     {{#if hasGateway}}{{#if (includes gatewayAuthTypes "AWS_IAM")}}"mcp-proxy-for-aws >= 1.1.0",
     {{/if}}{{/if}}
 ]
--- a/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
+++ b/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
@@ -4000,20 +4000,20 @@ def get_all_gateway_mcp_toolsets() -> list[MCPToolset]:
     {{#each gatewayProviders}}
     url = os.environ.get("{{envVarName}}")
     if url:
-        {{#if (eq authType "AWS_IAM")}}
+{{#if (eq authType "AWS_IAM")}}
         session = create_aws_session()
         auth = SigV4HTTPXAuth(session.get_credentials(), "bedrock-agentcore", session.region_name)
-        toolsets.append(MCPToolset(connection_params=StreamableHTTPConnectionParams(
+        toolsets.append(MCPToolset(tool_name_prefix="{{snakeCase name}}", connection_params=StreamableHTTPConnectionParams(
             url=url,
             httpx_client_factory=lambda **kwargs: httpx.AsyncClient(auth=auth, **kwargs)
         )))
-        {{else if (eq authType "CUSTOM_JWT")}}
+{{else if (eq authType "CUSTOM_JWT")}}
         token = _get_bearer_token_{{snakeCase name}}()
         headers = {"Authorization": f"Bearer {token}"} if token else None
-        toolsets.append(MCPToolset(connection_params=StreamableHTTPConnectionParams(url=url, headers=headers)))
-        {{else}}
-        toolsets.append(MCPToolset(connection_params=StreamableHTTPConnectionParams(url=url)))
-        {{/if}}
+        toolsets.append(MCPToolset(tool_name_prefix="{{snakeCase name}}", connection_params=StreamableHTTPConnectionParams(url=url, headers=headers)))
+{{else}}
+        toolsets.append(MCPToolset(tool_name_prefix="{{snakeCase name}}", connection_params=StreamableHTTPConnectionParams(url=url)))
+{{/if}}
     else:
         logger.warning("{{envVarName}} not set — {{name}} gateway tools unavailable")
     {{/each}}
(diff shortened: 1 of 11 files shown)


jariy17:
Great Fix! I like how we gateway_name to prefix tools. Is there a reason why we can't update our adk's dependencies all together? Its cleaner and its a free upgrade for us.

aidandaly24:
You're right — done in 23c76f7c. Bumped `google-adk >= 1.17.0` → `>= 1.35.0` (latest 1.x), kept the `< 2.0.0` ceiling.

My "keep it separate" reasoning doesn't hold for a vended template: the floor only affects freshly `create`d projects, so there's no lockfile-narrowing downside that a real library dep would have — it just means new ADK projects start on a current baseline. Confirmed `tool_name_prefix` is present in 1.35.0, and added a floor assertion to the regression suite + regenerated the snapshot.

Ceiling stays at `< 2.0.0` as you noted — ADK 2.x is a major and we haven't validated against it.

aidandaly24:
Simplified in 2f9caed9 — dropped the `{{#if hasGateway}}` conditional so all four frameworks now pin the gateway-capable SDK floor unconditionally (`strands-agents >= 1.15.0`, `langchain-mcp-adapters >= 0.2.0`, `openai-agents >= 0.16.0`; ADK already unconditional at `>= 1.35.0`). For a vended template the floor only affects freshly created projects, so a slightly newer baseline for non-gateway agents costs ~nothing and removes three Handlebars branches.

One thing to set expectations on re: "gateway added after the agent could still work" — the floor isn't the blocker there. `add gateway` only writes to `agentcore.json`; it doesn't re-render the agent's `client.py`, so an agent created before its gateways still has the no-gateway example wiring (no `MCPClient(prefix=...)` at all) regardless of the SDK version. Making order not matter needs `add gateway`/`deploy` to regenerate agent code — happy to take that on as a separate PR if you want it.

jariy17 (approved):
Lgtm! Great find

PR #1555, opened 2026-06-17
fix(harness): surface managed-memory heads-up on dev deploy + validate session storage path in TUI

aidandaly24:
## Description

Fixes two bugs in the gated managed-memory / harness TUI flows (both gated behind `ENABLE_GATED_FEATURES`).

### 1. Managed-memory heads-up missing on `agentcore dev` deploys

The regular `deploy` path shows a heads-up when a harness uses managed memory ("this harness automatically provisions a dedicated AgentCore Memory resource… provisioning can take 3-5 minutes…"). The **dev** deploy path did not: `useDevDeploy` called `handleDeploy` without an `onNotice` callback, so the notice (emitted via `onNotice`) was dropped.

Fix: thread `onNotice` through `useDevDeploy` into a `managedMemoryNotice` state, and render it in `DevScreen`'s "Deploying project resources…" view as a plain dim `Note:` (matching the deploy-screen convention). Verified live: the note now appears as the CFN apply starts and stays pinned through memory provisioning.

### 2. Session-storage mount path not validated in the add-harness wizard

The session-storage step only checked `value.startsWith('/')`, so a nested path like `/mnt/data/workplace/` passed Enter and then failed later at schema-write/deploy. The other mount-path steps (EFS, S3 Files) already validate with `validateBYOMountPath`.

I confirmed the constraint against the API: the smithy `MountPath` shape (`com.amazonaws.bedrockagentcorecontrol#MountPath`) is `pattern: ^/mnt/[a-zA-Z0-9._-]+/?$`, length 6-200 — exactly one segment under `/mnt` — and the `AWS::BedrockAgentCore::Harness` CFN resource schema mirrors it. So nested paths are genuinely invalid at the API; the CLI regex is correct, the TUI step just wasn't using it.

Fix: use `validateBYOMountPath` in the session-storage step's `customValidation`. `TextInput` already blocks Enter and renders a red error/✗ when validation fails. Verified live: `/mnt/data/workplace/` now shows a red error and Enter is blocked; `/mnt/data` shows ✓.

## Related Issue

Closes #

## Type of Change

- [x] Bug fix

## Testing

- [x] I ran `npm run test:unit` (added `useDevDeploy` tests for the `onNotice` → `managedMemoryNotice` wiring, set + unset; the mount-path validator already has direct coverage incl. nested-path rejection)
- [x] I ran `npm run lint` (no new warnings; the 2 remaining in `AddHarnessScreen.tsx` are pre-existing and outside the changed lines)
- [ ] If I modified `src/assets/` — n/a, no asset changes

Also verified **end-to-end live** (account 603141041947, ap-southeast-2, preview build) via the TUI harness:
- `agentcore dev` on a managed-memory harness now shows the `Note:` heads-up during the deploy phase (pinned through `CREATE_IN_PROGRESS`).
- `add harness` session-storage step: `/mnt/data/workplace/` → red error + Enter blocked; `/mnt/data` → ✓.

## Checklist

- [x] I have added any necessary tests that prove my fix is effective
- [ ] I have updated the documentation accordingly — n/a (gated/unreleased surface)

### DIFF
--- a/src/cli/tui/hooks/__tests__/useDevDeploy.test.tsx
+++ b/src/cli/tui/hooks/__tests__/useDevDeploy.test.tsx
@@ -36,10 +36,11 @@ vi.mock('../../../operations/deploy/change-detection', () => ({
 }));
 
 function Harness({ skip }: { skip?: boolean }) {
-  const { steps, isComplete, error } = useDevDeploy({ skip });
+  const { steps, isComplete, error, managedMemoryNotice } = useDevDeploy({ skip });
   return (
     <Text>
-      steps:{steps.length} isComplete:{String(isComplete)} error:{error ?? 'null'}
+      steps:{steps.length} isComplete:{String(isComplete)} error:{error ?? 'null'} notice:
+      {managedMemoryNotice ?? 'null'}
     </Text>
   );
 }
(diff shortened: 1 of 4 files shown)


tejaskash (approved):
LGTM

PR #1560, opened 2026-06-17
test(e2e): fix shard 4/5 stale paths and missing --wait

notgitika:
there are 3 failures in e2e shard 4. they are based on stale assumptions.

2 of them have the same root cause: path migration to a common jobs/ dir. those tests are looking at the wrong place for the run results file.

the other one is expecting json.result to be defined but since that was made async recently, the user has to pass --wait, otherwise it works async.

this PR aims to fix all of those 3 tests.

note: no CLI code is being touched here, only text fixes


notgitika:
should close https://github.com/aws/agentcore-cli/issues/1553 will verify post merge

jariy17 (approved):

nborges-aws (approved):

Hweinstock (approved):
nice! (will merge in post-release)

PR #1566, opened 2026-06-17
fix(recommendation): resolve --bundle-version LATEST to deployed versionId

### LINKED ISSUE #1565: run recommendation does not resolve --bundle-version LATEST (sends literal 'LATEST', 400)
## Summary

`agentcore run recommendation` does not resolve `--bundle-version LATEST` to a concrete version ID. The literal string `LATEST` is sent straight through to the config-bundle API path, which only accepts a version UUID, so the command fails with a 400 validation error.

By contrast, `agentcore run ab-test --control-version LATEST / --treatment-version LATEST` **does** resolve `LATEST` correctly. The two code paths are inconsistent.

## Version

`@aws/agentcore` `0.20.0` (also reproduces on the `0.19.0-d5745` prerelease build).

## Steps to reproduce

1. Create a project with a runtime and add a config bundle, then deploy so the bundle has a version.
2. Run a system-prompt recommendation referencing the bundle with `--bundle-version LATEST`:

```bash
agentcore run recommendation \
  --type system-prompt \
  -r <runtime> \
  -e Builtin.Helpfulness \
  --bundle-name PromptV1 \
  --bundle-version LATEST \
  --system-prompt-json-path systemPrompt \
  --lookback 1 \
  -n MyReco \
  --json
```

## Expected

`LATEST` resolves to the bundle's latest version ID (the same way `run ab-test` resolves it), and the recommendation job is created.

## Actual

The command fails with a 400. The literal `LATEST` is placed into the request URL (`.../configuration-bundles/<bundle-id>/versions/LATEST`) and rejected by the version-id regex validation:

```
Recommendation API error (400): {"message":"Config bundle API client error (400): 400 Client Error:
Bad Request for url: https://bedrock-agentcore-control.<region>.amazonaws.com/configuration-bundles/<bundle-id>/versions/LATEST
| body: {"message":"1 validation error detected: Value at 'versionId' failed to satisfy constraint:
Member must satisfy regular expression pattern: [a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}"}}
```

## Workaround

Pass the explicit version UUID instead of `LATEST` (e.g. from the deploy outputs or `agentcore config-bundle versions`):

```bash
agentcore run recommendation ... --bundle-name PromptV1 --bundle-version a7be96aa-8a83-47b8-8d98-0829b14c1d9b ...
```

With an explicit UUID the recommendation job is created successfully (status `PENDING`).

## Suggested fix

In the `run recommendation` bundle-resolution path, resolve `--bundle-version LATEST` (and likely a missing/omitted version) to the latest concrete version ID before constructing the config-bundle API request — mirroring the resolution already done in the `run ab-test` config-bundle path.

jariy17:
## What

`agentcore run recommendation --bundle-version LATEST` now resolves `LATEST` to the config bundle's deployed version UUID before calling the recommendation API.

Fixes #1565.

## Why

The recommendation path passed the literal string `LATEST` straight through to the config-bundle API (`.../configuration-bundles/<bundle-id>/versions/LATEST`), which only accepts a concrete version UUID. The call failed with:

```
400 ... Value at 'versionId' failed to satisfy constraint:
Member must satisfy regular expression pattern: [a-f0-9]{8}-[a-f0-9]{4}-...
```

The `run ab-test` config-bundle path already resolves `LATEST` (via `resolveConfigBundleVersion`), so the two code paths were inconsistent. Users had to manually pass an explicit version UUID as a workaround.

## What changed

- Added `resolveBundleVersionId(bundleArn, versionRef, deployedState)` in `recommendation/build-config.ts`. It returns an explicit version verbatim, expands `LATEST` to the bundle's deployed `versionId` (looked up by ARN from deployed state), and returns `undefined` when `LATEST` can't be resolved (bundle not deployed). This mirrors the ab-test helper.
- `recommendation/handler.ts` now resolves the version before building the request. If `LATEST` can't be resolved it surfaces a friendly `ResourceNotFoundError` ("Run `agentcore deploy` first, or pass an explicit version UUID.") instead of sending `LATEST` to the API.
- Added unit tests for `resolveBundleVersionId` (explicit passthrough, LATEST resolution, unresolved bundle, empty targets).

## Testing

- `npm run typecheck` — pass
- `eslint` on changed files — clean
- `vitest run src/cli/operations/jobs/recommendation/` — 42 passed (incl. 4 new)
- Manually verified against a deployed project (`@aws/agentcore` 0.20.0): `--bundle-version LATEST` previously returned a 400; the equivalent flow with an explicit UUID succeeds, and this change makes `LATEST` resolve to that same UUID.

### DIFF
--- a/src/cli/operations/jobs/recommendation/__tests__/resolve-bundle-version.test.ts
+++ b/src/cli/operations/jobs/recommendation/__tests__/resolve-bundle-version.test.ts
@@ -1,0 +1,38 @@
+import type { DeployedState } from '../../../../../schema';
+import { resolveBundleVersionId } from '../build-config';
+import { describe, expect, it } from 'vitest';
+
+const BUNDLE_ARN = 'arn:aws:bedrock-agentcore:us-east-1:123456789012:configuration-bundle/Proj-PromptV1-abc123';
+const VERSION_ID = 'a7be96aa-8a83-47b8-8d98-0829b14c1d9b';
+
+const deployedState = {
+  targets: {
+    default: {
+      resources: {
+        configBundles: {
+          PromptV1: { bundleArn: BUNDLE_ARN, versionId: VERSION_ID },
+        },
+      },
+    },
+  },
+} as unknown as DeployedState;
+
+describe('resolveBundleVersionId', () => {
+  it("expands 'LATEST' to the deployed versionId", () => {
+    expect(resolveBundleVersionId(BUNDLE_ARN, 'LATEST', deployedState)).toBe(VERSION_ID);
+  });
+
+  it('returns an explicit version verbatim (never touches deployed state)', () => {
+    const explicit = '11111111-2222-3333-4444-555555555555';
+    expect(resolveBundleVersionId(BUNDLE_ARN, explicit, deployedState)).toBe(explicit);
+  });
+
+  it("returns undefined when 'LATEST' cannot be resolved (bundle not deployed)", () => {
+    const unknownArn = 'arn:aws:bedrock-agentcore:us-east-1:123456789012:configuration-bundle/Proj-Unknown-zzz999';
+    expect(resolveBundleVersionId(unknownArn, 'LATEST', deployedState)).toBeUndefined();
+  });
+
+  it('returns undefined for LATEST when there are no deployed targets', () => {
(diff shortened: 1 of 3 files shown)


aidandaly24 (approved):
this looks good to me

PR #1573, opened 2026-06-17
fix(payment-connector): validate secret key formats client-side at add time

### LINKED ISSUE #1564: [Bug] add client side validation in payment connector
### Description

I entered the wrong payment connector secret key format which allowed me to "add" the payment connector (even showed me the green tick validaiton in TUI) but then deploy failed.

Add better client side validation in this case.

### Steps to Reproduce

1. be in the TUI
2. agentcore add payment-manager
3. agentcore add payment-connector 
4. then just type in random values for payment connector fields
5. hit agentcore deploy, which errors out with message

```
[done]    Validate project
  [done]    Check dependencies
  [done]    Build CDK project
  [error]   Creating payment infrastructure
            → Payment setup failed: Credential provider for "MyCdpConnector": Failed to create payment credential provider "MyPaymentManager-MyCdpConnector-cdp": Payment API error (400):
  {"fieldList":null,"message":"Invalid apiKeySecret format: Expected base64-encoded Ed25519 private key","reason":null}
```

### Expected Behavior

it should catch the client side validation earlier during the add step instead of deploy

### Actual Behavior

it errors out during deploy

### CLI Version

0.20.0

### Operating System

macOS

### Additional Context

_No response_

aidandaly24:
## Description

Adds client-side format validation for payment connector secret keys at `add` time, so a wrong-format key is caught immediately instead of failing later at `deploy`.

**Bug (#1564):** A user entered a wrong-format CoinbaseCDP `apiKeySecret` in the TUI, saw a green-tick "valid", and `add payment-connector` succeeded — but `deploy` then failed server-side with `Payment API error (400): Invalid apiKeySecret format: Expected base64-encoded Ed25519 private key`.

**Root cause:** Only the StripePrivy `authorizationPrivateKey` had any format validation, and it lived inline in the CLI action — the TUI never ran it. Every other secret field (including the CoinbaseCDP key fields) was accepted as any non-empty string on both surfaces, so invalid keys were written to `.env.local` and only rejected by the service at deploy.

**Fix:** A shared validator module (`src/cli/primitives/payment-validation.ts`) following the existing `validateBYOMountPath` pattern — each validator returns `true | string`, which plugs directly into both the CLI add path and the TUI `SecretInput` `customValidation` prop, so the two surfaces validate identically (the same approach used by the session-storage path validation in #1555).

Validated fields (cryptographic, server-enforced format):
- `apiKeySecret` (CoinbaseCDP) — base64-encoded **Ed25519** private key
- `walletSecret` (CoinbaseCDP) — base64-encoded **EC P-256** private key
- `authorizationPrivateKey` (StripePrivy) — base64-encoded EC P-256, accepting the optional `wallet-auth:` prefix (now via the shared `stripWalletAuthPrefix`)

Opaque identifier fields (`apiKeyId`, `appId`, `appSecret`, `authorizationId`) keep their non-empty checks, since they have no documented format and over-validating risks false rejections. Ed25519 and P-256 keys differ substantially in size (Ed25519 decodes to 32–64 bytes; P-256 PKCS8 to ~138), so each field uses its own decoded-byte band rather than one shared range — reusing the P-256 band for the Ed25519 field would have rejected every valid CDP key.

## Related Issue

Closes #1564

## Documentation PR

N/A — no doc changes; validation behavior is self-describing via the error messages.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I ran `npm run lint`

Notes:
- `typecheck` (0 errors), `lint` (0 errors), `test:unit` (5404 passed), `test:integ` (274 passed) all green. No `src/assets/` changes.
- Unit tests in `payment-validation.test.ts` exercise each validator against real generated Ed25519/P-256 keys, including cross-curve rejection (a P-256 key fails `validateApiKeySecret`, an Ed25519 key fails `validateWalletSecret`) and the `wallet-auth:` prefix.
- Integration fixtures (`add-remove-payment.test.ts`) now use real keys so the lifecycle tests exercise add/remove rather than tripping the new format check.
- Verified end-to-end against the built CLI: the exact #1564 repro (`--api-key-secret randomsecret`) is now rejected at `add` (`apiKeySecret must be a base64-encoded Ed25519 private key (unexpected length)`, exit 1, connector not written), while valid Ed25519 + P-256 keys are accepted.

## Checklist

---

### DIFF
--- a/integ-tests/add-remove-payment.test.ts
+++ b/integ-tests/add-remove-payment.test.ts
@@ -1,8 +1,20 @@
 import { createTestProject, readProjectConfig, runCLI } from '../src/test-utils/index.js';
 import type { TestProject } from '../src/test-utils/index.js';
+import { generateKeyPairSync } from 'node:crypto';
 import { readFile } from 'node:fs/promises';
 import { join } from 'node:path';
 import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+
+// The CLI now validates CDP secret formats at add time: apiKeySecret must be a
+// base64-encoded Ed25519 private key and walletSecret a base64-encoded EC P-256
+// private key. Use real keys so these lifecycle tests exercise add/remove rather
+// than tripping the format check.
+const CDP_API_KEY_SECRET = generateKeyPairSync('ed25519')
+  .privateKey.export({ type: 'pkcs8', format: 'der' })
+  .toString('base64');
+const CDP_WALLET_SECRET = generateKeyPairSync('ec', { namedCurve: 'P-256' })
+  .privateKey.export({ type: 'pkcs8', format: 'der' })
+  .toString('base64');
 
 describe('integration: add and remove payment managers and connectors', () => {
   let project: TestProject;
@@ -154,9 +166,9 @@ describe('integration: add and remove payment managers and connectors', () => {
           '--api-key-id',
           'test-key-id',
           '--api-key-secret',
-          'test-key-secret',
+          CDP_API_KEY_SECRET,
           '--wallet-secret',
-          'test-wallet-secret',
+          CDP_WALLET_SECRET,
           '--json',
         ],
         project.projectPath
@@ -202,9 +214,9 @@ describe('integration: add and remove payment managers and connectors', () => {
           '--api-key-id',
           'test-key-id-2',
           '--api-key-secret',
-          'test-key-secret-2',
+          CDP_API_KEY_SECRET,
           '--wallet-secret',
-          'test-wallet-secret-2',
+          CDP_WALLET_SECRET,
           '--json',
         ],
         project.projectPath
@@ -232,9 +244,9 @@ describe('integration: add and remove payment managers and connectors', () => {
           '--api-key-id',
           'x',
           '--api-key-secret',
-          'y',
+          CDP_API_KEY_SECRET,
           '--wallet-secret',
-          'z',
+          CDP_WALLET_SECRET,
           '--json',
         ],
         project.projectPath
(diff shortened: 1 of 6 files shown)


notgitika (approved):
```
                                   _o
                                  .#$?\
                                .'##$ ?$\
                              .'.###$  ?$$\
                            .' .####$   ?$$$\
                          .'  +#####$   `$$$$$\
                        .'   d######$    `$$$$$$\
                      .'   .########$     ?$$$$$$$\
                    .'    .#########$      ?$$$$$$$$\
                  .'     ,##########$       ?$$$$$$$$$\
                .'      +###########$       `?$$$$$$$$$$\
             ..'      .;############$        `$$$$$$$$$$$$\
             |'       ;    ````""""?$         ?$$$$??"""?$$$\
             |_/_____;._______._____$_________)?;__________?;X\___.-.
             /'_____;_|______,|ooooo$oo|o,,,,/..|..___      |`'+\.\.'
          .'||  ,ooooo|ooooodS|SSSSSSSS|SSS/'SSS|SSSSSSSSSSS|SSSS|/
      .  .; |~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~/'
.,;.;';`'(`;|',.;'.'.,,,+','+.,'+,.+,'+..^+,','+.,',.';.'.,+'`;'.';,:'';,:.'
;'.';,:'.,;.;';`'(`;|',.;'.'.,,,+','+.,'+,.+,'+..^+,','+.,',.';.'.,+'`.';,:'
```

PR #1575, opened 2026-06-18
fix: stop web-search TUI early exit when no gateway

nborges-aws:
## Description

Fix for issue causing the web-search interactive flow to kill the TUI process early if no gateway is present. Tested via installing tarball, process now stays alive and `NoGatewayMessage` renders as expected.

image

## Related Issue

Closes #

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I ran `npm run lint`

## Checklist

- [ ] I have added any necessary tests that prove my fix is effective or my feature works
- [ ] My changes generate no new warnings

---

terms of your choice.

### DIFF
--- a/src/cli/tui/screens/web-search/AddWebSearchScreen.tsx
+++ b/src/cli/tui/screens/web-search/AddWebSearchScreen.tsx
@@ -92,7 +92,7 @@ export function AddWebSearchScreen({
       onExit={onExit}
       helpText={helpText}
       headerContent={headerContent}
-      exitEnabled={isNameStep}
+      exitEnabled={isNameStep || (isGatewayStep && noGatewaysAvailable)}
     >
       <Panel>
         {isNameStep && (
--- a/src/cli/tui/screens/web-search/__tests__/AddWebSearchScreen.test.tsx
+++ b/src/cli/tui/screens/web-search/__tests__/AddWebSearchScreen.test.tsx
@@ -1,0 +1,107 @@
+import { AddWebSearchScreen } from '../AddWebSearchScreen';
+import type { AddWebSearchConfig } from '../types';
+import { render } from 'ink-testing-library';
+import React from 'react';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+const ENTER = '\r';
+const ESCAPE = '\x1B';
+const BACKSPACE = '\x7f';
+const delay = (ms = 50) => new Promise(resolve => setTimeout(resolve, ms));
+
+function makeProps(overrides: Partial<React.ComponentProps<typeof AddWebSearchScreen>> = {}) {
+  return {
+    onComplete: vi.fn<(config: AddWebSearchConfig) => void>(),
+    onExit: vi.fn(),
+    existingGatewayNames: [],
+    existingToolNames: [],
+    ...overrides,
+  };
+}
+
+// Walk past the name step by accepting the default and pressing Enter.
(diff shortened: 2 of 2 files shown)


src/cli/tui/screens/web-search/AddWebSearchScreen.tsx
```diff
@@ -92,7 +92,7 @@ export function AddWebSearchScreen({
       onExit={onExit}
       helpText={helpText}
       headerContent={headerContent}
-      exitEnabled={isNameStep}
+      exitEnabled={isNameStep || (isGatewayStep && noGatewaysAvailable)}
```
Hweinstock, line 95: Pasting my own understanding of why this fixes it, since I don't think its obvious. 

So before, exitEnabled was false on that page, meaning this [hook](https://github.com/aws/agentcore-cli/blob/main/src/cli/tui/hooks/useExitHandler.ts#L10) was disabled. Therefore, when there were no gateways, that screen in the wizard wasn't listening for any user inputs so nothing was keeping the node event loop alive. 

Relevant docs:
> An Ink app is a Node.js process, so it stays alive only while there is active work in the event loop (timers, pending promises, [useInput](https://github.com/vadimdemedes/ink/blob/master/readme.md#useinputinputhandler-options) listening on stdin, etc.). If your component tree has no async work, the app will render once and exit immediately.
https://github.com/vadimdemedes/ink/blob/master/readme.md#app-lifecycle


As a long term fix (OOS here), I feel like we need to build a common Wizard abstraction that handles this once for us. Exposing `exitEnabled` directly here will always carry the risk that someone disables exits, and forgets to wire up another hook in its place, causing the process to exit early.

Hweinstock (approved):
nice fix, verified locally as well since I was initially confused how this fixed it.

src/cli/tui/screens/web-search/AddWebSearchScreen.tsx
notgitika: Thanks! And I agree with the suggestion as a follow up

notgitika (approved):

notgitika:
```
                                   _o
                                  .#$?\
                                .'##$ ?$\
                              .'.###$  ?$$\
                            .' .####$   ?$$$\
                          .'  +#####$   `$$$$$\
                        .'   d######$    `$$$$$$\
                      .'   .########$     ?$$$$$$$\
                    .'    .#########$      ?$$$$$$$$\
                  .'     ,##########$       ?$$$$$$$$$\
                .'      +###########$       `?$$$$$$$$$$\
             ..'      .;############$        `$$$$$$$$$$$$\
             |'       ;    ````""""?$         ?$$$$??"""?$$$\
             |_/_____;._______._____$_________)?;__________?;X\___.-.
             /'_____;_|______,|ooooo$oo|o,,,,/..|..___      |`'+\.\.'
          .'||  ,ooooo|ooooodS|SSSSSSSS|SSS/'SSS|SSSSSSSSSSS|SSSS|/
      .  .; |~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~/'
.,;.;';`'(`;|',.;'.'.,,,+','+.,'+,.+,'+..^+,','+.,',.';.'.,+'`;'.';,:'';,:.'
;'.';,:'.,;.;';`'(`;|',.;'.'.,,,+','+.,'+,.+,'+..^+,','+.,',.';.'.,+'`.';,:'
```

PR #1599, opened 2026-06-22
test(e2e): add insights and online-insights lifecycle tests

notgitika:
This PR adds e2e tests for insights feature that was recently shipped. two independent `describe.sequential` suites in a single file (`e2e-tests/insights-lifecycle.test.ts`), each owning its own deployed agent and CFN stack so a deploy failure in one suite does not blank the other.

Suite A: `online-insights` lifecycle
`add online-insights → deploy → invoke → pause → resume → teardown`. Mirrors `evals-lifecycle.test.ts`. Verifies live `executionStatus` toggling through the control plane.

Suite B: `run insights` + recommendation chain
`deploy → invoke → run insights (async) → view list → view detail → archive → run insights --wait → run recommendation --from-insights → teardown`. Covers:
- Async submission of `run insights` (no `--wait`) and local job record
- `view insights` list + per-id detail
- `archive insights` round-trip (job no longer appears in `view insights`)
- `run insights --wait` reaching a terminal status
- `run recommendation --from-insights` chaining off the completed insights job

The chain step accepts either `success: true` or a service error containing `no sessions` / `completed_with_errors` / `failed` / `empty`, since real trace volume is not guaranteed seconds after invoke. Flag-parsing errors (e.g. `Unknown option`, `--from-insights`) still fail the test hard, so the wiring is verified even when the upstream job has nothing to learn from.

### DIFF
--- a/e2e-tests/insights-lifecycle.test.ts
+++ b/e2e-tests/insights-lifecycle.test.ts
@@ -1,0 +1,420 @@
+/**
+ * E2E tests for the Insights feature (Lens).
+ *
+ * Two independent sequential suites in this file:
+ *
+ *   A. online-insights lifecycle:
+ *        add online-insights → deploy → invoke → pause → resume → teardown
+ *
+ *   B. run-insights + recommendation chain:
+ *        deploy → invoke → run insights (async) → view → archive
+ *        → run insights --wait → run recommendation --from-insights → teardown
+ *
+ * Each suite owns its own deployed agent + CFN stack so a deploy failure in
+ * one suite does not blank the other.
+ */
+import { parseJsonOutput, retry } from '../src/test-utils/index.js';
+import {
+  baseCanRun,
+  hasAws,
+  installCdkTarball,
+  runAgentCoreCLI,
+  teardownE2EProject,
+  writeAwsTargets,
+} from './e2e-helper.js';
+import { randomUUID } from 'node:crypto';
+import { mkdir, rm } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+
+const canRun = baseCanRun && hasAws;
+const FAILURE_ANALYSIS = 'Builtin.Insight.FailureAnalysis';
+
+// ──────────────────────────────────────────────────────────────────────────────
+// Suite A: online-insights lifecycle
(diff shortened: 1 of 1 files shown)


e2e-tests/insights-lifecycle.test.ts
```diff
@@ -0,0 +1,420 @@
...
+  it.skipIf(!canRun)(
+    'view insights <id> returns detail for the async job',
+    async () => {
+      const result = await run(['view', 'insights', asyncJobId, '--json']);
+      expect(result.exitCode, `View insights detail failed: ${result.stderr}`).toBe(0);
+      const json = parseJsonOutput(result.stdout) as InsightsJobJson;
+      expect(json.success).toBe(true);
+      expect(json.id).toBe(asyncJobId);
+      expect(json.status).toBeTruthy();
+      expect(json.region).toBeTruthy();
```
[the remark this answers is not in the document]
notgitika: Good catch — confirmed `JobRecordBase` only stores `arn` and notes "region is NOT stored — parse from arn". Went with option 2 from your suggestion: dropped `region` from the `InsightsJobJson` interface and changed the detail assertion to `expect(json.arn).toMatch(/^arn:[^:]+:bedrock-agentcore:/)`, which proves the field is present and well-formed across partitions. Fixed in f2611c64.

e2e-tests/insights-lifecycle.test.ts
```diff
@@ -0,0 +1,420 @@
...
+        console.log('Deploy stderr:', result.stderr);
+      }
+      expect(result.exitCode, 'Deploy failed').toBe(0);
+      const json = parseJsonOutput(result.stdout) as { success: boolean };
+      expect(json.success).toBe(true);
+    },
+    600000
+  );
+
+  it.skipIf(!canRun)(
```
jariy17, line 123: I understand this is a "setup" test but we shouldn't fail the insight test suite if the invoke command breaks.

jariy17 (commented):
We shouldn't use tests to setup testing environments. Beside that, breaking down tests by cli commands is a great idea!

e2e-tests/insights-lifecycle.test.ts
notgitika: that makes sense. I will move the invoke call into beforeAll, catch failures, and conditionally skip the suite.

aidandaly24 (approved):

PR #1598, opened 2026-06-19
feat: bring harness flows into latest version of cli

avi-alpert:
## Summary

The `__PREVIEW__` / `isPreviewEnabled()` build-time flag gated harness and related flows (`export`, `add`/`remove tool`, `add`/`remove skill`) to preview-only builds. This PR cuts over so those flows are the **only** path: every preview-enabled branch is inlined as always-on, the GA-only branches are deleted, and the build-time machinery is removed.

Result: harness is now available in the standard (`@latest`) release — no preview build required.

Closes #1569

## What changed

- Remove `isPreviewEnabled()` + the `__PREVIEW__` esbuild define, the `BUILD_PREVIEW` wiring in `index.ts`, the `build:preview` npm script, and the `AGENTCORE_PREVIEW` CDK passthrough (`wrapper.ts` + vended `assets/cdk/bin/cdk.ts`).
- Inline all preview-gated branches across commands and TUI as always-on; delete the now-dead GA-only branches (including the old `create-prompt` flow).
- Dedupe `scripts/bundle.mjs` dual-tarball logic — GA and preview builds are now byte-identical.
- Fix the CI workflows that consumed the removed `-preview` tarball: `e2e-tests.yml`, `prerelease-tarball.yml`, `canary.yml`.
- Update `README.md`: harness section no longer says it requires `@aws/agentcore@preview`.
- Delete `preview-flag.test.ts`; update tests that mocked the flag.

## Notes
- **Preview publishing continues.** We keep publishing the binary to the `@preview` npm tag so existing `1.0.0-preview.XX` customers don't have to switch over

## Testing

- Unit suite, typecheck, lint, and asset snapshots all pass.
- Harness e2e tests (formerly preview-gated) now run unconditionally in the single build.

PR #1608, opened 2026-06-22
test(integ): expand harness config-shape coverage

tejaskash:
## Summary

Expands the harness integration test suite (`integ-tests/add-remove-harness.test.ts`) to cover the harness configuration surface that was previously untested at the integration layer. Follow-up to #1598, which made harness the default `create` path and brought the harness flows into the GA build.

All new cases are **local — no AWS**. They exercise the CLI flag → action → validate → spec-write wiring that schema unit tests don't reach.

## Coverage added

- **model**: `lite_llm` (+`--api-base`/`--additional-params`), bedrock `--api-format`, plus api-format provider-gating rejections (open_ai/gemini)
- **tools** (`add tool`): `remote_mcp`, `agentcore_code_interpreter`, `agentcore_gateway` (awsIam), `inline_function`, plus required-config rejections
- **skills** (`add skill`): path / s3 / git (+auth credential)
- **memory modes** (gated): managed + strategies, existing-by-name + tuning, plus by-arn / no-ref / gated-off rejections
- **truncation**: summarization
- **VPC** network-mode validation (no deploy)
- **CUSTOM_JWT** inbound authorizer + missing-config rejection

## Notes

- **Fast by design.** All cases share a single project (one `create` spawn instead of one per block) and rejection groups use `it.each` tables. Each `runCLI` is a real ~1.3s process spawn, so the suite is kept tight.
- **No duplication.** Pure schema refinements already covered by `src/schema/schemas/primitives/__tests__/harness.test.ts` (top-k / api-base provider gating, skill URI-scheme rejection) are intentionally not re-tested here.
- **Out of scope:** E2E additions (lite_llm deploy, harness-with-tool deploy, CUSTOM_JWT + `fetch access`) — tracked separately.

## Testing

- `vitest run --project integ integ-tests/add-remove-harness.test.ts` → **35 passed** (16 existing + 19 new).
- eslint, prettier, secretlint, and `tsc --noEmit` clean.

### DIFF
--- a/integ-tests/add-remove-harness.test.ts
+++ b/integ-tests/add-remove-harness.test.ts
@@ -8,6 +8,12 @@ async function readHarnessSpec(projectPath: string, harnessName: string) {
   return JSON.parse(await readFile(join(projectPath, `app/${harnessName}/harness.json`), 'utf-8'));
 }
 
+// Placeholder ARNs/keys for config-shape tests (never resolved against AWS).
+const OPENAI_KEY_ARN = 'arn:aws:secretsmanager:us-east-1:123456789012:secret:openai-key';
+const GEMINI_KEY_ARN = 'arn:aws:secretsmanager:us-east-1:123456789012:secret:gemini-key';
+const GATEWAY_ARN = 'arn:aws:bedrock-agentcore:us-east-1:123456789012:gateway/abc123';
+const MEMORY_ARN = 'arn:aws:bedrock-agentcore:us-east-1:123456789012:memory/abc123';
+
 describe('integration: harness add/remove lifecycle', () => {
   let project: TestProject;
   const harnessName = 'TestHarness';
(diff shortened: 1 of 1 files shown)


avi-alpert (dismissed):

tejaskash:
Good catch — verified and fixed in 26bb9e9.

You're right on all three points: `isGatedFeaturesEnabled()` is a strict `=== '1'`, `cleanSpawnEnv` only strips `AGENTCORE_TELEMETRY_DISABLED` (not `ENABLE_GATED_FEATURES`), so the gated-off case inherited the host value and would flip if a shell/CI had it exported.

I went with **option 1** (force the gate off for that one case via `env: { ENABLE_GATED_FEATURES: '' }`) rather than stripping it globally in `cleanSpawnEnv`. Reason: `add-remove-config-bundle.test.ts:113` deliberately *reads* `process.env.ENABLE_GATED_FEATURES` to predict the spawned CLI's behavior — a global strip would make the CLI always-off while that test still computed `expectedBranch='feature-branch'` in a gated-on shell, breaking it. Option 1 is self-contained and touches no shared infra.

Verified the memory-modes block passes both with the gate unset and with `ENABLE_GATED_FEATURES=1` exported (before the fix, the gated-off case failed in the latter).

avi-alpert (approved):

PR #1609, opened 2026-06-22
test(e2e): add harness E2E coverage for lite_llm, tools, and CUSTOM_JWT

tejaskash:
## Summary

Adds three real-AWS E2E scenarios for harness that the existing provider matrix (`harness-bedrock` / `harness-openai` / `harness-gemini`) doesn't cover. Companion to #1608 (integration-layer coverage); these prove behaviors that only surface at CloudFormation synth/deploy or runtime.

## What's added

- **`harness-litellm.test.ts`** — the `lite_llm` provider, routed at a Bedrock model so the deploy needs no third-party API key (LiteLLM's bedrock backend uses the runtime execution role's IAM). Deploy-only (`skipInvoke`) — the bedrock suite already proves invoke; this proves the lite_llm model config is accepted by CloudFormation.
- **`harness-with-tool.test.ts`** — a bedrock harness with an `agentcore_code_interpreter` tool attached via `add tool`. Tool wiring only fails at synth/deploy, so this is the layer that catches it; then invokes to confirm the harness still works. Code interpreter chosen because it needs no external ARN.
- **`harness-custom-jwt.test.ts`** — a harness with a `CUSTOM_JWT` authorizer backed by a Cognito user pool. Asserts `AuthorizerConfiguration` in the CFN template, SigV4 rejection, and bearer-token invoke. Mirrors the existing `byo-custom-jwt.test.ts` pattern.

Also extends `harness-e2e-helper.ts` with `modelId` / `apiBase` / `additionalParams` config (for the lite_llm case).

## Notes

- **Naturally gated.** All tests self-skip without AWS creds. The per-PR `e2e-tests.yml` workflow auto-runs any changed `harness-*.test.ts`, so these run once on this PR; afterward they only run in the sharded full suite (`e2e-tests-full.yml`).
- **`create` has no harness authorizer flags**, so the JWT test patches `harness.json` after create — same approach `byo-custom-jwt.test.ts` uses for `agentcore.json`.
- **Correction vs. original plan:** `fetch access` does **not** support harness (`FetchResourceType = 'gateway' | 'agent'`); the harness JWT-token path lives in `invoke` auto-fetch. The JWT test therefore exercises bearer-token **invoke**, not `fetch access`.
- **VPC deploy intentionally excluded** — too slow/flaky for E2E (VPC validation is covered locally in the integ PR).

## Testing

- `vitest list --project e2e` enumerates all cases correctly (lite_llm's invoke step correctly absent due to `skipInvoke`).
- `tsc --noEmit`, eslint, prettier, secretlint all clean.
- **Not run against AWS from this workstation** (no creds; real deploys) — validated for collection/compile only. CI will execute them on this PR.

### DIFF
--- a/e2e-tests/harness-custom-jwt.test.ts
+++ b/e2e-tests/harness-custom-jwt.test.ts
@@ -2,14 +2,13 @@
  * E2E test: a harness with CUSTOM_JWT inbound auth (Cognito).
  *
  * Creates a Cognito user pool as the OIDC provider, deploys a harness configured with a
- * CUSTOM_JWT authorizer, and verifies that:
+ * CUSTOM_JWT authorizer (added via `add harness` with the JWT + OAuth-credential flags, so
+ * the managed OAuth credential is registered the way a real user would), and verifies that:
  * - Deploy embeds AuthorizerConfiguration in the CloudFormation template
  * - A default SigV4 invocation is rejected (auth method mismatch)
  * - A bearer-token invocation is not rejected for auth reasons
+ * - `fetch access --type harness` mints a CUSTOM_JWT bearer token via the managed credential
  * - Status reports the harness as deployed
- *
- * `create` exposes no authorizer flags for the harness path, so the authorizer is written
- * into harness.json after create (mirrors byo-custom-jwt.test.ts patching agentcore.json).
  *
  * Requires: AWS credentials, npm, git.
  */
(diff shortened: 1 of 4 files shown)


tejaskash:
Updated and marked ready for review.

**Added** a `fetch access --type harness` step to `harness-custom-jwt.test.ts`, and reworked that test to set the harness up the real way — via `add harness --authorizer-type CUSTOM_JWT --discovery-url ... --allowed-audience ... --client-id ... --client-secret ...` — instead of patching `harness.json` directly. That flag path registers the managed OAuth credential + `.env.local` secret, which are the prerequisites for `fetch access` to mint a bearer token. The new step asserts the command returns a CUSTOM_JWT token and that the JWT's `iss`/`client_id` claims match the Cognito pool.

**Verified against real AWS** (us-west-2): the bedrock, litellm, and with-tool harness e2e tests pass end-to-end. The fetch-access flow itself was also verified live (CLI + TUI) on a deployed CUSTOM_JWT harness.

**Dependency:** the `fetch access --type harness` step exercises the feature in #1611, which isn't on main yet. The e2e suite runs manually / in the full sharded suite (not per-PR CI), so this doesn't gate anything — but the new step will only pass once #1611 merges. Suggest merging #1611 first.

avi-alpert (approved):

PR #1617, opened 2026-06-23
Ungate passthrough gateway targets

jariy17:
## What

Makes **passthrough gateway targets** generally available by removing the `isGatedFeaturesEnabled()` (ENABLE_GATED_FEATURES) checks that hid them.

## Changes
- `add/validate.ts` — always include `passthrough` in the valid target-type list; drop the `Passthrough targets are not yet available.` early return.
- `primitives/GatewayTargetPrimitive.ts` — always advertise passthrough options/description in `--help`; `gatePassthroughOption()` is now a pass-through (kept so the six call sites read consistently).
- `tui/screens/mcp/AddGatewayTargetScreen.tsx` — passthrough row is always selectable (no more "Coming soon").
- `add/__tests__/validate.test.ts` — assert passthrough is accepted **without** the flag, and that it appears in the invalid-type error.

Other gated features (knowledge-base, managed memory, aws-skills, previews) are intentionally untouched.

## Testing
- `npm run typecheck`, eslint, prettier, secretlint — pass (pre-commit hooks).
- `npm test src/cli/commands/add src/cli/primitives` — pass.
- `npm run bundle` + installed the local build globally.
- Non-interactive: `agentcore add gateway-target --type passthrough --passthrough-endpoint https://... --outbound-auth gateway-iam-role --signing-service execute-api` → `{"success":true}` **without** `ENABLE_GATED_FEATURES`.
- TUI (via tui-harness): `add → Gateway Target → Target Type` shows **Passthrough — Route to external HTTPS endpoint** (no "Coming soon"), selectable, and the wizard advances through Endpoint → Protocol → Stickiness → Gateway → Outbound Auth → Signing Service → Signing Region → Confirm.

### DIFF
--- a/src/cli/commands/add/__tests__/validate.test.ts
+++ b/src/cli/commands/add/__tests__/validate.test.ts
@@ -469,8 +469,8 @@ describe('validate', () => {
       expect(result.valid).toBe(true);
     });
 
-    // Passthrough is gated behind ENABLE_GATED_FEATURES
-    describe('passthrough feature flag', () => {
+    // Passthrough is generally available (no longer gated)
+    describe('passthrough target type', () => {
       const passthroughOpts: AddGatewayTargetOptions = {
         name: 'pt-target',
         type: 'passthrough',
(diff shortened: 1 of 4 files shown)


src/cli/commands/add/__tests__/validate.test.ts
```diff
@@ -469,40 +469,29 @@ describe('validate', () => {
       expect(result.valid).toBe(true);
     });
 
-    // Passthrough is gated behind ENABLE_GATED_FEATURES
-    describe('passthrough feature flag', () => {
+    // Passthrough is generally available (no longer gated)
```
notgitika: these comments are too specific to the ungating, can we remove them?

src/cli/commands/add/__tests__/validate.test.ts
```diff
@@ -469,40 +469,29 @@ describe('validate', () => {
...
         gateway: 'my-gateway',
         passthroughEndpoint: 'https://api.example.com',
       } as AddGatewayTargetOptions;
 
-      afterEach(() => {
-        delete process.env.ENABLE_GATED_FEATURES;
-      });
-
-      it('rejects passthrough when the flag is off', async () => {
+      it('allows passthrough without any feature flag', async () => {
```
notgitika: some of these tests are too specific to the gating, can we remove them now? or maybe remove them a week or so after release once we verify everything?

notgitika (dismissed):
LGTM; added nits

src/cli/commands/add/__tests__/validate.test.ts
jariy17: Done in f935cbd — removed the ungate-specific comment.

src/cli/commands/add/__tests__/validate.test.ts
jariy17: Stripped the delete process.env.ENABLE_GATED_FEATURES scaffolding in f935cbd, so the two assertions now read like any other target-type test (valid passthrough accepted; passthrough listed in the invalid-type error). Kept the assertions themselves for day-one coverage — happy to prune further post-release if you'd prefer.

notgitika (approved):
thanks for addressing comments!

PR #1619, opened 2026-06-23
fix: gate `agentcore run ingest` behind ENABLE_GATED_FEATURES

tejaskash:
## Problem

Knowledge bases (FMKB) are gated behind the `ENABLE_GATED_FEATURES` flag, but `agentcore run ingest` — which operates **exclusively** on knowledge bases — was reachable regardless of the flag. This was inconsistent with `add knowledge-base` / `remove knowledge-base`, which are properly gated, and effectively leaked an unlaunched FMKB feature in both the CLI and TUI.

## Fix

Match the existing FMKB gating pattern (`KnowledgeBasePrimitive.registerCommands`):

- **CLI** (`src/cli/commands/run/command.tsx`): mark the `ingest` subcommand as `hidden` from `--help` when the gate is off, and reject invocation early with the same `Knowledge bases are not yet available.` message used elsewhere.
- **TUI** (`src/cli/tui/screens/run-eval/RunScreen.tsx`): omit the "Ingest knowledge base" item from the Run menu when the gate is off.

## Testing

Added `src/cli/commands/run/__tests__/run-ingest-gating.test.ts` (drives the built CLI, mirrors `add-knowledge-base.test.ts`):

- `run ingest` exits 1 with "Knowledge bases are not yet available." when the gate is off
- `ingest` hidden from `run --help` when off, visible when `ENABLE_GATED_FEATURES=1`
- with the gate open, the command passes the gate and fails later on KB-name validation (proving the gate no longer blocks)

Verified locally:
- `npm run build` ✅ (CLI bundle builds; unrelated optional `@aws/agent-inspector` asset-copy warning only)
- New test suite: 4/4 passing ✅
- `npm run typecheck` ✅
- `eslint` + `prettier` on changed files ✅

### DIFF
--- a/src/cli/commands/run/__tests__/run-ingest-gating.test.ts
+++ b/src/cli/commands/run/__tests__/run-ingest-gating.test.ts
@@ -1,0 +1,66 @@
+import { runCLI } from '../../../../test-utils/index.js';
+import { randomUUID } from 'node:crypto';
+import { mkdir, rm } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+
+/**
+ * Knowledge bases (FMKB) are gated behind ENABLE_GATED_FEATURES. `agentcore run
+ * ingest` operates exclusively on knowledge bases, so it must be gated to match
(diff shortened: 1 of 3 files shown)


notgitika (approved):
LGTM! Thanks for adding a test file for this but please ensure that there is a comment or an issue that ensures cleanup of this file. 

Please address the automation findings too.

PR #1622, opened 2026-06-23
fix(e2e): broaden CUSTOM_JWT rejection regex to match service response

notgitika:
## What

Fixes the `harness-custom-jwt.test.ts` e2e test that has been failing on mainline since the harness flows were brought into the standard build (PR #1598).

## Root Cause

The test deploys a harness with a CUSTOM_JWT authorizer (Cognito), then invokes without an explicit `--bearer-token`. The CLI auto-attaches a bearer token via the managed OAuth credential. The Cognito `access_token` lacks a standard `aud` claim (Cognito uses `client_id` instead), so the service's `HarnessAuthValidator.validateTokenAudience()` in `YggdrasillDataPlaneService` rejects it with:

```
AgentCore API error (403): {"message":"*** missing required audience claim."}
```

The test's regex only matched `configured for CUSTOM_JWT`, `authoriz(ation|er).*mismatch`, or `different.*authorization` — none of which match the actual service response.

## Fix

Broaden the regex to also accept `missing required audience claim`. The test's intent (SigV4/invalid-auth is rejected with 403) is unchanged — the `exitCode !== 0` assertion still validates the rejection. The negative assertion on line 262 (`not.toMatch`) remains safe because a successful bearer-token invoke will never contain this message.

## Testing

- Verified the regex matches the actual error from the CI logs
- The negative assertion (`expect(output).not.toMatch(...)`) on the bearer-token success path is unaffected since a successful invoke returns the agent response, not an auth error

### DIFF
--- a/e2e-tests/harness-custom-jwt.test.ts
+++ b/e2e-tests/harness-custom-jwt.test.ts
@@ -34,7 +34,8 @@ import { afterAll, beforeAll, describe, expect, it } from 'vitest';
 const hasAws = hasAwsCredentials();
 const canRun = prereqs.npm && prereqs.git && hasAws;
 const region = process.env.AWS_REGION ?? 'us-east-1';
-const customJWTRejectMsgRegex = /configured for CUSTOM_JWT|[Aa]uthoriz(ation|er).*mismatch|different.*authorization/i;
+const customJWTRejectMsgRegex =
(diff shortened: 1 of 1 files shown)


notgitika:
this one makes more sense https://github.com/aws/agentcore-cli/pull/1624

PR #1639, opened 2026-06-25
fix(templates): key per-session agent state by context.session_id across all HTTP starter templates (#808, #809)

### LINKED ISSUE #808: OpenAI Agents, Google ADK, and LangGraph HTTP templates are stateless — no conversation history across invocations

### LINKED ISSUE #809: Strands HTTP template leaks conversation history across sessions and users

aidandaly24:
## Description

The HTTP starter templates rebuild agent/session state on every invocation, so within a single session there is no multi-turn recall (turn 2 cannot see turn 1) — which contradicts `agentcore invoke --session-id`. The Strands no-memory template is worse: it shares one module-level global `_agent` across every invocation in a process, so when one warm microVM serves more than one session, conversation history bleeds between sessions (issue #809). AgentCore Runtime routes a given session ID to the same microVM (microVM stickiness) and keeps that compute warm across invocations, so module-level Python state persists for the session lifecycle — but only if the template reuses the *same* Strands `Agent` object per session. A Strands `Agent` accumulates history on `self.messages`; recall happens only when the same Agent instance is reused, and a leak happens when one Agent is shared across sessions.

This PR keys in-process agent/session state by `context.session_id` in all five HTTP templates, bounded to **128 sessions with LRU eviction** so a long-lived microVM serving many sessions keeps per-session isolation without growing memory without limit:

- **strands**, **autogen** — per-session `OrderedDict` cache (`move_to_end` on hit, evict least-recently-used at the cap).
- **openaiagents** — `@lru_cache(maxsize=128)` over a per-session `SQLiteSession`.
- **googleadk** — module-scoped `InMemorySessionService` bounded via an LRU key set, evicting the oldest via `delete_session`.
- **langchain_langgraph** — module-scoped `InMemorySaver` bounded via an LRU thread set, evicting the oldest via `delete_thread`.

In-memory recall is best-effort: it survives within a session's warm compute but resets on cold start (idle/stop), and `NullConversationManager` does not truncate. For durable, cross-restart history, attach AgentCore Memory (the `hasMemory` template variant) or a managed conversation manager. This was bug-bashed with real deploy + invoke per framework: per-session recall works and there is no cross-session leak on all frameworks.

## Related Issue

Closes #808
Closes #809

## Documentation PR

N/A — starter-template behavior fix; no devguide change required.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I ran `npm run lint`

Additional validation: rendered each template across flag combinations and confirmed valid Python via `py_compile`; multi-agent bug bash with real deploy + invoke per framework verified within-session recall and cross-session isolation.

## Checklist

---

terms of your choice.


src/assets/python/http/autogen/base/main.py
```diff
@@ -91,23 +92,40 @@ def list_files(path: str) -> str:
...
+        _agents.move_to_end(session_id)
+        return _agents[session_id]
+    if len(_agents) >= 128:
+        _agents.popitem(last=False)
     # Get MCP Tools
     mcp_tools = await get_streamable_http_mcp_tools()
-
-    # Define an AssistantAgent with the model and tools
-    agent = AssistantAgent(
+    _agents[session_id] = AssistantAgent(
```
nborges-aws: Potential race condition here given the `await` between the existence check and the dict write. If two concurrent requests arrive for the same `session_id`, both pass the first check, both yield at the await block, both potentially could end up creating a fresh AssistantAgent. The second write would orphan the request who got the first agent.

Suggested fix: recheck after the await

      mcp_tools = await get_streamable_http_mcp_tools()
      if session_id not in _agents:
          _agents[session_id] = AssistantAgent(...)

src/assets/__tests__/googleadk-session-eviction.test.ts
nborges-aws, line 1: This this intentionally just a binary file?

src/assets/python/http/autogen/base/main.py
aidandaly24: Good catch — fixed in 6ed3afcb. Added a re-check after the `await get_streamable_http_mcp_tools()`: if a concurrent first-invocation for the same `session_id` already stored an agent, we keep theirs instead of overwriting (then `move_to_end` + return). No lock needed since the only shared mutation is the dict write, which is now idempotent per session.

src/assets/__tests__/googleadk-session-eviction.test.ts
aidandaly24: Not intentional — the test used literal NUL bytes (`\0`) as the key separator in the eviction simulation, which made git classify the file as binary. Fixed in 6ed3afcb by switching the separator to `::`, so it's now plain UTF-8 text and diffs/reviews normally.

nborges-aws (approved):

PR #1659, opened 2026-06-29
fix(deploy): scope TUI deploy to the picker-selected targets (#1267)

### LINKED ISSUE #1267: All targets are shown even if only 1 was selected
### Description

TUI shows all targets even if only one was selected for deployment

Image

### Steps to Reproduce

1. create an agent harness
2. add new targets to agentcore/aws_targets.json 
3. run `agentcore`
4. select `Deploy`
5. select one of the targets

### Expected Behavior

Only the selected target is show in the next screen (deploy screen)

### Actual Behavior

All targets are shown

Image

### CLI Version

1.0.0-preview.8

### Operating System

macOS

### Additional Context

_No response_

aidandaly24:
## Description

**#1267**

In the TUI deploy flow, when aws-targets.json has more than one target the multi-select picker lets the user choose a subset, but the choice is dropped: the deploy screen header lists every target (cosmetic), and far worse, the deploy itself runs against ALL synthesized stacks (all targets), not the selected one(s). A user who picks a single target silently deploys real infrastructure to every configured account/region. Only affects multi-target projects; single-target (the default) is correct.

## Fix

Thread the picker selection through preflight AND scope the deploy. (1) Display + context (what open PR #1302 does): add selectedTargets?: AwsDeploymentTarget[] to PreflightOptions/DeployFlowOptions; DeployScreen derives it from awsConfig.selectedTargetIndices over availableTargets and passes it down; useCdkPreflight filters preflightContext.awsTargets to the selected names after validateProject(). This fixes the header (DeployScreen.tsx:279) and makes awsTargets[0] reads (persist/teardown/transaction-search; checkBootstrapNeeded preflight.ts:307) reflect the user's choice. (2) ADDITIONALLY (NOT in PR #1302, and the load-bearing part): scope the actual deploy by passing a stacks selector into the TUI cdkToolkitWrapper.deploy() at useDeployFlow.ts:784, built from the selected target names mapped through toStackName(projectName, targetName) with StackSelectionStrategy.PATTERN_MUST_MATCH, e

## Related Issue

Closes #1267

## Documentation PR

N/A — bug fix; no devguide change required.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

How have you tested the change? (full validation in PR comments after bug bash)

- [x] I ran `npm run lint`
- [ ] If I modified `src/assets/`, I ran `npm run test:update-snapshots` and committed the updated snapshots — N/A, no `src/assets/` changes

## Checklist

---

terms of your choice.

### DIFF
--- a/src/cli/tui/screens/deploy/__tests__/useDeployFlow.targets.test.tsx
+++ b/src/cli/tui/screens/deploy/__tests__/useDeployFlow.targets.test.tsx
@@ -1,0 +1,217 @@
+/**
+ * Regression test for issue #1267.
+ *
+ * In the TUI deploy flow a multi-target project synthesizes one stack per configured target.
+ * The deploy was calling `cdkToolkitWrapper.deploy()` with no stacks selector, which the wrapper
+ * forwards as `{ stacks: undefined }` — toolkit-lib defaults that to ALL_STACKS and provisions
+ * infrastructure to EVERY configured account/region, ignoring the picker selection.
+ *
+ * These tests render `useDeployFlow` with a mocked preflight + CDK wrapper and assert the deploy
+ * is scoped to ONLY the selected target's stack name via PATTERN_MUST_MATCH.
+ */
+import { toStackName } from '../../../../commands/import/import-utils';
+import { useDeployFlow } from '../useDeployFlow';
+import { StackSelectionStrategy } from '@aws-cdk/toolkit-lib';
+import { render } from 'ink-testing-library';
+import React from 'react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+// ---- Module mocks -----------------------------------------------------------
+
+const deploySpy = vi.fn().mockResolvedValue(undefined);
+const diffSpy = vi.fn().mockResolvedValue(undefined);
+const disposeSpy = vi.fn().mockResolvedValue(undefined);
+
+const fakeWrapper = {
+  deploy: deploySpy,
+  diff: diffSpy,
+  dispose: disposeSpy,
+};
+
+// switchableIoHost: noop setters so the deploy effect can wire callbacks without crashing.
+const fakeIoHost = {
+  setOnRawMessage: vi.fn(),
+  setOnMessage: vi.fn(),
+  setVerbose: vi.fn(),
+};
+
+// preflightState is mutated per-test before render so the same mock can vary phase/context.
+let preflightState: any;
+
+vi.mock('../../../hooks', async () => {
+  const actual = await vi.importActual<any>('../../../hooks');
+  return {
+    ...actual,
+    useCdkPreflight: () => preflightState,
+  };
+});
+
+// Telemetry wrapper: just run the deploy closure so deploy() actually fires.
+vi.mock('../../../../telemetry/cli-command-run.js', () => ({
+  withCommandRunTelemetry: async (_cmd: string, _attrs: unknown, fn: () => Promise<unknown>) => fn(),
+}));
+
+// persistDeployedState polls getStackOutputs; make it throw fast so the (caught) post-deploy path
+// resolves immediately instead of polling for 15s. The deploy() call has already happened by then.
+vi.mock('../../../../cloudformation', async () => {
+  const actual = await vi.importActual<any>('../../../../cloudformation');
+  return {
+    ...actual,
+    getStackOutputs: vi.fn().mockRejectedValue(new Error('test: skip persist')),
+  };
+});
+
+// Managed-memory notice does a config read; short-circuit it.
+vi.mock('../../../../operations/deploy', async () => {
+  const actual = await vi.importActual<any>('../../../../operations/deploy');
+  return {
+    ...actual,
+    hasManagedMemoryHarness: vi.fn().mockResolvedValue(false),
+    setupTransactionSearch: vi.fn().mockResolvedValue({ success: true }),
+  };
+});
+
+// ---- Fixtures ---------------------------------------------------------------
+
+const TARGET_A = { name: 'prod-east', account: '111111111111', region: 'us-east-1' as const };
+const TARGET_B = { name: 'prod-west', account: '222222222222', region: 'us-west-2' as const };
+const PROJECT_NAME = 'myproj';
+
+function makePreflight(opts: { awsTargets: any[]; projectName?: string }) {
+  const stackNames = opts.awsTargets.map(t => toStackName(opts.projectName ?? PROJECT_NAME, t.name));
+  return {
+    phase: 'complete',
+    steps: [],
+    context: {
+      projectSpec: { name: opts.projectName ?? PROJECT_NAME, runtimes: [] },
+      awsTargets: opts.awsTargets,
+      isTeardownDeploy: false,
+      isFirstDeploy: true, // skip the pre-deploy diff branch
+    },
+    cdkToolkitWrapper: fakeWrapper,
+    stackNames,
+    switchableIoHost: fakeIoHost,
+    hasTokenExpiredError: false,
+    hasCredentialsError: false,
+    missingCredentials: [],
+    allCredentials: {},
+    identityKmsKeyArn: undefined,
+    startPreflight: vi.fn().mockResolvedValue(undefined),
+    confirmTeardown: vi.fn(),
+    cancelTeardown: vi.fn(),
+    confirmBootstrap: vi.fn(),
+    skipBootstrap: vi.fn(),
+    clearTokenExpiredError: vi.fn(),
+    clearCredentialsError: vi.fn(),
+    useEnvLocalCredentials: vi.fn(),
+    useManualCredentials: vi.fn(),
+    skipCredentials: vi.fn(),
+  };
+}
+
+// Harness component: mounts the hook so its effects fire under ink-testing-library.
+function Harness({ selectedTargets }: { selectedTargets?: any[] }) {
+  useDeployFlow({ isInteractive: true, selectedTargets });
+  return null as unknown as React.ReactElement;
+}
+
+async function flush() {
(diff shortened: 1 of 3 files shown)


src/cli/tui/screens/deploy/useDeployFlow.ts
```diff
@@ -779,8 +804,10 @@ export function useDeployFlow(options: DeployFlowOptions = {}): DeployFlowState
 
       try {
         // Run deploy - toolkit-lib handles CloudFormation orchestration
-        // Output goes to stdout via the switchable ioHost
-        await cdkToolkitWrapper.deploy();
+        // Output goes to stdout via the switchable ioHost.
+        // deployStacks restricts the deploy to the picker's selected targets; undefined (single
+        // target or plan path) lets the assembly's lone stack deploy as before.
+        await cdkToolkitWrapper.deploy({ stacks: deployStacks });
```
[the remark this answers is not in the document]
aidandaly24: Addressed in a37a9b5 by taking your Option 3 (resolve the active target in-hook, no dependency on the unmerged #1302).

Added a memoized `activeTarget = selectedTargets?.[0] ?? context?.awsTargets[0]` and routed the post-deploy bookkeeping through it:
- `persistDeployedState()` now derives `currentStackName` via `toStackName(projectName, activeTarget.name)` (instead of `stackNames[0]`) and polls `activeTarget.region` — so a non-first pick polls the stack that was actually deployed and writes deployed-state under the right target name.
- The teardown branch now resolves `targetName`/region from `activeTarget` instead of `awsTargets[0]`.
- Transaction-search setup now reads `activeTarget.region`/`.account`.

Fall-throughs are preserved: the plan path (no project name / no picker) and the no-selection case still fall back to `stackNames[0]` / `context.awsTargets[0]`, matching prior behavior. The TUI persists a single target; when several are picked we bookkeep the first selected one (the same target whose outputs we poll). New regression test asserts the persist path polls B's region + B's stack when B is the non-first pick.

src/cli/tui/screens/deploy/__tests__/useDeployFlow.targets.test.tsx
```diff
@@ -0,0 +1,217 @@
...
+  withCommandRunTelemetry: async (_cmd: string, _attrs: unknown, fn: () => Promise<unknown>) => fn(),
+}));
+
+// persistDeployedState polls getStackOutputs; make it throw fast so the (caught) post-deploy path
+// resolves immediately instead of polling for 15s. The deploy() call has already happened by then.
+vi.mock('../../../../cloudformation', async () => {
+  const actual = await vi.importActual<any>('../../../../cloudformation');
+  return {
+    ...actual,
+    getStackOutputs: vi.fn().mockRejectedValue(new Error('test: skip persist')),
```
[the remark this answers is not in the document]
aidandaly24: Good catch — closed the coverage gap in a37a9b5. The `getStackOutputs` mock is now a hoisted spy (so the test can keep a handle while still short-circuiting the long poll), and a new test `persist polls the SELECTED target stack/region, not awsTargets[0]/stackNames[0]` renders with `selectedTargets=[B]` on an `[A, B]` project and asserts the persist path calls `getStackOutputs(B.region, toStackName(project, B.name))` — and never A's region/stack. Before the hook fix this asserted on A's values, so the regression is now self-enforcing. Also added `scopes the pre-deploy diff to the selected target stack` to cover the diff-scoping fix.

aidandaly24:
Thanks for the thorough review. Both concerns are addressed in a37a9b5 — kept the change surgical and avoided depending on the unmerged #1302.

**1. Post-deploy bookkeeping now tracks the deployed target.** Added a memoized `activeTarget` (`selectedTargets?.[0] ?? context?.awsTargets[0]`) and routed `persistDeployedState` (stack name now via `toStackName(project, activeTarget.name)`, region via `activeTarget.region`), the teardown branch, and transaction-search through it instead of `awsTargets[0]` / `stackNames[0]`. The plan path and no-selection case still fall back to the prior values, so single-target and pre-synthesized flows are unchanged. This removes the state-corruption window for multi-target users who pick a non-first target.

**2. Diff path is now scoped.** All three `cdkToolkitWrapper.diff()` call sites — the pre-deploy diff, the standalone diff-mode run, and the on-demand `requestDiff` — now pass `{ stacks: deployStacks }`, so the diff mirrors exactly what will deploy. `deployStacks` is `undefined` (→ full assembly) when nothing is picked / on the plan path, preserving existing behavior.

Two new regression tests lock both fixes in: persist polls the selected target's region+stack (not `awsTargets[0]`), and the pre-deploy diff is scoped to the selected target's stack. Full local suite: `prettier --check`, `eslint src/`, `typecheck`, build, and the 7-test deploy unit file all green.

src/cli/tui/screens/deploy/useDeployFlow.ts
notgitika: stackNames is the full synthesized set, so after scoping the deploy to target B this still reads 
>"Deployed 2 stack(s): …-A, …-B" 

could derive it from the scoped set, e.g. `deployStacks?.patterns ?? stackNames.`

src/cli/tui/screens/deploy/useDeployFlow.ts
```diff
@@ -147,6 +157,34 @@ export function useDeployFlow(options: DeployFlowOptions = {}): DeployFlowState
   const identityKmsKeyArn = preSynthesized?.identityKmsKeyArn ?? preflight.identityKmsKeyArn;
   const allCredentials = preSynthesized?.allCredentials ?? preflight.allCredentials;
 
+  // Scope the deploy to the picker's selected targets. The vended CDK app synthesizes one stack
+  // per target, so an unscoped deploy resolves to ALL_STACKS and provisions every configured
+  // account/region — even the ones the user did not pick (see issue #1267). Mirrors CLI mode
+  // (commands/deploy/actions.ts), which patterns deploy() by toStackName(project, target).
+  // Skipped on the pre-synthesized plan path, which already targets a single stack.
+  const deployStacks = useMemo(() => {
```
notgitika, line 165: If selectedTargets is empty this returns undefined, which makes CDK deploy everything. Can't happen today (the picker blocks empty selections), but maybe fail safe here instead of falling back to all stacks, just in case that guard ever changes.

src/cli/tui/screens/deploy/useDeployFlow.ts
```diff
@@ -147,6 +157,34 @@ export function useDeployFlow(options: DeployFlowOptions = {}): DeployFlowState
...
+
+  // The target whose stack the post-deploy bookkeeping (persist state, teardown, transaction
+  // search) operates on. The deploy is now scoped to the picker selection, so this MUST track the
+  // deployed target rather than blindly using `awsTargets[0]`: when a multi-target project's user
+  // picks a non-first target, `awsTargets[0]`/`stackNames[0]` point at a stack that was never
+  // deployed, so persist would poll the wrong stack and write deployed-state under the wrong target
+  // (see issue #1267). Falls back to the resolved context target for the plan path (no picker) and
+  // when nothing was selected. The TUI persists a single target; if the user picked several, we
+  // bookkeep the first selected one (the same target whose outputs we poll below).
+  const activeTarget = useMemo(
```
notgitika, line 191: If someone picks two targets (say A and C), we deploy both stacks but only persist state + set up transaction search for A. So C deploys fine but shows up empty in status/invoke. Can we loop over all selected targets here, or at least warn when more than one is picked?

notgitika (changes requested):

src/cli/tui/screens/deploy/useDeployFlow.ts
aidandaly24: Good catch — fixed in d899da3d. The summary now reports the scoped set:

```ts
const deployedStackNames = deployStacks?.patterns ?? stackNames;
setDeployOutput(`Deployed ${deployedStackNames.length} stack(s): ${deployedStackNames.join(', ')}`);
```

So picking a single target prints `Deployed 1 stack(s): <that stack>` instead of listing every synthesized stack. Falls back to `stackNames` on the single-target / plan path (where `deployStacks` is undefined). Added a test asserting a pick of B reports only B's stack and never A's.

src/cli/tui/screens/deploy/useDeployFlow.ts
aidandaly24: Done in d899da3d. It now distinguishes "no picker at all" from "picker returned empty":

- `selectedTargets === undefined` (programmatic / non-interactive callers) → unscoped assembly, unchanged.
- `selectedTargets === []` → scopes to `PATTERN_MUST_MATCH` with **no patterns**, which toolkit-lib turns into a `NoStacksMatched` throw. So an empty pick now **fails safe** (deploys nothing) instead of falling back to ALL_STACKS if the picker's empty-selection guard ever changes.

Added a test asserting `[]` yields a defined selector with empty patterns (not `undefined`).

src/cli/tui/screens/deploy/useDeployFlow.ts
aidandaly24: You're right, and it's a real (pre-existing) gap — persist + transaction search have always tracked a single target. Took the "at least warn" option in d899da3d: when more than one target is picked, a deploy note calls out that deployed-state and transaction search were recorded only for the first selected target and tells the user to re-run `deploy` selecting the others.

The full per-target bookkeeping loop (persist + transaction-search per selected target) is a larger change I kept out of this deploy-scoping PR and tracked as a dedicated follow-up: #1697. Added tests asserting the warning fires on a 2-target pick and not on a single pick.

notgitika: okay yea that issue sounds good

notgitika (approved):

PR #1671, opened 2026-06-30
feat: thread vpcId for container build VPC placement (#297)

aidandaly24:
## Description

For `Container` build-type agents, the CLI's vended CDK app provisions a CodeBuild project and an
orchestrating Lambda to build the agent's container image at deploy time. Customers whose AWS
Organization enforces an SCP denying Lambda deployments outside approved VPCs cannot deploy any
Container-build agent — the build Lambda is created in no VPC and CloudFormation denies it.

This PR threads an optional **`networkConfig.vpcId`** through the CLI so the build infrastructure can
be placed in the same VPC as the runtime (the construct-side change lives in
**aws/agentcore-l3-cdk-constructs#297**). The CLI's Zod schema strips unknown keys by default, so
without this change a user-supplied `vpcId` would be silently dropped before reaching the CDK. The
field is required only for Container builds in VPC mode (CodeBuild's API cannot infer the VPC from
subnets); runtime and the build Lambda infer it from the subnets.

## Fix

| Change | File |
| --- | --- |
| Add optional `vpcId` to `NetworkConfigSchema`; require it for Container + `networkMode: VPC` (and for dockerfile harness builds in VPC mode); cap security groups at 5 for container builds in VPC mode | `src/schema/schemas/agent-env.ts`, `src/schema/schemas/primitives/harness.ts` |
| Tighten VPC/subnet/SG ID validation to the real AWS shape (lowercase hex, exactly 8 or 17 chars); the TUI inherits this, so malformed IDs are rejected inline | `src/cli/commands/shared/vpc-utils.ts`, schema files |
| Add `--vpc-id` validation and a `buildType` argument to `validateVpcOptions` | `src/cli/commands/shared/vpc-utils.ts` |
| Collect, persist, and round-trip `vpcId` across create / add-agent / import / export | `src/cli/commands/create/*`, `src/cli/operations/agent/generate/*`, import/export mappers |
| Collect `vpcId` in the create / add-agent / add-harness TUI wizards (shown only for Container/dockerfile + VPC), via the shared config mapper so all paths persist it | `src/cli/tui/screens/{generate,agent,harness}/*` |
| Add `--vpc-id` flag to `agentcore add harness` | `src/cli/primitives/HarnessPrimitive.ts` |

Non-VPC and CodeZip builds are unaffected: `vpcId` is never required or requested for them.

## Related Issue

No tracking issue — net-new, customer-driven feature (enterprise FSI customer blocked by an SCP that
denies Lambda outside approved VPCs).

**Depends on aws/agentcore-l3-cdk-constructs#297** — the construct-side VpcConfig wiring. This PR's
vended CDK app must compile against the published `@aws/agentcore-cdk` that contains #297, so #297
should merge and release first; this PR then bumps the dependency.

## Documentation PR

N/A.

## Type of Change

- [x] New feature

## Testing

Added unit tests (schema validation for `vpcId`/SG-limit/strict-ID-format; `vpc-utils` validator
tests; mapper round-trip tests proving `vpcId` persists from create/add-agent/add-harness into
`agentcore.json`). Full unit suite green; typecheck + lint clean. Validated interactively (the create
wizard now collects and persists `vpcId` and creates the project; malformed IDs are red-lined inline)
and end-to-end with a real `Container` + VPC deploy into a NAT-equipped VPC — the build Lambda and
CodeBuild both attach to the VPC, the image builds inside the VPC, and the agent invokes successfully.

- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.

### DIFF
--- a/src/cli/aws/agentcore-control.ts
+++ b/src/cli/aws/agentcore-control.ts
@@ -1,4 +1,5 @@
 import type { EvaluationLevel } from '../../schema/schemas/primitives/evaluator';
+import type { NetworkConfig } from '../../schema';
 import { resolveVpcIdFromSubnets } from '../commands/shared/vpc-utils';
 import { getCredentialProvider } from './account';
 import { controlPlaneEndpoint } from './stage-endpoint';
@@ -191,7 +192,7 @@ export interface AgentRuntimeDetail {
   description?: string;
   roleArn: string;
   networkMode: string;
-  networkConfig?: { subnets: string[]; securityGroups: string[]; vpcId?: string };
+  networkConfig?: NetworkConfig;
   protocol: string;
   runtimeVersion?: string;
   entryPoint?: string[];
@@ -231,13 +232,10 @@ export async function getAgentRuntimeDetail(options: GetAgentRuntimeOptions): Pr
   if (networkMode === 'VPC' && response.networkConfiguration?.networkModeConfig) {
     const subnets = response.networkConfiguration.networkModeConfig.subnets ?? [];
     const securityGroups = response.networkConfiguration.networkModeConfig.securityGroups ?? [];
-    // Only resolve the VPC ID for Container builds: that's the only case the schema requires it
-    // (CodeBuild can't infer the VPC). CodeZip+VPC runtimes don't need it, so we avoid the extra
-    // ec2:DescribeSubnets call (and the IAM it would demand) for them.
-    let vpcId: string | undefined;
-    if (isContainer && subnets.length > 0) {
-      vpcId = await resolveVpcIdFromSubnets(subnets, options.region);
-    }
+    // Resolve the VPC ID only for Container builds (the case that requires it — CodeBuild can't infer
+    // the VPC). CodeZip+VPC runtimes don't need it, so this avoids the extra ec2:DescribeSubnets call
+    // (and its IAM) for them. `isContainer` here comes from the runtime's containerConfiguration.
+    const vpcId = isContainer && subnets.length > 0 ? await resolveVpcIdFromSubnets(subnets, options.region) : undefined;
     networkConfig = { subnets, securityGroups, vpcId };
   }
 
--- a/src/cli/aws/agentcore-control.ts
+++ b/src/cli/aws/agentcore-control.ts
@@ -1,4 +1,5 @@
 import type { EvaluationLevel } from '../../schema/schemas/primitives/evaluator';
+import { resolveVpcIdFromSubnets } from '../commands/shared/vpc-utils';
 import { getCredentialProvider } from './account';
 import { controlPlaneEndpoint } from './stage-endpoint';
 import {
@@ -190,7 +191,7 @@ export interface AgentRuntimeDetail {
   description?: string;
   roleArn: string;
   networkMode: string;
-  networkConfig?: { subnets: string[]; securityGroups: string[] };
+  networkConfig?: { subnets: string[]; securityGroups: string[]; vpcId?: string };
   protocol: string;
   runtimeVersion?: string;
   entryPoint?: string[];
@@ -223,13 +224,16 @@ export async function getAgentRuntimeDetail(options: GetAgentRuntimeOptions): Pr
   const response = await client.send(command);
 
   const networkMode = response.networkConfiguration?.networkMode ?? 'PUBLIC';
-  const networkConfig =
-    networkMode === 'VPC' && response.networkConfiguration?.networkModeConfig
-      ? {
-          subnets: response.networkConfiguration.networkModeConfig.subnets ?? [],
-          securityGroups: response.networkConfiguration.networkModeConfig.securityGroups ?? [],
-        }
-      : undefined;
+  let networkConfig: AgentRuntimeDetail['networkConfig'];
+  if (networkMode === 'VPC' && response.networkConfiguration?.networkModeConfig) {
+    const subnets = response.networkConfiguration.networkModeConfig.subnets ?? [];
+    const securityGroups = response.networkConfiguration.networkModeConfig.securityGroups ?? [];
+    let vpcId: string | undefined;
+    if (subnets.length > 0) {
+      vpcId = await resolveVpcIdFromSubnets(subnets, options.region);
+    }
+    networkConfig = { subnets, securityGroups, vpcId };
+  }
 
   const isContainer = !!response.agentRuntimeArtifact?.containerConfiguration;
   const codeConfig = response.agentRuntimeArtifact?.codeConfiguration;
--- a/src/cli/aws/agentcore-control.ts
+++ b/src/cli/aws/agentcore-control.ts
@@ -1,5 +1,5 @@
-import type { EvaluationLevel } from '../../schema/schemas/primitives/evaluator';
 import type { NetworkConfig } from '../../schema';
+import type { EvaluationLevel } from '../../schema/schemas/primitives/evaluator';
 import { resolveVpcIdFromSubnets } from '../commands/shared/vpc-utils';
 import { getCredentialProvider } from './account';
 import { controlPlaneEndpoint } from './stage-endpoint';
@@ -235,7 +235,8 @@ export async function getAgentRuntimeDetail(options: GetAgentRuntimeOptions): Pr
     // Resolve the VPC ID only for Container builds (the case that requires it — CodeBuild can't infer
     // the VPC). CodeZip+VPC runtimes don't need it, so this avoids the extra ec2:DescribeSubnets call
     // (and its IAM) for them. `isContainer` here comes from the runtime's containerConfiguration.
-    const vpcId = isContainer && subnets.length > 0 ? await resolveVpcIdFromSubnets(subnets, options.region) : undefined;
+    const vpcId =
+      isContainer && subnets.length > 0 ? await resolveVpcIdFromSubnets(subnets, options.region) : undefined;
     networkConfig = { subnets, securityGroups, vpcId };
   }
 
--- a/src/cli/aws/agentcore-control.ts
+++ b/src/cli/aws/agentcore-control.ts
@@ -223,20 +223,23 @@ export async function getAgentRuntimeDetail(options: GetAgentRuntimeOptions): Pr
 
   const response = await client.send(command);
 
+  const isContainer = !!response.agentRuntimeArtifact?.containerConfiguration;
+  const codeConfig = response.agentRuntimeArtifact?.codeConfiguration;
+
   const networkMode = response.networkConfiguration?.networkMode ?? 'PUBLIC';
   let networkConfig: AgentRuntimeDetail['networkConfig'];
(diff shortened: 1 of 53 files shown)


aidandaly24:
Thanks for the thorough review — all four findings addressed:

1. **Regex duplication/divergence** — consolidated all VPC/subnet/SG ID patterns into a single source of truth in `src/schema/constants.ts` (canonical lowercase-hex 8/17). `auth.ts`, `agent-env.ts`, and `vpc-utils.ts` now import from it; `auth.ts` re-exports the names so `JwtConfigInput`/`AddHarnessScreen` keep working. The auth.ts subnet pattern was intentionally tightened (lax → hex 8/17) as part of unifying the source of truth. (`b0db7350`)
2. **Stale `agentcore.schema.v1.json`** — regenerated via `build:lib && build:schema`; now has the `vpcId` property and the tightened patterns. (`11301855`)
3. **Harness `--vpc-id` CLI validation** — `validateAddHarnessOptions` now derives the build type from the container flag (matching `parseContainerFlag`) and threads it into `validateVpcOptions`, giving the same friendly error as the agent path. (`d5f30aa1`)
4. **import/export `vpcId` regression** — rather than weaken the schema, both paths now resolve the vpcId from the first subnet via `ec2:DescribeSubnets` (a subnet uniquely determines its VPC), using a shared `resolveVpcIdFromSubnets` helper. Clear error naming `ec2:DescribeSubnets` on failure — no silent vpcId-less writes. Note: this adds an `ec2:DescribeSubnets` IAM requirement to import/export for VPC-mode agents. (`8dd2a1f6`)

Full suite green (5776 tests). The same consolidation landed in the companion CDK PR aws/agentcore-l3-cdk-constructs#297.

aidandaly24:
Re: regenerating `schemas/agentcore.schema.v1.json` — I initially committed the regenerated file, but the repo's `schema-check` CI job explicitly **rejects** schema changes outside release PRs ("schemas/ must not be modified directly — regeneration happens during the release workflow"). So I've reverted that commit; the Zod source changes (the actual driver) remain, and the published JSON schema will regenerate automatically during release. Net: the source-of-truth changes are in, the generated artifact is correctly left to the release pipeline.

aidandaly24:
Addressed the gating over-reach (`0d875609`): vpcId resolution on import/export now fires only when it's actually a **container build**, matching the schema requirement — not for all VPC-mode resources.
- **ARN import** (`agentcore-control.ts`): gated on `isContainer` — a CodeZip+VPC runtime no longer triggers a DescribeSubnets call.
- **Export** (`fetch-harness-spec.ts`): skipped for prebuilt-`containerUri` harnesses (they run no CodeBuild, so never need vpcId); a from-source export still resolves it. Added a test asserting a prebuilt-containerUri VPC harness does NOT resolve.
- **YAML import** already gated correctly on `build === 'Container'`.

Net: the `ec2:DescribeSubnets` call (and its IAM) is now demanded only for the container-build case the schema requires vpcId for. Full suite green (5784).

src/schema/schemas/primitives/harness.ts
```diff
@@ -633,6 +633,26 @@ export const HarnessSpecSchema = z
         path: ['networkConfig'],
       });
     }
+    if (data.networkMode === 'VPC' && data.dockerfile && !data.networkConfig?.vpcId) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message:
+          'networkConfig.vpcId is required for container (dockerfile) builds in VPC mode (CodeBuild cannot infer the VPC from subnets)',
```
tejaskash: This gates vpcId on `dockerfile` only, but `resolveBuildType` treats a `containerUri` as a Container build too. So `add harness --container <ecr-uri> --network-mode VPC` (no vpc-id) validates and persists, then in-project `export harness` maps it to an AgentEnvSpec that fails the Container+VPC vpcId check. Nothing on that path ever fills vpcId in — it's a dead-end config.

src/schema/schemas/agent-env.ts
```diff
@@ -371,6 +376,26 @@ export const AgentEnvSpecSchema = z
         path: ['networkConfig'],
       });
     }
+    if (data.networkMode === 'VPC' && data.build === 'Container' && !data.networkConfig?.vpcId) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        message:
+          'networkConfig.vpcId is required for Container builds in VPC mode (CodeBuild cannot infer the VPC from subnets)',
```
tejaskash: This makes vpcId mandatory for Container+VPC, but Container+VPC agents were already valid and deployable before. There's no backfill on load, so an existing `agentcore.json` with no vpcId now hard-fails at deploy/validate after upgrading. Worth auto-resolving vpcId on load (or at least documenting the migration).

src/cli/aws/agentcore-control.ts
```diff
@@ -222,18 +223,24 @@ export async function getAgentRuntimeDetail(options: GetAgentRuntimeOptions): Pr
...
+  let networkConfig: AgentRuntimeDetail['networkConfig'];
+  if (networkMode === 'VPC' && response.networkConfiguration?.networkModeConfig) {
+    const subnets = response.networkConfiguration.networkModeConfig.subnets ?? [];
+    const securityGroups = response.networkConfiguration.networkModeConfig.securityGroups ?? [];
+    // Only resolve the VPC ID for Container builds: that's the only case the schema requires it
+    // (CodeBuild can't infer the VPC). CodeZip+VPC runtimes don't need it, so we avoid the extra
+    // ec2:DescribeSubnets call (and the IAM it would demand) for them.
+    let vpcId: string | undefined;
+    if (isContainer && subnets.length > 0) {
+      vpcId = await resolveVpcIdFromSubnets(subnets, options.region);
```
tejaskash: This turns a pure control-plane read into one that calls `ec2:DescribeSubnets` and throws on denial. `import-runtime` is the only caller, so `agentcore import` of a Container+VPC runtime now needs ec2:DescribeSubnets where it didn't before. Probably fine, but worth calling out in the IAM docs.

src/cli/commands/add/validate.ts
```diff
@@ -1131,7 +1131,16 @@ export function validateAddHarnessOptions(options: AddHarnessCliOptions): Valida
...
-  const vpcResult = validateVpcOptions(options);
+  // Derive build type from --container: a dockerfile path means Container, absent/URI means CodeZip.
+  const containerValue = options.container;
+  const looksLikeDockerfile =
+    containerValue !== undefined &&
+    (containerValue.endsWith('Dockerfile') ||
+      containerValue.endsWith('.dockerfile') ||
+      containerValue.startsWith('./') ||
+      containerValue.startsWith('../'));
+  const harnessBuildType = looksLikeDockerfile ? 'Container' : undefined;
```
tejaskash: "Is this a container build?" is now decided ~5 different ways across the code: `build==='Container'`, `data.dockerfile`, this filename heuristic, `containerUri||dockerfile`, and `containerConfiguration`. The two bugs above are drift between them. A single `needsVpcId(spec)` predicate would keep these in sync.

src/cli/commands/export/fetch-harness-spec.ts
```diff
@@ -21,7 +22,18 @@ export async function fetchHarnessSpecByArn(
...
   const { harness } = await getHarness({ region, harnessId });
-  return mapApiHarnessToSpec(harness);
+  const result = mapApiHarnessToSpec(harness);
+  const nc = result.spec.networkConfig;
+  // Resolve the VPC ID for any Container build in VPC mode — that's exactly the case the schema
+  // requires it for (CodeBuild can't infer the VPC). A harness is a Container build whenever it
+  // carries a containerUri OR a dockerfile (matching export's resolveBuildType): even a prebuilt
+  // containerUri export emits a `FROM <uri>` Dockerfile stub that CodeBuild builds, so it needs
+  // vpcId too. Only a CodeZip harness (neither) genuinely skips resolution.
+  const isContainerBuild = !!result.spec.containerUri || !!result.spec.dockerfile;
```
tejaskash: This resolve-vpcId guard is hand-written three times (here, agentcore-control, import/actions), each computing `isContainerBuild` differently. Would be cleaner as one helper next to `resolveVpcIdFromSubnets`.

src/cli/aws/agentcore-control.ts
```diff
@@ -190,7 +191,7 @@ export interface AgentRuntimeDetail {
   description?: string;
   roleArn: string;
   networkMode: string;
-  networkConfig?: { subnets: string[]; securityGroups: string[] };
+  networkConfig?: { subnets: string[]; securityGroups: string[]; vpcId?: string };
```
tejaskash: This inline shape now matches the exported `NetworkConfig` type — reuse it instead (same in import/types.ts). Otherwise the next schema field won't propagate here. (AGENTS.md: prefer existing types over inline ones.)

src/cli/commands/shared/vpc-utils.ts
```diff
@@ -52,12 +54,42 @@ export function validateSecurityGroupIds(value: string): true | string {
...
+/**
+ * Resolve the VPC ID from the first subnet via ec2:DescribeSubnets.
+ * Throws a clear error naming the required permission if the call fails.
+ */
+export async function resolveVpcIdFromSubnets(subnetIds: string[], region: string): Promise<string> {
+  const ec2 = new EC2Client({ region, credentials: getCredentialProvider() });
+  let vpcId: string | undefined;
+  try {
+    const resp = await ec2.send(new DescribeSubnetsCommand({ SubnetIds: subnetIds }));
+    vpcId = resp.Subnets?.[0]?.VpcId;
```
tejaskash: Assumes all subnets share one VPC and just takes `Subnets[0]`. If the list spans VPCs (or the first is filtered by permissions) you'll silently record the wrong vpcId, and the error always names `subnetIds[0]` regardless of which subnet failed.

src/cli/commands/create/command.tsx
```diff
@@ -465,6 +466,7 @@ export const registerCreate = (program: Command) => {
     .option('--network-mode <mode>', 'Network mode (PUBLIC, VPC) [non-interactive]')
     .option('--subnets <ids>', 'Comma-separated subnet IDs (required for VPC mode) [non-interactive]')
     .option('--security-groups <ids>', 'Comma-separated security group IDs (required for VPC mode) [non-interactive]')
+    .option('--vpc-id <id>', 'VPC ID (required for Container builds with VPC mode) [non-interactive]')
```
tejaskash, line 470: This flag is registered but dropped on the harness path: `CreateHarnessProjectOptions` has no vpcId field and `handleCreateHarnessCLI`→`createProjectWithHarness`→`add()` never forwards it, even though `HarnessPrimitive.add` supports it. So `create --container ./Dockerfile --network-mode VPC --vpc-id …` discards the value, then the harness schema rejects the write for missing vpcId. The create-agent path (line 407) and `add harness` thread it fine — just this one doesn't.

tejaskash (approved):
LGTM

Issue #1674, opened 2026-06-30
E2E: harness-custom-jwt shard fails — leaked OAuth2 credential providers exhaust account quota


aidandaly24:
## Description

The `E2E Tests (Full Suite)` workflow fails intermittently on the shard that runs `e2e-tests/harness-custom-jwt.test.ts`. The first test (`deploys with CUSTOM_JWT authorizer configuration`) fails at deploy with:

```
{"success":false, ...,
 "error":"The number of agent identity Oauth2 credential providers in this account has reached its limit"}
```

and the remaining tests in the file then cascade (`No deployed targets found`, `expected 'local-only' to be 'deployed'`).

Example failed run: https://github.com/aws/agentcore-cli/actions/runs/28463247986/job/84356322175

## Root Cause

`harness-custom-jwt.test.ts` adds a CUSTOM_JWT harness with `--client-id/--client-secret`, which registers a **managed OAuth2 credential provider** (`<harness>-oauth`). That provider is created imperatively pre-deploy, *outside* the CloudFormation stack, so the test's `remove all` + teardown deploy destroys the stack but leaves the OAuth2 provider behind.

The E2E `globalSetup` stale-cleanup only reaps **ApiKey** credential providers (`ListApiKeyCredentialProvidersCommand`), not OAuth2 providers, and nothing else deletes them. So every run leaks one `E2eHrnsJwt<ts>-oauth` provider permanently. The account caps OAuth2 providers at **50** (quota `L-431051DC`); once full, every CUSTOM_JWT deploy fails with the quota error above.

Confirmed empirically: the E2E account was at exactly **50/50** OAuth2 providers, all named `E2eHrnsJwt<ts>-oauth`, accumulated over ~1 week. (Leaked providers have been manually reaped to unblock CI in the interim.)

## Fix (this PR)

- Add `cleanupStaleOAuth2CredentialProviders` to the E2E `globalSetup` hook so every run auto-reaps leaked `E2e`-prefixed OAuth2 providers older than 30 min (mirrors the existing ApiKey / recommendation / stack reapers).
- Add per-test teardown in `harness-custom-jwt.test.ts` `afterAll` to delete its own managed OAuth2 provider, so a single run self-cleans.

## Acceptance Criteria

- [ ] `globalSetup` deletes stale `E2e`-prefixed OAuth2 providers before tests start.
- [ ] `harness-custom-jwt.test.ts` deletes its `<harness>-oauth` provider on teardown.
[...]

PR #1675, opened 2026-06-30
fix(deploy): delete orphaned credential providers on teardown and removal

### LINKED ISSUE #1673: OAuth2 credential providers leak on teardown (CUSTOM_JWT harness / OAuth gateway) — exhausts 50-provider account quota
## Description

`agentcore deploy` registers a **managed OAuth2 credential provider** as an imperative pre-deploy step (outside the CloudFormation stack) whenever a CUSTOM_JWT harness — or an OAuth-outbound gateway — is configured with `--client-id/--client-secret`. Tearing the project down with `agentcore remove all` + `agentcore deploy` (or `remove` of the individual resource) destroys the CloudFormation stack but **never deletes the OAuth2 credential provider**. The provider is orphaned in the customer's account.

Each AWS account is capped at **50 OAuth2 credential providers** (Service Quotas code `L-431051DC`). Because nothing reaps them, repeated create/teardown cycles accumulate orphaned providers until the quota is exhausted, after which **every** subsequent CUSTOM_JWT/OAuth deploy fails with:

```
The number of agent identity Oauth2 credential providers in this account has reached its limit
```

This is the same root cause that recently broke the E2E suite (`harness-custom-jwt.test.ts`); that side was patched with test-scoped cleanup, but the **underlying product behavior — orphaning OAuth2 providers on teardown — is unfixed for real users.**

## Steps to Reproduce

1. `agentcore create --name MyProj --no-agent`
2. `agentcore add harness --name MyProj --authorizer-type CUSTOM_JWT --discovery-url <url> --allowed-clients <id> --client-id <id> --client-secret <secret>` (registers a managed OAuth credential named `MyProj-oauth`)
3. `agentcore deploy --yes` → a `MyProj-oauth` OAuth2 credential provider is created in AgentCore Identity
4. `agentcore remove all && agentcore deploy --yes` (teardown) → CloudFormation stack is destroyed
5. **Observe:** `aws bedrock-agentcore-control list-oauth2-credential-providers` still shows `MyProj-oauth`. It was never deleted.
6. Repeat the create/deploy/teardown cycle ~50 times (or share an account across many projects/CI) → every CUSTOM_JWT deploy now fails with the quota-limit error above.

## Expected Behavior

Teardown (`remove all` + teardown deploy, and individual resource `remove`) should delete the managed OAuth2 credential provider it created during deploy — the same way payment credential providers are already reaped — so no resource is orphaned in the customer's account.

## Actual Behavior

The OAuth2 credential provider is leaked permanently. Stack destroy does not touch it because it is created imperatively, outside the stack.

## Root Cause

OAuth2 credential providers are created imperatively pre-deploy:

- `src/cli/operations/deploy/pre-deploy-identity.ts` → `setupOAuth2Providers` → `createOAuth2Provider`
- `src/cli/operations/identity/oauth2-credential-provider.ts` → `CreateOauth2CredentialProviderCommand`
- Managed credential name = `computeManagedOAuthCredentialName(resourceName)` → `` `${name}-oauth` `` (`src/cli/primitives/credential-utils.ts`)

But the teardown path only cleans up **payment** credential providers — there is no OAuth2 equivalent:

```ts
// src/cli/commands/deploy/actions.ts (~L536)
if (context.isTeardownDeploy) {
  // Clean up imperative payment credential providers (CFN stack delete handles manager/connector/roles).
  // Harnesses are part of the CloudFormation stack, so stack destroy handles them.   <-- misleading
  ...
  await cleanupPaymentCredentialProviders({ region: target.region, payments: existingPayments });
  ...
  const teardown = await performStackTeardown(target.name);
}
```

The comment "Harnesses are part of the CloudFormation stack, so stack destroy handles them" is true for the harness *runtime* resource but **not** for its managed OAuth2 credential provider, which lives outside the stack. A grep confirms **nothing** in `src/` calls `DeleteOauth2CredentialProvider` — only ApiKey and Payment providers have delete paths.

## Proposed Fix

Add OAuth2 credential provider cleanup to the teardown path, mirroring `cleanupPaymentCredentialProviders`:

1. Add a `deleteOAuth2Provider` / `cleanupOAuth2CredentialProviders` operation in `src/cli/operations/deploy/pre-deploy-identity.ts` (or alongside the existing OAuth2 ops in `src/cli/operations/identity/oauth2-credential-provider.ts`) using `DeleteOauth2CredentialProviderCommand`.
2. Call it from the `isTeardownDeploy` branch in `src/cli/commands/deploy/actions.ts`, iterating the project's `OAuthCredentialProvider` credentials (resolving names via `computeManagedOAuthCredentialName` for managed harness/gateway creds). Make it best-effort like payment cleanup (warn, don't fail teardown).
3. Consider the same on individual-resource `remove` for a harness/gateway that owns a managed OAuth credential, so a single `remove` doesn't orphan the provider either.
4. Be careful not to delete a provider that another still-deployed resource in the project references (only reap providers owned by the resource(s) being torn down).

## Acceptance Criteria

- [ ] After `agentcore remove all` + teardown deploy of a CUSTOM_JWT harness project, `list-oauth2-credential-providers` shows the `<name>-oauth` provider is gone.
- [ ] Removing an individual harness/gateway that owns a managed OAuth credential deletes its OAuth2 provider.
- [ ] Cleanup is best-effort (a delete failure logs a warning and does not abort teardown).
- [ ] Cleanup never deletes a provider still referenced by another deployed resource in the project.
- [ ] Test coverage for the teardown cleanup path.

## Notes

- Quota `L-431051DC` ("Resource OAuth2 credential providers", default 50) is adjustable, but a limit increase only delays the leak — the durable fix is reaping on teardown.
- Related (test-side mitigation, separate PR): added `cleanupStaleOAuth2CredentialProviders` to the E2E `globalSetup` hook and per-test teardown in `harness-custom-jwt.test.ts`. That keeps CI green but does not fix the product behavior described here.

aidandaly24:
## Description

OAuth2 and API key credential providers are created imperatively as a **pre-deploy step, outside the CloudFormation stack** — managed OAuth creds for CUSTOM_JWT harnesses/gateways, API keys for non-Bedrock model providers. Nothing ever deleted them:

- Stack teardown only deletes in-stack resources.
- The teardown path reaped **payment** providers only (`cleanupPaymentCredentialProviders`) — there was no equivalent for OAuth2 or API key providers, and no `Delete*` wrapper for them existed in the CLI at all.

So both teardown flows leaked:

| Flow | Result before this PR |
|------|----------------------|
| `remove all` + teardown deploy | stack destroyed; OAuth2/ApiKey providers **orphaned** |
| `remove <credential>` + redeploy (others remain) | removed cred's provider **orphaned** |

Orphaned providers accumulate against the account's **50-provider** quotas (`L-431051DC` OAuth2 / `L-B04D9A86` API key). Once full, every CUSTOM_JWT deploy fails with:

```
The number of agent identity Oauth2 credential providers in this account has reached its limit
```

This broke the `E2E Tests (Full Suite)` CUSTOM_JWT shard (the account was at exactly 50/50, all `E2eHrnsJwt<ts>-oauth`), but it leaks the **same way for real users** — which is why this is fixed in the CLI, not just in the tests.

### Changes

- **`oauth2-credential-provider.ts` / `api-key-credential-provider.ts`** — add `deleteOAuth2Provider` / `deleteApiKeyProvider`. Idempotent: `ResourceNotFound` is treated as success. No secret deletion — the managed secret lives in the service-owned `bedrock-agentcore-identity!` namespace and is cascade-deleted with the provider (verified empirically).
- **`pre-deploy-identity.ts`** — add `reconcileCredentialProviders`: delete providers recorded in deployed state (`resources.credentials`) that the current spec no longer references, routing by ARN segment (`oauth2credentialprovider` / `apikeycredentialprovider`). Best-effort: failures are collected and logged, never thrown. The control client uses `maxAttempts: 5` so a throttled/transient delete is retried by the SDK rather than silently left as an orphan.
- **`deploy/actions.ts` (CLI) + `useDeployFlow.ts` (TUI)** — snapshot the prior recorded credentials before state is overwritten, then reconcile:
  - **Normal deploy** (partial `remove` + redeploy): reconcile right after deploy — the removed resource is already detached, so its provider deletes cleanly. Deletes just the diff.
  - **Teardown deploy** (`remove all`): reconcile **after `performStackTeardown`**, so the harness/agent runtime that referenced the provider is already gone (deleting it while the stack still references it can fail). Deletes all recorded providers.
  - Payment teardown cleanup is untouched.
- **`harness-custom-jwt.test.ts`** — new regression test asserting the managed `<harness>-oauth` provider exists while deployed and is **gone after teardown**. The post-teardown check retries the `GetOauth2CredentialProvider` call to tolerate read-after-delete eventual consistency. Reverts the earlier test-side band-aid (globalSetup reaper + manual `afterAll` delete).
- **`pre-deploy-identity.test.ts`** — unit tests for `reconcileCredentialProviders`: OAuth2/ApiKey ARN routing, retained-names skipped, payment ARNs untouched, error collection, teardown-deletes-all.

### Safety: imported / user-supplied providers are never deleted

`reconcileCredentialProviders` operates **only** on providers recorded in `deployed-state.resources.credentials`, which is written **only** from providers the CLI created this/a prior deploy (`setup*Providers` results). Imported OAuth providers are skipped at creation (no `discoveryUrl` → `status:'skipped'`, no ARN recorded), and the import flow never writes `resources.credentials`. So a pre-existing/shared provider the user supplied an ARN for is never reaped. Payment ARNs recorded in the same map fall through to their existing dedicated cleanup.

## Related Issue

Closes #1673

Related (CI symptom that surfaced this): #1674

## Documentation PR

N/A — bugfix in existing deploy/teardown behavior.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I ran `npm run lint`

Results:
- `typecheck` ✅ · `lint` ✅ (no new warnings) · `prettier` ✅
- `test:unit` ✅ 5691 passed (393 files) — includes 6 new `reconcileCredentialProviders` tests
- `test:integ` ✅ 314 passed, 1 skipped (32 files)
- **Live e2e** against the test account: the `harness-custom-jwt` suite (which exercises deploy → teardown → assert-provider-gone) passed **4 consecutive runs**, including a manual repro where the teardown log confirmed `Clean up unused credentials → SUCCESS` and the provider returned `ResourceNotFoundException` afterward.

### Note on the commit history

The first version of this fix ran the reconcile *before* stack teardown, which flaked the e2e teardown assertion in CI (a throttled delete against a still-referenced provider was best-effort-swallowed, orphaning it). The follow-up commit reorders the teardown reconcile to run after `performStackTeardown`, adds SDK retries, and makes the e2e assertion tolerant of eventual consistency. Net diff vs `main` is the complete, hardened fix.

## Checklist

---

terms of your choice.

### DIFF
--- a/e2e-tests/global-setup.ts
+++ b/e2e-tests/global-setup.ts
@@ -1,7 +1,4 @@
-import {
-  cleanupStaleCredentialProviders,
-  cleanupStaleOAuth2CredentialProviders,
-} from './utils/credential-provider-cleanup';
+import { cleanupStaleCredentialProviders } from './utils/credential-provider-cleanup';
 import { getLogger } from './utils/logger';
 import { cleanupStaleRecommendations } from './utils/recommendation-cleanup';
 import { cleanUpOldStacks } from './utils/stack-cleanup';
@@ -48,22 +45,6 @@ export default async function setup(_project: TestProject): Promise<() => void> 
   } catch (e) {
     logger.error(String(e));
     logger.warn(`failed to clean up all credential providers`);
-  }
-
-  // OAuth2 credential providers (managed OAuth creds for CUSTOM_JWT harnesses/gateways)
-  // are created imperatively pre-deploy and live outside the CDK stack, so `remove all`
-  // never reaps them. The account caps OAuth2 providers at 50 (quota L-431051DC); leaked
-  // providers accumulate until every CUSTOM_JWT deploy fails with a quota error. Reap them
-  // here the same way as stale stacks and recommendations.
-  logger.info(`cleaning up stale OAuth2 credential providers...`);
-  try {
-    await cleanupStaleOAuth2CredentialProviders(bedrockCPClient, logger.child('oauth2-credential-provider-cleanup'), {
-      minAgeMs: 30 * 60 * 1000,
-      prefix: 'E2e',
-    });
-  } catch (e) {
-    logger.error(String(e));
-    logger.warn(`failed to clean up all OAuth2 credential providers`);
   } finally {
     bedrockCPClient.destroy();
   }
--- a/e2e-tests/global-setup.ts
+++ b/e2e-tests/global-setup.ts
@@ -1,4 +1,7 @@
-import { cleanupStaleCredentialProviders } from './utils/credential-provider-cleanup';
+import {
+  cleanupStaleCredentialProviders,
+  cleanupStaleOAuth2CredentialProviders,
+} from './utils/credential-provider-cleanup';
 import { getLogger } from './utils/logger';
 import { cleanupStaleRecommendations } from './utils/recommendation-cleanup';
 import { cleanUpOldStacks } from './utils/stack-cleanup';
@@ -45,6 +48,22 @@ export default async function setup(_project: TestProject): Promise<() => void> 
   } catch (e) {
     logger.error(String(e));
     logger.warn(`failed to clean up all credential providers`);
+  }
+
+  // OAuth2 credential providers (managed OAuth creds for CUSTOM_JWT harnesses/gateways)
+  // are created imperatively pre-deploy and live outside the CDK stack, so `remove all`
+  // never reaps them. The account caps OAuth2 providers at 50 (quota L-431051DC); leaked
+  // providers accumulate until every CUSTOM_JWT deploy fails with a quota error. Reap them
+  // here the same way as stale stacks and recommendations.
+  logger.info(`cleaning up stale OAuth2 credential providers...`);
+  try {
+    await cleanupStaleOAuth2CredentialProviders(bedrockCPClient, logger.child('oauth2-credential-provider-cleanup'), {
+      minAgeMs: 30 * 60 * 1000,
+      prefix: 'E2e',
+    });
+  } catch (e) {
+    logger.error(String(e));
+    logger.warn(`failed to clean up all OAuth2 credential providers`);
   } finally {
     bedrockCPClient.destroy();
   }
--- a/e2e-tests/harness-custom-jwt.test.ts
+++ b/e2e-tests/harness-custom-jwt.test.ts
@@ -15,9 +15,11 @@
 import { computeManagedOAuthCredentialName } from '../src/cli/primitives/credential-utils.js';
 import { hasAwsCredentials, parseJsonOutput, prereqs, runCLI, stripAnsi } from '../src/test-utils/index.js';
 import { installCdkTarball, writeAwsTargets } from './e2e-helper.js';
-import { deleteOAuth2CredentialProvider } from './utils/credential-provider-cleanup.js';
-import { getLogger } from './utils/logger.js';
-import { BedrockAgentCoreControlClient } from '@aws-sdk/client-bedrock-agentcore-control';
+import {
+  BedrockAgentCoreControlClient,
+  GetOauth2CredentialProviderCommand,
+  ResourceNotFoundException,
+} from '@aws-sdk/client-bedrock-agentcore-control';
 import { CloudFormationClient, GetTemplateCommand } from '@aws-sdk/client-cloudformation';
 import {
   CognitoIdentityProviderClient,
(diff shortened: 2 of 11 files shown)


aidandaly24:
Pushed a hardening commit (`536d04de`) after the first version flaked the e2e teardown assertion in CI:

**What happened:** the e2e regression test (`teardown deletes the managed OAuth2 credential provider`) failed once in CI — the provider survived teardown — even though it passed locally. Root cause: the reconcile ran *before* `performStackTeardown`, so a transient/throttled `DeleteOauth2CredentialProvider` (the harness runtime still referencing it, under concurrent e2e shard load) could fail and was best-effort-swallowed, leaving the provider orphaned.

**Hardening:**
1. On a teardown deploy, reconcile now runs **after** `performStackTeardown` (stack/harness gone → no reference) — both CLI and TUI paths.
2. Reconcile control client uses `maxAttempts: 5` so throttled deletes are retried, not silently dropped.
3. The e2e assertion now retries the GET to tolerate read-after-delete eventual consistency.

**Verification:** the originally-flaky `harness-custom-jwt` e2e test now passes **4 consecutive live runs** against the test account; full unit (5691) + integ (314) green.

PR #298, opened 2026-02-12
chore: bump bedrock-agentcore minimum to >= 1.3.0

jariy17:
## Summary
- Bumps `bedrock-agentcore` floor pin from `>= 1.0.3` to `>= 1.3.0` across all 6 Python framework templates
- Updates snapshot tests to match

## Why
v1.3.0 includes important changes since 1.0.3:
- **Removed deprecated APIs**: `save_turn()` and `process_turn()` methods were removed (verified no template code uses these)
- **Bug fixes**: `download_file`/`download_files` binary content crash, workload identity endpoint URLs, bytes serialization, empty text messages breaking Converse API
- **Features**: Bidirectional streaming, episodic memory strategies, memory metadata support, WebSocket session_id support, code interpreter convenience methods

Bumping the floor pin ensures new projects start with a consistent, stable baseline.

## Test plan
- [x] Verified no template code uses removed `save_turn()`/`process_turn()` APIs
- [x] Unit tests pass (292/292)
- [x] Snapshot tests updated and passing (6 snapshots updated)

PR #1681, opened 2026-07-02
fix: update KB and WS connector devEx

nborges-aws:
## Description

Update CLI connector DevEx for web-search and knowledge base. This PR restructures how connector targets are stored and managed.

Web-search moves from standalone targetType to connector
  - Removed `agentcore add web-search` / `agentcore remove web-search` top-level shortcuts
  - Removed web-search from `--type` valid values; added to `--connector`
  - Canonical path: `agentcore add gateway-target --type connector --connector web-search`

configurations[] array replaces flat typed fields
  - Connector targets now store a configurations array matching the CFN wire format
  - Each entry: `{ name, parameterValues, parameterOverrides }`
  - CLI bespoke flags (`--exclude-domains`, `--knowledge-base-id`) translate to the correct parameterValues shape via a connector translator module

bedrock-agentic-retrieve removed as standalone connector
  - AgenticRetrieveStream is now a configuration entry within bedrock-knowledge-bases
  - Removed auto-upsert logic (upsertAgenticRetrieveTarget)
  - Single `--knowledge-base-id` flag produces both Retrieve and AgenticRetrieveStream configurations

KB --gateway flag removed
  - Gateway wiring is now a separate operation via `agentcore add gateway-target --connector bedrock-knowledge-bases`
  - Removed gateway step from KB TUI wizard
  - Removed `appendConnectorTargets` dead code

--name optional for KB paths
  - agentcore add knowledge-base: defaults to kb-quick-start-xxxxx (console convention)
  - `agentcore add gateway-target --connector bedrock-knowledge-bases`:defaults to kb-{gateway}-{kb-id}
  - Collision-safe (checks existing names, sanitizes underscores)

  6. TUI fallback for agentcore add gateway-target
  - Bare command (no flags) now drops into the TUI wizard instead of erroring
  - Flag detection uses Object.entries(rawOptions) — no hand-maintained allowlist

## Related Issue

Closes #

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Other (please describe): devEx rework

## Testing

- [x] I ran `npm run lint`

## Checklist

- [ ] I have added any necessary tests that prove my fix is effective or my feature works
- [ ] My changes generate no new warnings

---

terms of your choice.


src/cli/primitives/KnowledgeBasePrimitive.ts
```diff
@@ -601,17 +561,14 @@ export class KnowledgeBasePrimitive extends BasePrimitive<AddKnowledgeBaseOption
           }
 
           await runCliCommand('add.knowledge-base', !!cliOptions.json, async () => {
-            if (!cliOptions.name) {
-              throw new ValidationError('A --name is required for `agentcore add knowledge-base`.');
-            }
+            const resolvedName = cliOptions.name ?? `kb-quick-start-${Math.random().toString(36).slice(2, 7)}`;
```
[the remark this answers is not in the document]
nborges-aws: The randomness here is intentional. Follows console convention.

for 1) -> Producing a new KB in this case would be the expectation. If the first attempt succeeds, name is printed in the output and user can proceed. If attempt fails -- nothing is persisted from first attempt, so running again and getting a fresh resource is warranted.

for 2) -> This is the correct behavior. `agentcore add knowledge-base` is creating a new resource per run. If a user creates a KB, and wants to append on the next command, they can pas sin the name output from first run. Default behavior creating two disjointed KBs is the expectation.

src/cli/primitives/GatewayTargetPrimitive.ts
```diff
@@ -439,6 +432,38 @@ Target types and their options:
           process.exit(1);
         }
 
+        const userPassedAnyFlag =
+          !!cliOptions.name ||
+          !!cliOptions.type ||
+          !!cliOptions.connector ||
+          !!cliOptions.endpoint ||
+          !!cliOptions.json;
```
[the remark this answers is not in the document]
nborges-aws: expanded `||` chain to avoid unintentional drop into TUI

src/cli/commands/add/validate.ts
```diff
@@ -394,7 +394,9 @@ export async function validateAddGatewayTargetOptions(options: AddGatewayTargetO
     options.language =
       (matchEnumValue(TargetLanguageSchema, options.language) as typeof options.language) ?? options.language;
 
-  if (!options.name) {
+  const kbConnectors = ['bedrock-knowledge-bases', 'bedrock-agentic-retrieve'];
+  const nameOptional = options.type === 'connector' && kbConnectors.includes(options.connector ?? '');
```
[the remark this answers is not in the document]
nborges-aws: This is intentional, I've updated PR description to clarify

src/cli/commands/add/validate.ts
```diff
@@ -682,13 +645,13 @@ export async function validateAddGatewayTargetOptions(options: AddGatewayTargetO
         error: `Invalid --connector "${options.connector}". Valid: ${validConnectors}`,
       };
     }
-    if (!options.knowledgeBaseId || options.knowledgeBaseId.length === 0) {
+    if (options.connector !== 'web-search' && (!options.knowledgeBaseId || options.knowledgeBaseId.length === 0)) {
       return {
         valid: false,
         error: `--knowledge-base-id is required for --connector ${options.connector}`,
       };
     }
```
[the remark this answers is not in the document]
nborges-aws: Added explicit check & rejection

tejaskash (commented):
Automated correctness + cleanup review (high effort). 10 findings; the first three are potential merge blockers because they cross the CLI↔L3-CDK boundary or reverse a deliberate gating decision. Inline comments below. One additional finding could not be anchored inline because its file is not in this diff:

- **`e2e-tests/httpgateway-all-targets.test.ts:290`** — the assertion `expect(types.has('webSearch')).toBe(true)` still expects a `webSearch` target, but the CLI now writes web-search as `targetType: 'connector'` (`connectorId: 'web-search'`). This test will fail; it wasn't updated to the new shape.

src/cli/primitives/GatewayTargetPrimitive.ts
```diff
@@ -1404,7 +1285,8 @@ Target types and their options:
 
     const target: AgentCoreGatewayTarget = {
       name: config.name,
-      targetType: 'webSearch',
+      targetType: 'connector',
+      connectorId: 'web-search',
```
tejaskash, line 1227: **[Blocker — cross-repo contract break]** This now writes web-search targets as `targetType: 'connector'` + `connectorId: 'web-search'`, but the synth/deploy engine lives in the separate `@aws/agentcore-cdk` L3 construct (pinned `^0.1.0-alpha.19` in `src/assets/cdk/package.json`), which has **no handler for this shape**:

- Its `ConnectorIdSchema` enum only allows `bedrock-knowledge-bases` / `bedrock-agentic-retrieve`, so the project fails schema validation at deploy.
- Even past that, `Gateway.ts` routes every `targetType === 'connector'` into a branch that handles only the two KB connectors and ends in `throw new Error('Unknown connectorId "web-search"')`.
- The dedicated `targetType === 'webSearch'` synth branch (InvokeWebSearch grant + `domainFilter`) is now dead for CLI-authored targets.

Net: `agentcore add gateway-target --connector web-search` succeeds but the next `agentcore deploy` fails. The L3 construct must be updated in lockstep (route `connectorId: 'web-search'` on the connector path) and released before this ships.

src/schema/schemas/mcp.ts
```diff
@@ -20,7 +20,6 @@ export const GatewayTargetTypeSchema = z.enum([
   'httpRuntime',
   'connector',
   'passthrough',
-  'webSearch',
 ]);
 export type GatewayTargetType = z.infer<typeof GatewayTargetTypeSchema>;
```
tejaskash, line 24: **[Blocker — backward-compat break]** Dropping `'webSearch'` from `GatewayTargetTypeSchema` with no migration means any existing `agentcore.json` that already contains a `targetType: 'webSearch'` target (written by a prior/preview CLI) now fails `AgentCoreProjectSpecSchema.parse` in `readProjectSpec`. Every command that loads the project — `status`, `deploy`, `add`, `remove`, `export` — throws `ConfigValidationError` with no upgrade path. Consider keeping `'webSearch'` accepted on read (normalize to the connector shape on load) even if the CLI no longer writes it.

src/cli/primitives/GatewayTargetPrimitive.ts
```diff
@@ -302,21 +297,18 @@ export class GatewayTargetPrimitive extends BasePrimitive<AddGatewayTargetOption
       .option('--type <type>', typeDescription)
       .option(
         '--connector <id>',
-        'Connector id (for connector type): bedrock-knowledge-bases or bedrock-agentic-retrieve [non-interactive]'
+        'Connector id (for connector type): bedrock-knowledge-bases, bedrock-agentic-retrieve, or web-search [non-interactive]'
```
tejaskash: **[Blocker — confirm intent]** This PR removes *all* `ENABLE_GATED_FEATURES` gating on web-search (the `isGatedFeaturesEnabled` import, the `'not yet available'` throws, the `hidden: !isGatedFeaturesEnabled()` shortcuts, and the TUI `'Coming soon'`/disabled state). Commit #1625 (`chore: re-gate web search`) deliberately re-gated this. Meanwhile `KnowledgeBasePrimitive` still gates its `add` command. Given finding #1 (web-search can't currently deploy), ungating it by default looks like an unintended leak rather than a coordinated launch. Please confirm this is intentional.

src/schema/schemas/mcp.ts
```diff
@@ -800,10 +799,10 @@ export const AgentCoreGatewayTargetSchema = z
           });
```
tejaskash: **[Correctness — lost validation]** The deleted `targetType === 'webSearch'` block used to reject `knowledgeBaseId`, `knowledgeBaseIds`, `httpRuntime`, and `passthrough` on a web-search target. In the new `connector` branch, the KB-field rejections live only inside the `connectorId === BEDROCK_KNOWLEDGE_BASES` / `BEDROCK_AGENTIC_RETRIEVE` guards, and `httpRuntime`/`passthrough` aren't rejected at all — so `{ targetType: 'connector', connectorId: 'web-search', knowledgeBaseId: 'ABCDE12345' }` (or a `passthrough`/`httpRuntime` block) now passes validation silently and is carried into synth. Add an explicit `connectorId === 'web-search'` guard that rejects `knowledgeBaseId`/`knowledgeBaseIds` (and ideally the sub-configs).

src/cli/primitives/GatewayTargetPrimitive.ts
```diff
@@ -633,11 +641,44 @@ Target types and their options:
...
+            // KB connectors require --knowledge-base-id
             const kbRefs = cliOptions.knowledgeBaseId ?? [];
             if (kbRefs.length === 0) {
               throw new ValidationError(`--knowledge-base-id is required for --connector ${connectorId}.`);
             }
 
+            const resolvedName =
+              cliOptions.name ??
+              (connectorId === 'bedrock-knowledge-bases'
+                ? `kb-${cliOptions.gateway}-${kbRefs[0]}`
```
tejaskash: **[Correctness — invalid auto-name reaches deploy]** `kb-${cliOptions.gateway}-${kbRefs[0]}` embeds the raw KB reference. KB names permit underscores (`/^[a-zA-Z][a-zA-Z0-9_-]{0,47}$/`), but gateway tool names forbid them (`ToolNameSchema` = `/^[a-zA-Z][a-zA-Z0-9-]*$/`). The target is written with only `z.string().min(1)` validation, so `--knowledge-base-id my_kb` yields `kb-gw-my_kb`, persists locally, then is rejected server-side at `agentcore deploy` — a confusing failure long after `add` succeeded. Long `gateway`+`kb` names can also exceed the 128-char limit. Validate the derived name against `ToolNameSchema` (and sanitize/truncate) before writing. This is newly introduced — the base code required an explicit `--name`.

src/cli/primitives/GatewayTargetPrimitive.ts
```diff
@@ -439,6 +432,49 @@ Target types and their options:
           process.exit(1);
         }
 
+        const userPassedAnyFlag =
```
tejaskash: **[Correctness — incomplete flag detection]** This allowlist omits several real non-interactive flags: `--tool-schema-file`, `--credential-name`, `--stage`, `--oauth-client-id/secret/discovery-url/scopes`, `--tool-filter-path/methods`, `--schema-s3-account`, `--description`. Invoking with only one of those sets `userPassedAnyFlag = false`, so in a TTY the command silently drops into the interactive `AddFlow` (discarding the flags), and in CI `requireTTY()` throws — instead of validating and erroring on the missing `--type`. Consider deriving "any flag passed" from the presence of any option key rather than an explicit hand-maintained OR-chain.

src/cli/primitives/KnowledgeBasePrimitive.ts
tejaskash: **[Correctness — silent data merge on collision]** The auto-generated name has no uniqueness check. On a collision with an existing KB, `add()` finds that KB and routes to `appendToExisting()`, silently merging the new data sources into an unrelated KB instead of creating the intended new one — with no error surfaced. `generateUniqueName(base, existingNames)` already exists in `tui/utils` and the project's KB names are available here; use it to guarantee uniqueness.

src/cli/commands/add/validate.ts
```diff
@@ -682,13 +645,19 @@ export async function validateAddGatewayTargetOptions(options: AddGatewayTargetO
...
-    if (!options.knowledgeBaseId || options.knowledgeBaseId.length === 0) {
+    if (options.connector === 'web-search' && options.knowledgeBaseId && options.knowledgeBaseId.length > 0) {
+      return {
+        valid: false,
+        error: '--knowledge-base-id is not applicable for --connector web-search',
+      };
+    }
+    if (options.connector !== 'web-search' && (!options.knowledgeBaseId || options.knowledgeBaseId.length === 0)) {
```
tejaskash, line 654: **[Correctness — lost early error]** The connector `irrelevant` list (just below) dropped `['schema','--schema']` and `['schemaS3Account','--schema-s3-account']`, both of which were in the deleted `WEB_SEARCH_DISALLOWED_OPTIONS`. So `--connector web-search --schema s3://...` no longer gets the clean early CLI error `'--schema is not applicable'`, and a bare `--schema-s3-account` (no `--schema`) slips past the CLI entirely. Lower severity since a materialized `schemaSource` may still trip the schema layer, but the early feedback is gone.

src/cli/primitives/KnowledgeBasePrimitive.ts
```diff
@@ -218,18 +205,9 @@ export class KnowledgeBasePrimitive extends BasePrimitive<AddKnowledgeBaseOption
...
-      // --gateway: append the connector targets — one Retrieve per KB plus the
-      // shared gateway-scoped agentic-retrieve target (this KB gets appended
-      // to its knowledgeBaseIds[]).
-      if (options.gateway) {
-        this.appendConnectorTargets(project, options.gateway, kb.name, kb.name);
-      }
-
       await this.writeProjectSpec(project);
 
       if (!options.json) for (const w of warnings) console.warn(w);
```
tejaskash, line 206: **[Cleanup — dead code + defunct field]** Removing the `--gateway` wiring here leaves `appendConnectorTargets` and its only callee `upsertRetrieveTarget` (~lines 238–278) with no callers (`git grep appendConnectorTargets` returns only the definition), and the `kb.gateway` schema field (`primitives/knowledge-base.ts`) is now write-dead — only read in remove/preview branches that are unreachable for CLI-created projects. `noUnusedLocals` is `false`, so nothing flags this. Also: an existing project whose KB already has `gateway` set no longer gets its connector target re-ensured on a subsequent `add` (that `if (existing.gateway) appendConnectorTargets(...)` block was deleted). Delete the dead cluster + field, or confirm the legacy remove-cascade still matters.

src/cli/primitives/GatewayTargetPrimitive.ts
nborges-aws: Intentional. Ungated web-search experience is shipping with this release

src/schema/schemas/mcp.ts
nborges-aws: Web-search was behind `ENABLE_GATED_FEATURES` and never released. Zero customer impact

src/cli/primitives/GatewayTargetPrimitive.ts
nborges-aws: Updates to [CDK PR #298](https://github.com/aws/agentcore-l3-cdk-constructs/pull/298) handles this

src/schema/schemas/mcp.ts
nborges-aws: added explicit rejections for passthrough and httpRuntime

src/cli/primitives/GatewayTargetPrimitive.ts
nborges-aws: Added logic to sanitize & truncate name before writing

src/cli/primitives/GatewayTargetPrimitive.ts
nborges-aws: Replaced OR-chain with a generic check as recommended

src/cli/primitives/KnowledgeBasePrimitive.ts
nborges-aws: updated logic to validate no collision with existing name before proceeding

src/cli/commands/add/validate.ts
nborges-aws: I've added `--schema/--schema-s3-account` to the connector irrelevant list

src/cli/primitives/KnowledgeBasePrimitive.ts
nborges-aws: valid. deleted dead code in latest commit

src/cli/commands/add/validate.ts
```diff
@@ -437,11 +437,11 @@ export async function validateAddGatewayTargetOptions(options: AddGatewayTargetO
   }
   options.type = mappedType;
 
-  // --exclude-domains is webSearch-target-only. Reject it on every other target type.
-  if (mappedType !== 'webSearch' && options.excludeDomains) {
```
jariy17, line 441: nit: We should consider a cleaner way to do flag validation in our rf.

src/cli/primitives/GatewayTargetPrimitive.ts
```diff
@@ -439,6 +432,35 @@ Target types and their options:
...
         }
 
+        const userPassedAnyFlag = Object.entries(rawOptions).some(
+          ([, v]) => v !== undefined && v !== false && !(Array.isArray(v) && v.length === 0)
+        );
+        if (!userPassedAnyFlag) {
+          try {
+            requireTTY();
+            const [{ render }, { default: React }, { AddFlow }] = await Promise.all([
+              import('ink'),
```
jariy17, line 442: nit: why are you import within a function?

jariy17 (approved):
LGTM just small nits:
  ▎ few low-pri cleanups from the #1681 review, none are blockers:
  ▎ - parameterValues/parameterOverrides are optional in the zod schema (mcp.ts:473) but 
  ▎ required in the ConfigurationEntry type — should reconcile those
  ▎ - no tests for the kb-{gateway}-{kb-id} auto-name path or the bare-command TUI fallback 
  ▎ detection
  ▎ - the reverse-index KB-walk in status/action.ts is duplicated (~L265 and ~L563), might be 
  ▎ worth extracting

PR #1682, opened 2026-07-02
fix(vpc): correct create-harness VPC validation, dry-run persistence, and TUI review display (follow-up to #1671)

### LINKED ISSUE #1691: create --network-mode VPC fails for harness projects: '--subnets is required' even when --subnets is passed
## Summary

Creating a harness project in VPC mode with `agentcore create` fails with `--subnets is required when network mode is VPC`, even when `--subnets` (and `--security-groups` / `--vpc-id`) are supplied. The same VPC configuration works through the agent path of `create` and through `add harness`, so a VPC harness is a supported configuration that simply can't be reached via `create`.

## Reproduction

On `@aws/agentcore@0.22.0`:

```bash
mkdir repro && cd repro
echo "FROM public.ecr.aws/lambda/python:3.12" > Dockerfile

agentcore create --name vh --container ./Dockerfile \
  --network-mode VPC \
  --subnets subnet-0123456789abcdef0 \
  --security-groups sg-0123456789abcdef0 \
  --vpc-id vpc-0123456789abcdef0 \
  --model-provider bedrock --dry-run
```

**Actual:** exit 1, `--subnets is required when network mode is VPC` (even though `--subnets` was passed).

**Expected:** the project is created (or, for `--dry-run`, the preview succeeds). When `--vpc-id` is omitted for a container build, the expected error is `--vpc-id is required for Container builds with --network-mode VPC`.

## Scope

- Affects: `agentcore create` for harness projects (any `--container` value) in `--network-mode VPC`.
- Not affected: `agentcore create` for agent projects (`--build Container --network-mode VPC ...` works), and `agentcore add harness --network-mode VPC ...` (works and persists `networkConfig.vpcId`).

## Root cause

`validateCreateHarnessOptions` validates the VPC options, but the two call sites in `src/cli/commands/create/command.tsx` (the `--dry-run` path and the main path) do not pass `container`, `subnets`, `securityGroups`, or `vpcId` into it. Those fields read as undefined inside the validator, so the "subnets required" check triggers even when they were provided, and the intended `--vpc-id` validation is never reached.

## Fix

Forward `container`, `subnets`, `securityGroups`, and `vpcId` into `validateCreateHarnessOptions` at both call sites, bringing `create` for harnesses to parity with `add harness` and the agent path.

aidandaly24:
## Description

Follow-up to #1671. Three issues were found in the container-build VPC support after that PR merged; this PR corrects them.

### 1. `agentcore create --network-mode VPC` for a harness project (#1691)

`validateCreateHarnessOptions` validates the VPC options (`--subnets`, `--security-groups`, `--vpc-id`, and the build type derived from `--container`), but the two call sites in `command.tsx` (the dry-run path and the main path) were not passing those fields into it. As a result the validator saw them as undefined and returned `--subnets is required when network mode is VPC` even when the flags were supplied, so a VPC-mode harness could not be created through `agentcore create`, and the intended errors (missing `--vpc-id`, malformed `--vpc-id`, security-group limit) were never reached. The same VPC configuration works through the agent path of `create` and through `add harness`, so this path was simply unreachable.

Fix: forward `container`, `subnets`, `securityGroups`, and `vpcId` into `validateCreateHarnessOptions` at both call sites, bringing it to parity with `add harness` and the agent path.

The existing unit tests called `validateCreateHarnessOptions` directly with a fully-populated object, so they did not cover the wiring in `command.tsx`. Added integration tests that run the command through `runCLI --dry-run` to exercise that path.

### 2. `deploy --dry-run` writing to config files

The VPC-ID backfill on deploy resolves a missing `networkConfig.vpcId` from the subnets and writes it to `agentcore.json` / `harness.json` so the CDK synth subprocess (which reads config from disk) can use it. This write also happened under `--dry-run` and `--diff`, which are meant to preview without changing anything, so a preview left the config files modified.

Fix: `backfillContainerVpcIds` now takes a `persist` flag. In preview modes it still writes the resolved value to disk for synth, then returns a `restore()` callback that reverts the files to their original contents afterward, so the working tree is unchanged. The plan and diff paths call `restore()`.

### 3. TUI review screen showing a stale VPC ID

In the interactive `agentcore create` wizard, selecting Container + VPC, entering a VPC ID, then going back and switching the build to CodeZip left the VPC ID set. The review screen then displayed a VPC ID line for a CodeZip agent, and the value could be written into `agentcore.json`.

Fix: guard the review-screen render and the schema-mapper write on `buildType === 'Container'`, so a VPC ID only appears and persists for container builds.

## Related Issue

Closes #1691 (issue #1 above). Issues #2 and #3 were found during follow-up testing of #1671 and do not have separate tracking issues.

## Type of Change

- [x] Bug fix

## Testing

- [x] I ran `npm run test:unit` — full unit suite passing (6120 tests)
- [x] I ran `npm run lint`
- [x] Added integration tests that run `agentcore create` through the CLI dispatch for the VPC harness path, plus tests for the dry-run restore behavior

Verified end-to-end on a build of this branch (rebased onto latest `main`):
- `create --container ./Dockerfile --network-mode VPC --subnets ... --security-groups ... --vpc-id ...` now succeeds and persists `networkConfig.vpcId` to `harness.json` (previously exited 1 with the misleading `--subnets is required`).
- Omitting `--vpc-id` now returns `--vpc-id is required for Container builds with --network-mode VPC`.
- A malformed `--vpc-id` returns `Invalid VPC ID format`.

New/updated tests:
- `create.test.ts`: `harness VPC dispatch (--dry-run)` — accepts a full set of VPC flags, returns the expected error when `--vpc-id` is missing, and rejects a malformed `--vpc-id`.
- `backfill-vpc-id.test.ts`: `dry-run (persist=false) restore` — writes the resolved VPC ID for synth and then reverts the file; confirms a real deploy leaves the value in place.

## Checklist

- [x] I have added tests that cover the fixes


src/cli/operations/deploy/backfill-vpc-id.ts
```diff
@@ -49,10 +74,17 @@ export async function backfillContainerVpcIds(
...
     if (isContainerBuild(harness) && harness.networkMode === 'VPC' && nc && nc.subnets.length > 0 && !nc.vpcId) {
       nc.vpcId = await resolveVpcIdFromSubnets(nc.subnets, region);
+      await snapshot(() => configIO.getHarnessConfigPath(entry.name));
       await configIO.writeHarnessSpec(entry.name, harness);
       backfilled.push(entry.name);
     }
   }
 
-  return { backfilled };
+  const restore = async (): Promise<void> => {
```
tejaskash: **Correctness — a throw inside this function after a partial write leaves the working tree dirty, defeating PATCH 2's "restore on every exit path" guarantee.**

`restore` is only handed back on the successful `return { backfilled, restore }`. On `deploy --dry-run` with a Container+VPC runtime **and** a Container+VPC harness both missing vpcId (or two such harnesses): the runtime resolves and agentcore.json is snapshotted+rewritten with the vpcId, then the harness loop's `resolveVpcIdFromSubnets` (line 76) throws (subnets span multiple VPCs, DescribeSubnets throttled/denied). The function rejects before returning, so in actions.ts `const backfill = await ...` (423) throws, `previewRestore = backfill.restore` (426) never runs, and the finally-block restore is skipped. agentcore.json is left dirtied.

The added tests only exercise a single-resource fixture, so they never hit a post-write throw. Root-cause fix: own the rollback in the caller, or wrap the loops in `try/finally` here, so a partial write is always revertible — not just on a clean return.

src/cli/operations/deploy/backfill-vpc-id.ts
```diff
@@ -49,10 +74,17 @@ export async function backfillContainerVpcIds(
...
+    for (const [path, content] of originals) {
+      await writeFile(path, content, 'utf-8');
```
tejaskash: **Robustness — restore() is non-atomic.** The writes run sequentially with no per-file `try/catch`; if the first `writeFile` rejects (file concurrently removed, transient EPERM), the remaining files are never reverted (partial restore) and the rejection propagates out of handleDeploy's finally, also skipping `restoreEnv?.()`. Make it best-effort/total: attempt every path and aggregate errors.

src/cli/operations/deploy/backfill-vpc-id.ts
```diff
@@ -2,31 +2,55 @@ import type { ConfigIO } from '../../../lib';
...
+  persist = true
 ): Promise<BackfillVpcIdResult> {
   const backfilled: string[] = [];
+  // path -> original bytes, captured before the first rewrite of each file (preview mode only, so
+  // restore() can revert them). Never touched on a real deploy (persist=true).
+  const originals = new Map<string, string>();
+
+  // Snapshot a file's current bytes before we rewrite it — but only in preview mode. Computes the
+  // path lazily so a real deploy never calls the path getter (keeps the persist=true path minimal).
+  const snapshot = async (getPath: () => string): Promise<void> => {
```
tejaskash: **Simplification.** The lazy `getPath: () => string` closure buys nothing: the first statement is already `if (persist) return` before `getPath()` is ever called, and the getters are trivial string returns. The comment "so a real deploy never calls the path getter" implies a cost that the guard already eliminates. Simplify to `snapshot(path: string)`, called as `await snapshot(configIO.getAgentConfigPath())`.

src/cli/commands/deploy/actions.ts
```diff
@@ -981,6 +992,10 @@ export async function handleDeploy(options: ValidatedDeployOptions): Promise<Dep
     if (toolkitWrapper) {
       await toolkitWrapper.dispose();
     }
+    // Revert the preview-mode vpcId backfill on every exit path (success, early return, or throw).
+    if (previewRestore) {
```
tejaskash, line 1003: **Robustness — restore ordering in `finally`.** `previewRestore()` runs *after* `toolkitWrapper.dispose()` (line 993). If `dispose()` throws — common after a synth/bootstrap failure on a creds-less preview, the exact path this PR targets — the finally block aborts before `if (previewRestore) await previewRestore()`, so agentcore.json stays dirty **and** `restoreEnv?.()` (999) never runs. Wrap `dispose()` in `try/catch`, or run `previewRestore` before `dispose`.

src/cli/commands/create/command.tsx
```diff
@@ -169,7 +169,11 @@ async function handleCreateHarnessCLI(options: CreateOptions): Promise<void> {
         modelProvider: options.modelProvider,
         modelId: options.modelId,
         apiKeyArn: options.apiKeyArn,
+        container: options.container,
```
tejaskash: **Simplification — this duplication is the shape that caused this PR's blocker.** This 12-field options object is byte-identical to the telemetry-wrapped call site at line 223. The original bug (59bbde0a) was forwarding VPC fields at one path but not the other; two verbatim copies guarantee the next field-add can be applied to one and forgotten on the other, reopening the same "validation sees undefined" blocker. Hoist one `const harnessValidationInput = { name, projectName, ... }` above the `if (options.dryRun)` branch and pass it to both `validateCreateHarnessOptions` calls.

tejaskash (commented):
Code review (high-effort, recall-biased). 8 findings; 5 posted inline below, 3 land in files this PR doesn't touch so they're summarized here.

**Correctness**
- **[#2 — HarnessPrimitive.ts:354, not in diff]** The CLI create/add-harness write path has no `buildType === 'Container'` guard on the vpcId write — the same stale-vpcId leak defect #3 fixes in schema-mapper.ts and GenerateWizardUI.tsx, left unfixed on this third path. `agentcore create --name X --network-mode VPC --subnets ... --security-groups ... --vpc-id ...` with **no** `--container` (a CodeZip harness) sets `harnessBuildType=undefined`, so `validateVpcOptions` never rejects the stray `--vpc-id`, and `...(options.vpcId && { vpcId: options.vpcId })` at HarnessPrimitive.ts:354 leaks a Container-only field into a CodeZip harness.json. Add the same guard here.

**Altitude**
- **[#5 — useGenerateWizard.ts:148, not in diff]** `setBuildType` clears `dockerfile` but not `vpcId` on a Container→CodeZip switch; this PR instead guards two *read* sites. That's symptom-patching stale state. `setNetworkMode`→PUBLIC already clears `vpcId` (line 269), proving the source-clear pattern exists. There are 4+ read sites of `config.vpcId`; clearing it in `setBuildType` when leaving Container (one line) makes all reads correct by default instead of requiring the buildType guard at every present and future consumer.
- **[#7 — vpc-utils.ts:104, not in diff]** `validateVpcOptions` gained an **optional** `buildType?` param, and the vpcId requirement fires only when `'Container'` is passed. A caller that omits `buildType` gets zero vpcId validation and no error — the same forwarding-failure class this PR is fixing (59bbde0a forgot to forward fields). Make `buildType` required so the type system forces every caller to supply it.

Notable clears: `snapshot()` timing is correct (captures old on-disk bytes before the in-memory spec is written); the `persist=true` path is a genuine no-op; the redundant `ec2:DescribeSubnets`-per-resource calls are pre-existing (feature commit), not introduced here; no AGENTS.md rule violations in the new code.

aidandaly24:
Thanks @tejaskash — all five addressed in `295c184b`:

1. **Partial-write rollback (correctness).** The resolve/write loops are now wrapped so that if a later resource's `resolveVpcIdFromSubnets` throws after an earlier file was already rewritten in a preview, the written files are reverted before the error propagates — a failed dry-run/diff no longer leaves the tree dirty even though `restore` isn't handed back. Added a multi-resource regression test (runtime writes, harness resolve throws → `agentcore.json` asserted byte-for-byte original).
2. **`restore()` is now best-effort/total.** Each file is reverted in its own try/catch; failures are aggregated into one error at the end, so one bad write no longer skips the rest.
3. **Ordering in `finally`.** Each cleanup step (`dispose()`, `previewRestore()`, `restoreEnv?.()`) now runs in its own try/catch, so a throwing `dispose()` can't skip the rollback or env restore.
4. **Simplified `snapshot()`** to take a path directly (the `if (persist) return` guard already runs first); the backfill test mock now stubs the `ConfigIO` path getters to match the real interface.
5. **De-duplicated the create-harness validation input** into a single `harnessValidationInput` object shared by both call sites — removing the two-copies drift that caused the original blocker.

Full unit suite green (6145).

tejaskash (approved):
LGTM

PR #1694, opened 2026-07-07
ci: add API reference docs generation workflow

notgitika:
Adds a workflow that auto-generates the CLI command reference as AsciiDoc on each release (by scraping `agentcore <cmd> --help`, grouped by category) and uploads it as a build artifact for the docs team to publish.

Files: workflow + extractor (`scripts/extract-cli-model.mjs`) + renderer (`scripts/render_adoc.py`). Runs on `release` and `workflow_dispatch`.

### DIFF
--- a/.github/workflows/publish-api-docs.yml
+++ b/.github/workflows/publish-api-docs.yml
@@ -1,0 +1,82 @@
+# =============================================================================
+# publish-api-docs.yml  —  agentcore-cli
+# =============================================================================
+# Auto-generates the CLI command reference as AsciiDoc (.adoc) on every release
+# by scraping `agentcore <cmd> --help` for all commands (grouped by category),
+# and uploads it as a build artifact for the docs team to publish.
+#
+# Pipeline: install released CLI -> scrape --help into doc-model JSON
+#           -> render .adoc -> upload artifact.
+#
+# NOTE: this generates and publishes the docs as an artifact only; landing them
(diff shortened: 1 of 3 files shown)


notgitika:
Thanks for the careful review — addressed in the latest push:

**Version pin (workflow):** Fixed. Now installs the exact release tag (`@aws/agentcore@<tag>`) instead of `@latest`, with a fallback to `@latest` only for manual `workflow_dispatch` runs.

**`eval` / `recommendations` phantoms:** Confirmed against the installed CLI and removed both from `GROUPS`. Also added a **fail-loud guard** in `entryForCommand`: if a command's `--help` doesn't produce a self-describing usage line (`Usage: agentcore <name> ...`), the extractor throws instead of emitting an empty stub — so phantoms/drift break the build rather than silently generating bad docs.

**`discoverCommands()` returning `[]`:** I couldn't reproduce this against the **published** binary — `agentcore --help` does emit the `Commands:` block, and the discovery correctly surfaced `evals`/`help` as leftovers in my run. It looks like the built artifact falls through to Commander's text help in a non-TTY/CI context even though the source wires an Ink handler. Given that, plus the new fail-loud guard as a stronger backstop, I've left discovery as-is. Happy to switch to `agentcore help commands` if you'd prefer to not rely on top-level `--help` at all.

aidandaly24 (approved):

Issue #1440, opened 2026-06-02
`agentcore dev` can leave orphaned processes occupying ports


avi-alpert:
### Description

When agentcore dev is killed programmatically (SIGTERM/SIGKILL), the underlying uvicorn reloader and worker processes are not terminated. They get reparented to PID 1 and continue holding the port. Subsequent agentcore dev runs then must fall back to a different port.

Interactive Ctrl+C is unaffected (SIGINT hits the entire process group).

Impact: Any agent/scripted/CI usage of agentcore dev leaks processes and accumulates port bindings until manual cleanup.

### Steps to Reproduce

agentcore create --name BugRepro --defaults --json
cd BugRepro
agentcore dev --logs &
DEV_PID=$!
sleep 10
kill $DEV_PID
wait $DEV_PID 2>/dev/null

### Port is still held by orphaned uvicorn:

lsof -ti :8080
### Returns PIDs with PPID=1 (reparented)

ps -p $(lsof -ti :8080) -o pid,ppid,command
### Expected Behavior

Killing the agentcore process should terminate its child process tree.

### Actual Behavior

Running agentcore dev again now prints Port 8080 in use, using 8081 - the orphaned process blocks the default port, and each leaked run claims another.

### CLI Version

0.14.1

### Operating System

macOS

PR #1695, opened 2026-07-07
fix: reap orphaned dev-server child on process exit

### LINKED ISSUE #1690: Bug: `agentcore dev` leaves orphaned Uvicorn process on Ctrl+C
## Summary

When stopping `agentcore dev --logs` with Ctrl+C, the CLI's Node process exits but does not terminate the Uvicorn subprocess it spawned. The Uvicorn process (and its children) become orphaned under PID 1 and continue running indefinitely, holding the port. This happens **every time**, not intermittently.

## Impact

- The orphaned server retains in-memory state (e.g., Strands Agent conversation history), making it appear that sessions persist across restarts.
- On next `agentcore dev` start, the port is occupied, forcing fallback to a different port (e.g., 8081).
- The CLI may still route prompts to the old port, talking to the stale server instance without the user's knowledge.
- Accumulated orphaned processes consume system resources.

### Steps to Reproduce

1. Start the dev server:
   ```bash
   agentcore dev --no-browser --logs
   ```

2. Confirm the process tree:
   ```bash
   ps -ef | grep uvicorn
   # Shows: agentcore (node) → uvicorn reloader → uvicorn worker
   ```

3. Press Ctrl+C in the terminal running `agentcore dev`.

4. Verify the orphaned processes:
   ```bash
   lsof -i :8080
   # Uvicorn processes still listening on the port

   ps -ef | grep <uvicorn_pid>
   # PPID is now 1 (orphaned under launchd/init)
   ```

### Expected Behavior

Expected behaviour is that the Uvicorn process listening on 8080 is also killed.

### Actual Behavior

Ctrl+C should terminate both the `agentcore` CLI process and the entire Uvicorn subprocess tree cleanly, releasing the port.

### CLI Version

0.22.0

### Operating System

macOS

## Observed Process Tree

**Before Ctrl+C:**
```
PID 15652  node /opt/homebrew/bin/agentcore dev --no-browser --logs   (PPID: 7376, shell)
  PID 15656  python3 ... uvicorn main:app --reload --host 127.0.0.1 --port 8080   (PPID: 15652)
    PID 15657  python3 ... multiprocessing.resource_tracker   (PPID: 15656)
    PID 15658  python3 ... multiprocessing.spawn              (PPID: 15656)
```

**After Ctrl+C:**
```
PID 15652  — GONE (agentcore node process exited)
PID 15656  python3 ... uvicorn ... --port 8080   (PPID: 1, orphaned under launchd)
  PID 15657  python3 ... resource_tracker          (PPID: 15656)
  PID 15658  python3 ... spawn                     (PPID: 15656)
```
## Workaround (for users)

Until the fix is released, users can work around this by manually killing orphaned processes:

```bash
# Kill any process on port 8080
lsof -ti :8080 | xargs kill

# Or use SIGQUIT instead of SIGINT (more forceful)
# Press Ctrl+\ instead of Ctrl+C
```

hagaym1:
## Description

`agentcore dev` spawns the dev server **detached** (its own process group, added in #1438) and relies on the dev command's `SIGINT`/`SIGTERM` handler (`onShutdownSignal`, registered with `process.once`) to call `server.kill()` and tree-kill the group. But `setupAltScreenCleanup()` (`src/cli/cli.ts`, called first thing in `main()`) registers a **global** `process.on('SIGINT'/'SIGTERM')` handler that calls `process.exit(0)`. Node runs signal listeners in registration order, and this global handler is registered **before** the dev command's — so it fires first, `process.exit(0)` terminates the process, and `server.kill()` never runs. The detached child is then reparented to PID 1 and keeps holding the port.

This reproduces on **every** stop — both interactive `Ctrl+C` (SIGINT) and programmatic `kill $PID` (SIGTERM) — for both Python (uvicorn) and TypeScript/Node dev servers. #1438 correctly fixed the reparenting mechanics (`detached` spawn + process-group tree-kill in `kill()`), but the compensating handler it added is preempted by the pre-existing global `process.exit(0)`, so the orphan is back in 0.22.0 (and now also on `Ctrl+C`, since the detached child no longer receives the terminal's group SIGINT directly).

### Fix

Register a `process.on('exit')` reaper in `DevServer.start()` that sends `SIGKILL` to the detached child's process group. The `'exit'` event **always** fires on `process.exit()`, so the child is reaped no matter which path exits the process (the global cleanup handler, an uncaught exception in `index.ts`, or a normal exit). `kill()`'s graceful `SIGTERM`-then-`SIGKILL` path is unchanged and still runs when it is reached; this is a last-resort net for when it isn't. The listener is removed when the child exits on its own, to avoid a stale reaper firing at final process exit. POSIX-only, matching the POSIX-only `detached` spawn (on Windows the child is not detached and dies with the parent).

## Related Issue

Closes #1690

## Documentation PR

N/A

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Added unit tests in `dev-server.test.ts` covering the reaper: it `SIGKILL`s the child's process group on exit, is skipped when there is no pid, is removed once the child exits on its own, and swallows `ESRCH` when the group is already gone.

Verified end-to-end on macOS with a TypeScript agent, before vs. after this change, mirroring the repro in #1690 / #1440:

```bash
agentcore dev --logs --port 9137 &
DEV_PID=$!
# wait for the port to bind, then:
kill $DEV_PID
sleep 4
lsof -ti :9137
# before this change: orphaned node still listening (reparented to PPID 1)
# after this change:  empty — port freed
```

- [x] I ran `npm run lint`
- [ ] If I modified `src/assets/`, I ran `npm run test:update-snapshots` and committed the updated snapshots (N/A — no asset changes)

> Note on the local runs: `typecheck`, `lint`, and the added `dev-server` unit tests all pass. The remaining `test:unit` / `test:integ` failures on my machine are environmental and outside the changed `dev-server` area: the unit failures are AWS SSO credential-resolution tests (`resolveSSOCredentials`, need live AWS credentials), and the integ failures are Python-project `create` tests that require the `uv` toolchain, which isn't installed in my sandbox (`'uv' is required for Python projects`). The repo's own `integ-tests/dev-server.test.ts` is `skipIf(!hasUv)` and skips cleanly here.

## Checklist

---

terms of your choice.

### DIFF
--- a/src/cli/operations/dev/__tests__/dev-server.test.ts
+++ b/src/cli/operations/dev/__tests__/dev-server.test.ts
@@ -1,7 +1,7 @@
 import type { DevConfig } from '../config.js';
 import { DevServer, type DevServerCallbacks, type DevServerOptions, type SpawnConfig } from '../dev-server.js';
 import { EventEmitter } from 'events';
-import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { type MockInstance, afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 const mockSpawn = vi.fn();
 vi.mock('child_process', () => ({
@@ -56,6 +56,8 @@ describe('DevServer', () => {
   let options: DevServerOptions;
   let server: TestDevServer;
   let mockChild: ReturnType<typeof createMockChildProcess>;
+  let onceSpy: MockInstance;
+  let removeListenerSpy: MockInstance;
 
   beforeEach(() => {
     onLog = vi.fn<DevServerCallbacks['onLog']>();
@@ -65,11 +67,19 @@ describe('DevServer', () => {
     server = new TestDevServer(config, options);
     mockChild = createMockChildProcess();
     mockSpawn.mockReturnValue(mockChild);
+    onceSpy = vi.spyOn(process, 'once').mockReturnValue(process);
+    removeListenerSpy = vi.spyOn(process, 'removeListener').mockReturnValue(process);
   });
 
   afterEach(() => {
+    vi.restoreAllMocks();
     vi.clearAllMocks();
   });
+
+  function getRegisteredExitHandler(): (() => void) | undefined {
+    const call = onceSpy.mock.calls.find(([event]) => event === 'exit');
+    return call?.[1] as (() => void) | undefined;
+  }
 
   describe('start()', () => {
     it('calls spawn with correct cmd, args, cwd, env, and stdio when prepare succeeds', async () => {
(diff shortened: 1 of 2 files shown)


hagaym1:
This directly closes the gap left by #1438. That PR made the dev child `detached` (its own process group) and added a SIGINT/SIGTERM handler to tree-kill it — but the global `process.on('SIGINT'/'SIGTERM')` in `setupAltScreenCleanup` (registered first, in `main()`) calls `process.exit(0)` and runs before the dev command's handler, so `server.kill()` never runs and the child is orphaned on every stop (Ctrl+C and `kill $PID` alike, Python and TypeScript). The fix reaps the child's process group in a `process.on('exit')` hook, which always fires on `process.exit()`.

Root cause, mechanism, and a before/after A/B are in the description; unit tests added; `typecheck`/`lint` pass locally.

@avi-alpert (author of #1438) / @jariy17 — could you take a look? And could a maintainer apply `safe-to-review` so CI and the security review can run? Happy to adjust anything.

tejaskash (approved):
LGTM, thank you for your contribution

PR #1716, opened 2026-07-08
ci: gate fork-PR E2E/tarball on PR author (fix pwn-request + checkout@v7 fork block)

aidandaly24:
## Description

`e2e`, `pr-tarball`, and `pr-security-review` fail at checkout for **fork PRs**. Root cause: #1710 bumped `actions/checkout` v6→v7, and v7 refuses to check out fork-PR code in a `pull_request_target` workflow by default (the "pwn request" protection). Since PRs here (including from authorized maintainers) are routinely opened from forks, this hard-fails those jobs at checkout.

Fixing it also surfaced a pre-existing flaw in `e2e`/`pr-tarball`: their `authorize` gate keyed on **`github.actor`** (who *triggered* the run), not the PR **author** (whose *code* runs). On `pull_request_target`, a trusted actor pushing to a fork PR would pass the gate and run untrusted fork code with the AWS role / write token — a real pwn-request.

## Fix

All three `pull_request_target` workflows now use the same, safe combination — **gate on the PR author, then opt into the fork checkout**:

1. **Gate on `github.event.pull_request.user.login`** (the PR author):
   - `e2e` / `pr-tarball`: author must be in `AUTHORIZED_USERS` (curated list — appropriate since these deploy real AWS resources / publish releases). *(This replaces the previous `github.actor` gate — the actual pwn-request fix.)*
   - `pr-security-review`: already gated on author write/admin via `getCollaboratorPermissionLevel` (unchanged).
2. **`allow-unsafe-pr-checkout: true`** on the checkout step. This is safe *because* of (1): external/untrusted authors skip at the gate, so only trusted authors' code is ever checked out — and their fork PRs (the normal path here) run again instead of hard-failing.

The key correctness point: `allow-unsafe-pr-checkout` is only dangerous with an **actor**-based gate (trusted actor + untrusted author's code). With an **author**-based gate it's safe, since the code's author is the thing being trusted.

Other `pull_request_target` workflows were checked and are fine (`codeql.yml`, `pr-ai-review.yml` use the default merge-ref, not fork head; `pr-size.yml`/`pr-title.yml` have no checkout).

## Type of Change

- [x] Bug fix (CI) — also closes a pre-existing pull_request_target pwn-request in e2e/pr-tarball

## Related

Regression surfaced by #1710 (checkout v7 bump).

## Checklist

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -85,6 +85,10 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
+          # This job runs only after the `authorize` gate confirms the actor is in AUTHORIZED_USERS,
+          # so checking out the PR head (incl. fork PRs) is intentional. checkout@v7 blocks fork
+          # checkout in pull_request_target by default; the authorize gate is our pwn-request guard.
+          allow-unsafe-pr-checkout: true
       - uses: actions/setup-node@v6
         with:
           node-version: '20.x'
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -88,6 +88,11 @@ jobs:
         with:
           ref: ${{ github.event.pull_request.head.sha }}
           fetch-depth: 0
+          # Safe because the `authorize` job above gates on the PR AUTHOR being in AUTHORIZED_USERS:
+          # external authors skip at the gate, so only trusted authors' code is ever checked out here.
+          # Needed because authorized maintainers routinely open PRs from their own forks, and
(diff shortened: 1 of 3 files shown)


notgitika (changes requested):
Since we usually open PRs from forks (authorized users included), e2e and pr-tarball will now hard-fail at checkout on the normal path. `authorize` passes on the PR author, then checkout@v7 refuses the fork head with no `allow-unsafe-pr-checkout`. So authorized fork PRs go red and need a manual workflow_dispatch.

The gate is already on the PR author being in AUTHORIZED_USERS, so the pwn-request is closed adding allow-unsafe-pr-checkout: true here too (like you did for security review) would let authorized fork PRs run normally while external authors still skip at the gate.

notgitika (approved):
Thanks for addressing the comment and making this more secure!

PR #1719, opened 2026-07-09
fix: pass through max tokens for non-mantle Bedrock models

### LINKED ISSUE #1718: Setting `maxTokens` in `harness.json` does nothing on non-mantle models
### Description

The `maxTokens` setting in `harness.json` isn't properly parsed into `AgentRenderConfig`, so it never is passed to Strands. Therefore, setting this property has no effect.
Tested with Claude Sonnet 5, where the model would run out of tokens on relatively simple tasks (with thinking) through using up tokens.

### Steps to Reproduce

1. Set `maxTokens` to a value
2. Run `agentcore dev`
3. Observe the "Max output tokens" entry in "Harness Settings"

### Expected Behavior

Observe that the "Max output tokens" entry in "Harness Settings" is populated with the correct value

### Actual Behavior

Observe that the "Max output tokens" entry in "Harness Settings" remained empty

### CLI Version

0.23.0

### Operating System

Linux

### Additional Context

_No response_

Tommypop2:
## Description

Ensures that the `maxTokens` entry in the configuration is passed through to the agent

## Related Issue

Closes #1718

## Documentation PR

N/A

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I ran `npm run test:unit` and `npm run test:integ` -> 1 failure but this also failed when I ran `npm run test` on the current `master`
- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,12 +1,12 @@
 {
   "name": "@aws/agentcore",
-  "version": "0.20.2",
+  "version": "0.23.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@aws/agentcore",
-      "version": "0.20.2",
+      "version": "0.23.0",
       "hasInstallScript": true,
       "license": "Apache-2.0",
       "dependencies": {
--- a/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
+++ b/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
@@ -6199,7 +6199,7 @@ from strands.models.bedrock import BedrockModel
 
 def load_model() -> BedrockModel:
     """Get Bedrock model client using IAM credentials."""
-    return BedrockModel(model_id="{{#if modelId}}{{modelId}}{{else}}global.anthropic.claude-sonnet-4-5-20250929-v1:0{{/if}}")
+    return BedrockModel(model_id="{{#if modelId}}{{modelId}}{{else}}global.anthropic.claude-sonnet-4-5-20250929-v1:0{{/if}}"{{#if modelMaxTokens}}, max_tokens={{modelMaxTokens}}{{/if}})
 {{/if}}
 {{/if}}
 {{#if (eq modelProvider "Anthropic")}}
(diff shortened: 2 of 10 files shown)


tejaskash:
Hi @Tommypop2 , the export fix is correct - the exported `load.py` dropped `model.maxTokens` for non-Mantle `BedrockModel`, and this threads it through.

But I traced #1718 through the managed-harness path and I don't think this PR closes it. The issue's repro is `agentcor
e dev` → "Max output tokens" in Harness Settings, and that field never touches `load.py`. `load.py` only exists on the `agentcore export` → code-based-agent path (export converts a managed harness into a standalone Python agent you deploy yourself). A managed harness interprets its model config server-side, it never renders `load.py`.

On the managed path, `model.maxTokens` is already wired end to end. 

So the "field is empty" symptom isn't reproduced by the current code on the managed path. Before merging as "Closes #1718", can you confirm:
  1. Did you repro via `agentcore dev` (managed harness) or `export` + deploy (code-based agent)?
  2. What CLI / agent-inspector version?
  3. Was the value under `model.maxTokens`, not the top-level execution `maxTokens`, in `harness.json`?

If it's a managed-harness issue it's likely either an older UI version or server-side (harness service not honoring `BedrockModelConfig.MaxTokens`), and should stay open.

Tommypop2:
Hi!
I encounted and tested with `agentcore dev` as that's where I initially found the error. Happy to test more on a deployed agent to verify the fix works for both. (Though I'd appreciate some guidance on reliably using enough tokens as consistently getting the model to error even with the lower token limit was sometimes unreliable even in dev)
The value was under `model.maxTokens` as follows:
```json
"model": {
    "provider": "bedrock",
    "modelId": "global.anthropic.claude-sonnet-5",
    "maxTokens": 32000
  },
```
The changes did seem to fix the issue, though I guess that can't be true if the `agentcore dev` flow doesn't use `load.py`. I am creating the agents in typescript (rather than the declarative harnesses), but keeping the `harness.json` so the CLI commands can still work.

I'm on CLI version v0.23.0.

Thanks!

Tommypop2:
Done some more testing:
When I was testing, I was still running `agentcore deploy -y` so I'm assuming that's what meant that the `maxTokens` in dev was updated.
I have a fix that I've tested to actually update the `maxTokens` value in dev without deployment (tested by setting a small `maxTokens` value then prompting and looking for token overflow). I've added it to this PR for now, but happy to split them into separate PRs and issues if needed

Thanks :)

tejaskash (approved):
Nice, this is the right fix, and it matches what I traced. I'm approving and merging this PR, I'll send a followup for the other model providers. Thank you for your contribution!

PR #1729, opened 2026-07-10
fix(dev): apply model config for all providers in dev, not just Bedrock

tejaskash:
## Description

Follow-up to #1719. That PR made `agentcore dev` default the harness invoke model from the local spec, so `agentcore.json` model settings (notably `maxTokens`) apply in dev without a redeploy — but the default only fired for `provider === 'bedrock'`. OpenAI / Gemini / LiteLLM harnesses still silently fell back to the **deployed** model config, so a local `maxTokens` edit did nothing for them until the next `agentcore deploy`.

This extends the same fix to all four providers.

## Changes

- Extract `buildModelOverride()` in `harness-utils.ts`, replacing the inline Bedrock-only branch. It maps each provider to its own invoke-API config key — `bedrockModelConfig` / `openAiModelConfig` / `geminiModelConfig` / `liteLlmModelConfig` — carrying only the fields that config accepts:
  - common: `modelId`, `temperature`, `topP`, `maxTokens`
  - Bedrock: `+ apiFormat`
  - OpenAI: `+ apiKeyArn`, `+ apiFormat` (drops `converse_stream`, which the OpenAI config rejects)
  - Gemini: `+ apiKeyArn`, `+ topK`
  - LiteLLM: `+ apiKeyArn`, `+ apiBase`, `+ additionalParams`
- `buildInvokeOptions` now defaults from the spec for any provider (was `else if (specModel?.provider === 'bedrock')`).
- Tests: replace the old "sends no override for non-bedrock providers" case (which asserted the bug) with mapping coverage for OpenAI, Gemini, and LiteLLM.

## Type of Change

- [x] Bug fix

## Testing

- [x] `npm run typecheck`
- [x] `npm run lint` (eslint + prettier clean on changed files)
- [x] `npx vitest run src/cli/operations/dev/web-ui/__tests__/harness-utils.test.ts` → 7/7 pass

> Note: verified the invoke API's config shapes accept `maxTokens` for all four providers. Runtime honoring for OpenAI/Gemini/LiteLLM is server-side and not verified end-to-end here — this wires the config through consistently, matching the Bedrock path.

### DIFF
--- a/src/cli/operations/dev/web-ui/__tests__/harness-utils.test.ts
+++ b/src/cli/operations/dev/web-ui/__tests__/harness-utils.test.ts
@@ -33,14 +33,60 @@ describe('buildInvokeOptions — model resolution', () => {
     expect(opts.skills).toEqual([{ path: './skills' }]);
   });
 
-  it('sends no default override for non-bedrock providers', () => {
+  it('defaults an OpenAI model with maxTokens + apiKeyArn (dropping converse_stream apiFormat)', () => {
+    const openAiModel: HarnessModel = {
+      provider: 'open_ai',
+      modelId: 'gpt-4.1',
+      apiKeyArn: 'arn:aws:secretsmanager:us-east-1:123:secret/openai',
+      maxTokens: 2048,
+      temperature: 0.5,
+    };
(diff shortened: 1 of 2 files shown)


jesseturner21 (approved):
LGTM

PR #1748, opened 2026-07-13
feat: add gh workflows setup

Hweinstock:
built on https://github.com/aws/agentcore-cli/pull/1747. 
Direct port of https://github.com/AlexanderRichey/agentcore-cli/pull/5. 
### Problem 
The refactor is missing CI and a pattern for GH workflows. There are no automated checks on PRs, or on merge to main. 

The existing CLI does not have a consistent pattern for workflows, which results in a large amount of duplication. Ex. e2e-tests logic duplicated across push-pr, and release duplicates much of release main. 

### Solution
- Organize workflows into a hierarchy of orchestrators and jobs (see README for more info). At a high level, jobs are re-usable logic and orchestrators are the wiring up of jobs (ex. CI, release, etc.). 
- add simple CI with unit tests, lint, format, build checks, typechecks, and audit. 
- coverage manged by CodeCov. (also has flaky test support https://about.codecov.io/blog/find-failing-and-flaky-tests-with-codecov-test-analytics/). 

### Testing 
Running commands locally I can confirm all pass except test and typecheck (which are also failing upstream). 

### Notes
- sharding is left as a follow-up, but confirmed the existing pattern in the CLI is supported in bun. 

### Future Work
- potentially add hooks on workflows that posts CWL metrics. We can then leverage this for ticketing (for canary failures or CI failures on main) and dashboards to track health over time. (depends if migrating to codebuild gives us this for free)

PR #1756, opened 2026-07-14
fix(ci): address failing typechecker

Hweinstock:
Built on https://github.com/aws/agentcore-cli/pull/1748. 

### Problem 
The typechecker is failing, see https://github.com/aws/agentcore-cli/pull/1748 as an example. 

The issues are isolated to the TUI code, and mostly resolve around missing undefined checks.

### Solution
- add safe assertions where its simple to see that undefined is not possible. 
- zip column up with columnWidths to avoid managing two seperate arrays. 
- use local variables with undefined guards for better type narrowing in TS. 

### Testing 
`bun run typecheck`. 
Also passing in CI.

### DIFF
--- a/src/components/ui/data-table/DataTable.tsx
+++ b/src/components/ui/data-table/DataTable.tsx
@@ -154,14 +154,14 @@ export function DataTable<T extends Record<string, unknown>>({
     { isActive: focus },
   );
 
-  // Column widths
-  const colWidths = columns.map((col) => {
-    if (col.width) return col.width;
+  // default the width of each column
+  const columnsWithWidths = columns.map((col) => {
+    if (col.width !== undefined) return { ...col, width: col.width };
     const maxData = pageData.reduce((max, row) => {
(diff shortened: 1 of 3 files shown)


nborges-aws (dismissed):

src/components/ui/data-table/DataTable.tsx
```diff
@@ -154,14 +154,14 @@ export function DataTable<T extends Record<string, unknown>>({
     { isActive: focus },
   );
 
-  // Column widths
-  const colWidths = columns.map((col) => {
-    if (col.width) return col.width;
+  // default the width of each column
+  const columnsWithWidths = columns.map((col) => {
+    if (col.width !== undefined) return { ...col, width: col.width };
```
Hweinstock, line 159: we explicitly check undefined to fix the edge case where the width is explicitly set to 0

avi-alpert (approved):

tejaskash (approved):
Looks good to me!

PR #1759, opened 2026-07-14
fix(deploy): pin vended aws-cdk-lib and enforce cloud-assembly schema compat

jesseturner21:
## Description

The vended CDK template declared `aws-cdk-lib` with a caret range (`^2.248.0`), so freshly created projects resolve whatever version npm serves. When upstream `aws-cdk-lib` bumps its cloud-assembly schema major (as 2.258.0 did with schema v54), new projects write manifests that the CLI's bundled readers (`@aws-cdk/toolkit-lib`, `@aws-cdk/cdk-assets-lib`) cannot parse, and `agentcore deploy` fails with `AssemblyVersionMismatch`. Readers accept older schemas but hard-fail on newer ones, so an unpinned template can always drift into this breakage between CLI releases.

This PR pins the template and makes the writer/reader coupling impossible to break silently:

- **Pin**: `src/assets/cdk/package.json` now exact-pins `aws-cdk-lib` to `2.261.0` (writes schema v54, matching the bundled readers). The root `aws-cdk-lib` devDependency is exact-pinned to the same version as a lockstep anchor; `npm-shrinkwrap.json` regenerated.
- **Enforcement**: new unit test `src/assets/__tests__/cdk-schema-compat.test.ts` asserts (1) the template pin is exact, (2) template pin === root devDep === installed copy, and (3) each bundled reader's max supported schema major >= the schema major the pinned `aws-cdk-lib` writes. Any one-sided bump fails CI with an actionable message.
- **Maintenance**: new `scripts/sync-template-cdk.mjs` re-syncs the root devDep, template pin, and asset snapshots in one command. On Dependabot `aws-cdk` group PRs: `npm ci && node scripts/sync-template-cdk.mjs`. Documented in `docs/testing/unit-tests.md`, with a pointer comment in `dependabot.yml`.

Existing projects (already scaffolded with the caret) are unaffected by this change; newly created projects get the exact pin.

## Related Issue

Closes #

## Documentation PR

N/A — docs updated in-repo (`docs/testing/unit-tests.md`).

## Type of Change

- [x] Bug fix

## Testing

- [x] I ran `npm run lint`

Additional verification:
- Built the CLI and ran `agentcore create`: the generated project vends `"aws-cdk-lib": "2.261.0"` and its install resolves `@aws-cdk/cloud-assembly-schema@54.11.0` (major 54, readable by both bundled readers).
- Negative tests: a drifted template pin fails the lockstep assertions; simulating a schema-55 writer fails both reader-compat assertions with the AssemblyVersionMismatch explanation and sync-script instructions.
- Sync script is idempotent (reports nothing to do on a synced tree) and correctly re-pins a drifted template.

## Checklist

---

terms of your choice.

### DIFF
--- a/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
+++ b/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
@@ -646,7 +646,7 @@ exports[`Assets Directory Snapshots > CDK assets > cdk/cdk/package.json should m
   },
   "dependencies": {
     "@aws/agentcore-cdk": "^0.1.0-alpha.19",
-    "aws-cdk-lib": "^2.248.0",
+    "aws-cdk-lib": "2.261.0",
     "constructs": "^10.0.0"
   }
 }
--- a/src/assets/__tests__/cdk-schema-compat.test.ts
+++ b/src/assets/__tests__/cdk-schema-compat.test.ts
@@ -1,0 +1,103 @@
+/**
+ * Guards against cloud-assembly schema drift between the vended CDK template
+ * and the CLI's bundled readers (@aws-cdk/toolkit-lib, @aws-cdk/cdk-assets-lib).
+ *
+ * The template's aws-cdk-lib determines the schema version a user project
+ * *writes*; the bundled readers determine the schema version the CLI can
+ * *read* during deploy. Readers accept older schemas but hard-fail on newer
+ * ones (AssemblyVersionMismatch), so the template pin must never emit a
+ * schema newer than the readers support. See ticket V2240408418 for the
+ * outage this prevents.
+ *
+ * If any test here fails after a dependency bump, run:
+ *
+ *   node scripts/sync-template-cdk.mjs
+ */
+import * as fs from 'fs';
+import { createRequire } from 'module';
+import * as path from 'path';
+import { describe, expect, it } from 'vitest';
+
+const REPO_ROOT = path.resolve(__dirname, '../../..');
+const HOW_TO_FIX = 'Run `node scripts/sync-template-cdk.mjs` to re-sync the template pin and snapshots.';
+const EXACT_VERSION = /^\d+\.\d+\.\d+$/;
+const READERS = ['@aws-cdk/toolkit-lib', '@aws-cdk/cdk-assets-lib'];
+
+const requireFromRoot = createRequire(path.join(REPO_ROOT, 'package.json'));
+
+function readJson(filePath: string): Record<string, any> {
+  return JSON.parse(fs.readFileSync(filePath, 'utf-8'));
+}
+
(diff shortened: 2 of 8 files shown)


src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
```diff
@@ -646,7 +646,7 @@ exports[`Assets Directory Snapshots > CDK assets > cdk/cdk/package.json should m
   },
   "dependencies": {
     "@aws/agentcore-cdk": "^0.1.0-alpha.19",
-    "aws-cdk-lib": "^2.248.0",
+    "aws-cdk-lib": "2.261.0",
```
Hweinstock, line 649: if a customer creates a project with this version, is it their responsibility for updating it going forward? What happens if aws-cdk-lib releases new features we depend on for a new feature, but the customers existing project is on an older version?

src/assets/__tests__/cdk-schema-compat.test.ts
```diff
@@ -0,0 +1,103 @@
...
+ * *writes*; the bundled readers determine the schema version the CLI can
+ * *read* during deploy. Readers accept older schemas but hard-fail on newer
+ * ones (AssemblyVersionMismatch), so the template pin must never emit a
+ * schema newer than the readers support. See ticket V2240408418 for the
+ * outage this prevents.
+ *
+ * If any test here fails after a dependency bump, run:
+ *
+ *   node scripts/sync-template-cdk.mjs
+ */
```
Hweinstock, line 15: it looks like we have similar comments in the dependabot.yml, the script to update, the readme, and the test here. Does it make sense to establish a source of truth and link to it?

src/assets/__tests__/cdk-schema-compat.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  if (!match) {
+    throw new Error(
+      `aws-cdk-lib's @aws-cdk/cloud-assembly-schema range "${range}" is not caret-style; ` +
+        `this test's major-version parsing is no longer sound and must be updated.`
+    );
+  }
+  return Number(match[1]);
+}
+
+describe('vended CDK template / bundled cloud-assembly reader compatibility', () => {
```
Hweinstock, line 69: Is there a way to test this at a higher level? This feels highly specific to the implementation rather than the behavior.

src/assets/cdk/package.json
```diff
@@ -24,7 +24,7 @@
   },
   "dependencies": {
     "@aws/agentcore-cdk": "^0.1.0-alpha.19",
-    "aws-cdk-lib": "^2.248.0",
+    "aws-cdk-lib": "2.261.0",
     "constructs": "^10.0.0"
```
Hweinstock, line 28: should we pin other dependencies as well? Even though we own the cdk package, we don't currently have backwards-compat testing gating release, and wondering if there's a risk we break existing projects.

PR #1760, opened 2026-07-14
ci: auto-assign a round-robin reviewer to new PRs

notgitika:
External PRs currently land in a shared queue with no named reviewer, so many sit unreviewed.

This PR adds a workflow that requests one reviewer per new non-draft PR when the PR author is not in the existing `AUTHORIZED_USERS` roster. Reviewers are selected round-robin from that roster. PRs authored by someone in `AUTHORIZED_USERS` are left unassigned, and PRs that already have a reviewer are skipped.

**Safety:** follows the metadata-only `pull_request_target` pattern from `pr-size.yml` (needed so fork PRs can be assigned; safe because no PR code is checked out or executed, only PR metadata is read and a reviewer requested).

Future work: We can look into adding a CODEOWNERS file and then editing this workflow to make sure the right folks with context are tagged in PRs. I will also be looking into a simple starter system to assignees to get pinged on our slack


.github/workflows/pr-assign-reviewer.yml
```diff
@@ -0,0 +1,67 @@
+name: Assign PR Reviewer (round-robin)
+
+# Auto-assigns one reviewer to each new external PR, so community contributions
```
Hweinstock, line 3: Are there ideas in the comment here that we don't feel like are expressed clearly below in the implementation, if so why?

.github/workflows/pr-assign-reviewer.yml
```diff
@@ -0,0 +1,67 @@
...
+
+            // Don't re-assign if a reviewer was already requested (e.g. manual, or a re-run).
+            if ((pr.requested_reviewers?.length ?? 0) > 0 || (pr.requested_teams?.length ?? 0) > 0) {
+              console.log('Reviewer(s) already requested; leaving as-is.');
+              return;
+            }
+
+            // Stateless round-robin: PR number modulo pool size. Deterministic, no state
+            // file to keep in sync, and spreads load evenly as PR numbers increase.
+            const reviewer = pool[pr.number % pool.length];
```
Hweinstock, line 59: clever :)

.github/workflows/pr-assign-reviewer.yml
```diff
@@ -0,0 +1,67 @@
+name: Assign PR Reviewer (round-robin)
```
Hweinstock, line 1: broader q: is there any prior art you were able to find in your research? Does the strands team have anything?

I'm wondering about what happens if the assigned reviewer isn't available (PTO), or busy with another project. Also is there a clear path to overriding the assigned reviewer here if I want someone else to review?

Also wondering if a no-code solution might be simpler. For example, on-call responsible for external PR review for their week, or we delegate these are part of sprint work.

notgitika: I think this makes sense as well 🤔

.github/workflows/pr-assign-reviewer.yml
notgitika: Nope, just a basic description :)

notgitika:
punted for now, will reopen

PR #835, opened 2026-04-14
feat: add auto-instrumentation to langchain agent template

avi-alpert:
## Description

Adds auto-instrumentation to the LangChain/LangGraph agent templates so that traces emitted by LangChain agents are compatible with AgentCore's observability and evaluation features out of the box.

**Changes:**
- Import and call `LangchainInstrumentor().instrument()` at module level in both A2A and HTTP LangChain template `main.py` files
- Add `opentelemetry-instrumentation-langchain` to the template `pyproject.toml` dependencies for both A2A and HTTP variants
- Update asset snapshots

**Why:** Without the LangChain OpenTelemetry instrumentor, traces from LangChain agents are emitted under a generic scope that the Evaluate API does not recognize. Running `agentcore run eval` against a LangChain agent would fail with:

> Error: Provided input has no spans with supported scope. Currently supported scopes are ['strands.telemetry.tracer', 'opentelemetry.instrumentation.langchain', 'openinference.instrumentation.langchain']. Please verify that you are using the supported agentic-framework and instrumentation-library

With this change, LangChain agent traces are emitted under the `opentelemetry.instrumentation.langchain` scope, and `agentcore run eval` succeeds.

**Reference:** [Configuring custom observability in AgentCore](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/observability-configure.html#observability-configure-custom)

## Type of Change

- [x] New feature

## Testing

Verified by running `agentcore run eval` against a LangChain agent — evaluation completes successfully instead of returning the unsupported-scope error.

- [x] I ran `npm run test:unit` and `npm run test:integ`
- [x] I ran `npm run typecheck`
- [x] I ran `npm run lint`
- [x] If I modified `src/assets/`, I ran `npm run test:update-snapshots` and committed the updated snapshots

## Checklist

- [x] I have read the CONTRIBUTING document
- [ ] I have added any necessary tests that prove my fix is effective or my feature works
- [ ] I have updated the documentation accordingly
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the
terms of your choice.

Issue #140, opened 2026-02-06
Enhance observability in Google ADK and Langchain


notgitika:
## Description
Observability should be enabled by default in new projects so users get monitoring and tracing out of the box without needing to configure it manually.

## Category
Feature Request

## Acceptance Criteria
- [ ] New projects created with `agentcore create` have observability enabled by default
- [ ] Users can opt out of observability if desired
- [ ] Documentation reflects the default-on behavior

notgitika:
Verification update: LangChain is not a remaining issue. The current template includes the AWS OpenTelemetry distro and explicit LangChain instrumentation from #552, #835, and #836. A live deployed invocation completed a tool call and emitted 22 spans, including `invoke_agent`, `chat`, `execute_task`, and `execute_tool`. The remaining defect is specific to the Google ADK templates and is addressed by #1773.

PR #1773, opened 2026-07-15
fix(templates): enable Google ADK observability

notgitika:
## Description

Google ADK HTTP and AG-UI projects currently install the generic OpenTelemetry distro and exporter. Although their container entrypoint runs through `opentelemetry-instrument`, live AgentCore verification produced no spans.

This change:

- replaces the generic OpenTelemetry dependencies with `aws-opentelemetry-distro >= 0.17.0`
- keeps the dependency open-ended so the resolver can select the newest version compatible with Google ADK
- adds regression coverage across HTTP, AG-UI, and A2A Google ADK templates
- updates the affected asset snapshots

AWS distro 0.18.x currently requires OpenTelemetry 1.42.x while current Google ADK releases cap OpenTelemetry at 1.41.1. Fresh resolution therefore selects AWS distro 0.17.1 and OpenTelemetry 1.40.0 without requiring a stale upper pin.

## Framework Scope

LangChain is already working and is not changed by this PR. Its template includes the AWS OpenTelemetry distro and explicit LangChain instrumentation from #552, #835, and #836. A live deployed LangChain invocation completed a tool call and emitted 22 spans, including `invoke_agent`, `chat`, `execute_task`, and `execute_tool` operations.

The remaining issue is specific to the Google ADK HTTP and AG-UI dependency templates.

## Verification

Live runtime verification used a generated Google ADK HTTP project:

- Before the change: the invocation reached Google ADK but emitted zero spans to `aws/spans`.
- After replacing only the two generic OpenTelemetry dependencies with AWS distro 0.17.0: the same invocation emitted 18 spans.
- The emitted spans included Google ADK `invoke_agent` and `generate_content` operations, plus correlated request/error telemetry; `traces get` returned 15 correlated runtime records.
- The available Gemini key was invalid, so the model request ended with an authentication error. This still verified telemetry initialization and export on the ADK execution path; a successful model/tool invocation remains an additional end-to-end check.

Dependency and repository verification:

- fresh `uv pip compile` resolution for both changed templates selected `aws-opentelemetry-distro==0.17.1`, `google-adk==1.35.2`, and OpenTelemetry API/SDK 1.40.0
- 131 focused asset, snapshot, and regression tests passed
- `npm run build` passed
- the full unit run passed 5,911 tests; six unrelated failures caused by starting tests before the concurrent build produced `dist/cli/index.mjs`, plus one TUI timing failure, all passed when the two failed files were rerun after the build
- pre-commit typecheck, lint, formatting, and secret checks passed

## Related Issue

Addresses #140

## Documentation PR

Not applicable; this corrects generated runtime dependencies without changing the user-facing configuration.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] I have added tests that prove the fix is effective
- [x] No documentation change is required

---

### DIFF
--- a/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
+++ b/src/assets/__tests__/__snapshots__/assets.snapshot.test.ts.snap
@@ -2518,8 +2518,7 @@ dependencies = [
     "fastapi >= 0.115.12",
     "google-adk >= 1.16.0, < 2.0.0",
     "google-genai >= 1.0.0, < 2.0.0",
-    "opentelemetry-distro",
-    "opentelemetry-exporter-otlp",
+    "aws-opentelemetry-distro >= 0.17.0",
     "uvicorn >= 0.34.3",
 ]
 
@@ -4196,8 +4195,7 @@ description = "AgentCore Runtime Application using Google ADK"
 readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
-    "opentelemetry-distro",
-    "opentelemetry-exporter-otlp",
+    "aws-opentelemetry-distro >= 0.17.0",
     "google-adk >= 1.35.0, < 2.0.0",
     "google-genai >= 1.0.0, < 2.0.0",
     "bedrock-agentcore >= 1.0.3",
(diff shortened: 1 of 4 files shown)


jariy17 (approved):

jariy17:
This is only an issue because we didn't upgrade our Google ADK to v2. Once we upgrade, this shouldn't be a problem.

notgitika:
> This is only an issue because we didn't upgrade our Google ADK to v2. Once we upgrade, this shouldn't be a problem. 

Yes, you are correct! Hopefully the refactor addresses the root cause altogether

PR #1777, opened 2026-07-16
feat: manage vended CDK project dependencies with minor-version pinning

### LINKED ISSUE #1540: feat: CLI should manage dependencies of created projects.
### Problem 
The CLI currently creates projects with unpinned dependencies. Ex. see https://github.com/aws/agentcore-cli/blob/ef1e0667e3c764ac56fcdaeeb33bec82c81ec04c/src/assets/cdk/package.json#L25-L29. 

This allows customers to manage dependencies themselves, and always pull down the latest new features. 

However, it creates a risk where any dependencies (outside of our control) could make breaking changes that affect the template code, or the ability for the CLI to interact with the project. 

### Definition of Done
Propose a plan for how to manage those dependencies (on behalf of the customer) to avoid the risks outlined above. 
- consider both new and existing projects. 
- consider the feasibility across different templates (ex. what about BYO, python templates, typescript templates, etc). 
- consider how the CLI should manage this with minimal complexity. 
- should this be controllable? If so how?

jesseturner21:
### Problem

The CLI vends projects with caret (`^`) dependency ranges in `agentcore/cdk/package.json`, so any upstream minor release (`@aws/agentcore-cdk`, `aws-cdk-lib`, …) can land in a customer's project without any CLI change — breaking template code or the CLI's ability to interact with the project. The CLI needs to own these versions on the customer's behalf while staying out of the way of dependencies the customer added themselves.

### Solution

**Pin to minor, let patches float.** Managed dependencies are written as tilde ranges (`~x.y.z`); `@aws/agentcore-cdk` (L3 constructs, prerelease-versioned) is pinned exactly. A dependency is "managed" iff its name appears in the CLI's vended asset `src/assets/cdk/package.json` — that asset, baked into each CLI release, is the single source of truth for both the managed list and the known-good versions. Everything else in the customer's file is never touched.

On every `agentcore deploy` (headless and TUI), a preflight step:

1. **Skew check** — if the project declares a managed dep with a higher minor/major than this CLI expects (e.g. a teammate on a newer CLI updated it), throw: *"This project requires a newer version of the AgentCore CLI. Run `npm install -g @aws/agentcore@latest` and retry."* Higher patch is fine — that's the point of tilde.
2. **Migrate** — projects created before pinning (caret ranges) are rewritten to the pinned form with a one-time explanation.
3. **Sync** — managed deps are updated to the current CLI's values; user-added deps are left byte-for-byte untouched; non-semver specifiers (`file:` bundled-tarball overrides, `git:`, …) are skipped with a warning, never rewritten.
4. **Install (only if something changed)** — run an incremental `npm install` in `agentcore/cdk/` that reconciles the edited package.json against the existing lockfile; `node_modules` and the lockfile are never deleted, so user-added deps keep their resolved versions and installs stay warm. The install also runs as a self-heal when `node_modules` is missing even with a matching manifest. If the install fails, the pre-rewrite package.json is restored so a retry recomputes the same plan and re-attempts the install. No-op deploys stay fast.
5. **Report** — print a summary of every change (`old → new` per dep, plus "Dependencies you added yourself were not changed.").

**Preview modes:** `--dry-run` and `--diff` run the sync in check-only mode — the notice reports what a real deploy would change (future tense), but nothing is written and no install runs.

**Teardown:** teardown deploys are never blocked by the sync — both skew and write/install failures downgrade to warnings, so dependency pinning can never prevent destroying a cost-incurring stack.

**Opt-out:** `agentcore config disableDependencyManagement true` skips migrate/sync/reinstall entirely; the skew check still runs but downgrades to a warning and deploy proceeds — an escape hatch if we ever ship a bad pin, and support for teams with their own dependency tooling.

### High-level implementation

- **`src/lib/dependency-management/` (new, self-contained)** — one deep module behind a single function, `syncManagedDependencies({vendedPackageJsonPath, projectDir, disabled})`, designed to lift into the planned CLI refactor unchanged (zero imports from `src/cli/`). All notices/warnings are returned as data on the result; callers only display them, so the headless and TUI paths can't drift.
- **Vended asset pinned** — `src/assets/cdk/package.json` converted to the policy form, with tilde bases bumped to what today's caret ranges actually resolve to (e.g. `constructs ^10.0.0` → `~10.7.0`, not `~10.0.0`) so migration never silently downgrades an existing project.
- **Deploy integration** — `ensureManagedDependencies()` adapter (`src/cli/operations/deploy/dependency-sync.ts`) resolves the vended asset via `getTemplatePath` and reads the global-config opt-out; called as a "Sync CDK dependencies" step between validate and build in `handleDeploy` (`actions.ts`) and TUI preflight (`useCdkPreflight.ts`).
- **Config** — `disableDependencyManagement` added to `GlobalConfigSchemaStrict`; the generic `agentcore config <key> <value>` command handles it with no command changes.

## Related Issue

Closes #1540

## Documentation PR

N/A — behavior is self-describing via the deploy-time notices; can follow up in agent-docs if wanted.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

40 new unit tests in `src/lib/dependency-management/__tests__/` (16 sync + 12 policy + 12 semver), plus new `useDevDeploy` cases for the dev-flow notice/warnings:

- [x] I ran `npm run test:unit` and `npm run test:integ` (unit: 411 files / 5956 tests pass; integ not run — no integ coverage touched)
- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.


src/lib/dependency-management/semver.ts
```diff
@@ -0,0 +1,94 @@
+/**
```
Hweinstock, line 1: what is the gap between the functionality we have here, and what we could get from https://github.com/npm/node-semver?

src/lib/dependency-management/types.ts
```diff
@@ -0,0 +1,91 @@
...
+  notice: string | null;
+}
+
+export interface SyncManagedDependenciesOptions {
+  /** Path to the CLI's vended CDK package.json — source of truth for managed names + versions. */
+  vendedPackageJsonPath: string;
+  /** The user project's CDK directory (contains package.json, node_modules, lockfile). */
+  projectDir: string;
+  /** From global config `disableDependencyManagement`. When true: no writes, skew becomes a warning. */
+  disabled?: boolean;
```
Hweinstock, line 74: does this duplicate the treatSkewAsWarning input below?

src/cli/commands/deploy/actions.ts
```diff
@@ -448,6 +513,7 @@ export async function handleDeploy(options: ValidatedDeployOptions): Promise<Dep
         success: false,
         error: stackSelection.error,
         logPath: logger.getRelativeLogPath(),
```
Hweinstock: nit: the threading of this result into so many places feels fragile. Is there a way we can simplify this? One idea could be factory function for the result that always includes this variable from closure.

e2e-tests/dependency-sync.test.ts
```diff
@@ -0,0 +1,450 @@
+import { parseJsonOutput, spawnAndCollect } from '../src/test-utils/index.js';
```
Hweinstock, line 1: appreciate the e2e tests! Were you able to test this on old projects manually? wonder if that's worth a bug bash because this feels decently high risk.

src/lib/dependency-management/sync.ts
```diff
@@ -0,0 +1,170 @@
...
+      reinstalled: false,
+      applied: false,
+    });
+    return result;
+  }
+
+  if (hasPlannedChanges) {
+    // Mutate the parsed manifest in place so user key order and unknown fields survive.
+    for (const change of plan.changes) {
+      project[change.section]![change.name] = change.to;
```
Hweinstock: Is this `!` assertion assuming the package.json has specific sections? Is there a chance any of those sections are missing?

src/cli/commands/deploy/actions.ts
```diff
@@ -261,6 +277,48 @@ export async function handleDeploy(options: ValidatedDeployOptions): Promise<Dep
...
+      dependencySyncResult = await ensureManagedDependencies(context.cdkProject, {
+        checkOnly: isPreview,
+        treatSkewAsWarning: context.isTeardownDeploy,
+      });
+    } catch (syncErr) {
+      if (!(context.isTeardownDeploy && syncErr instanceof DependencySyncError)) {
+        // The sync itself is the failure: attach a minimal 'failed' result before rethrowing
+        // so the outer catch's failure result still carries dep_sync_* telemetry.
+        dependencySyncResult = failedSyncResult();
+        throw syncErr;
```
Hweinstock, line 310: why exactly do we want to throw here instead of return a failure result like the other cases?

Hweinstock (commented):
nice work! I like how most of the functionality was isolated to the module. Few questions, but overall makes sense to me. I do think this is worth bug-bashing since it feels pretty high-risk.

jariy17 (dismissed):
It looks good to me

src/lib/dependency-management/types.ts
jesseturner21: It is different because `disabled` will NOT write dependencies and `treatSkewAsWarning` will still write the dependencies and npm install

e2e-tests/dependency-sync.test.ts
jesseturner21: Good call! going to set up a bugbash tomorrow

src/lib/dependency-management/sync.ts
jesseturner21: good callout i updated the line to `(project[change.section] ??= {})[change.name] = change.to;`

src/lib/dependency-management/semver.ts
jesseturner21: Good call out, node-semver fully covers the parse/compare half, so I switched to it for that. What it does not give us is whether it was written as tilde/caret/exact, which the sync needs to know.

src/cli/commands/deploy/actions.ts
jesseturner21: Good call, added a `fail(error)` helper inside `handleDeploy` that does all failure steps.

src/cli/commands/deploy/actions.ts
jesseturner21: the throw is caught in the handleDeploy [here](https://github.com/aws/agentcore-cli/pull/1777/changes#diff-a083dd16ac9003e657fd55a639c155777d7dfd953ba49c97de2c19dbc9211e59R1024-R1028)
and catching locally would just duplicate the log/finalize/return steps at this site

Hweinstock (approved):
LGTM

PR #1787, opened 2026-07-17
fix(deploy): fail fast on AWS account mismatch

### LINKED ISSUE #761: Better error handling during deploy for valid creds but for a different account
### Description

When you run deploy and you have valid creds but for a different account (not your target for the project)
It starts the deploy properly and the "publish assets" step fails with a generic error.

We should have this check as the first step instead of after doing everything else

### Steps to Reproduce

have a different account in aws-targets.json and have valid creds for a separate account.

### Expected Behavior

should fail in the first step itselff

### Actual Behavior

fails after several steps

### CLI Version

0.5.1

### Operating System

macOS

### Additional Context

_No response_

notgitika:
## Summary

- validate that active AWS credentials belong to the selected deployment target account
- run the check during preflight, before CDK build, synth, and asset publishing
- apply the same validation to CLI and TUI teardown deploys after confirmation
- preserve existing credential validation behavior for callers without a deployment target

The account comparison is integrated into `validateAwsCredentials`, so credential validity and target ownership are checked with one STS `GetCallerIdentity` request. Mismatches now name both account IDs and the selected target instead of surfacing later as a generic CDK asset publishing error.

## Testing

- `npx vitest run --project unit src/cli/aws/__tests__/account-extended.test.ts src/cli/operations/deploy/__tests__/preflight.test.ts src/cli/commands/deploy/__tests__/deploy-teardown.test.ts`
- `npm run typecheck`
- focused ESLint and Prettier checks
- `npm run build`

Fixes #761

### DIFF
--- a/src/cli/tui/hooks/useCdkPreflight.ts
+++ b/src/cli/tui/hooks/useCdkPreflight.ts
@@ -441,7 +441,7 @@ export function useCdkPreflight(options: PreflightOptions): PreflightResult {
         // Validate AWS credentials (deferred for teardown deploys until after confirmation)
         if (preflightContext.isTeardownDeploy) {
           try {
-            await validateAwsCredentials();
+            await validateAwsCredentials(preflightContext.awsTargets[0]);
           } catch (err) {
             const errorMsg = formatError(err);
             logger.endStep('error', errorMsg);
--- a/src/cli/tui/hooks/useCdkPreflight.ts
+++ b/src/cli/tui/hooks/useCdkPreflight.ts
@@ -1,6 +1,6 @@
 import { ConfigIO, SecureCredentials, toError } from '../../../lib';
 import { AwsCredentialsError, UserCancellationError } from '../../../lib/errors/types';
-import type { DeployedState } from '../../../schema';
+import type { AwsDeploymentTarget, DeployedState } from '../../../schema';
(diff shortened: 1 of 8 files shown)


src/cli/tui/hooks/useCdkPreflight.ts
```diff
@@ -441,7 +441,7 @@ export function useCdkPreflight(options: PreflightOptions): PreflightResult {
         // Validate AWS credentials (deferred for teardown deploys until after confirmation)
         if (preflightContext.isTeardownDeploy) {
           try {
-            await validateAwsCredentials();
+            await validateAwsCredentials(preflightContext.awsTargets[0]);
```
nborges-aws: Would this validation be an issue if the selected target is different than the first target? I see above we're doing `selectedTarget ?? awsTargets[0]` but not here.

notgitika: Good catch. The TUI was not passing its selected target into preflight, so both normal and deferred teardown validation could fall back to the first configured target. I threaded the first selected target through `useCdkPreflight` and now use it consistently for credential validation and target-specific preflight work. Added a regression test covering selection of a non-first target.

nborges-aws (approved):

PR #1789, opened 2026-07-17
feat(config): implement global config to wire up config command

Hweinstock:
### Problem 
The CLI is missing a way to view and control global configuration. There is an existing `config` command, but its not wired up. 

### Solution
- implement a new top level `globalConfig/` module that defines an accessor for global config. The global config is decoupled from the source of the data via an injected dependency. 
- move `src/router/schemas.tsx` to top level `/parsing` directory so that logic can be re-used. 
- global config is type-safe based on the global config schema using json dot notation. For example, valid keys are `telemetry`, `telemetry.enabled`, etc. All of which are derived automatically from the schema definition. 
- wire up config command to be output JSON responses. 

### Testing 
- wrote unit tests at the config command level to test e2e, rather than testing implementation directly. 
- e2e testing using existing config from CLI also works. 
Ex. 
```
> bun run start config
{
  "telemetry": {
    "enabled": false,
    "endpoint": "https://telemetry.agentcore.aws.dev"
  },
}
> bun run start config telemetry.endpoint 
"https://telemetry.agentcore.aws.dev"

> bun run start config telemetry.endpoint https://example.com
"https://example.com"
```
### Future Work
- add error types for telemetry, and easier reasoning (few todo comments on this). 
- add logic to setup installationId for telemetry (getOrPut).

### DIFF
--- a/src/globalConfig/accessor.tsx
+++ b/src/globalConfig/accessor.tsx
@@ -8,7 +8,7 @@ import {
 import type { Logger } from "../logging";
 import z from "zod";
 import { globalConfigFileSchema } from "./types";
-import { getDefaultGlobalConfig, resolveConfig } from "./config";
+import { DEFAULT_GLOBAL_CONFIG, resolveConfig } from "./config";
 
 type DefaultGlobalConfigAccessorConfig = {
   logger: Logger;
@@ -46,7 +46,7 @@ export class DefaultGlobalConfigAccessor implements GlobalConfigAccessor {
 
     // if no installationId is present, generate one and merge it into the file data
     if (!configFileData.installationId) {
-      configFileData.installationId = getDefaultGlobalConfig().installationId;
+      configFileData.installationId = DEFAULT_GLOBAL_CONFIG.installationId;
       this.logger.info(`no installationId found, persisting one`);
 
       try {
--- a/src/globalConfig/accessor.tsx
+++ b/src/globalConfig/accessor.tsx
@@ -8,7 +8,7 @@ import {
 import type { Logger } from "../logging";
 import z from "zod";
 import { globalConfigFileSchema } from "./types";
-import { DEFAULT_GLOBAL_CONFIG, resolveConfig } from "./config";
+import { DEFAULT_GLOBAL_CONFIG, applyOverrides } from "./config";
 
 type DefaultGlobalConfigAccessorConfig = {
   logger: Logger;
@@ -60,14 +60,14 @@ export class DefaultGlobalConfigAccessor implements GlobalConfigAccessor {
       }
     }
 
-    this.cachedConfig = resolveConfig(configFileData);
+    this.cachedConfig = applyOverrides(DEFAULT_GLOBAL_CONFIG, configFileData);
     return this.cachedConfig;
   }
 
   public async set(newConfig: GlobalConfig): Promise<GlobalConfig> {
     this.logger.child({ newConfig, method: "set" }).debug("writing global config");
 
-    const configDiff = diff(newConfig, resolveConfig({}));
+    const configDiff = diff(newConfig, DEFAULT_GLOBAL_CONFIG);
     await this.writeToConfigFile(configDiff);
     this.cachedConfig = newConfig;
     return this.cachedConfig;
--- a/src/globalConfig/accessor.tsx
+++ b/src/globalConfig/accessor.tsx
@@ -1,0 +1,129 @@
+import {
+  type DeepPartial,
+  type GlobalConfig,
+  type GlobalConfigAccessor,
+  type GlobalConfigFileData,
+  type ReadWriteJson,
+} from "./types";
+import type { Logger } from "../logging";
+import z from "zod";
+import { globalConfigFileSchema } from "./types";
+import { getDefaultGlobalConfig, resolveConfig } from "./config";
+
+type DefaultGlobalConfigAccessorConfig = {
+  logger: Logger;
+  json: ReadWriteJson;
+  filePath: string;
+};
+
+/**
+ * Implements {@link GlobalConfigAccessor} accepting overrides from the given file.
+ *
+ * @param config - The logger and json datasource to be used by the config. .
+ * @returns A {@link GlobalConfigAccessor} instance.
+ */
+export class DefaultGlobalConfigAccessor implements GlobalConfigAccessor {
+  private cachedConfig: GlobalConfig | undefined;
+  private readonly json: ReadWriteJson;
+  private readonly filePath: string;
+  private readonly logger: Logger;
+
+  constructor(config: DefaultGlobalConfigAccessorConfig) {
+    this.json = config.json;
+    this.filePath = config.filePath;
+    this.logger = config.logger.child({ configFilePath: this.filePath });
+    this.logger.info(`creating global config accessor`);
+  }
+
+  public async get(): Promise<GlobalConfig> {
+    const logger = this.logger.child({ method: "get" });
+    logger.debug("reading global config");
+
+    if (this.cachedConfig) return this.cachedConfig;
+    logger.debug("no config cached, reading from source");
+
+    const configFileData = await this.readConfigFile();
+
+    // if no installationId is present, generate one and merge it into the file data
+    if (!configFileData.installationId) {
+      configFileData.installationId = getDefaultGlobalConfig().installationId;
+      this.logger.info(`no installationId found, persisting one`);
+
+      try {
+        await this.writeToConfigFile(configFileData);
+      } catch (e) {
+        const error = e instanceof Error ? e : new Error(String(e));
+        this.logger
+          .child({ errorName: error.name, errorMessage: error.message })
+          .warn(`failed to write initial config file data`);
+        // best effort
+      }
+    }
+
+    this.cachedConfig = resolveConfig(configFileData);
+    return this.cachedConfig;
+  }
+
+  public async set(newConfig: GlobalConfig): Promise<GlobalConfig> {
+    this.logger.child({ newConfig, method: "set" }).debug("writing global config");
+
+    const configDiff = diff(newConfig, resolveConfig({}));
+    await this.writeToConfigFile(configDiff);
+    this.cachedConfig = newConfig;
+    return this.cachedConfig;
+  }
+
+  private async writeToConfigFile(data: GlobalConfigFileData): Promise<GlobalConfigFileData> {
+    const dataParseResult = globalConfigFileSchema.safeParse(data);
+    if (!dataParseResult.success) {
+      // TODO: mark this as a client-source error.
+      throw new TypeError(z.prettifyError(dataParseResult.error));
+    }
+
+    await this.json.write(this.filePath, dataParseResult.data);
+    return data;
+  }
+
+  private async readConfigFile(): Promise<GlobalConfigFileData> {
+    try {
+      return await this.json.read(this.filePath, globalConfigFileSchema);
+    } catch (e) {
+      if (isFileNotFoundError(e)) return {};
+
+      const error = e instanceof Error ? e : new Error(String(e));
+      this.logger
+        .child({ errorName: error.name, errorMessage: error.message })
+        .warn(`failed to read global config file`);
+      throw e;
+    }
+  }
+}
+
+function isFileNotFoundError(e: unknown): boolean {
+  return e instanceof Error && "code" in e && e.code === "ENOENT";
+}
+
+/** Recursively diffs two objects, comparing leaf values by reference equality. Returns only the fields in a that differ from b */
+function diff<T extends Record<string, unknown>>(a: T, b: T): DeepPartial<T> {
+  const result: Record<string, unknown> = {};
+
+  for (const key of Object.keys(a)) {
+    const aVal = a[key];
+    const bVal = b[key];
+
+    if (isRecord(aVal) && isRecord(bVal)) {
+      const nested = diff(aVal, bVal);
+      if (Object.keys(nested).length > 0) {
+        result[key] = nested;
+      }
+    } else if (aVal !== bVal) {
+      result[key] = aVal;
+    }
+  }
+
+  return result as DeepPartial<T>;
+}
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === "object" && value !== null && !Array.isArray(value);
+}
--- a/src/globalConfig/config.tsx
+++ b/src/globalConfig/config.tsx
@@ -1,40 +1,27 @@
 import type { GlobalConfig, GlobalConfigFileData } from "./types";
+
 /**
  * Default values for the global config. Includes a unique installationId for each process.
  */
-export const getDefaultGlobalConfig = once<GlobalConfig>(() => ({
+export const DEFAULT_GLOBAL_CONFIG: GlobalConfig = {
   telemetry: {
     enabled: true,
     audit: false,
     endpoint: "https://telemetry.agentcore.aws.dev",
   },
   installationId: crypto.randomUUID(),
-}));
+};
 
 /**
- * Applies the given overrides from given {@link GlobalConfigFileData} to {@link getDefaultGlobalConfig} and returns the merged result
+ * Applies the given overrides from given {@link GlobalConfigFileData} to {@link DEFAULT_GLOBAL_CONFIG} and returns the merged result
  */
 export function resolveConfig(globalConfigFile: GlobalConfigFileData): GlobalConfig {
-  const defaults = getDefaultGlobalConfig();
   return {
     telemetry: {
-      enabled: globalConfigFile.telemetry?.enabled ?? defaults.telemetry.enabled,
-      audit: globalConfigFile.telemetry?.audit ?? defaults.telemetry.audit,
-      endpoint: globalConfigFile.telemetry?.endpoint ?? defaults.telemetry.endpoint,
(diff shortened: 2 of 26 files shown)


src/globalConfig/accessor.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+function buildKeys(schema: z.ZodObject, prefix: string = ""): string[] {
+  return Object.entries(schema.shape).flatMap(([key, value]) => {
+    const path = prefix ? `${prefix}.${key}` : key;
+    const unwrapped = inspectZodSchema(value as z.ZodType).innerType;
+    return unwrapped instanceof z.ZodObject ? [path, ...buildKeys(unwrapped, path)] : [path];
+  });
+}
+
+function buildSchemas(schema: z.ZodObject, prefix: string = ""): Map<string, z.ZodType> {
```
Hweinstock: we could probably combine this with `buildKeys` above into a single traversal, but found it more readable this way.

src/parsing/index.tsx
```diff
@@ -32,9 +69,26 @@ function baseType(schema: z.ZodType): string | undefined {
...
+}
+
+/**
+ * Peels optional/default/nullable/readonly wrappers off a schema to determine
+ * the option shape, any default, and the underlying inner type.
+ *
+ * @param schema - The zod schema to inspect.
+ * @returns An inspection describing the schema's structure.
+ */
+export function inspectZodSchema(schema: z.ZodType): ZodSchemaInspection {
```
Hweinstock: this is pretty close to a generalization of `basetype` helper above now that we extract inner type. I definitely think we can unify them, but left it OOS for this PR.

src/handlers/config/handler.tsx
```diff
@@ -21,8 +22,31 @@ export const createConfigHandler = (io: AppIO) =>
...
-      // TODO: implment a real handler that update the global config via injected config accessor.
-      io.stdout.write(`updating config key=${args.key}, value=${args.value}\n`);
+    handle: async (ctx, _flags, args) => {
+      const globalConfigAccessor = ctx.require(GlobalConfigKey);
+      const jsonRenderer = ctx.require(JsonRendererKey);
+
+      // print entire config when key is missing.
+      if (!args.key) {
+        const fullConfig = await globalConfigAccessor.toObject();
+        if (fullConfig !== undefined) jsonRenderer.renderJson(fullConfig);
```
Hweinstock: this mirrors the behavior with git config AFAICT, if the value is unset, we don't print anything.

Hweinstock:
typecheck fixed in https://github.com/aws/agentcore-cli/pull/1756#top

Hweinstock:
rebased onto create command work.

AlexanderRichey (changes requested):
I feel that the level of complexity this PR introduces is out of proportion with the functionality gained. Let's sync on this and see if we can develop a simpler approach.

AlexanderRichey:
I have some prior art here that might be useful to look at: https://github.com/AlexanderRichey/yagss/blob/main/internal/builder/config.go

Hweinstock:
converted to draft while I rework based on offline discussion from above, will take out of draft when ready for another review.

Hweinstock:
Reworked the implementation to simplify:
- config access is now done directly on a typed object rather than a type safe datastore interface. 
- partially invalid configs are no longer supported. 
- mid-execution config updates are no longer respected. 
- no longer leverages shared coercion functions from router, instead leverages zod preprocessing.

Hweinstock:
taking back into draft to address some comments on https://github.com/aws/agentcore-cli/pull/1794, that are relevant here as well. Specifically, 
- use type instead of interface for non-extended shape. 
- prefer classes over direct definitions.

AlexanderRichey (changes requested):
This is a nice improvement over the initial rev, but I still think there's a lot of room to simplify. Let me sketch what I have in mind.

First, `ReadWriteJson` needs to become generic to justify itself as an interface. The current `JsonDataSource` is way too specific—and it's name is a bit confusing since it can only be used to read and write one file. Here's an implementation, thanks to Claude, that should handle everything we need to.

```typescript
import { readFileSync, writeFileSync } from "node:fs";
import { z } from "zod";

class DeserializationError extends Error {
  constructor(path: string, options?: { cause?: unknown }) {
    super(`Failed to deserialize JSON at "${path}"`, options);
    this.name = "DeserializationError";
  }
}

interface ReadWriteJson {
  read<T>(path: string, schema: z.ZodType<T>): T;
  write<T>(path: string, data: T): void;
}

class JsonFileStore implements ReadWriteJson {
  read<T>(path: string, schema: z.ZodType<T>): T {
    const contents = readFileSync(path, "utf-8");

    let parsed: unknown;
    try {
      parsed = JSON.parse(contents);
    } catch (cause) {
      // The file wasn't valid JSON.
      throw new DeserializationError(path, { cause });
    }

    const result = schema.safeParse(parsed);
    if (!result.success) {
      // The file was valid JSON but didn't match the shape of T.
      throw new DeserializationError(path, { cause: result.error });
    }
    return result.data;
  }

  write<T>(path: string, data: T): void {
    const contents = JSON.stringify(data, null, 2);
    writeFileSync(path, contents, "utf-8");
  }
}
```

With this, the implementation of `GlobalConfigAccessor` should become straightforward.

```typescript
// ---------------------------------------------------------------------------
// Global config schema and types
// ---------------------------------------------------------------------------

export const globalConfigSchema = z.object({
  telemetry: z
    .object({
      enabled: z.boolean().optional(),
      endpoint: z.string().optional(),
      audit: z.boolean().optional(),
    })
    .optional(),
  installationId: z.uuid().optional(),
});

/** Inferred shape of the global config from {@link globalConfigSchema}. */
export type GlobalConfig = z.infer<typeof globalConfigSchema>;

/** Reads and writes global CLI configuration. */
export interface GlobalConfigAccessor {
  /** Returns the current global config. */
  get(): Promise<GlobalConfig>;
  /** Validates and persists a new config. Throws on invalid shape. */
  set(updated: GlobalConfig): Promise<GlobalConfig>;
}

// ---------------------------------------------------------------------------
// Accessor implementation
// ---------------------------------------------------------------------------

export class DefaultGlobalConfigAccessor implements GlobalConfigAccessor {
  constructor(
    private readonly store: ReadWriteJson,
    private readonly path: string,
  ) {}

  async get(): Promise<GlobalConfig> {
    try {
      return this.store.read(this.path, globalConfigSchema);
    } catch (err) {
      if (isFileNotFound(err)) {
        // No config yet (first run, fresh install). An empty object is a
        // valid GlobalConfig because every field in the schema is optional.
        // Or throw some error here.
        return {};
      }
      throw err;
    }
  }

  async set(updated: GlobalConfig): Promise<GlobalConfig> {
    // Even though `updated` is typed as GlobalConfig, we re-validate at
    // runtime: the type is only a compile-time guarantee, and callers can
    // defeat it (e.g. a value cast with `as`, or data that originated as
    // `unknown`). parse() throws a ZodError on invalid shape, satisfying
    // the interface's "throws on invalid shape" contract.
    const validated = globalConfigSchema.parse(updated);
    this.store.write(this.path, validated);
    return validated;
  }
}

function isFileNotFound(err: unknown): boolean {
  return (
    typeof err === "object" &&
    err !== null &&
    "code" in err &&
    (err as { code?: unknown }).code === "ENOENT"
  );
}
```

Now, all you'd have to do is wire this stuff up in the handler chain using DI.

```typescript
const store = new JsonFileStore();
const accessor = new DefaultGlobalConfigAccessor(store, "/path/to/config.json");
```

Note: We might want to adjust `ReadWriteJson` so that it's async, which would be more consistent with the whole codebase.

Hweinstock:
Aligned implementation closer with proposal. swapped to async, changed some names, and swapped the `set` argument to be untyped to avoid having to lie to the typesystem to pass the parameter.

src/globalConfig/accessor.tsx
```diff
@@ -0,0 +1,61 @@
...
+    try {
+      this.cachedConfig = await this.json.read(this.filePath, globalConfigSchema);
+      return this.cachedConfig;
+    } catch (e) {
+      if (isFileNotFoundError(e)) return {};
+      throw e;
+    }
+  }
+
+  public async set(newConfig: Record<string, unknown>): Promise<GlobalConfig> {
```
AlexanderRichey: Why not use the `GlobalConfig` type here? That would give us type-safety and any non-GlobalConfig value passed would fail at compile time.

src/globalConfig/index.tsx
```diff
@@ -0,0 +1,4 @@
+export { type GlobalConfigAccessor, type GlobalConfig, type ReadWriteJson } from "./types";
+export { globalConfigSchema, isValidKey } from "./schemas";
+export { DefaultGlobalConfigAccessor } from "./accessor";
+export { FsReadWriteJson } from "./json";
```
AlexanderRichey, line 3: This should probably be in its own `json` package. We can handle that in a follow up.

src/globalConfig/schemas.tsx
```diff
@@ -0,0 +1,110 @@
+import z from "zod";
+
+/** Zod boolean schema that coerces string values (`"true"`, `"1"`, `"false"`, `"0"`) to booleans. */
```
AlexanderRichey: Is this the current behavior? If not, I don't think we should support this. JSON has bools.

src/globalConfig/schemas.tsx
```diff
@@ -0,0 +1,110 @@
...
+const coercedBoolean = z.preprocess((val) => {
+  if (typeof val === "string") {
+    if (val === "true" || val === "1") return true;
+    if (val === "false" || val === "0") return false;
+  }
+  return val;
+}, z.boolean());
+
+/** Creates a Zod object schema that coerces JSON strings into objects before validation. */
+const coerceObject = <TShape extends z.ZodRawShape>(shape: TShape) =>
```
AlexanderRichey: I'm not sure I'm following why we need this. We read a JSON string with JSON.parse, which should parse nested objects just fine, if that's what this is intended to handle.

src/globalConfig/schemas.tsx
```diff
@@ -0,0 +1,110 @@
...
+ * Schema for the global config file. {@link GlobalConfigAccessor} is built from this definition.
+ * All fields should be optional; missing fields should be overridden with reasonable defaults in the consuming code.
+ */
+export const globalConfigSchema = z.object({
+  telemetry: coerceObject({
+    enabled: coercedBoolean.optional(),
+    endpoint: z.string().optional(),
+    audit: coercedBoolean.optional(),
+  }).optional(),
+  installationId: z.uuid().optional(),
```
AlexanderRichey: What's the argument in favor of using coerceObject and coercedBoolean over standard object and bool?

src/globalConfig/schemas.tsx
```diff
@@ -0,0 +1,110 @@
...
+    if (getDef(inner).type === "object") {
+      for (const nested of buildKeySet(inner as z.ZodObject, fullPath)) {
+        paths.add(nested);
+      }
+    }
+  }
+
+  return paths;
+}
+
```
AlexanderRichey: I'm not following what all this code is for.

src/globalConfig/types.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+/** Inferred shape of the global config from {@link globalConfigSchema}. */
+export type GlobalConfig = z.infer<typeof globalConfigSchema>;
+
+/** Reads and writes global CLI configuration. */
+export interface GlobalConfigAccessor {
+  /** Returns the current global config. */
+  get(): Promise<GlobalConfig>;
+  /** Validates and persists a new config. Throws on invalid shape. */
+  set(newConfig: Record<string, unknown>): Promise<GlobalConfig>;
```
AlexanderRichey: Same as above comment. This should accept a type.

src/handlers/config/handler.tsx
AlexanderRichey: Shouldn't the options be enumerated here? Enumerating the options here would be far simpler than the current `isValidKey()` code.

AlexanderRichey (changes requested):
This is looking really good. The main things blocking for me at the moment are (1) supporting (what I think are unnecessary) bool types, when JSON has a native bool type, and (2) adding a whole bunch of code to support `isValidKey()` when simpler options seem to be available.

src/handlers/config/handler.tsx
Hweinstock: I like that idea!

src/globalConfig/types.tsx
Hweinstock: I was originally thinking its more "honest" since we already validate it internally to accept unknown. This also avoids the cast at the config handler call site. However, I think I agree that its easier to type this, and cast for exceptions so that we get the static check in as many places as possible.

src/globalConfig/accessor.tsx
Hweinstock: answered below, but I'm aligned with typing it here.

src/globalConfig/index.tsx
Hweinstock: +1, I figure we pull it out once we have another consumer.

src/globalConfig/schemas.tsx
Hweinstock: discussed in person, this was for coercion, but has a bad side effect of accepting odd looking configs.

Hweinstock:
Taking into draft while I address the comments and look to simplify.

src/globalConfig/schemas.tsx
Hweinstock: removed based on offline conversation

src/globalConfig/schemas.tsx
Hweinstock: answered offline.

Hweinstock:
one other approach I considered is duplicating the schema into:
```
const CONFIG_KEY_TYPES = {
    "telemetry": "object",
    "telemetry.enabled": "boolean",
    "telemetry.endpoint": "string",
    "telemetry.audit": "boolean",
    "installationId": "string",
  };
```

Then using that instead of walking the zod schema to validate keys and paths. It would remove some complexity at the cost of maintaining two configs. 

However, I feel like building this at runtime is easier and avoids the drift (but does introduce more complexity). Especially since if we go with the explicit duplicate mapping, we'd probably want a unit test verifying they don't drift which might result in writing the same code anyway.

Hweinstock:
Main changes:
- now treats the file as a source of overrides, rather than source of truth. 
- maintains a default global config in memory and a resolve function to allow file values to override (env vars too at some point). 
- all coercion happens in the handler by walking the default config value.

src/globalConfig/config.tsx
```diff
@@ -0,0 +1,40 @@
+import type { GlobalConfig, GlobalConfigFileData } from "./types";
+/**
+ * Default values for the global config. Includes a unique installationId for each process.
+ */
+export const getDefaultGlobalConfig = once<GlobalConfig>(() => ({
```
AlexanderRichey: Any reason why we can't have this?

```typescript
const DEFAULT_GLOBAL_CONFIG =  {
  telemetry: {
    enabled: true,
    audit: false,
    endpoint: "https://telemetry.agentcore.aws.dev",
  },
  installationId: crypto.randomUUID(),
} as GlobalConfig
```

src/globalConfig/accessor.tsx
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    this.cachedConfig = resolveConfig(configFileData);
+    return this.cachedConfig;
+  }
+
+  public async set(newConfig: GlobalConfig): Promise<GlobalConfig> {
+    this.logger.child({ newConfig, method: "set" }).debug("writing global config");
+
+    const configDiff = diff(newConfig, resolveConfig({}));
```
AlexanderRichey: What's the upshot of doing this? Could this cause fields to be omitted to the written config file?

AlexanderRichey: Oh I think I get it now. The idea is that you write only the fields that differ from the default config. A little hard to follow since I'd expect `resolveConfig` to resolve the config and not return the default, but I get the approach.

AlexanderRichey (dismissed):

src/globalConfig/config.tsx
Hweinstock: I originally wasn't confident that loading the module in multiple places could cause different installationIds here. It looks like this isn't the case, and module resolution should handle this for us , but thought it was clearer to make this explicit in the code. 

However, it sounds like maybe this isn't adding clarity and instead only adds complexity. I'll swap back to the constant.

src/globalConfig/accessor.tsx
Hweinstock: That makes sense. I agree the naming `resolveConfig` is misleading. I've renamed it `applyOverrides` and accept the defaults its overriding to make this clearer.

AlexanderRichey (approved):

src/handlers/config/handler.tsx
```diff
@@ -1,28 +1,134 @@
...
+export const createConfigHandler = () =>
   createHandler({
     name: "config",
     description: "read/write global config values",
     arguments: [
       argument(
         "key",
         "config key in JSON path notation (e.g. telemetry.enabled)",
-        z.string().optional(),
+        z.enum(getKeys(DEFAULT_GLOBAL_CONFIG)).optional(),
```
nborges-aws, line 21: Should we filter out installationId from the returned keys here? Otherwise users technically could directly write the installationId. Probably not a huge risk/issue but worth a thought

nborges-aws (approved):

src/handlers/config/handler.tsx
Hweinstock: discussed offline.

PR #1793, opened 2026-07-17
feat(project): scaffold `agentcore project create` with --template flag

tejaskash:
## What

First horizontal slice of the CLI project-commands re-org: introduces the `project` namespace with a stubbed `create` leaf.

## Why

Per the project-commands restructure, project (`create`, `add`, `deploy`, ...) move under `agentcore project`

## Scope 
- CLI scaffolding only. 

## Tests
- `create` throws not-implemented
- known `--template` value accepted (parses, then hits the stub)
- unknown `--template` value rejected by the enum

PR #1794, opened 2026-07-20
feat(project): scaffold project manager interface

Hweinstock:
### Problem 
We are currently missing a way to create and find existing projects on the local filesystem. 

### Solution
- add ProjectManager interface to core interface with stub implementations.
- migrate core input to a config object instead of positional arguments. 
- design re-usable `withProject` middleware for commands that requires an existing project. 

Note: this PR does not implement e2e functionality, but sets up the scaffolding for future contributions. 

We'll likely want a working create command before filling in all of these gaps. 

### Verification
- added placeholder unit tests. 
- running `agentcore project create --> Error: ProjectManager.create is not implemented yet`

### DIFF
--- a/src/core/index.tsx
+++ b/src/core/index.tsx
@@ -22,12 +22,12 @@ export type {
   CreateIamClient,
 } from "./types";
 
-interface CoreClientConfig {
+type CoreClientConfig = {
   createControlClient: CreateControlClient;
   createDataClient: CreateDataClient;
   createIamClient: CreateIamClient;
   logger: Logger;
-}
+};
 
 // CoreClient is the single entry point to the Bedrock AgentCore APIs. It owns the
 // underlying SDK clients (one per config, created on demand from the injected
--- a/src/core/index.tsx
+++ b/src/core/index.tsx
@@ -12,7 +12,7 @@ import type {
 } from "./types";
 import type { Logger } from "../logging";
 import type { ProjectManager } from "../handlers/project/types";
-import { createProjectManager } from "./project";
+import { FsProjectManager } from "./project";
 
 export type {
   AwsClients,
@@ -55,7 +55,7 @@ export class CoreClient implements AwsClients {
     this.createIamClient = config.createIamClient;
     this.logger = config.logger;
 
-    this.projectManager = createProjectManager({
+    this.projectManager = new FsProjectManager({
       logger: this.logger.child({ module: "projectManager" }),
     });
   }
--- a/src/core/index.tsx
+++ b/src/core/index.tsx
@@ -10,6 +10,9 @@ import type {
   CreateDataClient,
   CreateIamClient,
 } from "./types";
+import type { Logger } from "../logging";
+import type { ProjectManager } from "../handlers/project/types";
+import { createProjectManager } from "./project";
 
 export type {
   AwsClients,
@@ -18,6 +21,13 @@ export type {
   CreateDataClient,
   CreateIamClient,
 } from "./types";
+
+interface CoreClientConfig {
+  createControlClient: CreateControlClient;
+  createDataClient: CreateDataClient;
+  createIamClient: CreateIamClient;
+  logger: Logger;
+}
 
 // CoreClient is the single entry point to the Bedrock AgentCore APIs. It owns the
 // underlying SDK clients (one per config, created on demand from the injected
@@ -28,15 +38,27 @@ export class CoreClient implements AwsClients {
   private dataClients = new Map<string, BedrockAgentCoreClient>();
   private iamClients = new Map<string, IAMClient>();
 
+  private readonly createControlClient: CreateControlClient;
+  private readonly createDataClient: CreateDataClient;
+  private readonly createIamClient: CreateIamClient;
+  private logger: Logger;
+
   // Feature-scoped sub-clients. Access as e.g. `coreClient.harness.getHarness(...)`.
   readonly harness: HarnessClient = new HarnessClient(this);
   readonly runtime: RuntimeClient = new RuntimeClient(this);
 
-  constructor(
-    private readonly createControlClient: CreateControlClient,
-    private readonly createDataClient: CreateDataClient,
-    private readonly createIamClient: CreateIamClient,
-  ) {}
+  readonly projectManager: ProjectManager;
+
+  constructor(config: CoreClientConfig) {
+    this.createControlClient = config.createControlClient;
+    this.createDataClient = config.createDataClient;
+    this.createIamClient = config.createIamClient;
+    this.logger = config.logger;
+
+    this.projectManager = createProjectManager({
+      logger: this.logger.child({ module: "projectManager" }),
+    });
+  }
 
   // control returns the control-plane client for `config`, creating and caching it
   // on first use.
--- a/src/core/project/manager.tsx
+++ b/src/core/project/manager.tsx
@@ -1,20 +1,26 @@
-import type { ProjectManager } from "../../handlers/project/types";
+import type {
+  CreateProjectInput,
+  FindProjectInput,
+  Project,
+  ProjectManager,
+} from "../../handlers/project/types";
 import type { Logger } from "../../logging";
 
-type CreateProjectManagerConfig = {
+type ProjectManagerConfig = {
   logger: Logger;
 };
 
 /**
- * Creates a {@link ProjectManager} that relies on the local file system to manage access to projects.
+ * An implementation of {@link ProjectManager} that relies on the local file system to manage access to projects.
  */
-export function createProjectManager(_config: CreateProjectManagerConfig): ProjectManager {
-  return {
-    find: (_input) => {
-      throw new Error(`ProjectManager.find is not implemented yet`);
-    },
-    create: (_input) => {
-      throw new Error(`ProjectManager.create is not implemented yet`);
-    },
-  };
+export class FsProjectManager implements ProjectManager {
+  constructor(_config: ProjectManagerConfig) {}
+
+  public find(_input: FindProjectInput): Promise<Project> {
+    throw new Error(`ProjectManager.find is not implemented yet`);
+  }
+
+  public create(_input: CreateProjectInput): Promise<Project> {
+    throw new Error(`ProjectManager.create is not implemented yet`);
+  }
 }
--- a/src/core/project/manager.tsx
+++ b/src/core/project/manager.tsx
@@ -1,9 +1,9 @@
 import type { ProjectManager } from "../../handlers/project/types";
 import type { Logger } from "../../logging";
 
-interface CreateProjectManagerConfig {
+type CreateProjectManagerConfig = {
   logger: Logger;
-}
+};
 
 /**
  * Creates a {@link ProjectManager} that relies on the local file system to manage access to projects.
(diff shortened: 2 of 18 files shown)


tejaskash (commented):
I like that its available and injected through core.

src/handlers/project/types.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Manages project lifecycle: creation (scaffolding) and discovery
+ */
+export interface ProjectManager {
+  /** Scaffold a new AgentCore project from the given template. */
+  create(input: CreateProjectInput): Promise<Project>;
+
+  /** Locate an existing AgentCore project. Returns undefined if no project can be found. */
+  find(input: FindProjectInput): Promise<Project | undefined>;
```
tejaskash: Are these the only methods that are supported in `ProjectManager` , I would imagine that build and deploy are also members of this class.

tejaskash (dismissed):
LGTM, 1 clarification, not blocking

src/handlers/project/types.ts
Hweinstock: good question, I was envisioning the `ProjectManager` to specifically control access to projects, then operations that act on an individual project would live in the Project interface. 

With this model, build and deploy would live on `Project` instead of the `Manager`. 

I'm not a huge fan of the name `ProjectManager`, since I am not sure it communicates this well, so I'm open to other ideas.

tejaskash: What do you think about find being separate and Project also owning create? 


```typescript
class Project {
    scaffold(); 
    add();
    remove();
    build();
    deploy();
}
```

Hweinstock: Would scaffold be a static method? I guess I'm wondering how I go from nothing to a Project. Would it look like:
```
const p new Project(...)
p.scaffold(...)

const p = findProject(...)
p.add(...)
```
If so whats the behavior if I try to interact with the project before scaffolding, or scaffold when its already scaffolded?

src/core/project/manager.tsx
```diff
@@ -0,0 +1,20 @@
+import type { ProjectManager } from "../../handlers/project/types";
+import type { Logger } from "../../logging";
+
+interface CreateProjectManagerConfig {
```
AlexanderRichey: I think this should be a type and not an interface. Interfaces are used for things that can have multiple implementations, whereas types require specific, concrete/instantiated entities. That's what we need here for the input params for this constructor, so a standard type seems more appropriate.

src/core/project/manager.tsx
```diff
@@ -0,0 +1,20 @@
...
+  logger: Logger;
+}
+
+/**
+ * Creates a {@link ProjectManager} that relies on the local file system to manage access to projects.
+ */
+export function createProjectManager(_config: CreateProjectManagerConfig): ProjectManager {
+  return {
+    find: (_input) => {
+      throw new Error(`ProjectManager.find is not implemented yet`);
```
AlexanderRichey: I know we're just setting up the scaffolding here, but I tend to prefer class-based implementations of interfaces. They usually give you more flexibility going forward and, generally speaking, more developers are familiar with class-based patterns.

src/core/index.tsx
```diff
@@ -18,6 +21,13 @@ export type {
   CreateIamClient,
 } from "./types";
 
+interface CoreClientConfig {
```
AlexanderRichey: Like the above, this should be a type.

src/handlers/project/create/index.ts
```diff
@@ -1,9 +1,12 @@
 import z from "zod";
 import { createHandler, flag } from "../../../router";
+import { PROJECT_TEMPLATES, type ProjectManager } from "../types";
 
-export const PROJECT_TEMPLATES = ["placeholder"] as const;
+interface CreateProjectHandlerConfig {
```
AlexanderRichey: I think this should be a type too. Or just make the `ProjectManager` a standard argument. I suspect we'll want to pass `io` as well here.

src/handlers/project/index.ts
```diff
@@ -1,10 +1,15 @@
 import { Router } from "../../router";
 import { createCreateProjectHandler } from "./create";
+import type { ProjectManager } from "./types";
 
-export function createProjectHandler(): Router {
+interface ProjectHandlerConfig {
```
AlexanderRichey: Same as above. Let's make this a type. I think we'll want `io` on here too as we move forward.

src/handlers/project/types.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export interface CreateProjectInput {
+  /** The project template to scaffold from. */
+  template: (typeof PROJECT_TEMPLATES)[number];
+}
+
+export interface FindProjectInput {
+  /** A path to search from when locating the project root. */
+  filePath: string;
+}
```
AlexanderRichey: These two should be types as well.

src/handlers/project/types.ts
```diff
@@ -0,0 +1,28 @@
+/** Available project templates for scaffolding new AgentCore projects. */
+export const PROJECT_TEMPLATES = ["placeholder"] as const;
```
AlexanderRichey: I tend to think this way of typing this stuff out would be easier to work with.

```typescript
export type Template = "barebones" | "strands" | "langgraph";

export const TEMPLATE_BAREBONES: Template = "barebones";
```

src/handlers/project/types.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export interface FindProjectInput {
+  /** A path to search from when locating the project root. */
+  filePath: string;
+}
+
+/**
+ * Exposes the ability to configure, develop, and deploy a resolved AgentCore project.  
+ */
+export interface Project {}
```
AlexanderRichey: Should project be abstract or concrete? I tend to think it should be concrete. However it might implement some or other abstract interface.

src/handlers/project/types.ts
tejaskash: Those are good points, and there's only hacky ways to overcome them. I'm okay with the name `ProjectManager` and I'm convinced that `scaffold` should be in `ProjectManager`

AlexanderRichey: I tend to agree with Tejas' general direction here. I think we'll want something like this. But it's not clear to me what all the types should look like yet.

```typescript
interface ProjectManager {
  create(input: CreateInput): Promise<Project>;
  resolve(): Promise<Project>;
  add(input: AddInput): Promise<Project>;
  remove(input: RemoveInput): Promise<Project>;
  build(input: BuildInput): AsyncGenerator<BuildEvent ...>;
  deploy(input: DeployInput): AsyncGenerator<DeployEvent ...>;
}
```

Maybe we can just start by defining `create(...)` and `resolve(...)` in this PR and add the other methods as they become clearer.

It's also not clear to me what `Project` should be just yet, so we can start with something minimal like this:

```typescript
type Project {
  name: string
}
```

AlexanderRichey (changes requested):

src/core/project/manager.tsx
Hweinstock: I've always preferred interfaces to types since they tend to provide better error messages, and leave the door open if we want to extend them later on. The typescript language docs appear to recommend interfaces as the default [here](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces), but I do not have a strong preference here, especially since its very unlikely we'll want to extend these shapes. 

I'll make the swap to types here.

src/core/project/manager.tsx
Hweinstock: I definitely agree that more developers are familiar with classes, but I personally find them overly verbose in TS. One specific example is that when implementing an interface with an object, the compiler is able to infer the types (i.e. don't need type annotations on find here), which I find to be more readable and avoid defining the same types twice. 

I tend to only reach for classes if I need inheritance or `instanceof` checks, but I don't think either is relevant here.

src/handlers/project/types.ts
Hweinstock: I tried to keep the same types from https://github.com/aws/agentcore-cli/pull/1793 since I didn't see this as part of the scope of this PR. 

However, I do actually like this style of deriving the type from an array/object because it gives greater flexibility by allowing the values to be enumerated at runtime. What do you think of swapping to an object so that we can have named templates too (like `TEMPLATE_BAREBONES` above)?

Ex. 
```
const TEMPLATES = {
   BAREBONES: 'barebones', 
   STRANDS: 'strands,
} as const;

type Template = (typeof TEMPLATES)[keyof typeof TEMPLATES] 
```

src/handlers/project/types.ts
Hweinstock: My understanding is that we need to define an interface here to preserve the existing dependency inversion pattern. If that interface isn't `Project`, what would it be?

I could imagine a capability split like `Deployable`, `Buildable`, etc. but that feels overly complex for what we're doing. 

I could also imagine a java-like pattern of `IProject` or `ProjectImpl`, but that doesn't feel like idiomatic typescript. 

Are you suggesting we break the dependency inversion pattern here, or is there a naming/structure option I'm missing here?

src/handlers/project/types.ts
tejaskash: I definitely planned to have an enum or type for this. This is just a placeholder for scaffolding

src/core/project/manager.tsx
Hweinstock: chatted with @tejaskash, sounds like I'm in the minority. swapping implementation to classes.

src/core/project/manager.tsx
AlexanderRichey: There's an important distinction between types and interfaces that I think we should respect here. Types hold data of a determinate form. They're not implemented by anything else—they are the thing. They are, in this way, concrete. Interfaces, in contrast, are abstract; they don't have a concrete existence. They must be implemented and instantiated by something else that is concrete. In the case here, what we want is a concrete input with data, which makes a type the right tool for the current job. Making this an interface would bring in a bunch of unnecessary functionality and, as you said, make it possible to extend, which is precisely what we wouldn't want in this case.

src/handlers/project/types.ts
AlexanderRichey: Nice! This approach looks good to me. 👍

src/handlers/project/types.ts
AlexanderRichey: > My understanding is that we need to define an interface here to preserve the existing dependency inversion pattern. If that interface isn't Project, what would it be?

I'm not sure this argument follows exactly. If you look at the interfaces defined for harness, for example, you'll see that the methods take types as input and output objects. The relevant interface is `ProjectManager`, not project itself.

> Are you suggesting we break the dependency inversion pattern here, or is there a naming/structure option I'm missing here?

I'm suggesting that project has concrete thing, with concrete data, so it's not an abstract entity in the same way an interface is.

Hweinstock: Ahh I think I understand. My original plan was:

> With the project manager and project abstraction, we can start filling in the gaps (ex. add resources, deploy, dev, etc.) can all be exposed through the new **Project** interface.

However, based on your comments, it sounds like we're looking for Project to be pure data, and not deliver functionality itself with all functionality on the project manager class. 

Let me address this to align with your comment below.

Hweinstock:
Updated implementation to treat project as data instead of an interface. ProjectManager will host all project functionality.

AlexanderRichey (approved):

tejaskash (approved):

PR #1797, opened 2026-07-20
feat: add runtime read-only control-plane commands

aidandaly24:
## Description

Adds project-free, command-line access to inspect AgentCore Runtimes:

- `runtime get` and `runtime list`
- `runtime version get` and `runtime version list`
- `runtime endpoint get` and `runtime endpoint list`

The commands follow the Harness handler/Core structure, use the existing JSON output path, and expose one-page pagination through `--max-results` and `--next-token`. Endpoint lookup uses `--qualifier`, mapped to the control-plane `endpointName`.

Runtime appears in the root TUI command menu, and bare Runtime groups print help until their TUI routes are implemented. Runtime TUI workflows, invocation, mutations, project or deployment resolution, tags and policies, agent-card operations, logs, and traces are outside this change.

## Related Issue

N/A

## Documentation PR

N/A. README command examples are included in this PR.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Verified from current HEAD `73434193`:

- [x] `bun test --coverage --coverage-reporter=lcov` (254 passed, 0 failed)
- [x] `bun test --coverage` (254 passed, 0 failed; 93.64% functions, 96.11% lines; Runtime Core and handlers 100%)
- [x] `bun test src/handlers/runtime src/core` (30 passed, 0 failed)
- [x] `bun run lint:check`
- [x] Changed-surface `prettier --check`
- [x] `bun run typecheck`
- [x] `bun run build`
- [x] `bun run compile` (all six targets)

Additional verification from Runtime-handler-equivalent commit `22634e54`:

- [x] `bun run format:check`
- [x] `bun audit`
- [x] `bun pm pack`
- [x] Built help smokes for all nine Runtime command and group routes
- [x] Built CLI rejection smokes for the excluded `runtime logs` and `runtime traces` groups
- [x] All six live read commands and a continuation-token request in `us-west-2`

The committed Runtime fixtures were recorded with `AWS_PROFILE=e2e-test` from shared E2E account `685197708687`. The dedicated Runtime `agentcore_cli_runtime_read_only_fixture-wZ7V4Q6vhx` is retained with versions `1` and `2` plus `DEFAULT` and `runtimeReadOnlyFixture` endpoints for repeatable team recording. The fixture set contains no deploy-account identifiers or credential material.

Live read-command verification used an existing Runtime and did not mutate it. Subsequent commits change root TUI visibility and tests/fixtures only; the Runtime handlers and Core are unchanged from that live verification.

## Checklist

---


src/components/RouterScreen.tsx
```diff
@@ -57,7 +57,10 @@ export function RouterScreen({ ctx, path }: RouterScreenProps) {
 
   const command = resolveCommand(ctx.require(CommandKey), path);
   const options: Option[] = useMemo(
-    () => command.commands.map((c) => ({ name: c.name(), description: c.description() })),
+    () =>
+      command.commands
+        .filter((command) => command.name() !== "runtime")
```
AlexanderRichey: I understand that we haven't built the TUI for Runtime yet, but still I don't think it's worth introducing this edge case. After we build the TUI in the next CR, the original code will work as intended. Even now, I think what would happen is that the help screen would be shown, which is also fine for now.

src/handlers/runtime/version/index.tsx
```diff
@@ -0,0 +1,12 @@
+import { Router } from "../../../router";
+import type { AppIO, Core } from "../../types";
+import { createHelpDefault } from "../help";
+import { createGetRuntimeVersionHandler } from "./get";
+import { createListRuntimeVersionsHandler } from "./list";
+
+export function createRuntimeVersionHandler(core: Core, io: AppIO): Router {
+  return new Router("version", "inspect AgentCore Runtime versions")
+    .default(createHelpDefault(io))
```
AlexanderRichey, line 9: This makes sense for now, but it will be replaced by `withTui()`.

src/components/RouterScreen.test.tsx
```diff
@@ -15,6 +15,7 @@ describe("menu rendering", () => {
     const frame = r.lastFrame()!;
     expect(frame).toContain("harness");
     expect(frame).toContain("manage agentcore harnesses");
+    expect(frame).not.toContain("runtime");
```
AlexanderRichey: I think we want the main screen to show runtime.

src/handlers/runtime/runtime.test.tsx
```diff
@@ -0,0 +1,354 @@
...
+      expect(result.stdout).toContain("Commands:");
+      expect(result.runtime.calls).toEqual([]);
+    },
+  );
+});
+
+describe("runtime control reads", () => {
+  test("gets a Runtime and renders the complete response", async () => {
+    const result = await run(["runtime", "get", "--id", "runtime-1"]);
+
```
AlexanderRichey, line 79: Is there a reason why we're not doing golden file testing for some of these cases? I think that would be both simpler and more robust.

AlexanderRichey (changes requested):
Great work on this. I think we should make two changes before merging. First, I think runtime should show in the RouterScreen, even though the TUI isn't there yet. I don't think it's worth it to carve out a special case for runtime when, in the next CR, we'll have to reverse it. Second, I think we should streamline the tests to be more consistent. We now have `run()`, which manually sets up mocks, and `runFixture()`, which returns golden file mocks. Let's only use the golden file testing pattern, if possible. So `run()` should become what `runFixture()` is and we'll just have to generate fixtures for the couple missing cases. You might have to just create a bunch of runtimes, versions, and endpoints to make this work easily when running the tests in record mode.

src/handlers/runtime/runtime.test.tsx
aidandaly24: This test now uses `matchGolden()`

src/core/runtime.tsx
```diff
@@ -0,0 +1,91 @@
...
+    nextToken: string | undefined,
+    maxResults: number | undefined,
+    options: CoreOptions,
+  ): Promise<ListAgentRuntimesResponse> {
+    return this.clients
+      .control(toClientConfig(options))
+      .send(new ListAgentRuntimesCommand({ nextToken, maxResults }));
+  }
+
+  async listRuntimeVersions(
```
Hweinstock, line 62: I see this follows the existing pattern, but what do think of avoiding positional arguments for functions with many arguments? I've found that using input objects makes it easier to handle optional args, and easier to read at the callsite.

src/handlers/runtime/endpoint/get/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const createGetRuntimeEndpointHandler = (core: Core) =>
+  createHandler({
+    name: "get",
+    description: "get a Runtime endpoint",
+    flags: [
+      flag("id", "the ID of the Runtime", z.string().optional()),
+      flag("qualifier", "the endpoint name (qualifier)", z.string().optional()),
+    ],
+    handle: async (ctx, flags) => {
+      if (!flags.id) {
```
Hweinstock, line 16: I remember this being brought up, but I can't remember. Why exactly can't we make these fields required in the schema to have zod validate this for us?

AlexanderRichey (dismissed):

src/handlers/runtime/runtime.test.tsx
```diff
@@ -0,0 +1,236 @@
...
+describe("runtime control reads", () => {
+  test("gets a service-valid long Runtime ID through the real Core", async () => {
+    const stdout = await run(["runtime", "get", "--id", FIXTURE_RUNTIME_ID]);
+
+    matchGolden(FIXTURES, "get.golden.json", stdout);
+    expect(JSON.parse(stdout).agentRuntimeId).toBe(FIXTURE_RUNTIME_ID);
+  });
+
+  test("lists two Runtime pages with Harness pagination names", async () => {
```
Hweinstock: what do you mean by harness pagination names here?

src/handlers/runtime/runtime.test.tsx
```diff
@@ -0,0 +1,236 @@
...
+      "--max-results",
+      "1",
+      "--next-token",
+      first.nextToken,
+    ]);
+    matchGolden(FIXTURES, "version-list-page-2.golden.json", secondPage);
+    expect(JSON.parse(secondPage).agentRuntimes).toHaveLength(1);
+  });
+
+  test("maps the endpoint qualifier for a service-valid long Runtime ID", async () => {
```
Hweinstock: what is a long runtime Id?

src/core/runtime.tsx
aidandaly24: I agree an input object would make this method easier to read in isolation, particularly with the optional pagination fields. I kept the positional signature intentionally because these Runtime methods directly mirror the existing Harness Core interface. Changing only Runtime would leave two different conventions for otherwise parallel APIs. I would prefer to keep them consistent in this PR and, if we decide input objects are the better convention, refactor both Core clients together.

src/handlers/runtime/endpoint/get/index.tsx
aidandaly24: Technically, I think we could make them required today for Runtime because there is no TUI yet (will be the next PR). The reason not to is how the router supports future TUI commands.

A required schema:
```
z.string()
```
becomes a mandatory Commander option. Commander rejects a bare command before middleware or the handler runs:
```
agentcore runtime endpoint get
```
But once runtime has TUI support, that bare command must reach `withTuiOnEmptyFlagsAndArgs`, which opens the selector screen. So, Harnerss uses optional schemas and validates inside the headless handler. I kept the same schema and headless validation pattern as the next change will be the TUI.

src/handlers/runtime/runtime.test.tsx
aidandaly24: “Long” was unclear. I meant a service-generated Runtime ID exceeding the 48-character Runtime name limit. Valid IDs can exceed that limit because the service appends a generated suffix. Thanks for this call out I renamed the test and explicitly assert that the fixture ID is longer than 48 characters.

src/handlers/runtime/runtime.test.tsx
aidandaly24: I meant here that Runtime uses the same user-facing pagination flags as Harness, `--max-results` and `--next-token`. The test name is unclear because this is the standard pagination, so I renamed it to paginates Runtime list with `--max-results` and `--next-token`.

aidandaly24:
> ## [Codecov](https://app.codecov.io/gh/aws/agentcore-cli/pull/1797?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aws) Report
> 
> ❌ Patch coverage is `85.62092%` with `44 lines` in your changes missing coverage. Please review. ✅ Project coverage is 93.25%. Comparing base ([`d4c31b1`](https://app.codecov.io/gh/aws/agentcore-cli/commit/d4c31b109c81d0ee0866d22c30664da67dc3f250?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aws)) to head ([`68afc22`](https://app.codecov.io/gh/aws/agentcore-cli/commit/68afc22d18b15a536519f5dddcfd267571b8ae68?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aws)).
> [Files with missing lines](https://app.codecov.io/gh/aws/agentcore-cli/pull/1797?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aws) 	Patch % 	Lines
> [src/testing/TestCoreClient.tsx](https://app.codecov.io/gh/aws/agentcore-cli/pull/1797?src=pr&el=tree&filepath=src%2Ftesting%2FTestCoreClient.tsx&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aws#diff-c3JjL3Rlc3RpbmcvVGVzdENvcmVDbGllbnQudHN4) 	21.42% 	[44 Missing ⚠️ ](https://app.codecov.io/gh/aws/agentcore-cli/pull/1797?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=aws)
> Additional details and impacted files
> ...

Investigated the Codecov gap. All 44 uncovered changed lines are in the private `TestRuntimeClient` stub, not production Runtime code. The Runtime CLI tests intentionally use the fixture-backed real CoreClient, and no Runtime TUI tests consume this fake yet. `src/core/runtime.tsx` and every Runtime handler have 100% line coverage. The stub currently exists only so `TestCoreClient` satisfies the expanded Core interface. Adding tests that only invoke its constant-return methods would not provide meaningful behavioral coverage. It will be exercised naturally by the Runtime TUI tests in the following PR.

AlexanderRichey (approved):

src/handlers/runtime/endpoint/get/index.tsx
Hweinstock: thanks that makes sense.

Hweinstock (approved):

PR #1802, opened 2026-07-22
feat: add Runtime read-only TUI

aidandaly24:
## Description

Adds an interactive, read-only TUI for the Runtime control-plane commands introduced in #1797.

The Runtime TUI provides route-based flows for:

- Listing and inspecting Runtimes
- Listing and inspecting immutable Runtime versions
- Listing and inspecting Runtime endpoints

Runtime follows the established Harness interaction model. Bare Runtime groups and leaves open their corresponding TUI routes, operation flags keep commands headless, and `--json` suppresses interactive rendering. Nested endpoint and version flows preserve route history so Escape returns through the selectors and command menus in order.

This change also introduces a shared token-paged table picker used by Runtime and Harness. It centralizes loading, empty, error, retry, filtering, responsive columns, transition locking, pagination, resize resets, and Escape behavior while keeping each resource's API calls, columns, row mapping, sorting, labels, and destinations in its adapter.

Runtime detail and headless output omit SDK `$metadata`. The shared table filter accepts pasted text delivered as a single terminal input event.

Runtime invocation, mutations, deployment or project resolution, logs, traces, and other resource families remain outside this change.

## Related Issue

N/A

## Documentation PR

N/A. README command and interaction documentation is included in this PR.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Verified from current HEAD `023f2af`:

- [x] `bun test` (319 passed, 0 failed)
- [x] `bun run typecheck`
- [x] `bun run lint:check`
- [x] `bun run format:check`
- [x] Manual terminal and TUI-harness verification of Runtime, version, and endpoint navigation

## Checklist

---


src/components/HarnessEndpointPicker.tsx
```diff
@@ -0,0 +1,103 @@
+import { useNavigate } from "react-router";
```
aidandaly24, line 1: I wouldn't be against sorting these into folders since I could picture this getting unwieldy as time goes on. The reason I held off was because I didn't know if maybe we just sort the primitive ones or if maybe we just want to sort into a folder called "pickers".

src/components/TokenPagedTablePicker.tsx
```diff
@@ -0,0 +1,134 @@
...
+  getValue: (row: TRow) => string | undefined;
+  onSelect: (value: string) => void;
+  onBack: () => void;
+  loadingMessage: string;
+  errorMessage: (error: Error) => string;
+  emptyMessage: string;
+  emptyPageMessage: string;
+}
+
+export function TokenPagedTablePicker<TItem, TRow extends Record<string, unknown>>({
```
aidandaly24: `TokenPagedTablePicker` centralizes the interaction state shared by Runtime and Harness resource pickers like terminal-derived page sizing, continuation-token history, loading/retry/empty states, page transitions, resize resets, filtering, selection, and keyboard navigation.

Resource-specific adapters still own their API calls, query keys, row mapping, columns, sorting, labels, selectors, and navigation destinations. This removes duplicated pagination state machines without forcing different resources into one data model, and gives the shared behavior one focused test suite.

src/handlers/runtime/withoutSdkMetadata.ts
```diff
@@ -0,0 +1,7 @@
+export function withoutSdkMetadata(data: unknown): unknown {
```
aidandaly24: This PR strips SDK `$metadata` at the Runtime handler boundary so both headless and TUI output remain resource-focused. I think we should decide whether transport metadata removal belongs in Core, JsonRenderer, or another shared normalization layer, then apply this contract consistently across resource families. What do you think, it could be done in this PR or a followup?

src/components/ui/data-table/DataTable.tsx
```diff
@@ -62,6 +66,7 @@ export function DataTable<T extends Record<string, unknown>>({
   showFooter = true,
   emptyMessage = "No data",
   focus = true,
+  selectionResetKey,
```
AlexanderRichey, line 66: What's this for?

src/components/ui/data-table/DataTable.tsx
```diff
@@ -190,12 +205,14 @@ export function DataTable<T extends Record<string, unknown>>({
           <Text color={theme.colors.muted}>
             {searchMode ? (
               <Text>
-                <Text color={theme.colors.primary}>/ Filter: </Text>
+                <Text color={theme.colors.primary}>/ {searchPlaceholder}: </Text>
```
AlexanderRichey: What makes changing this necessary?

src/components/ui/key-hint/KeyHint.tsx
```diff
@@ -15,15 +17,48 @@ export interface KeyHintProps {
...
+
+  for (const candidate of candidates) {
+    const nextWidth = itemWidth(candidate.item) + (selected.size === 0 ? 0 : ITEM_GAP);
+    if (width + nextWidth > columns) continue;
+    selected.add(candidate.index);
+    width += nextWidth;
+  }
+
+  return keys.filter((_, index) => selected.has(index));
+}
```
AlexanderRichey: These changes add a lot of complexity. I see the upshot, but I'm not sure that now is the time to add all this. Let me pull these changes down and test the behavior a bit.

src/components/HarnessEndpointPicker.tsx
AlexanderRichey: I actually prefer a flat directory of components! I tend to think a flat directory is easier to maintain over time and that it makes it easier to find what you're looking for.

src/components/HarnessEndpointPicker.tsx
```diff
@@ -0,0 +1,103 @@
...
+      queryKey={["harness-endpoints", opts.region, harnessId]}
+      loadPage={async (token, pageSize) => {
+        const response = await core.harness.listHarnessEndpoints(harnessId, token, pageSize, opts);
+        return {
+          items: response.endpoints ?? [],
+          nextToken: response.nextToken,
+        };
+      }}
+      toRow={toRow}
+      columns={(terminalColumns) => {
```
AlexanderRichey: We're adding a whole lot of complexity here. I tend to think we shouldn't worry about this problem now. When we do take on this problem, we should solve it in a general way and not in a way that will require every user of DataTable to solve the problem again, as I think the current approach would.

src/components/JsonDetail.tsx
```diff
@@ -43,6 +56,7 @@ export function JsonDetail({ breadcrumb, isPending, error, data, loadingLabel }:
       breadcrumb={breadcrumb}
       keyHints={[
         { key: "↑↓/kj", label: "navigate" },
+        ...(error && onRetry ? [{ key: "r", label: "retry" }] : []),
```
AlexanderRichey, line 59: Should this be "retry" or "reload"?

src/components/HarnessVersionPicker.tsx
```diff
@@ -0,0 +1,88 @@
...
+          nextToken: response.nextToken,
+        };
+      }}
+      toRow={toRow}
+      columns={(terminalColumns) => {
+        const showStatus = terminalColumns >= 60;
+        const showCreatedAt = terminalColumns >= 90;
+        const statusWidth = showStatus ? 20 : 0;
+        const createdAtWidth = showCreatedAt ? 30 : 0;
+        const versionWidth = Math.max(8, terminalColumns - 2 - statusWidth - createdAtWidth);
```
AlexanderRichey: Same here. I like the idea of a component that handles this stuff, but I think we need to simplify the approach here a bit.

src/components/RouterScreen.test.tsx
```diff
@@ -1,8 +1,156 @@
...
+  overrides: Partial<GetAgentRuntimeResponse> = {},
+): GetAgentRuntimeResponse {
+  return {
+    agentRuntimeArn: "arn:aws:bedrock-agentcore:us-east-1:123456789012:runtime/runtime-123",
+    agentRuntimeName: "checkout",
+    agentRuntimeId: "runtime-123",
+    agentRuntimeVersion: "7",
+    createdAt: new Date("2026-07-19T01:02:03.000Z"),
+    lastUpdatedAt: new Date("2026-07-20T12:34:56.000Z"),
+    roleArn: "arn:aws:iam::123456789012:role/runtime-role",
```
AlexanderRichey: All of this stuff seems out of scope for RouterScreen tests. We should be testing the behavior of the router screen and not all this other stuff here.

src/handlers/runtime/withoutSdkMetadata.ts
AlexanderRichey: I agree that we shouldn't show `$metadata`. Let's handle globally in a follow up.

AlexanderRichey (changes requested):
Looks good! Blocking things are:

- Simplifying the column handling where `TokenPagedTablePicker` is used. We might want to do something similar to what you have here in a later PR, but for now let's try to keep things as simple as possible.
- Cleaning up some of the tests, particularly the ones for the `RouterScreen`.

src/components/ui/data-table/DataTable.tsx
aidandaly24: This handles terminal-height resizes. When pageSize changes, pagination resets but DataTable would otherwise retain its selected row and internal page. This resets those states while preserving filter and sort state.

src/components/JsonDetail.tsx
aidandaly24: I intentionally used "retry" because this action only appears after a failed query and invokes refetch to retry that request. I could switch it to "reload" if that makes more sense or you think would be more consistent for other places where we may want to use the "r" letter.

src/components/HarnessVersionPicker.tsx
aidandaly24: The version picker now passes static columns without terminal-specific width calculations.

src/handlers/runtime/withoutSdkMetadata.ts
aidandaly24: Makes sense. I’ll remove the Runtime-specific handling from this PR and address $metadata globally in a follow-up. I'll add a backlog item for it.

src/components/ui/data-table/DataTable.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
```
Hweinstock, line 1: unrelated to change here but noticed the path has `data-table`. Should we swap to `dataTable` to be consistent in another PR?

AlexanderRichey (dismissed):
This is looking good! Let's merge this one. We have one follow up to handle which is to get the column lengths right in the data tables. Let's discuss offline on a strategy for this.

src/components/HarnessEndpointPicker.tsx
```diff
@@ -0,0 +1,86 @@
...
+  description?: string;
+  // onSelect receives the chosen endpoint's name.
+  onSelect: (endpointName: string) => void;
+  // onEscape overrides what esc does (default: pop back in history). Hosts
+  // that embed the picker as an overlay (e.g. the chat's ctrl+t endpoint
+  // switch) pass a closer instead.
+  onEscape?: () => void;
+}
+
+// HarnessEndpointPicker fetches a harness's endpoints and renders them as a navigable
```
Hweinstock: Patrick had a nice comment a bit ago about standardizing on TSDoc format. Do we think its worth trying to follow that standard?

> I would recommend using TSDoc format consistently so you can get auto generated docs. I've spotted some functions throughout that just use //.

https://github.com/AlexanderRichey/agentcore-cli/pull/3#discussion_r3539673821

src/components/HarnessPicker.tsx
```diff
@@ -56,69 +50,35 @@ export function HarnessPicker({
...
-  });
-
-  const nextToken = list.data?.nextToken;
-  // Pagination surfaces only once a response reports more pages (nextToken).
-  const paginated = paging.pageIndex > 0 || nextToken !== undefined;
+  const goBack = () => navigate("/" + breadcrumb.slice(0, -1).join("/"));
 
   return (
-    <Layout
+    <TokenPagedTablePicker
```
Hweinstock: I really like this abstraction btw. One nit is the name `TokenPagedTablePicker` feels kinda odd to me. What do we think of `PaginatedTablePicker`? (also open to original name if we think it makes more sense)

Hweinstock (dismissed):

src/components/HarnessEndpointPicker.tsx
aidandaly24: Updated the exported Harness picker documentation to TSDoc for consistency. I left implementation-only comments as regular comments.

src/components/HarnessPicker.tsx
aidandaly24: I agree the name is currently a mouthful. Updating to your suggestion it is better.

aidandaly24:
> We have one follow up to handle which is to get the column lengths right in the data tables. Let's discuss offline on a strategy for this.

Sounds good. For the follow-up, my initial direction is to make DataTable own terminal-aware column sizing rather than having each picker calculate widths. Column definitions could declare sizing intent such as minimum/maximum width and priority, while DataTable allocates the available width, truncates predictably, and hides optional columns only when necessary. We can align offline on the exact format and validate it.

AlexanderRichey (approved):

Hweinstock (approved):

Issue #1801, opened 2026-07-21
bug: logger initialization fails on compiled binary of refactor branch.


Hweinstock:
### Problem 
Running the CLI as a compiled binary fails to run with `error: unable to determine transport target for "pino-roll"`. 

### Root Cause
When specifying `pino-roll` in the transport config, pino attempts to pull in `pino-roll` at runtime from a worker thread. This works when built for node since its able to find it in the local `node_modules`, but fails when compiled as an executable since the worker thread can't resolve the dependency. 

transport specification: https://github.com/aws/agentcore-cli/blob/aae8c4b0184d6cba0267fd9e6446bbe8ac0f471d/src/logging/fileLogger.tsx#L49

pino resolution logic: https://github.com/pinojs/pino/blob/6b344980eae3ebed904fc87caf4bba0ab9dbe946/lib/transport.js#L273
error thrown from directly below this. 

### Potential Solutions
- avoid pino-roll entirely and write to an file. (pros: simple, cons: [unbounded](https://github.com/anthropics/claude-code/issues/16093))
- move the pino-roll stream into the main thread (pros: keeps rolling, cons: complexity since we now manually setup the stream + logging in same thread, pino-roll doesn't vend types). 
- invesitgate alternative logging solutions like [`winston-daily-rotate-file`](https://github.com/winstonjs/winston-daily-rotate-file/blob/master/daily-rotate-file.js) which appears to support the same functionality, but runs in the main thread by default.

PR #1804, opened 2026-07-22
fix(binary): move logging into main thread for binary compatibility. 

Hweinstock:
### Problem 
https://github.com/aws/agentcore-cli/issues/1801

The logger is failing to initialize when the CLI is compiled to a binary. See the issue for a root cause. 

### Solution
- move the logger into the main thread by instantiating the stream with the logger. 
- revert back to using console.error at the top level to surface errors. 
- add a CI job that compiles the binary on each platform to add protection against this going forward. 

### Testing 
- CI job is passing. 
- local binary works now after the fix.


.github/workflows/compile.yml
```diff
@@ -0,0 +1,44 @@
...
+            binary: ./dist/bin/agentcore-linux-x64
+          - name: Windows
+            runner: '["codebuild-agentcore-e2e-${{ github.run_id }}-${{ github.run_attempt }}", "image:windows-1.0"]'
+            target: windows-x64
+            binary: ./dist/bin/agentcore-windows-x64.exe
+          # CodeBuild does not support macOS.
+          - name: macOS
+            runner: '["macos-latest"]'
+            target: darwin-arm64
+            binary: ./dist/bin/agentcore-darwin-arm64
```
AlexanderRichey, line 32: Would a simpler way of doing all this just be to call `make`? Bun can cross-compile.

AlexanderRichey (approved):

Hweinstock:
closed in favor of https://github.com/aws/agentcore-cli/pull/1808

.github/workflows/compile.yml
Hweinstock: I think we need to do it individually since the matrix is parallelizing across runners so each runner needs its own binary.

PR #1808, opened 2026-07-22
fix(logger): migrate to winston for simpler executable support

Hweinstock:
### Problem 
https://github.com/aws/agentcore-cli/issues/1801

This PR goes with option 3 from the issue link. 

The original attempt lives in https://github.com/aws/agentcore-cli/pull/1804, however there were some issues in that PR:
- a bug that causes logs to inconsistently show up after executing commands due to `flush` behavior on pino due how we manually manage the output stream. 
- manually maintaining a type definition for an external dependency. 
- CI job organization wasn't clear (compile vs build?). 

### Solution 
- switch to winston rotating logger since it avoids the complexity of handling the stream ourselves. Winston puts it in the main thread by default. 
- switch to organizing CI checks into `build` and `check` to mirror what [`cargo`](https://github.com/rust-lang/cargo) does. Specifically,`check` does static analysis without producing any artifacts, whereas `build` compiles, packs, and produces the build artifacts. 

### Testing
- CI now verifies the compiled binary is working. 
- logging tests passes, and verified locally:

```
rm -rf ~/.agentcore/logs
./dist/bin/agentcore-linux-x64 config k1 k2
cat ~/.agentcore/logs/output-2026-07-22.log

{"level":"debug","msg":"executing command","time":1784752392752,"cliSessionId":"7458c0f4-216a-464e-b68f-28441d229d1e","commandPath":"/agentcore/config","flags":{},"args":{"k
ey":"k1","value":"k2"}}
{"level":"debug","msg":"command executed successfully","time":1784752392753,"cliSessionId":"7458c0f4-216a-464e-b68f-28441d229d1e","commandPath":"/agentcore/config"}

```

### Notes
- try to not to squash merge since this PR includes a few components that may be useful as separate commits in the commit history.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -55,7 +55,7 @@ process.exit(
         .error();
       throw e;
     } finally {
-      await rootLogger.flush();
+      await rootLogger.end();
     }
   }),
 );
--- a/src/index.ts
+++ b/src/index.ts
@@ -50,7 +50,6 @@ process.exit(
       await rootHandler.route(argv);
     } catch (e) {
       const error = e instanceof Error ? e : new Error(String(e));
-      io.stderr.write(`${error.name}: ${error.message}\n`);
       rootLogger
         .child({ errorName: error.name, errorMessage: error.message, stack: error.stack ?? "" })
         .error();
--- a/src/logging/fileLogger.tsx
+++ b/src/logging/fileLogger.tsx
@@ -42,8 +42,8 @@ function wrapWinstonLogger(
  * @returns A {@link AsyncLogger} that writes to a rotating file via winston.
  */
 export function createFileLogger(config: FileLoggerConfig): AsyncLogger {
-  const maxSizeInMB = config.maxSizeInMB ?? 10;
-  const maxFileCount = config.maxFileCount ?? 5;
+  const maxSizeInMB = config.maxSizeInMB ?? 5;
+  const maxFileCount = config.maxFileCount ?? 10;
   const bindings = config.bindings ?? {};
 
   const transport = new DailyRotateFile({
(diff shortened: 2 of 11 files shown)


src/logging/fileLogger.tsx
```diff
@@ -9,54 +10,61 @@ export interface FileLoggerConfig {
...
  * Creates a logger that writes structured JSON to a rotating file.
  *
  * @param config - Logger configuration (file path, rotation limits, level).
- * @returns A {@link AsyncLogger} that writes to a rotating file via pino.
+ * @returns A {@link AsyncLogger} that writes to a rotating file via winston.
  */
 export function createFileLogger(config: FileLoggerConfig): AsyncLogger {
-  const maxSizeInMB = config.maxSizeInMB ?? 10;
-  const maxFileCount = config.maxFileCount ?? 5;
+  const maxSizeInMB = config.maxSizeInMB ?? 5;
```
Hweinstock, line 46: the new behavior is that it rotates the file each day or when the size cap it hit. Before we rotated only when the size limit was hit. This swap ensures we maintain roughly the past 10 days but still no more than 50MB.

AlexanderRichey (dismissed):

src/index.ts
```diff
@@ -50,7 +50,6 @@ process.exit(
       await rootHandler.route(argv);
     } catch (e) {
       const error = e instanceof Error ? e : new Error(String(e));
-      io.stderr.write(`${error.name}: ${error.message}\n`);
```
Hweinstock, line 58: moved to top-level to ensure that logger/io init failures are logged to console.

tejaskash (approved):

AlexanderRichey (approved):

PR #1809, opened 2026-07-22
feat(project): tree-based scaffolding for `agentcore project create`

tejaskash:
## Summary

Implements `agentcore project create` with a tree-based scaffolding subsystem. A project is modeled as a **file tree** (`dir`/`file` nodes) . The writer is unaware of the underlying runtime, but here's where files are actually read from: 

- **Node / dev / npm dist** → reads assets from `src/assets/` on disk
- **compiled standalone binary** → reads from `Bun.embeddedFiles`

`scripts/build.ts` embeds assets by *naming* (`agentcore-assets/[dir]/[name].[ext]`) via a file-loader plugin, so no per-file `with { type: "file" }` import attributes are needed. 

## Testing

- `bun run typecheck` — clean
- `oxlint` — clean
- `bun test` — 263 pass
- Verified end to end on both the compiled binary (embedded assets) and the node bundle (mirrored `dist/assets/`): `project create --project-name MyAgent` scaffolds `./MyAgent/`, and a re-run refuses to overwrite


src/assets/cdk/README.md
```diff
@@ -0,0 +1,26 @@
...
+- `npm run test` run unit tests
+- `npx cdk synth` emit the synthesized CloudFormation template
+- `npx cdk deploy` deploy this stack to your default AWS account/region
+- `npx cdk diff` compare deployed stack with current state
+
+## Usage
+
+You typically don't need to interact with this directory directly. The AgentCore CLI handles synthesis and deployment:
+
+```bash
```
Hweinstock, line 26: Should these commands include the new project prefix?

src/assets/cdk/lib/cdk-stack.ts
```diff
@@ -0,0 +1,249 @@
...
+      new AgentCoreMcp(this, 'Mcp', {
+        projectName: spec.name,
+        mcpSpec,
+        agentCoreApplication: this.application,
+        credentials,
+        projectTags: spec.tags,
+      });
+    }
+
+    // Create payment infrastructure via CFN constructs
```
Hweinstock, line 127: for my own understanding, what is special about payments that requires all this special logic?

src/assets/cdk/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "devDependencies": {
+    "@types/jest": "^29.5.14",
+    "@types/node": "^24.10.1",
+    "jest": "^29.7.0",
+    "ts-jest": "^29.2.5",
+    "aws-cdk": "2.1126.0",
+    "prettier": "^3.4.2",
+    "typescript": "~5.9.3"
+  },
+  "dependencies": {
```
Hweinstock, line 25: I think @jesseturner21 has a recent [PR](https://github.com/aws/agentcore-cli/pull/1777) that pinned these. Should we be building off that?

src/io/atomicWrite.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  await Promise.all(dirs.splice(0).map((d) => rm(d, { recursive: true, force: true })));
+});
+
+async function tempDir(): Promise<string> {
+  const d = await mkdtemp(join(tmpdir(), "atomic-write-"));
+  dirs.push(d);
+  return d;
+}
+
+test("writes contents and leaves no temp file", async () => {
```
Hweinstock, line 18: do we think direct tests on `atomicWrite` are warranted? I believe the pattern so far has been to test as e2e as possible, and avoiding tests that couple to the current implementation. I also see how having these is valuable because there are properties we want to verify here (ex. doesn't leave the temp file behind). Curious to others' thoughts.

src/assetManager/types.ts
```diff
@@ -0,0 +1,13 @@
+export const ASSET_NAMES = ["cdk"] as const;
+
+export type AssetName = (typeof ASSET_NAMES)[number];
+
+export type AssetVariables = Record<string, unknown>;
+export interface AssetFile {
+  relativePath: string;
+  text(): Promise<string>;
+}
+export interface EmbeddedFile {
```
Hweinstock: is it worth adding some docstrings here to clarify the difference between an `AssetFile` and `EmbeddedFile`? Its not obvious to me atm.

src/assetManager/manager.ts
```diff
@@ -0,0 +1,108 @@
+import Handlebars from "handlebars";
```
Hweinstock: should the file name start with lowercase? My understanding was that only react components have upper case file names.

src/assetManager/manager.ts
```diff
@@ -0,0 +1,108 @@
...
+      this.embeddedFiles.length > 0
+        ? this.listEmbeddedFiles(asset)
+        : await this.listFileSystemFiles(asset);
+
+    if (assetFiles.length === 0) {
+      throw new Error(`Asset '${asset}' does not exist`);
+    }
+
+    const resolvedDestination = resolve(destination);
+    for (const file of assetFiles) {
```
Hweinstock: should we treat the rendering of files as a transaction or is this up to the caller?

src/assetManager/manager.ts
```diff
@@ -0,0 +1,108 @@
...
+): string {
+  const bundledRoot = resolve(moduleDirectory, "assets");
+  if (existsSync(bundledRoot)) {
+    return bundledRoot;
+  }
+
+  return resolve(moduleDirectory, "../assets");
+}
+
+export class AssetManager {
```
Hweinstock: does it make sense to decouple template rendering from asset management? I'm wondering if we'll ever want to work with static assets that aren't part of our project templates.

src/assets/cdk/lib/cdk-stack.ts
tejaskash: @aidandaly24 can help here, I asked him to chime in

src/assets/cdk/README.md
tejaskash: I'll add a TODO in here to update it when we actually know the commands

src/assets/cdk/package.json
tejaskash: Fixed.

src/io/atomicWrite.test.ts
tejaskash: I think keeping it could be valuable for the same reasons you pointed out. I've left it in for now, but will let @AlexanderRichey comment on this too.

src/assetManager/types.ts
tejaskash: Done

src/assetManager/manager.ts
tejaskash: Done

src/assetManager/manager.ts
tejaskash: Created #1827 so this doesn't fall through. Will let the team comment on what we should do here.

src/assetManager/manager.ts
tejaskash: We haven't so far. We haven't needed to update the templates or the rendering system much since it was first created.

Hweinstock: fair point, its likely premature to pull these out as separate concepts.

src/assetManager/manager.ts
Hweinstock: could we get away with something simple, like removing the top-level folder when we hit an error?

src/assets/cdk/lib/cdk-stack.ts
aidandaly24: Since the generated project still consumes `@aws/agentcore-cdk`, could this payment orchestration live in that package instead of being emitted into every generated stack? The package already owns `AgentCoreApplication`, the manager and connector constructs, and the runtime constructs, so it has the context needed to create the payment resources, bind eligible runtimes, add IAM grants and environment variables, and emit the outputs. The generated stack could then pass `spec` and the resolved credential map to `AgentCoreApplication`, keeping payment behavior centralized in the versioned CDK package.

I think we could also model the default connector explicitly? With one connector we can infer it, but with multiple connectors we should require a `defaultConnector`. Using `payment.connectors[0]` makes configuration order affect runtime behavior.

src/assets/cdk/bin/cdk.ts
```diff
@@ -0,0 +1,192 @@
...
+async function main() {
+  // Config root is parent of cdk/ directory. The CLI sets process.cwd() to agentcore/cdk/.
+  const configRoot = path.resolve(process.cwd(), '..');
+  const configIO = new ConfigIO({ baseDir: configRoot });
+
+  const spec = await configIO.readProjectSpec();
+  const targets = await configIO.readAWSDeploymentTargets();
+
+  // The vended CDK project compiles against the published @aws/agentcore-cdk
+  // schema type, which may lag the CLI's own AgentCoreProjectSpec (e.g. payments,
```
Hweinstock: is this still true now that we're pinning the version? (likely OOS here, but curious)

Hweinstock (dismissed):

src/assetManager/manager.ts
```diff
@@ -0,0 +1,111 @@
...
+    private readonly embeddedFiles: readonly EmbeddedFile[] = [],
+    private readonly sourceRoot?: string,
+  ) {}
+
+  async render(
+    asset: AssetName,
+    destination: string,
+    variables: AssetVariables = {},
+  ): Promise<void> {
+    const assetFiles =
```
AlexanderRichey: So each "asset" has a list of files associated with it? Is asset the right word here? Do we really mean something like `projectScaffold`?

src/assetManager/manager.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+export class AssetManager {
+  private handlebars = Handlebars.create();
+  constructor(
+    private readonly embeddedFiles: readonly EmbeddedFile[] = [],
+    private readonly sourceRoot?: string,
+  ) {}
+
+  async render(
```
AlexanderRichey: Would a more specific name for this be `renderAndWriteProjectScaffold`? It's a bit of a mouthful, but it might communicate what it does more clearly. I had to read every line of this to know what it's supposed to do.

src/assetManager/manager.ts
```diff
@@ -0,0 +1,111 @@
...
+    const resolvedDestination = resolve(destination);
+    for (const file of assetFiles) {
+      const outputPath = this.resolveOutputPath(resolvedDestination, file.relativePath);
+      const rendered = this.handlebars.compile(await file.text(), {
+        noEscape: true,
+        strict: true,
+      })(variables);
+
+      await mkdir(dirname(outputPath), { recursive: true });
+      await atomicWrite(outputPath, rendered);
```
AlexanderRichey: Do we need to worry about writing atomically here? It seems that we just need to make sure that the output path doesn't already exist. Then we could simply call `writeFile`.

src/assetManager/types.ts
```diff
@@ -0,0 +1,26 @@
...
+  text(): Promise<string>;
+}
+
+/**
+ * A raw file embedded in the compiled binary, as exposed by `Bun.embeddedFiles`.
+ * `name` is the full build-time virtual path (e.g.
+ * `agentcore-assets/src/assets/cdk/bin/cdk.ts`), which the manager strips down
+ * to an `AssetFile.relativePath`.
+ */
+export interface EmbeddedFile {
```
AlexanderRichey: Should both of these be interfaces or types? They seem a little more like types to me in that they're concrete and not meant to be implemented by anything else.

src/assets/cdk/bin/cdk.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  // Read deployed state for credential ARNs (populated by pre-deploy identity setup)
+  let deployedState: Record<string, unknown> | undefined;
+  try {
+    deployedState = JSON.parse(fs.readFileSync(path.join(configRoot, '.cli', 'deployed-state.json'), 'utf8'));
+  } catch {
+    // Deployed state may not exist on first deploy
+  }
+
+  if (targets.length === 0) {
```
AlexanderRichey: Should we move this up to right after line 29. I think it'd be easier to follow if we fail up there since this check doesn't seem to be dependent on anything between line 29 and here.

src/assets/cdk/bin/cdk.ts
```diff
@@ -0,0 +1,192 @@
...
+  const targets = await configIO.readAWSDeploymentTargets();
+
+  // The vended CDK project compiles against the published @aws/agentcore-cdk
+  // schema type, which may lag the CLI's own AgentCoreProjectSpec (e.g. payments,
+  // harnesses, gateway fields). Cast once so those fields are reachable.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const specAny = spec as any;
+
+  // Extract MCP configuration from project spec.
+  // Gateway fields are stored in agentcore.json but may not yet be on the
```
AlexanderRichey: This sentence looks incomplete.

src/assets/cdk/bin/cdk.ts
```diff
@@ -0,0 +1,192 @@
...
+        // Full spec + dir drive the AWS::BedrockAgentCore::Harness CFN resource.
+        spec: harnessSpec,
+        harnessDir,
+      });
+    } catch (err) {
+      throw new Error(
+        `Could not read harness.json for "${entry.name}" at ${harnessPath}: ${err instanceof Error ? err.message : err}`
+      );
+    }
+  }
```
AlexanderRichey, line 102: I think the above would be easier to follow if we break things down into discrete steps.

```typescript
const mcpSpecs = await resolveMcpSpecs(...)
const connectorParametersByFile = await resolveConnectorParametersByFile(...)
const harnessConifgs = await resolveHarnessConifgs(...)
```

src/assets/cdk/bin/cdk.ts
```diff
@@ -0,0 +1,192 @@
...
+      | undefined;
+    const targetResources = targetState?.[target.name]?.resources as Record<string, unknown> | undefined;
+    const credentials = targetResources?.credentials as
+      | Record<string, { credentialProviderArn: string; clientSecretArn?: string }>
+      | undefined;
+
+    // Payment credential provider ARNs live in the same credentials map as identity credentials
+    const paymentCredentials = credentials;
+
+    const paymentSpec = specAny.payments?.length
```
AlexanderRichey, line 150: Is there a reason why this is nested under the above for-loop? It seems like this and the above stuff about `targetState` could be pulled up and put behind a helper function like the others.

```typescript
const paymentSpec = await resolvePaymentSpec(...)
```

src/assets/cdk/lib/cdk-stack.ts
```diff
@@ -0,0 +1,249 @@
...
+  return entrypointFile.endsWith('.py');
+}
+
+/**
+ * CDK Stack that deploys AgentCore infrastructure.
+ *
+ * This is a thin wrapper that instantiates L3 constructs.
+ * All resource logic and outputs are contained within the L3 constructs.
+ */
+export class AgentCoreStack extends Stack {
```
AlexanderRichey, line 93: Doesn't this belong in `@aws/agentcore-cdk`? It seems like we're making a lot of complexity the user's concern by putting all this in outputted files. It might also be a bit harder for us to control if we want to make our own changes to this since we'd have to overwrite changes the user may have made.

src/io/atomicWrite.test.ts
AlexanderRichey: These tests seem useful to me since there are specific behaviors `atomicWrite` is responsible for.

AlexanderRichey (commented):
I didn't see any blocking issues in this PR, but I have a number of questions whose answers I'd like to understand before approving.

src/assetManager/types.ts
tejaskash: This has been moved to a new implementation

src/assetManager/manager.ts
tejaskash: This has been moved to a new implementation

src/assetManager/manager.ts
tejaskash: This has been moved to a new implementation

src/assets/cdk/bin/cdk.ts
tejaskash: This has been moved to a new implementation

src/assetManager/manager.ts
tejaskash: This has been moved to a new implementation

src/assetManager/manager.ts
tejaskash: This has been moved to a new implementation

src/assetManager/manager.ts
tejaskash: This has been moved to a new implementation

src/assetManager/types.ts
tejaskash: This has been moved to a new implementation

src/io/atomicWrite.test.ts
tejaskash: Kept

src/assets/cdk/lib/cdk-stack.ts
tejaskash: I agree, this is just a simple copy of what we had in the Old CLI. Refactoring this will involve an L3 CDK PR and I think that is out of scope of this change. I am happy to work on a followup PR that addresses this.

src/assets/cdk/bin/cdk.ts
tejaskash: I agree, this is just a simple copy of what we had in the Old CLI. Refactoring this will involve an L3 CDK PR and I think that is out of scope of this change. I am happy to work on a followup PR that addresses this.

src/assets/cdk/lib/cdk-stack.ts
tejaskash: I agree, this is just a simple copy of what we had in the Old CLI. Refactoring this will involve an L3 CDK PR and I think that is out of scope of this change. I am happy to work on a followup PR that addresses this.

src/project/compose.ts
```diff
@@ -0,0 +1,75 @@
...
+  template: ProjectTemplate,
+  src: Source,
+): Promise<ProjectNode> {
+  const { appDir, assetDir } = TEMPLATES[template];
+  return dir(".", [
+    dir("agentcore", [
+      dir("cdk", await expandDir(src, "cdk")),
+      file("aws-targets.json", async () => json([])),
+    ]),
+    file("agentcore.json", async () => json(agentcoreSpec(name, template))),
```
aidandaly24: Is this supposed to be in the `agentcore/` directory? I tried creating a project and it creates it at the project root.

src/project/templates.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+export const TEMPLATES: Record<ProjectTemplate, Template> = {
+  [PROJECT_TEMPLATES.HELLO_WORLD_PYTHON]: {
+    appDir: "hello-world",
+    assetDir: "templates/hello-world-python",
+    spec: {
+      runtimes: [
+        {
+          name: "hello-world",
```
aidandaly24: nit: but this isn't a valid runtime name right?

src/handlers/project/types.ts
```diff
@@ -1,11 +1,24 @@
...
+
 /** Available project templates for scaffolding new AgentCore projects. */
 export const PROJECT_TEMPLATES = {
-  BAREBONES: "barebones",
+  HELLO_WORLD_PYTHON: "hello-world-python",
 } as const;
 
 export type ProjectTemplate = (typeof PROJECT_TEMPLATES)[keyof typeof PROJECT_TEMPLATES];
 
+export const ProjectNameSchema = z
```
aidandaly24, line 74: This validation differs from the pinned CDK `ProjectNameSchema`. For example, the CLI accepts `test`, but the project schema rejects it as a reserved name, so creation succeeds and the generated project later fails validation. I think we should reuse the shared schema or match its reserved-name validation here. Let me know what you think.

aidandaly24 (commented):
A few comments:

scripts/build.ts
```diff
@@ -0,0 +1,99 @@
...
+    process.exit(1);
+  }
+}
+
+async function bundle(): Promise<void> {
+  const result = await Bun.build({
+    entrypoints: [ENTRYPOINT],
+    outdir: DIST,
+    target: "node",
+    minify: true,
```
Hweinstock: do we need to tell bun to not minify identifiers? [By default I think it will shorten identifiers](https://bun.com/docs/bundler/minifier), which will make stack traces harder to follow, and potentially erase error names at runtime for telemetry.

scripts/build.ts
```diff
@@ -0,0 +1,99 @@
...
+  });
+  reportAndExit(result);
```
Hweinstock: nit: should we rename this to make it clear that it doesn't always exit? looking at the implementation it looks like it only exits if there is an error.

src/project/compose.test.ts
```diff
@@ -0,0 +1,59 @@
+import { afterEach, describe, expect, test } from "bun:test";
```
Hweinstock: what exactly differentiates `src/core/project` from `src/project`? Is this functionality used outside of core?

src/project/source.ts
```diff
@@ -0,0 +1,80 @@
+import { existsSync } from "node:fs";
+import { readFile, readdir } from "node:fs/promises";
+import { dirname, join, relative, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+
+/**
+ * Provides an interface for abstracting bun executable asset reading
+ * and Node asset reading
+ */
+export interface Source {
```
Hweinstock: Is this specific to Assets? If so, should we call it `AssetSource`?

src/project/source.ts
```diff
@@ -0,0 +1,80 @@
...
+  /** Returns a thunk that reads the text of the asset at `assetPath`. */
+  read(assetPath: string): () => Promise<string>;
```
Hweinstock: does it make more sense to inline the deferred function? When I see `source.read()`I expect to see data as the output, not another function.

src/project/tree.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (node.kind === "dir") {
+    await mkdir(path, { recursive: true });
+    for (const child of node.children) {
+      await writeTree(child, path);
+    }
+    return;
+  }
+
+  if (existsSync(path)) {
+    throw new Error(`Refusing to overwrite existing file: ${path}`);
```
Hweinstock: should we define a modeled error for this? maybe `ProjectAlreadyExistsError` or something?

src/core/project/tree.ts
```diff
@@ -0,0 +1,53 @@
+import { existsSync } from "node:fs";
```
Hweinstock, line 1: i really like this abstraction!

.gitignore
```diff
@@ -33,3 +33,5 @@ report.[0-9]_.[0-9]_.[0-9]_.[0-9]_.json
 
 # Finder (MacOS) folder config
 .DS_Store
+
+.agentreview
```
Hweinstock, line 37: out of curiosity, what is this from?  a claude code skill or something?

src/project/source.test.ts
```diff
@@ -0,0 +1,46 @@
+import { afterEach, describe, expect, test } from "bun:test";
```
Hweinstock: is it possible to test all of this through the manager? Wondering if that will make it easier to change the implementation of rendering, without losing guarentees on its behavior?

General question for all the tests in project. Do we still get solid coverage just by testing `create`?

scripts/build.ts
```diff
@@ -0,0 +1,99 @@
...
+    minify: true,
+    root: REPO_ROOT,
+    naming: { asset: ASSET_NAMING },
+    plugins: [assetLoaderPlugin()],
+  });
+  reportAndExit(result);
+  console.log(`Compiled ${target} → ${outfile} (${assets.length} assets embedded)`);
+}
+
+const [command, target] = process.argv.slice(2);
```
AlexanderRichey: Can we use the code we have in Runnable here and below?

.gitignore
tejaskash: Its a VS Code extension I created that I use to do a local code review of my diff before pushing PRs

src/assets/cdk/README.md
tejaskash: Addressed

src/assets/cdk/bin/cdk.ts
tejaskash: Addressed

src/assets/cdk/bin/cdk.ts
tejaskash: Addressed

src/assets/cdk/bin/cdk.ts
tejaskash: Addressed

src/project/compose.ts
tejaskash: Great find! Addressed

src/project/templates.ts
tejaskash: thanks for catching this, addressed

scripts/build.ts
tejaskash: Thanks for catching this, addressed

scripts/build.ts
tejaskash: Thanks for catching this, addressed

src/project/tree.ts
tejaskash: Addressed

src/project/source.ts
tejaskash: Addressed

src/project/source.ts
tejaskash: Addressed

src/project/compose.test.ts
tejaskash: Moved under src/core

src/handlers/project/types.ts
tejaskash: Thanks for catching this, addresed

src/core/project/source.ts
```diff
@@ -0,0 +1,76 @@
...
+// Prefix the build script gives every embedded asset name at compile time.
+const EMBEDDED_PREFIX = "agentcore-assets/src/assets/";
+
+// Embedded files carry a name property that Bun's types widen to Blob.
+type NamedBlob = Blob & { readonly name: string };
+
+const embeddedBlobs = () => Bun.embeddedFiles as readonly NamedBlob[];
+
+/** Reads assets embedded in the compiled standalone executable. */
+export const embeddedSource: AssetSource = {
```
Hweinstock: should we use a class for this to be consistent with existing implementations of interfaces in the codebase?

src/core/project/templates.ts
```diff
@@ -0,0 +1,37 @@
+import { PROJECT_TEMPLATES, type ProjectTemplate } from "../../handlers/project/types";
+
+interface TemplateSpec {
```
Hweinstock: should this be a type? (same with Template below) I think we're making an effort to use types for concrete data and interfaces for extendable functionality.

src/core/project/tree.ts
```diff
@@ -0,0 +1,62 @@
...
+});
+
+export const file = (name: string, bytes: () => Promise<string>): FileNode => ({
+  kind: "file",
+  name,
+  bytes,
+});
+
+/** Thrown when scaffolding would overwrite a file that already exists. */
+export class ProjectFileExistsError extends Error {
```
Hweinstock: should we extend the base class to get proper classification? 

https://github.com/aws/agentcore-cli/blob/950518515062f8a2ccc9e0f4218512b8dfd13f49/src/errors/errors.tsx#L14

src/io/atomicWrite.ts
```diff
@@ -0,0 +1,14 @@
+import { randomUUID } from "node:crypto";
```
Hweinstock, line 1: does this make sense in the existing io module? I think we already have some logic there for reading json files, and reading files from user input.

src/handlers/project/types.ts
```diff
@@ -1,11 +1,91 @@
...
+
 /** Available project templates for scaffolding new AgentCore projects. */
 export const PROJECT_TEMPLATES = {
-  BAREBONES: "barebones",
+  HELLO_WORLD_PYTHON: "hello-world-python",
 } as const;
 
 export type ProjectTemplate = (typeof PROJECT_TEMPLATES)[keyof typeof PROJECT_TEMPLATES];
 
+// A generated project's Python package is named after the
```
Hweinstock, line 10: maybe OOS since it requires some refactoring, but would a prefix on the python project name provide a stronger guarantee? My understanding is that if any template dependency adds a dependency (or transitive dependency) with the same name, it will stop working. 

I'm somewhat concerned this will drift from the actual dependencies, and we could run into issues.

src/core/project/source.ts
tejaskash: Addressed

src/core/project/templates.ts
tejaskash: Addressed

src/core/project/tree.ts
tejaskash: Addressed

src/io/atomicWrite.ts
tejaskash: Addressed

src/handlers/project/types.ts
tejaskash: Lets discuss

src/assets/templates/hello-world-python/pyproject.toml
```diff
@@ -0,0 +1,19 @@
...
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "hello-world"
+version = "0.1.0"
+description = "AgentCore Runtime application using the Strands SDK"
+readme = "README.md"
+requires-python = ">=3.10"
+dependencies = [
```
Hweinstock, line 11: in the real templates i think we should pin these.

src/core/project/source.ts
```diff
@@ -0,0 +1,82 @@
...
+export class EmbeddedAssetSource implements AssetSource {
+  private blobs(): readonly NamedBlob[] {
+    return Bun.embeddedFiles as readonly NamedBlob[];
+  }
+
+  public async read(assetPath: string): Promise<string> {
+    const name = `${EMBEDDED_PREFIX}${assetPath}`;
+    const blob = this.blobs().find((f) => f.name === name);
+    if (!blob) {
+      throw new Error(`Embedded asset not found: ${assetPath}`);
```
Hweinstock, line 33: should we throw a modeled error here?

Hweinstock (approved):
LGTM! one small error comment, but we can take that as a follow-up. I don't think its blocking 

I also found one small edge case when manually testing, that could be addressed as a followup. If I create a project within a project, it appears to lift the child up to the root. For example, 

```
bun run start project create --project-name root
cd root 
bun run start project create --project-name child
cd .. 
```

I see both root and child at the same level. I would have expected child inside root.

src/assets/templates/hello-world-python/pyproject.toml
tejaskash: Agreed!

src/core/project/source.ts
tejaskash: We should, I'll fold it into the fix for this: 
> I also found one small edge case when manually testing, that could be addressed as a followup. If I create a project within a project, it appears to lift the child up to the root. For example,

bun run start project create --project-name root
cd root 
bun run start project create --project-name child
cd .. 
I see both root and child at the same level. I would have expected child inside root.

notgitika (approved):

PR #1811, opened 2026-07-22
feat(identity): add identity scaffolding and CRUDL operations for api key

nborges-aws:
## Description

Adds identity scaffolding, as well as command-line CRUDL for Identity API key credential providers:
  - `identity api-key-credential-provider create`
  - `identity api-key-credential-provider get`
  - `identity api-key-credential-provider list`
  - `identity api-key-credential-provider update`
  - `identity api-key-credential-provider delete`

Follows the structure established by harness and runtime for commands. Create/update accept `--name` and `--api-key`, get/delete accept just `name`, list accepts `--max-results` and `--next-token`. Similar to runtime, identity will appear in the TUI command menu at the root, but prints help pending TUI routing. 

Additional change to move generic default help handler created for runtime to `src/handlers/help.tsx` to avoid duplication. This PR updates runtime call sites to point to new help handler. 

## Testing
Tested via fixture tests against live AWS resources, as well as manual dogfooding via running the build artifact in CLI.

## Related Issue

N/A

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Verified from current HEAD:

 - [x] bun test (274 pass, 0 fail)
- [x] bun test src/handlers/identity src/handlers/runtime src/core (39 passed, 0 failed)
- [x] bun run lint:check
- [x] bun run format:check (prettier --check)
- [x] bun run typecheck (clean in src/)
- [x] bun run build

Additional verification:

- [x] All five CRUDL commands verified against live AWS (us-west-2)
- [x] Manual dogfooding via built CLI artifact (bun dist/index.js identity api-key-credential-provider {create,get,list,update,delete})
- [x] Help output confirmed for bare identity and identity api-key-credential-provider groups

The committed identity fixtures were recorded with `RECORD=1` against a live account in us-west-2. The RECORD run creates two providers (agentcore-cli-identity-fixture, agentcore-cli-identity-fixture-2), exercises pagination, then deletes both. The fixture set contains no credential material (API key values are not returned by the service).

## Checklist

---

terms of your choice.


src/handlers/identity/__fixtures__/create-2.golden.json
```diff
@@ -0,0 +1,8 @@
+{
+  "apiKeySecretArn": {
+    "secretArn": "arn:aws:secretsmanager:us-west-2:314146320088:secret:bedrock-agentcore-identity!default/apikey/agentcore-cli-identity-fixture-2-7fd24230-KEL3mK"
```
aidandaly24: Could these be regenerated using the E2E account. 685197708687

src/handlers/identity/api-key-credential-provider/create/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+import type { Core } from "../../../types";
+import { coreOptsFromCtx } from "../../../utils";
+
+export const createCreateApiKeyCredentialProviderHandler = (core: Core) =>
+  createHandler({
+    name: "create",
+    description: "create an API key credential provider",
+    flags: [
+      flag("name", "the name of the API key credential provider", z.string().optional()),
+      flag("api-key", "the API key value", z.string().optional()),
```
aidandaly24, line 13: Adding --api-key puts the plaintext credential into the generic flags object. The root `withLogging` middleware currently logs all flags at DEBUG, and production file logging is always configured at DEBUG, so create/update will persist the key under `~/.agentcore/logs`.

We should probably mark this flag as sensitive and update `withLogging` to redact sensitive flag values before binding `{ flags, args }`. We could cover this generically in `withLogging.test.ts`, verifying that sensitive values are redacted while ordinary flags remain visible.

The same thing should effect update as well.

aidandaly24, line 13: In the doc I had wrote, the Identity contract supports managed keys through source-aware `--api-key` values (inline, `file://path`, or `-`) and external secrets through  a structured-reference flag containing the secret ID and JSON key. Could we support both mutually exclusive input forms here?

I sent you the doc line over slack. This could also be a follow up because I think this PR is well structured where it is now. Also tagging.

src/handlers/identity/identity.test.tsx
```diff
@@ -0,0 +1,216 @@
...
+import { join } from "node:path";
+import { CoreClient } from "../../core";
+import { createSilentLogger, fixtureFactories, matchGolden, testIO } from "../../testing";
+import { createRootHandler } from "../index";
+
+const REGION = "us-west-2";
+const FIXTURES = join(import.meta.dir, "__fixtures__");
+
+// Record with AWS_PROFILE=YOUR_PROFILE RECORD=1 bun test src/handlers/identity/identity.test.tsx
+// The test account must have two providers pre-created for pagination tests.
```
aidandaly24: I believe RECORD=1 creates both fixed provider names in the tests below, so pre-creating them would cause those create calls to fail with ConflictException. I think we should instead state that neither provider should exist before recording, or clean up stale providers before the create tests.

aidandaly24 (commented):
A few comments, but it looks good to me and matches what I did for Runtime well. One small thing also is we probably want to update the README.md.

src/handlers/identity/api-key-credential-provider/create/index.tsx
nborges-aws: Good call. I'll make the change to redact sensitive values in a separate PR to keep separation of concerns. I'll rebase the change into this PR after its merged and update the api-key flag.

nborges-aws: Thanks for the context! I'll do this in a follow-up PR

src/handlers/identity/identity.test.tsx
nborges-aws: Good call, never updated comment after I changed test to create both providers

src/handlers/identity/api-key-credential-provider/create/index.tsx
AlexanderRichey: Yeah, we'll have to accept the key via stdin so that it doesn't get into the user's shell history.

src/handlers/identity/api-key-credential-provider/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+import type { AppIO, Core } from "../../types";
+import { createHelpDefault } from "../../help";
+import { createCreateApiKeyCredentialProviderHandler } from "./create";
+import { createDeleteApiKeyCredentialProviderHandler } from "./delete";
+import { createGetApiKeyCredentialProviderHandler } from "./get";
+import { createListApiKeyCredentialProvidersHandler } from "./list";
+import { createUpdateApiKeyCredentialProviderHandler } from "./update";
+
+export function createApiKeyCredentialProviderHandler(core: Core, io: AppIO): Router {
+  return new Router("api-key-credential-provider", "manage API key credential providers")
```
AlexanderRichey, line 11: I wonder if we could shorten this to `api-key`. `api-key-credential-provider` is so looooong.

AlexanderRichey (approved):
Looks good! Let's merge this in but follow up with the fix to move the api key input to stdin right away.

src/handlers/identity/api-key-credential-provider/create/index.tsx
aidandaly24: Currently working on the Runtime invoke, and I have to do something similar with the keys for the bearer token. I am currently using this function which I have within the Runtime handlers `request.ts`, but it should eventually be a util we use:
```TypeScript
async function readSource(
  source: string,
  input?: NodeJS.ReadStream,
  signal?: AbortSignal,
): Promise<Uint8Array> {
  signal?.throwIfAborted();
  if (source.startsWith("file://")) {
    try {
      return await readFile(source.slice("file://".length), { signal });
    } catch (error) {
      if ((error as Error)?.name === "AbortError") throw error;
      throw new UsageError("Unable to read request source file");
    }
  }
  if (source !== "-") return new TextEncoder().encode(source);
  if (!input) throw new UsageError("stdin is not available for this request source");

  const stream = signal ? addAbortSignal(signal, input) : input;
  return buffer(stream);
}
```

and it's used like so:
```TypeScript
export async function resolveRuntimeInvokeSources(
  sources: { payload: string; bearerToken?: string },
  stdin?: NodeJS.ReadStream,
  signal?: AbortSignal,
): Promise<{ payload: Uint8Array; bearerToken?: string }> {
  if (sources.payload === "-" && sources.bearerToken === "-") {
    throw new UsageError("Payload and bearer token cannot both read from stdin");
  }

  const payload = await readSource(sources.payload, stdin, signal);
  if (sources.bearerToken === undefined) return { payload };
  const token = await readSource(sources.bearerToken, stdin, signal);
  return { payload, bearerToken: new TextDecoder().decode(token) };
}
```

aidandaly24 (approved):

PR #1812, opened 2026-07-22
ci: emit canary heartbeat metric for external failure alarm

jesseturner21:
## Summary

Adds a heartbeat step to the canary workflow: every run emits a `Success = 1|0` metric to CloudWatch (`AgentCoreCLI/Canary` namespace) via `if: always()`. A CloudWatch alarm in the canary's AWS account watches the metric and alerts the team when the canary fails — or stops running — for 30+ minutes.

### Why a heartbeat instead of a failure-triggered push

A metric pushed only on failure goes silent in exactly the scenarios we most need to know about: runner never starts, broken credentials, silently disabled cron. The alarm treats **missing data as breaching**, so with a heartbeat, silence is itself the signal.

### Notes for reviewers

- The alarm and its alerting action live in the canary's AWS account (already deployed and verified end-to-end); this PR only adds the metric emission.
- Single-run flakes keep going through the existing `Canary Failure` GitHub issue + Slack path, which carries the per-failure detail (variant, commit, run URL). The alarm only fires after 30+ minutes of failure or silence.
- The alarm period is coupled to the `*/15` cron — there's a comment at the cron line; if the schedule changes, the alarm period must change with it.
- The heartbeat step needs `cloudwatch:PutMetricData` on the workflow's AWS role (already granted, scoped to the `AgentCoreCLI/Canary` namespace).

### DIFF
--- a/.github/scripts/emit-canary-heartbeat.sh
+++ b/.github/scripts/emit-canary-heartbeat.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+# Emits a per-run heartbeat metric for the canary workflow. A CloudWatch
+# alarm in the AWS account the canary deploys to watches this metric and
+# alerts the team when the canary fails or stops reporting.
+#
+# Called from canary.yml with `if: always()`: every matrix cell reports
+# Success=1 or Success=0 on every run. The alarm treats missing data as
+# breaching, so silence (dead runner, broken credentials, disabled cron)
+# trips it just like an explicit failure does — which is why this step
+# must never be made conditional on failure only.
+#
+# Usage: emit-canary-heartbeat.sh <job-status>
+#   job-status: GitHub's ${{ job.status }} — "success", "failure", or "cancelled"
+set -euo pipefail
+
+# The alarm lives in us-east-1. Always emit there — even when the canary is
+# manually dispatched against another region — so an off-region run can't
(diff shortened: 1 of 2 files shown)


notgitika (approved):
Smart approach! LGTM :)

.github/scripts/emit-canary-heartbeat.sh
```diff
@@ -0,0 +1,39 @@
...
+#
+# Usage: emit-canary-heartbeat.sh <job-status>
+#   job-status: GitHub's ${{ job.status }} — "success", "failure", or "cancelled"
+set -euo pipefail
+
+# The alarm lives in us-east-1. Always emit there — even when the canary is
+# manually dispatched against another region — so an off-region run can't
+# starve the alarm into a false missing-data breach.
+readonly MONITOR_REGION=us-east-1
+readonly NAMESPACE=AgentCoreCLI/Canary
```
notgitika, line 20: qq can we make this an env variable instead? this would prevent anyone configuring it themselves to have this value that could mess with our alarms

jesseturner21: That is true but in the current design anyone could change the alarm by directly changing the .sh file where these envs would be imported. So there would be little benefit by importing the env vars.

PR #1818, opened 2026-07-23
fix(dev): support multiple A2A runtime ports

### LINKED ISSUE #1657: [Feature Request] Allow for multiple A2A agents to run simultaneously with agentcore dev
### Description

Currently when you have multiple A2A agents set in agentcore.json when you run agentcore dev it fails with `Bind for :::9000 failed: port is already allocated`.

This request is to remove the hardcoded port 9000 and potentially allow the port to be specified directly in the agentcore.json

### Acceptance Criteria

- When multiple A2A agents are set in agentcore.json there should be no port conflicts
- Optionally allow for a use specified port to be set in agentcore.json

### Additional Context

_No response_

notgitika:
## Summary

- Allocate A2A development ports from `9000 + runtime index`, while honoring an explicit `--port` for the selected runtime.
- Apply the same port resolution across direct logs mode, terminal UI, web UI, and prompt invocation.
- Map each container's unique host port to its internal A2A port and advertise the reachable host URL in the agent card.
- Pass `PORT` and `AGENTCORE_RUNTIME_URL` to CodeZip A2A processes.
- Keep MCP on its framework-required fixed port and retain existing explicit-port conflict behavior.

Closes #1657.

## SDK dependency

Container A2A runtimes work independently because each container still binds internally to `9000`. CodeZip A2A runtimes require aws/bedrock-agentcore-sdk-python#593 so `serve_a2a` honors the CLI-provided `PORT`; that PR is stacked on the A2A v1 migration in aws/bedrock-agentcore-sdk-python#591.

Agent Inspector requires no change: it routes start, invocation, and agent-card requests through the CLI backend by agent name, and the backend stores each runtime's resolved port.

## Testing

- `npm run build`
- `npm test` (415 files, 5,980 tests passed)
- `npm run typecheck`
- `npm run lint -- --no-cache` (no errors; existing warnings only)
- Focused dev-port suite (74 tests passed)
- Prettier and pre-commit checks

### DIFF
--- a/src/cli/commands/dev/command.tsx
+++ b/src/cli/commands/dev/command.tsx
@@ -16,8 +16,8 @@ import {
   callMcpTool,
   createDevServer,
   findAvailablePort,
-  getAgentPort,
   getDevConfig,
+  getDevPort,
   getDevSupportedAgents,
   getEndpointUrl,
   invokeAgent,
@@ -265,7 +265,6 @@ export const registerDev = (program: Command) => {
               let invokePort = port;
               let targetAgent = invokeProject?.runtimes[0];
               if (opts.runtime && invokeProject) {
-                invokePort = getAgentPort(invokeProject, opts.runtime, port, portExplicit);
                 targetAgent = invokeProject.runtimes.find(a => a.name === opts.runtime);
               } else if (invokeProject && invokeProject.runtimes.length > 1 && !opts.runtime) {
                 const names = invokeProject.runtimes.map(a => a.name).join(', ');
@@ -275,12 +274,12 @@ export const registerDev = (program: Command) => {
               }
 
               const protocol = targetAgent?.protocol ?? 'HTTP';
+              if (targetAgent && invokeProject) {
+                invokePort = getDevPort(invokeProject, targetAgent.name, protocol, port, portExplicit);
+              }
               recorder.set({
                 agent_protocol: standardize(AgentProtocol, protocol.toLowerCase()),
               });
-
-              if (protocol === 'A2A') invokePort = 9000;
-              else if (protocol === 'MCP') invokePort = 8000;
 
               if (protocol === 'MCP') {
                 await handleMcpInvoke(invokePort, invokePrompt, opts.tool, opts.input, headers);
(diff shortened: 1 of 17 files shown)


src/cli/commands/dev/command.tsx
```diff
@@ -405,31 +404,23 @@ export const registerDev = (program: Command) => {
...
+                  `Runtime "${config.agentName}" is at index ${idx}; using port ${targetPort} (pass --port ${targetPort} to override).`
                 );
               }
-              const actualPort = await findAvailablePort(fixedPort);
-              if ((isA2A || isMcp) && actualPort !== fixedPort) {
-                throw new ValidationError(
-                  `Port ${fixedPort} is in use. ${config.protocol} agents require port ${fixedPort}.`
-                );
+              const actualPort = await findAvailablePort(targetPort);
+              if (isMcp && actualPort !== targetPort) {
```
jariy17: Now, A2A startup could shift from 9000 to 9001. A later agentcore dev --runtime <name> "<prompt>" recomputes the original port and still targets 9000. That can fail or send the prompt to the wrong local process. Can we still fail fast?

src/cli/operations/dev/web-ui/handlers/start.ts
```diff
@@ -90,12 +92,12 @@ export function resolveAgentTargetPort(args: {
...
 }): number {
   const { protocol, agentName, agentIndex, uiPort, agentBasePort, selectedAgent } = args;
-  if (protocol === 'A2A') return 9000;
-  if (protocol === 'MCP') return 8000;
   const safeIndex = agentIndex >= 0 ? agentIndex : 0;
+  if (protocol === 'MCP') return MCP_DEFAULT_PORT;
   if (agentBasePort !== undefined && agentName === selectedAgent) {
     return agentBasePort;
   }
+  if (protocol === 'A2A') return A2A_DEFAULT_PORT + safeIndex;
```
jariy17, line 101: with one A2A runtime, agentcore dev --port 9100 falls through here and uses 9000. browser mode only sets selectedAgent when --runtime is provided, even though the only runtime is implicitly selected.

can we default selectedAgent to the sole supported runtime?
```
  selectedAgent:
    agentName ?? (supportedAgents.length === 1 ? supportedAgents[0]?.name : undefined)
```

src/cli/operations/dev/web-ui/handlers/start.ts
```diff
@@ -120,14 +122,11 @@ async function doStartAgent(
   const agentIndex = ctx.options.agents.findIndex(a => a.name === agentName);
```
jariy17: ctx.options.agents contains only dev-supported agents, but `getDevPort()` calculates the offset from the full project.runtimes list. if an unsupported runtime appears before an A2A runtime, browser mode assigns 9000 while terminal mode assigns 9001.

Can we carry the original runtime index into AgentInfo or reuse the shared port resolver here? we should also add a test with an unsupported runtime before the A2A runtime.

jariy17 (changes requested):
Pretty good just need some changes.

src/cli/commands/dev/command.tsx
notgitika: Addressed in ad872c21. A2A and MCP now share `requiresExactDevPort()`, and browser, TUI, and `--logs` startup paths all fail fast when the computed port is occupied, so invocation cannot recompute a different port.

src/cli/operations/dev/web-ui/handlers/start.ts
notgitika: Addressed in ad872c21. Browser mode now implicitly selects the sole supported runtime when `--runtime` is omitted, so an explicit `--port` applies to it. Added a regression test covering sole-runtime selection.

src/cli/operations/dev/web-ui/handlers/start.ts
notgitika: Addressed in ad872c21. `AgentInfo` now carries the original project runtime index through initial load and reload; browser startup uses it with a compatibility fallback. Added a regression test with an unsupported runtime before the A2A runtime, verifying port 9001.

jariy17 (approved):

PR #1819, opened 2026-07-23
fix: source aware api-key flag + redact api key value from logging

nborges-aws:
## Description

Expand api-key-credential-provider create/update operations with full source-aware flag surface. 

Changes:
- Source-aware `--api-key`: accepts inline values, file://path, or stdin so credentials never appear in shell history
- External secrets: `--api-key-secret-arn` + `--api-key-secret-json-key` for existing Secrets Manager entries
- Mutual exclusivity: `--api-key` and the pair of external secret cannot be mixed; the secret pair requires both values
- Tags: `--tags` on create (JSON object)
- Sensitive flag: `--api-key` marked sensitive so values are redacted from debug logs
- readSource / readSourceText utility for source-aware flag values. Follows @aidandaly24 pattern for runtime invoke, but moved to shared location and adds convenience method for decoding to text value. 
- added unit tests for source utilities, and mutual exclusivity tests for source-aware create/update identity
- updated fixture tests after latest changes

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.


src/handlers/identity/api-key-credential-provider/create/index.tsx
```diff
@@ -1,33 +1,74 @@
...
+      flag("api-key", "the API key value (inline, file://path, or -)", z.string().optional(), {
+        sensitive: true,
+      }),
+      flag("api-key-secret-arn", "existing Secrets Manager secret ARN", z.string().optional()),
+      flag(
+        "api-key-secret-json-key",
+        "JSON key containing the API key in the secret",
+        z.string().optional(),
+      ),
+      flag("tags", "tags as key=value (repeatable) or JSON object", z.string().optional()),
```
aidandaly24: This advertises repeatable `key=value` tags or a JSON object, but `z.string()` plus `parseJsonFlag` only accepts one JSON value. For example, `--tags env=prod` currently fails as invalid JSON. Could we implement and test both documented forms, including validating that the JSON form is a string map?

src/handlers/identity/api-key-credential-provider/create/index.tsx
```diff
@@ -1,33 +1,74 @@
...
   createHandler({
     name: "create",
     description: "create an API key credential provider",
     flags: [
       flag("name", "the name of the API key credential provider", z.string().optional()),
-      flag("api-key", "the API key value", z.string().optional()),
+      flag("api-key", "the API key value (inline, file://path, or -)", z.string().optional(), {
+        sensitive: true,
+      }),
+      flag("api-key-secret-arn", "existing Secrets Manager secret ARN", z.string().optional()),
```
aidandaly24: In the doc discussion, the preferred external-secret representation was one structured-reference flag rather than separate ARN and JSON-key flags. Could we use something like `--api-key-secret-reference '{"secretId":"...","jsonKey":"..."}'` for create/update, mutually exclusive with `--api-key`? We should also have a successful external-reference request-mapping test, I think the current tests only cover invalid combinations.

Update should first retrieve the provider and preserve its existing apiKeySecretSource. A MANAGED provider should accept only --api-key, while an EXTERNAL provider should accept only the structured secret reference. I think we should reject mismatched input locally and test both directions instead of deriving the source mode from the supplied flags.

src/handlers/source.tsx
```diff
@@ -0,0 +1,46 @@
+import { readFile } from "node:fs/promises";
+import { buffer } from "node:stream/consumers";
+import { addAbortSignal } from "node:stream";
+
+// readSource resolves a source-aware flag value to its raw bytes.
+// Supports three forms:
+//   - "file://path" reads the file at path
+//   - "-" reads stdin to EOF
+//   - anything else is returned as-is (inline value encoded as UTF-8)
+export async function readSource(
```
aidandaly24: I already implemented a version of source-aware input on another branch and I think it has some features we can use. Can we use that implementation (pasted below) as the shared behavior? The shared reader should remain byte-oriented so Runtime payloads are not decoded, while text-only credentials use strict UTF-8 decoding:

```ts
import { readFile } from "node:fs/promises";
import { addAbortSignal } from "node:stream";
import { buffer } from "node:stream/consumers";

export async function readSource(
  source: string,
  stdin?: NodeJS.ReadStream,
  signal?: AbortSignal,
): Promise<Uint8Array> {
  signal?.throwIfAborted();

  if (source.startsWith("file://")) {
    const path = source.slice("file://".length);
    try {
      return await readFile(path, { signal });
    } catch (error) {
      if (error instanceof Error && error.name === "AbortError") throw error;
      throw new TypeError(`unable to read source file: ${path}`);
    }
  }

  if (source !== "-") {
    return new TextEncoder().encode(source);
  }
  if (!stdin) {
    throw new TypeError("stdin is not available for this source");
  }

  return buffer(signal ? addAbortSignal(signal, stdin) : stdin);
}

export async function readSourceText(
  source: string,
  stdin?: NodeJS.ReadStream,
  signal?: AbortSignal,
): Promise<string> {
  const bytes = await readSource(source, stdin, signal);
  try {
    return new TextDecoder("utf-8", { fatal: true }).decode(bytes);
  } catch {
    throw new TypeError("source must contain valid UTF-8");
  }
}
```

Runtime can import these functions while retaining its request-specific dual-stdin validation. Identity can use `readSourceText` for `--api-key`. What do you think? I can also make the change in my Runtime invoke PR so not a problem really just a suggestion.

src/handlers/identity/types.tsx
```diff
@@ -3,14 +3,29 @@ import type {
   DeleteApiKeyCredentialProviderResponse,
   GetApiKeyCredentialProviderResponse,
   ListApiKeyCredentialProvidersResponse,
+  SecretSourceType,
   UpdateApiKeyCredentialProviderResponse,
 } from "@aws-sdk/client-bedrock-agentcore-control";
 import type { CoreOptions } from "../../core/types";
 
+export interface CreateApiKeyCredentialProviderInput {
```
AlexanderRichey: I think these should be types and not interfaces. Types hold values whereas interfaces are more abstract and don't have a concrete existence on their own—they need to be implemented by something else.

Would it make sense to import the types directly from the AWS SDK? That's what we did when defining the harness and runtime types. This will make the code a bit easier to maintain over time, I think.

src/handlers/source.tsx
AlexanderRichey: Want to merge in @nborges-aws code first and then add yours on top @aidandaly24 ? The signatures are identical, so this shouldn't cause any conflicts.

AlexanderRichey (changes requested):
Looks good! Only one blocking comment about using types over interfaces as input objects in Core, in the given context.

aidandaly24 (dismissed):
Thanks for making the update

src/handlers/source.tsx
aidandaly24: Yup that works for me

src/handlers/identity/api-key-credential-provider/create/index.tsx
```diff
@@ -1,33 +1,60 @@
...
 
-export const createCreateApiKeyCredentialProviderHandler = (core: Core) =>
+export const createCreateApiKeyCredentialProviderHandler = (core: Core, io: AppIO) =>
   createHandler({
     name: "create",
     description: "create an API key credential provider",
     flags: [
       flag("name", "the name of the API key credential provider", z.string().optional()),
-      flag("api-key", "the API key value", z.string().optional()),
+      flag("api-key", "the API key value (inline, file://path, or -)", z.string().optional(), {
```
AlexanderRichey, line 16: I don't think we should allow inline input here. That would be a security issue. You'd be able to retrieve secrets by running `history`.

We probably want to include some more detailed help text showing how to use this command safely with stdin. E.g., with the secret in the clipboard:

```sh
pbpaste | agentcore identity api-key create --name myKey --api-key -
```

or

```sh
cat mySecret.txt | agentcore identity api-key create --name myKey --api-key -
```

src/handlers/source.tsx
AlexanderRichey: As discussed in the office, @aidandaly24 will merge in the new `SourceResolver` and then @nborges-aws will rebase against those changes.

src/handlers/identity/api-key-credential-provider/create/index.tsx
nborges-aws: discussed offline. Decided we will allow inline if user chooses

aidandaly24 (approved):
Thanks!

AlexanderRichey (approved):

PR #1820, opened 2026-07-23
feat(runtime): add Runtime invoke command

aidandaly24:
## Description

Adds headless invocation for AgentCore Runtimes through `agentcore runtime invoke`.

IAM requests use the AgentCore data-plane SDK. CUSTOM_JWT requests use the generated Runtime data endpoint with an explicit bearer token. Payloads and bearer tokens can come from inline values, files, or stdin, with validation preventing both from consuming stdin at once.

The command maps the public Runtime invocation contract, including endpoint qualifiers, content negotiation, Runtime and MCP session IDs, ordered application headers, MCP metadata, and tracing headers. Application headers are validated against the Runtime allowlist and cannot override reserved protocol headers.

Raw output streams exact response bytes to stdout and writes response metadata to stderr. Binary responses can be written directly to a file, while `--json` emits one buffered response envelope with UTF-8 or base64 body encoding. Partial responses retain bytes already received and report whether the stream completed.

`SIGINT` propagates cancellation through Runtime lookup, transport, and response output. Usage failures return exit code 2, interruptions return 130, and transport or stream failures avoid exposing payloads, tokens, headers, or arbitrary upstream error causes. Invoke payloads, application headers, and bearer tokens are marked sensitive for command logging.

This change also adds shared abortable-stream handling used by Runtime and Harness, extends the Core Runtime client with IAM and CUSTOM_JWT invocation transport, and documents IAM, CUSTOM_JWT, MCP, binary, and JSON workflows.

The interactive Runtime invoke console has moved to the stacked `feat/runtime-invoke-tui` branch and is not part of this PR. Runtime creation and mutation, deployment or project resolution, cross-account invocation, Runtime ARNs, version targeting, custom request paths, and the Runtime `/commands` route also remain outside this change.

## Related Issue

N/A

## Documentation PR

N/A. README command, invocation, output, and MCP documentation is included in this PR.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Testing

Verified on current PR head `7a3520d8`:

- [x] `bun test` (422 passed, 0 failed)
- [x] `bun run typecheck`
- [x] `bun run lint:check`
- [x] `bun run format:check`
- [x] `bun run build`
- [x] `bun audit`
- [x] `agentcore runtime invoke --help`
- [x] Missing `--payload` exits with usage status 2
- [x] GitHub builds and unit tests on Linux, Windows, and macOS
- [x] Live IAM invocation against a newly deployed public HTTP Runtime streamed incremental SSE chunks, reported `complete=true`, and exited with status 0

## Checklist

---


src/components/RuntimeEndpointPicker.tsx
```diff
@@ -27,6 +27,7 @@ export interface RuntimeEndpointPickerProps extends ScreenProps {
   breadcrumb: string[];
   description?: string;
   onSelect: (qualifier: string) => void;
+  onEscape?: () => void;
```
Hweinstock: do we want to allow this to be overridden? My understanding is the `esc` consistently means go back to the last page right now in the tui, and by allowing an override we open the door to potentially unexpected behavior.

src/core/runtime.tsx
```diff
@@ -10,17 +10,164 @@ import {
...
+} from "../handlers/runtime/types";
+import type { AwsClients, CoreFetch, CoreOptions } from "./types";
+import { abortable } from "./abortable";
 import { toClientConfig } from "./utils";
 
+async function* emptyBody(): AsyncGenerator<Uint8Array> {}
+
 export class RuntimeClient implements CoreRuntimeClient {
-  constructor(private readonly clients: AwsClients) {}
+  constructor(
```
Hweinstock, line 28: should we be injecting the logger here? I see a few places below that might benefit from some detailed debug logs

src/core/runtime.tsx
```diff
@@ -10,17 +10,164 @@ import {
...
+      const endpoint = client.config.endpointProvider({
+        Region: options.region,
+        Endpoint: options.endpointUrl,
+      });
+      const url = new URL(endpoint.url);
+      if (url.protocol !== "https:") {
+        throw new TypeError("CUSTOM_JWT requires an HTTPS endpoint");
+      }
+      url.pathname = `${url.pathname.replace(/\/?$/, "/")}runtimes/${encodeURIComponent(runtimeId)}/invocations`;
+      url.search = new URLSearchParams({
```
Hweinstock: does it makes sense to expand the responsibility of `CoreFetch` to be a generic HttpClient where we can move some of this logic? I imagine we'll hit other cases where we want to do something similar to here.

src/handlers/runtime/invoke/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+          throw new UsageError("required option '--payload <payload>' not specified");
+        }
+        let path = `${ctx.require(PathKey)}/${encodeURIComponent(flags.id)}`;
+        if (flags.qualifier !== undefined) {
+          path += `/${encodeURIComponent(flags.qualifier)}`;
+        }
+        try {
+          await renderTuiAt(path, ctx, core, io);
+        } catch (error) {
+          throw error instanceof TypeError &&
```
Hweinstock: is it worth defining an explicit error for this case? wondering if depending on the message might be brittle.

src/handlers/runtime/invoke/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+            : error;
+        }
+        return;
+      }
+
+      const jsonOutput = ctx.require(JsonKey);
+      if (jsonOutput && flags["output-file"] !== undefined) {
+        throw new UsageError("--json cannot be used with --output-file");
+      }
+      if (flags["output-file"] === "") {
```
Hweinstock: could this be enforced in the schema?

src/handlers/runtime/invoke/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+      }),
+      flag("mcp-session-id", "the MCP session ID", z.string().optional()),
+      flag("mcp-protocol-version", "the MCP protocol version", z.string().optional()),
+      flag("mcp-method", "the MCP method", z.string().optional()),
+      flag("mcp-name", "the MCP tool, resource, or prompt name", z.string().optional()),
+      flag("trace-id", "the X-Ray trace ID", z.string().optional()),
+      flag("trace-parent", "the W3C trace parent", z.string().optional()),
+      flag("trace-state", "the W3C trace state", z.string().optional()),
+      flag("baggage", "the W3C baggage", z.string().optional()),
+      flag("output-file", "the response output file", z.string().optional()),
```
Hweinstock: q: whats the difference between this and manually piping `agentcore runtime invoke ... >> output.txt`?

src/handlers/runtime/invoke/request.ts
```diff
@@ -0,0 +1,165 @@
+import { readFile } from "node:fs/promises";
```
Hweinstock: should this request/response live outside of handlers? I kind of thought the original idea is that handlers are lightweight and mirror the command structure, with functionality mostly coming from core or injected through context. 

@AlexanderRichey curious to your thoughts here.

src/handlers/runtime/invoke/request.ts
```diff
@@ -0,0 +1,165 @@
...
+    const value = header.slice(separator + 1).trim();
+    try {
+      validateHeaderName(name);
+    } catch {
+      throw new UsageError(`Invalid HTTP header name: ${name}`);
+    }
+    try {
+      validateHeaderValue(name, value);
+    } catch {
+      throw new UsageError(`Invalid header value for ${name}`);
```
Hweinstock: is there a way we can give a hint to the user about why its invalid? Also, I think wiring up logging here could be valuable as well.

src/runnable/index.tsx
```diff
@@ -27,9 +31,19 @@ export async function runWithExitCode(
   try {
     await fn(argv);
     return ExitCode.SUCCESS;
-  } catch (e) {
-    const error = e instanceof Error ? e : new Error(String(e));
-    console.error(`${error.name}: ${error.message}`);
+  } catch (error) {
```
Hweinstock: Would it be simpler to define our own error base class that carries the exit code directly, and then leverage it here?

src/runnable/index.tsx
```diff
@@ -27,9 +31,19 @@ export async function runWithExitCode(
...
+    if (error instanceof CommanderError) {
```
AlexanderRichey, line 60: I see the need for something that controls the exit code, but I'm not sure this is how we ought to do it. I think we should define something like `AgentCoreError` that is a custom `Error` type that has an optional `exitCode` field. This should probably come in a separate PR.

src/router/handler.tsx
```diff
@@ -7,6 +7,7 @@ import type { Context, ContextKey } from "./context";
 // Flag<string, unknown>.
 export interface Flag<N extends string = string, T = unknown> {
   name: N;
+  short?: string;
```
AlexanderRichey: I think we should hold off on adding this feature for now. This will end up adding a lot more for us to maintain.

src/handlers/runtime/types.tsx
```diff
@@ -7,8 +7,51 @@ import type {
 } from "@aws-sdk/client-bedrock-agentcore-control";
 import type { CoreOptions } from "../../core/types";
 
+export interface RuntimeInvokeRequest {
```
AlexanderRichey: These should be types and not interfaces. Types hold data whereas interfaces are abstract, don't hold data themselves, and have to be implemented by something else.

src/runnable/index.tsx
```diff
@@ -27,9 +31,19 @@ export async function runWithExitCode(
   try {
     await fn(argv);
     return ExitCode.SUCCESS;
-  } catch (e) {
-    const error = e instanceof Error ? e : new Error(String(e));
-    console.error(`${error.name}: ${error.message}`);
```
Hweinstock: I think we want to keep this `console.error`, it ensures that any initialization errors are logged to the user.

src/router/router.test.ts
```diff
@@ -264,7 +264,40 @@ test("applies a schema default for an omitted flag", async () => {
   expect(seen).toEqual({ count: 7 });
 });
 
-test("reports invalid input via command.error (throws under exitOverride)", async () => {
+test("a short alias preserves encounter order for repeated variadic values", async () => {
```
Hweinstock: what exactly are we verifying here?

src/handlers/runtime/invoke/screen.tsx
```diff
@@ -0,0 +1,382 @@
...
+          byteCount += snapshot.byteLength;
+          responseText += decoder.decode(snapshot, { stream: true });
+          updateExchange({
+            response: responseText,
+            byteCount,
+          });
+        }
+        responseText += decoder.decode();
+        updateExchange({ response: responseText });
+        let text: string | undefined;
```
AlexanderRichey: There's a whole lot going on here. I tend to think there's a way to simplify all this. Let's find some time next week to discuss.

src/core/runtime.tsx
aidandaly24: I'll add some logging here, agreed.

src/core/runtime.tsx
aidandaly24: I think that would be a good use case for CoreFetch. I agree that this logic could be moved over I'll scope that out.

AlexanderRichey (dismissed):
I didn't have time to read the whole thing, but left some initial feedback in the comments.

src/components/RuntimeEndpointPicker.tsx
aidandaly24: Good point. I agree callers should not override the Escape key itself. Would you support replacing this with a required semantic `onBack` callback?

```tsx
interface RuntimeEndpointPickerProps extends ScreenProps {
  // ...
  onBack: () => void;
}
```

`PaginatedTablePicker` would remain the only component binding Escape, and Escape would always invoke `onBack`. Route-backed screens would navigate to their explicit parent, while embedded target pickers would close their local picker state.

I'd apply this consistently to `RuntimePicker`, `RuntimeEndpointPicker`, and `HarnessEndpointPicker`. I don't think always using `navigate(-1)` is sufficient because direct TUI deep links and embedded pickers may not have a preceding history entry.

src/core/runtime.tsx
aidandaly24: I scoped this out, and most of the surrounding behavior appears Runtime-specific; `CoreFetch` itself is already the generic injectable transport. Were you thinking the shared client would own only fetch/error handling, or also the Runtime URL, headers, and response mapping?

I'm hesitant to move the latter into a generic client, but I'm open to extracting the CUSTOM_JWT path into a Runtime-specific transport if the goal is to simplify `RuntimeClient`.

I'd also be happy to introduce a generic HTTP client once there is another consumer of the same behavior. Gateway invoke may become one, for example. Having that second implementation would give us a concrete shared contract rather than guessing now about which authentication, request, and error-handling policies are actually common.

src/handlers/runtime/invoke/index.tsx
aidandaly24: In a shell, `> output.bin` is effectively equivalent. `>>` differs because it appends. The intent behind `--output-file` was to provide an explicit binary-response destination for callers that invoke the CLI without a shell, while keeping metadata on stderr. The same file-streaming path is also needed by the TUI.

That said, the flag is convenience rather than a unique capability. Would you prefer that we remove it from the headless command and document stdout redirection instead? The TUI would retain its File response destination.

src/handlers/runtime/invoke/request.ts
aidandaly24: I've been treating the handler directory as the command-owned application layer, rather than limiting it to the route entrypoint itself. `request.ts` handles CLI sources, validation, and defaults, while `response.ts` handles stdout/stderr, files, TTY behavior, and JSON presentation. `RuntimeClient` owns transport.

This matches Harness, where request/response transformation lives under `handlers/harness/invoke`, while `core/harness.tsx` only owns SDK transport and abort wrapping.

Moving these concerns into Core would make Core aware of CLI input and output policy. Were you thinking they should move into Core, or into a separate command-independent Runtime module outside `handlers`? `normalizeRuntimeInvokeRequest` is the one boundary I could see revisiting, but the other helpers appear command-specific.

src/runnable/index.tsx
aidandaly24: Agreed. I’ll keep this PR scoped to structural `exitCode` handling for the local Runtime and router usage errors.

src/handlers/runtime/invoke/screen.tsx
aidandaly24: Agreed. Since this review, `9ccab9b` aligned the Runtime console more closely with Harness: payload editing moved into `RuntimePayloadInput`, and scrolling, key hints, status presentation, and pre-stream error handling now follow the Harness interaction model. Advanced options are already isolated in `RequestOptionsScreen`. The console still owns several responsibilities, and I agree there may be a cleaner split. Let’s discuss the intended boundary next week, and I’ll make a focused change from there.

src/runnable/index.tsx
aidandaly24: Update: #1834 has merged, so after rebasing this PR now uses `AgentCoreCLIError` directly. Runtime and router validation errors use `InputValidationError` with exit code `2`, and `runWithExitCode` reads exit codes from the shared base class. Addressed in `2b8ad055`.

src/runnable/index.tsx
aidandaly24: Yes. The shared `AgentCoreCLIError` introduced by #1834 now provides this. This PR was rebased and migrated onto it in `2b8ad055`.

src/runnable/index.tsx
aidandaly24: Addressed. Unreported failures still reach `console.error`; only Commander output and response failures that were already reported are suppressed.

src/core/runtime.tsx
```diff
@@ -10,17 +10,197 @@ import {
...
+
 export class RuntimeClient implements CoreRuntimeClient {
-  constructor(private readonly clients: AwsClients) {}
+  constructor(
+    private readonly clients: AwsClients,
+    private readonly fetch: CoreFetch,
+    private readonly logger: Logger,
+  ) {}
+
+  async invokeRuntime(
```
jariy17, line 34: nit: it seems to me like the `bearerToken` logic should be shifted into its own
  function (`invokeRuntimeWithBearerToken`), or even just a helper. Something like:

```suggestion
        async invokeRuntime(...) {
        if (bearerToken !== undefined) return this.invokeRuntimeWithBearerToken(...);
        else return this.invokeRuntimeWithIam(...);
      }
```

src/core/runtime.tsx
```diff
@@ -10,17 +10,197 @@ import {
...
+      url.pathname = `${url.pathname.replace(/\/?$/, "/")}runtimes/${encodeURIComponent(runtimeId)}/invocations`;
+      url.search = new URLSearchParams({
+        accountId: request.accountId,
+        qualifier: request.qualifier,
+      }).toString();
+      const headers = new Headers(applicationHeaders);
+      try {
+        headers.set("Authorization", `Bearer ${bearerToken}`);
+      } catch {
+        throw new TypeError("Invalid bearer token");
```
jariy17: please use `AgentCoreCLIError` here, maybe `InputValidationError` — this is invalid user input rather than an internal failure, so we get `source: USER` and the right exit code for free.

src/core/runtime.tsx
```diff
@@ -10,17 +10,197 @@ import {
...
+          ["X-Amzn-Bedrock-AgentCore-Runtime-User-Id", request.runtimeUserId],
+          ["X-Amzn-Trace-Id", request.traceId],
+          ["traceparent", request.traceParent],
+          ["tracestate", request.traceState],
+          ["baggage", request.baggage],
+        ] as const) {
+          if (value !== undefined) headers.set(name, value);
+        }
+      } catch {
+        throw new TypeError("Invalid Runtime request header");
```
jariy17: please use AgentCoreCLIError here, maybe InputValidationError — this is invalid user input rather than an internal failure, so we get source: USER and the right exit code for free.

src/core/runtime.tsx
```diff
@@ -10,17 +10,197 @@ import {
...
+          .child({
+            errorName:
+              error instanceof TypeError
+                ? "TypeError"
+                : error instanceof Error
+                  ? "Error"
+                  : typeof error,
+          })
+          .debug("Runtime invocation transport failed");
+        throw new Error("Runtime invocation failed");
```
jariy17: please use AgentCoreCLIError here, maybe InputValidationError — this is invalid user input rather than an internal failure, so we get source: USER and the right exit code for free.

src/core/runtime.tsx
```diff
@@ -10,17 +10,197 @@ import {
...
+      }
+      if (!response.ok) {
+        logger
+          .child({ httpStatusCode: response.status })
+          .debug("Runtime invocation returned a non-success response");
+        await response.body?.cancel().catch(() => undefined);
+        throw new Error(`HTTP ${response.status}`);
```
jariy17: please use AgentCoreCLIError here

src/handlers/runtime/invoke/invoke.test.tsx
```diff
@@ -0,0 +1,452 @@
...
+    contentType: "application/json",
+    body: body(Uint8Array.from([0, 255]), Uint8Array.from([10, 1])),
+  });
+  const output = captureIO();
+
+  await runCommand(core, output.io, args);
+  return { core, output };
+}
+
+describe("runtime invoke", () => {
```
jariy17, line 83: invoke uses TestCoreClient rather than the __fixtures__ + matchGolden pattern used by the read commands — reasonable given data-plane invocations aren't cleanly recordable, but a golden test for the raw/JSON stdout envelope would lock in the output contract

src/io/source.ts
```diff
@@ -6,7 +6,7 @@ const FILE_PREFIX = "file://";
 const STDIN = "-";
 
 export type SourceResolverConfig = {
-  stdin: NodeJS.ReadStream;
+  stdin?: NodeJS.ReadStream;
```
jariy17: Why is this optional now, when we throw if it's null anyway? Let's keep it required so we catch this issue during build time and not runtime.

jariy17 (dismissed):
Overall good, all of the nits can be a follow up CR.

src/core/runtime.tsx
aidandaly24: Addressed in c3639f73. `invokeRuntime` now dispatches by auth mode, with the existing transport logic localized in private `invokeRuntimeWithCustomJwt` and `invokeRuntimeWithIam` methods.

jariy17 (dismissed):

src/core/runtime.tsx
```diff
@@ -10,17 +10,216 @@ import {
...
+      headers.set("Authorization", `Bearer ${bearerToken}`);
+    } catch {
+      throw new TypeError("Invalid bearer token");
+    }
+    try {
+      for (const [name, value] of [
+        ["Content-Type", request.contentType],
+        ["Accept", request.accept],
+        ["Mcp-Session-Id", request.mcpSessionId],
+        ["X-Amzn-Bedrock-AgentCore-Runtime-Session-Id", request.runtimeSessionId],
```
notgitika: could we generate a runtime session id for `CUSTOM_JWT` requests when `--session-id` is omitted? 

The IAM SDK generates one automatically, but this raw fetch path appears to omit the header entirely.

src/handlers/runtime/invoke/errors.ts
```diff
@@ -0,0 +1,25 @@
+import { AgentCoreCLIError, InputValidationError } from "../../../errors";
+
+export class UsageError extends InputValidationError {
+  constructor(message?: string, options?: ErrorOptions) {
+    super(message, { ...options, exitCode: 2 });
+  }
+}
```
notgitika: I like this. however, I notice that in the `index.tsx` file below this, zod flag validation is handled by the shared router, which throws `InputValidationError` instead. 

I guess my question is should these schema validation failures return usage exit code 2? Invalid Runtime IDs and empty output paths currently become `InputValidationError` with exit code 1, while other invalid invoke arguments use UsageError. I also saw this behavior in the test files.

How can we make this consistent?

README.md
```diff
@@ -154,6 +155,85 @@ Source-aware values: any field flag documented as such accepts the value inline,
...
+Responses with `text/event-stream`, `application/x-ndjson`,
+`application/ndjson`, or `application/json-seq` content types stream exact bytes
+to stdout as they arrive. Other responses are buffered and written once after
+the body completes. Response metadata is written to stderr.
+
+`--output-file` streams either kind of response directly to disk and is required
+for binary output when stdout is a terminal. `--json` supports non-streaming
+responses only; it buffers one response and emits a metadata envelope without
+interpreting the customer body. If a streaming response fails, bytes already
+written remain available. A failed non-streaming response writes no partial body.
```
notgitika: raw mode currently streams every response type, and `--json` also buffers streaming responses and the tests appear to explicitly confirm both behaviors. So is this outdated?

jariy17 (approved):

notgitika (approved):
LGTM

PR #1822, opened 2026-07-23
feat(eval): add eval evaluator CLI commands (CRUD + LLaJ/code-based)

jariy17:
## What

Adds the imperative `agentcore eval evaluator` command surface (CLI only). Covers LLM-as-a-Judge and code-based evaluator create/update plus type-agnostic get/list/delete, following the DevX Evaluations/Optimization refactor doc and the existing `identity` handler conventions.

## Command structure

```
agentcore eval                              # evaluate and optimize AgentCore agents
└── evaluator                               # manage AgentCore evaluators
    ├── llm-as-a-judge                       # LLM-as-a-Judge evaluators
    │   ├── create
    │   └── update
    ├── code-based                           # code-based (Lambda-backed) evaluators
    │   ├── create
    │   └── update
    ├── get                                  # get an evaluator by id (type-agnostic)
    ├── list                                 # list evaluators (client-side --type filter)
    └── delete                               # delete an evaluator by id (requires --yes)
```

### `eval evaluator llm-as-a-judge create`
| Flag | Required | Notes |
|---|---|---|
| `--name` | ✓ | evaluator name |
| `--level` | ✓ | `SESSION \| TRACE \| TOOL_CALL` |
| `--model` | ✓ | Bedrock model id used to judge |
| `--instructions` | ✓ | source-aware: inline, `file://<path>`, or `-` (stdin) |
| `--rating-scale` | ✓* | preset: `1-5-quality \| 1-3-simple \| pass-fail \| good-neutral-bad` |
| `--rating-scale-json` | ✓* | raw `RatingScale` JSON (source-aware). Mutually exclusive with `--rating-scale`; exactly one required |
| `--kms-key-arn` | | customer-managed KMS key ARN |
| `--tags` | | JSON object (source-aware) |
| `--client-token` | | idempotency token |

### `eval evaluator llm-as-a-judge update`
| Flag | Required | Notes |
|---|---|---|
| `--id` | ✓ | evaluator id |
| `--instructions` | | source-aware |
| `--model` | | |
| `--rating-scale` / `--rating-scale-json` | | mutually exclusive |
| `--kms-key-arn` | | |
| `--client-token` | | |

Fields left unset are preserved: `UpdateEvaluator` replaces the whole `evaluatorConfig` union, so the client does get-then-merge over the current definition.

### `eval evaluator code-based create`
| Flag | Required | Notes |
|---|---|---|
| `--name` | ✓ | evaluator name |
| `--level` | ✓ | `SESSION \| TRACE \| TOOL_CALL` |
| `--lambda-arn` | ✓ | Lambda that scores a session |
| `--timeout` | | seconds (1–300). No CLI default; the service applies its own (60s) |
| `--kms-key-arn` | | |
| `--tags` | | JSON object (source-aware) |
| `--client-token` | | |

### `eval evaluator code-based update`
| Flag | Required | Notes |
|---|---|---|
| `--id` | ✓ | evaluator id |
| `--lambda-arn` | | |
| `--timeout` | | |
| `--kms-key-arn` | | |
| `--client-token` | | |

### `eval evaluator get`
| Flag | Required | Notes |
|---|---|---|
| `--id` | ✓ | evaluator id |

### `eval evaluator list`
| Flag | Required | Notes |
|---|---|---|
| `--next-token` | | pagination token (server-side) |
| `--max-results` | | max items (server-side) |
| `--type` | | filters the returned page client-side: `Builtin \| code-based \| llm-as-a-judge`. The ListEvaluators API paginates only, so the filter is page-local |

### `eval evaluator delete`
| Flag | Required | Notes |
|---|---|---|
| `--id` | ✓ | evaluator id |
| `--yes` | ✓ | confirm deletion (required in this headless/JSON-only path) |

## Notes / decisions
- **Source-aware field values** (`--instructions`, `--rating-scale-json`, `--tags`): inline, `file://<path>`, or `-` for stdin, following the AWS CLI `file://` convention. One flag reads stdin per command.
- **Rating scale** offers presets (the common defaults) and a raw-JSON escape hatch (what the API supports directly).
- **`--timeout`** has no CLI default; the service default applies when omitted.
- **Dependency inversion**: `CoreEvalClient` is declared next to the handlers (`src/handlers/eval/types.tsx`) and implemented in `src/core/eval.tsx`.

## Testing
- `bun test` — full suite green (356 tests).
- `bun run build` + bundle smoke (`node dist/index.js eval evaluator --help`) — works.
- New tests: handler behavior + validation/source/list-filter via `TestCoreClient`, `EvalClient` get-then-merge unit tests, and source-resolver tests.

---

TUI flows for these commands will come in a later PR.


src/handlers/eval/evaluator/code-based/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+      flag("client-token", "idempotency token", z.string().optional()),
+    ],
+    handle: async (ctx, flags) => {
+      if (!flags["name"]) throw new TypeError("required option '--name <name>' not specified");
+      if (!flags["level"]) throw new TypeError("required option '--level <level>' not specified");
+      if (!flags["lambda-arn"]) {
+        throw new TypeError("required option '--lambda-arn <lambda-arn>' not specified");
+      }
+
+      const source = new SourceResolver(io);
```
AlexanderRichey: Both @aidandaly24 and @nborges-aws wrote something that does the same thing in their PRs. Let's sync to align on a plan to avoid duplication here. Someone will have to merge first and the others will have to rebase their PRs against the changes, but it should be pretty easy.

src/handlers/eval/evaluator/code-based/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+          tags,
+          clientToken: flags["client-token"],
+        },
+        coreOptsFromCtx(ctx),
+      );
+      ctx.require(JsonRendererKey).renderJson(response);
+    },
+  });
+
+export const createCodeBasedUpdateHandler = (core: Core) =>
```
AlexanderRichey: I think we should pull this into its own file/dir. The pattern is documented here: https://github.com/aws/agentcore-cli/tree/refactor#adding-a-new-handler

src/handlers/eval/evaluator/delete/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const createDeleteEvaluatorHandler = (core: Core) =>
+  createHandler({
+    name: "delete",
+    description: "delete an evaluator by id",
+    flags: [
+      flag("id", "the ID of the evaluator to delete", z.string().optional()),
+      // This branch is headless/JSON only (no TUI prompt yet), so confirmation is
+      // required up front. `-y` shorthand is not wired: the router flag layer only
+      // supports long option names today.
+      flag("yes", "confirm deletion (required in non-interactive mode)", z.boolean().optional()),
```
AlexanderRichey: We don't have this convention for the other CRUDL commands. Let's leave this out to be consistent.

src/handlers/eval/evaluator/llm-as-a-judge/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+          tags,
+          clientToken: flags["client-token"],
+        },
+        coreOptsFromCtx(ctx),
+      );
+      ctx.require(JsonRendererKey).renderJson(response);
+    },
+  });
+
+export const createLlmAsAJudgeUpdateHandler = (core: Core, io: AppIO) =>
```
AlexanderRichey: Same here. I think this should follow the same convention: https://github.com/aws/agentcore-cli/tree/refactor#adding-a-new-handler

src/handlers/eval/types.tsx
```diff
@@ -0,0 +1,62 @@
...
+  UpdateEvaluatorResponse,
+} from "@aws-sdk/client-bedrock-agentcore-control";
+import type { CoreOptions } from "../../core/types";
+
+// LlmAsAJudgeUpdate carries the fields a caller may change on an LLM-as-a-Judge
+// evaluator. Any field left undefined is preserved from the existing evaluator:
+// the AgentCore UpdateEvaluator API replaces the whole evaluatorConfig union, and
+// the llmAsAJudge arm requires instructions + ratingScale + modelConfig together,
+// so a partial update is only possible by merging over the current definition.
+export interface LlmAsAJudgeUpdate {
```
AlexanderRichey: These should be types and not interfaces. Interfaces are abstract and need to be implemented by something else, whereas types are more concrete and can hold data themselves.

AlexanderRichey (dismissed):
Really good work here. Left a few comments. Main blocking things are using types vs interfaces appropriately and handling duplication of source reading.

src/handlers/eval/evaluator/code-based/index.tsx
jariy17: I'll wait till @nborges-aws's [PR](https://github.com/aws/agentcore-cli/pull/1819) to get merged

src/handlers/eval/types.tsx
jariy17: Changed `LlmAsAJudgeUpdate` and `CodeBasedUpdate` to type aliases (they just carry data). Left `CoreEvalClient` as an interface since `EvalClient` implements it.

src/handlers/eval/evaluator/code-based/index.tsx
jariy17: I'll update the PR to follow this convention

src/handlers/eval/evaluator/delete/index.tsx
jariy17: Removed it now

src/handlers/eval/evaluator/llm-as-a-judge/index.tsx
jariy17: I'll follow this convention

src/core/eval.tsx
```diff
@@ -0,0 +1,138 @@
...
+    id: string,
+    update: LlmAsAJudgeUpdate,
+    options: CoreOptions,
+  ): Promise<UpdateEvaluatorResponse> {
+    const control = this.clients.control(toClientConfig(options));
+    const current = await control.send(new GetEvaluatorCommand({ evaluatorId: id }));
+    const existing =
+      current.evaluatorConfig && "llmAsAJudge" in current.evaluatorConfig
+        ? current.evaluatorConfig.llmAsAJudge
+        : undefined;
```
aidandaly24: If its `undefined` should we throw? Since there should definitely be an existing, right?

We could reject immediately when the current config is not llmAsAJudge. Supplying all LLM fields currently converts a code-based evaluator. The code-based path has the same issue with `--lambda-arn`so I would change both.

src/core/eval.tsx
```diff
@@ -0,0 +1,138 @@
...
+      throw new TypeError(
+        `Evaluator "${id}" is not an LLM-as-a-Judge evaluator or is missing configuration required to update it`,
+      );
+    }
+
+    const evaluatorConfig: EvaluatorConfig = {
+      llmAsAJudge: {
+        instructions,
+        ratingScale,
+        modelConfig: { bedrockEvaluatorModelConfig: { modelId } },
```
aidandaly24: This rebuilds the model config with only `modelId`, could this not silently remove existing `inferenceConfig` and `additionalModelRequestFields`. Could we preserve the existing `bedrockEvaluatorModelConfig`, override only modelId, and maybe add a regression test if you agree?

README.md
```diff
@@ -64,6 +64,17 @@ agentcore                          # interactive TUI
...
+│   └── evaluator                  # manage AgentCore evaluators
+│       ├── llm-as-a-judge         # LLM-as-a-Judge evaluators
+│       │   ├── create             # create (instructions + rating scale + model)
+│       │   └── update             # update (merged over the existing config)
+│       ├── code-based             # code-based (Lambda-backed) evaluators
+│       │   ├── create             # create (Lambda ARN + optional timeout)
+│       │   └── update             # update (merged over the existing config)
+│       ├── get                    # get an evaluator by id (type-agnostic)
+│       ├── list                   # list evaluators (client-side --type filter)
+│       └── delete                 # delete an evaluator by id (requires --yes)
```
aidandaly24: Wasn't `--yes` removed from the delete handler?

src/handlers/eval/evaluator/evaluator.test.tsx
```diff
@@ -0,0 +1,375 @@
...
+} from "../../../testing";
+import { createRootHandler } from "../../index";
+import { ratingScaleFromPreset } from "../ratingScale";
+
+const REGION = "us-west-2";
+
+// run drives the real router (parsing → middleware → handler) against a
+// TestCoreClient so we can assert on the exact request a handler built, plus the
+// captured stdout. Optional `stdin` seeds the in-memory input stream for `-`.
+function run(core: TestCoreClient, args: string[], stdin?: string) {
```
aidandaly24: Is there a reason we're using `TestCoreClient` for these handler tests? For ordinary headless CRUD flows, Harness, Runtime, and Identity use the real `CoreClient` with fixture-backed SDK clients and golden output. `TestCoreClient` is reserved for interactive or streaming behavior. Following that pattern here would exercise Eval's SDK command construction and avoid adding a separate test seam.

aidandaly24 (changes requested):
A few things:

src/handlers/eval/evaluator/evaluator.test.tsx
aidandaly24: And for golden output generate them using the account 685197708687 for parity w what we have for Identity and Runtime.

src/core/eval.test.ts
```diff
@@ -0,0 +1,102 @@
+import { test, expect } from "bun:test";
```
Hweinstock: are we able to test this functionality through `src/handlers/eval/evaluator/evaluator.test.tsx` in a more e2e style? Looking at the existing integrations, it looks like we don't test core implementations directly, but instead test them through the handlers.

src/handlers/eval/evaluator/code-based/create/index.tsx
```diff
@@ -0,0 +1,61 @@
+import z from "zod";
```
Hweinstock: should we follow the camelCase convention for file paths?

src/handlers/eval/evaluator/code-based/create/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+      flag("kms-key-arn", "customer managed KMS key ARN for evaluator data", z.string().optional()),
+      flag(
+        "tags",
+        "tags to apply (JSON object of key/value strings; inline, file://<path>, or - for stdin)",
+        z.string().optional(),
+      ),
+      flag("client-token", "idempotency token", z.string().optional()),
+    ],
+    handle: async (ctx, flags) => {
+      if (!flags["name"]) throw new TypeError("required option '--name <name>' not specified");
```
Hweinstock: should we swap these to the modeled input validation errors? This change might have been done before that was merged yesterday, but it should safe us some effort in the future.

src/core/eval.tsx
```diff
@@ -0,0 +1,138 @@
...
+    const instructions = update.instructions ?? existing?.instructions;
+    const ratingScale = update.ratingScale ?? existing?.ratingScale;
+    const modelId =
+      update.model ??
+      (existing?.modelConfig && "bedrockEvaluatorModelConfig" in existing.modelConfig
+        ? existing.modelConfig.bedrockEvaluatorModelConfig?.modelId
+        : undefined);
+
+    if (!instructions || !ratingScale || !modelId) {
+      throw new TypeError(
```
Hweinstock: should we throw a modeled exception here?

src/handlers/eval/evaluator/list/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+export const createListEvaluatorsHandler = (core: Core) =>
+  createHandler({
+    name: "list",
+    description: "list evaluators",
+    flags: [
+      flag("next-token", "pagination token returned by a previous request", z.string().optional()),
+      flag("max-results", "maximum number of items to return", z.number().optional()),
+      flag(
+        "type",
+        `filter the returned page by type (${TYPE_FILTERS.join(" | ")})`,
```
Hweinstock: could this client side filtering result in some behavior where we get empty responses (but there is data on another page)? I'm wondering if it would be simpler to only support api-level filters.

src/handlers/eval/evaluator/llm-as-a-judge/sharedFlags.tsx
```diff
@@ -0,0 +1,39 @@
+import z from "zod";
```
Hweinstock, line 1: i believe the strands team avoids generic 'utils' or 'common' files/modules because they end up becoming a dumping ground of functionality that makes it difficult to find things (which I've definitely experience on past projects). What do you think of following the same pattern here and naming this more specifically for the functionality it provides (like sharedFlags)?

src/core/eval.test.ts
jariy17: If the rest of handlers are doing this, I'm fine with it.

src/core/eval.tsx
jariy17: yes, Ill used the modeled exceptions introduced by ur pr.

src/handlers/eval/evaluator/list/index.tsx
jariy17: Yes, it would. I wanted agentcore cli to be more than just aws cli wrapper so adding client side filters make sense to me. I'll just remove it for now until it's a customer ask.

src/core/eval.tsx
jariy17: Nice catch!, i'll take override the existing bedrockEvaluatorModelConfig rather than creating a new object.

src/core/eval.tsx
jariy17: Yes, this makes sense, I'll throw a InputValidationError

src/handlers/eval/evaluator/code-based/create/index.tsx
jariy17: Yes, I'll throw an InputValidationError

src/handlers/eval/evaluator/llm-as-a-judge/sharedFlags.tsx
jariy17: I agree with this sentiment.

src/handlers/eval/evaluator/evaluator.test.tsx
jariy17: Yes, I'll push a change to follow this fixture pattern

README.md
jariy17: Yes, I forgot about this, I remove this

src/handlers/eval/evaluator/llm-as-a-judge/sharedFlags.tsx
aidandaly24: I like this pattern

src/handlers/eval/evaluator/code-based/create/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+export const createCodeBasedCreateHandler = (core: Core, io: AppIO) =>
+  createHandler({
+    name: "create",
+    description: "create a code-based (Lambda-backed) evaluator",
+    flags: [
+      flag("name", "the name of the evaluator", z.string().optional()),
+      flag("level", `evaluation level (${LEVELS.join(" | ")})`, z.enum(LEVELS).optional()),
+      flag("lambda-arn", "ARN of the Lambda function that scores a session", z.string().optional()),
+      // No default; the service applies its own timeout (60s) when omitted.
+      flag("timeout", "Lambda timeout in seconds (1-300)", z.number().optional()),
```
aidandaly24: It seems `--timeout` is not enforcing the documented `1-300` range. I tested `0`, `301`, and `1.5`, and all three reached the Core client. This should use `z.number().int().min(1).max(300)` for both create and update, with boundary tests.

src/handlers/eval/evaluator/evaluator.test.tsx
```diff
@@ -0,0 +1,404 @@
...
+    const stdout = await run(["eval", "evaluator", "delete", "--id", llajId]);
+    matchGolden(FIXTURES, "llaj-delete.golden.json", stdout);
+  });
+
+  test("deletes the code-based evaluator", async () => {
+    const stdout = await run(["eval", "evaluator", "delete", "--id", codeBasedId]);
+    matchGolden(FIXTURES, "code-based-delete.golden.json", stdout);
+  });
+
+  test("propagates ResourceNotFoundException from get", async () => {
```
aidandaly24, line 333: This is not actually testing `ResourceNotFoundException`. The recorded fixture contains a `ValidationException` because `missing-evaluator-000` is not a valid evaluator ID. We should use a valid nonexistent ID and assert the error name, as Runtime and Identity do.

README.md
```diff
@@ -114,8 +125,35 @@ agentcore identity api-key-credential-provider get --name my-provider
...
 agentcore identity api-key-credential-provider update --name my-provider --api-key <new-key>
 agentcore identity api-key-credential-provider delete --name my-provider
+
+# Manage evaluators
+# Create an LLM-as-a-Judge evaluator with a rating-scale preset.
+agentcore eval evaluator llm-as-a-judge create \
+  --name order-support-quality \
+  --level SESSION \
+  --model us.anthropic.claude-sonnet-4-5 \
+  --instructions "Judge whether the order-support agent answered correctly." \
```
aidandaly24: nit: doesn't this require an allowed placeholder like `{context}` like 
```
--instructions "Judge from {context} whether the agent answered correctly." \
```

aidandaly24 (commented):
A few more comments

src/handlers/eval/evaluator/code-based/create/index.tsx
jariy17: Yeah, that makes sense. My instinct was to let the service throw the ValidationException, but harness/exec establishes the client-side pattern (z.number().min(1).max(3600)), so I'll follow it — z.number().int().min(1).max(300) on both create and update. However, I won't add boundary tests — that would be testing the shared flag layer's zod validation, which is already covered there

src/core/eval.tsx
```diff
@@ -0,0 +1,156 @@
...
+  ): Promise<UpdateEvaluatorResponse> {
+    const control = this.clients.control(toClientConfig(options));
+    const current = await control.send(new GetEvaluatorCommand({ evaluatorId: id }));
+
+    // Reject a type mismatch before merging: UpdateEvaluator replaces the whole
+    // evaluatorConfig union, so merging into the wrong arm would silently convert
+    // a code-based evaluator into an LLM-as-a-Judge one.
+    if (!current.evaluatorConfig || !("llmAsAJudge" in current.evaluatorConfig)) {
+      throw new InputValidationError(`Evaluator "${id}" is not an LLM-as-a-Judge evaluator`, {
+        meta: { evaluatorId: id },
```
Hweinstock, line 48: oooh, i like the use of meta!

src/handlers/eval/evaluator/llm-as-a-judge/sharedFlags.test.tsx
```diff
@@ -0,0 +1,61 @@
+import { describe, expect, test } from "bun:test";
```
Hweinstock, line 1: is it possible to get coverage for this through the handler tests? The advantage being that we can change the implementation of how we resolve the rating scale as long as it doesn't impact e2e behavior.

Hweinstock:
one thing I noticed testing e2e is that the help pages no longer tell us whats required since we aren't using commander's required option:
```
Usage: agentcore eval evaluator code-based create [options]

create a code-based (Lambda-backed) evaluator

Options:
  --name <name>                  the name of the evaluator
  --level <level>                evaluation level (SESSION | TRACE | TOOL_CALL)
  --lambda-arn <lambda-arn>      ARN of the Lambda function that scores a session
  --timeout <timeout>            Lambda timeout in seconds (1-300)
  --kms-key-arn <kms-key-arn>    customer managed KMS key ARN for evaluator data
  --tags <tags>                  tags to apply (JSON object of key/value strings; inline, file://<path>, or - for stdin)
  --client-token <client-token>  idempotency token
  -h, --help                     display help for command
```

I think this is an issue with a lot of the work we've done though, not specific to here. 

the apis are definitely a little awkward, but the commands themselves work great!

Hweinstock (approved):

aidandaly24 (approved):

src/handlers/eval/evaluator/llm-as-a-judge/sharedFlags.test.tsx
jariy17: Sure ill do it as a follow up

PR #1826, opened 2026-07-23
fix: preserve wrapped CLI help in API docs

notgitika:
## Summary
- join wrapped Commander argument and option descriptions during extraction
- normalize generated AsciiDoc for AWS style and China rebranding checks
- add a regression test for wrapped help output

Addresses generated documentation feedback from Amazon CR-290947859.

## Testing
- `node --test scripts/extract-cli-model.test.mjs`
- regenerated the v0.24.2 CLI reference

### DIFF
--- a/scripts/extract-cli-model.test.mjs
+++ b/scripts/extract-cli-model.test.mjs
@@ -1,9 +1,8 @@
-import assert from "node:assert/strict";
-import test from "node:test";
+import { normalizeParameterDescription, parseHelp } from './extract-cli-model.mjs';
+import assert from 'node:assert/strict';
+import test from 'node:test';
 
-import { normalizeParameterDescription, parseHelp } from "./extract-cli-model.mjs";
-
-test("parseHelp preserves wrapped argument and option descriptions", () => {
+test('parseHelp preserves wrapped argument and option descriptions', () => {
   const parsed = parseHelp(`Usage: agentcore invoke [options] [prompt]
 
 Invoke an agent.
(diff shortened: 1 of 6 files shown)


notgitika:
@strandly-the-agent review

scripts/render_adoc.py
```diff
@@ -55,15 +55,44 @@
...
+        "This feature is in preview and might change in future releases.",
+    )
+    return re.sub(r"\bAWS\b", "{aws}", text)
+
+
+def normalize_param_description(text):
+    """Normalize recurring parameter-description style issues."""
+    text = normalize_style(text).strip()
+    substitutions = (
+        (r"^Optional\b", "An optional"),
```
strandly-the-agent: 🟡 This global rewrite breaks valid plural and mass-noun descriptions from the renderer's other supported sources. I reproduced `Optional tags.` → `An optional tags.` and `Optional additional task metadata` → `An optional additional task metadata`; every parameter reaches this rule via `render_params()`. Could we preserve `Optional`, or make the rewrite noun-aware, and add an exact renderer test for those cases plus the intended `{aws}` output?

scripts/extract-cli-model.test.mjs
```diff
@@ -0,0 +1,32 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import { parseHelp } from "./extract-cli-model.mjs";
+
+test("parseHelp preserves wrapped argument and option descriptions", () => {
```
strandly-the-agent: 🟡 The regression test passes with `node --test`, but no repository test target runs it: the unit project only discovers `src/**/*.test.ts(x)`, and the package test scripts invoke Vitest. I confirmed Vitest reports this file as “No test files found.” Please wire it into a CI-invoked script/workflow or convert/move it into the discovered suite so this regression remains enforced.

strandly-the-agent (changes requested):
**TL;DR — changes requested:** the wrapped-help fix works, but the shared renderer can corrupt `Optional …` prose and the new regression test is not part of any automated test target.

✅ Reviewed `642f2f5`; direct Node test **1/1**, Python compile, build, and end-to-end generation (**6 groups / 30 entries / 139 params**) passed.
🔴 `format:check` still fails on the new test file; CI already identifies the formatting fix.
⚠️ The separate `ai-review` failure occurred during runner setup before review execution; the remaining build/test/security checks pass.

<details><summary>Review breakdown</summary>

- **Correctness/docs:** one deterministic cross-source wording regression.
- **Test quality:** the parser regression test passes directly but is absent from package/CI test targets.
- **Suppressed:** the memory-strategy wording concern did not reproduce against real inputs; no other escaping or parser defect was found.

No API or model-context pass was needed: these are internal documentation-generation scripts.
</details>

scripts/render_adoc.py
notgitika: Fixed in e9c9afd6. Optional descriptions now consistently use "The optional", with exact renderer regression coverage for singular, plural, and mass-noun cases.

scripts/extract-cli-model.test.mjs
notgitika: Fixed in e9c9afd6. Added the test:docs package target and run it from the Node 20 build job, so both extractor and renderer regressions are enforced in CI.

scripts/render_adoc.py
notgitika: fixed

scripts/extract-cli-model.test.mjs
notgitika: fixed

nborges-aws (approved):

PR #1832, opened 2026-07-27
feat(io): centralize shared IO utilities

aidandaly24:
## Description

This PR introduces a shared `src/io` package for CLI input and output concerns:

- Moves `AppIO`, `ReadWriteJson`, and `FsReadWriteJson` into `src/io` and migrates existing consumers.
- Adds a class-based `SourceResolver` for inline values, `file://` sources, and stdin, with byte and strict UTF-8 text resolution.
- Enforces a single stdin consumer and provides consistent `SourceResolutionError` failures.
- Supports optional `AbortSignal` cancellation for file and stdin reads, following the cancellation behavior introduced in #1820.

The `io` package has no dependency on handlers or global config. Existing exports were removed rather than temporarily re-exported so dependent PRs can rebase onto the intended package boundary.

## Related Issue

N/A - this is maintainer-coordinated refactor work and does not have an associated issue.

## Documentation PR

N/A

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Other (please describe): Internal package refactor

## Testing

- [x] I ran `bun test` (345 passed, 0 failed)
- [x] I ran `bun run typecheck`
- [x] I ran `bun run lint:check`
- [x] I ran `bun run format:check`
- [x] I ran `bun run build`
- [x] `src/assets/` was not modified, so snapshots do not require updates

## Checklist

- [x] Any dependent changes have been merged and published, or no published dependency is required

---

terms of your choice.

### DIFF
--- a/src/io/source.test.ts
+++ b/src/io/source.test.ts
@@ -1,0 +1,187 @@
+import { afterEach, describe, expect, test } from "bun:test";
+import { mkdtemp, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { PassThrough, Readable } from "node:stream";
+
+import { SourceResolutionError, SourceResolver } from "./source";
+
+const tempDirectories: string[] = [];
+
+async function tempFile(contents: string | Uint8Array): Promise<string> {
+  const directory = await mkdtemp(join(tmpdir(), "agentcore-source-"));
+  tempDirectories.push(directory);
+  const path = join(directory, "source");
+  await writeFile(path, contents);
+  return path;
+}
+
+function stdin(...chunks: Uint8Array[]): NodeJS.ReadStream {
+  return Readable.from(chunks) as NodeJS.ReadStream;
+}
+
+afterEach(async () => {
+  await Promise.all(
+    tempDirectories.splice(0).map((directory) => rm(directory, { recursive: true, force: true })),
+  );
+});
+
+describe("SourceResolver bytes", () => {
+  test("passes undefined through", async () => {
+    const resolver = new SourceResolver({ stdin: stdin() });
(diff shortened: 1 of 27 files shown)


AlexanderRichey (approved):

src/io/source.ts
```diff
@@ -0,0 +1,89 @@
...
+
+const FILE_PREFIX = "file://";
+const STDIN = "-";
+
+export type SourceResolverConfig = {
+  stdin: NodeJS.ReadStream;
+  signal?: AbortSignal;
+};
+
+export class SourceResolutionError extends TypeError {
```
jariy17, line 13: Did we discuss about making all Errors, "AgentCoreErrors"? If so, we need to change this after @Hweinstock merges the AgentCore PR.

src/io/source.test.ts
```diff
@@ -0,0 +1,187 @@
...
+    });
+    const resolution = resolver.resolveBytes("payload", "inline");
+
+    await expect(resolution).rejects.toMatchObject({ name: "AbortError" });
+    await expect(resolution).rejects.not.toBeInstanceOf(SourceResolutionError);
+  });
+
+  test("preserves AbortError from file resolution", async () => {
+    const path = await tempFile("payload");
+    const controller = new AbortController();
```
jariy17, line 149: Can we update this test to abort after the file read starts? Aborting before resolveBytes() means throwIfAborted() exits early, so this never verifies that the signal reaches readFile().

jariy17 (commented):
Overall, great CR! Left a few comments. Can we also update the README.md about this new SourceResolver so our agents try to use this?

jariy17 (approved):

Issue #1790, opened 2026-07-17
feat: implement typed errors on refactor branch.


Hweinstock:
### Problem 
We currently throw untyped errors in the refactor. This makes it difficult to classify errors for telemetry in the future, and also leads to brittle logic to determine the source of errors based on message. 

### Definition of Done
create an error classification system that makes it easy to:
- determine the category of the error (ex. credentials expired, missing file, etc.).
- determine the "source" of the error (is it the user, is it the CLI internally, is it a service endpoint?). 
- extend with new categories of errors.

PR #1834, opened 2026-07-27
feat(errors): add top-level errors module for definitions and classification. 

Hweinstock:
### Problem
See https://github.com/aws/agentcore-cli/issues/1790

There is no consistent format for custom errors or way to override the exitCode. 

### Solution 
Define a top level errors module that has exposes two pieces of functionality:
- a base `AgentCoreCLIError` to build new errors from, that accepts metadata and "source" (used in telemetry). 
- a `classify` function that takes an arbitary value at the top-level and turns it into the correct error (with support for common AWS SDK exceptions). 

### Verification 
- Added unit tests at runnable for exitCode functionality, and for classify directly.  
- Tested e2e by adding failing api calls in the body and observing the logs. 

```
{"level":"error","msg":"","time":1785158184077,"cliSessionId":"98973677-d14f-477b-9650-17e7a97bc88b","error":{"name":"AgentCoreCLIError","message":"User: {...}
 is not authorized to perform: bedrock-agentcore:GetAgentRuntime","stack":"AgentCoreCLIError: User: {...} is not authorized to perform: bedrock-agentcore:GetAg
entRuntime\n    at classify ({...}/agentcore-cli/src/errors/classify.tsx:16:16)\n    at <anonymous> ({...}/agentcore-cli/src/index.ts:70:21)\n    at async runWithExitCode
 ({...}/agentcore-cli/src/runnable/index.tsx:24:11)\n    at processTicksAndRejections (native:7:39)","exitCode":1,"meta":{"httpStatusCode":403,"requestId":"ffb5c1cf-379f-46a1-919d-a4c0f3f36aaa
","attempts":1,"totalRetryDelay":0},"source":"user"}}
```

### Notes
this PR sets up the infrastructure, but there are more call sites to migrate. Migrating will come as a follow-up to keep the PR small and avoid churn with other open PRs.


Hweinstock:
audit fail is fixed in github.com/aws/agentcore-cli/pull/1833

src/index.ts
```diff
@@ -64,11 +65,9 @@ process.exit(
       // Handle the request
       await rootHandler.route(argv);
     } catch (e) {
-      const error = e instanceof Error ? e : new Error(String(e));
-      rootLogger
-        .child({ errorName: error.name, errorMessage: error.message, stack: error.stack ?? "" })
-        .error();
-      throw e;
+      const error = classify(e);
```
Hweinstock: we classify here, instead of the top-level since we'll have the telemetry client in scope here, but not at the top-level.

src/errors/classify.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { AgentCoreCLIError } from "./errors";
+
+/** Converts any thrown value into an {@link AgentCoreCLIError}, preserving known CLI errors. */
+export function classify(error: unknown): AgentCoreCLIError {
+  if (error instanceof AgentCoreCLIError) return error;
+
+  if (ServiceException.isInstance(error)) {
+    const httpStatusCode = error.$metadata.httpStatusCode;
+    const source =
+      httpStatusCode !== undefined && httpStatusCode >= 400 && httpStatusCode < 500
```
nborges-aws: Any chance something like `429: ThrottlingException` shows up here? Just thinking about 400s that might not truly be user errors.

src/errors/errors.tsx
```diff
@@ -0,0 +1,43 @@
+import { ERROR_SOURCE, type ErrorSource } from "./types";
+
+export interface AgentCoreCLIErrorOptions extends ErrorOptions {
+  /** The source fo the error. See {@link ErrorSource} for more information */
+  source?: ErrorSource;
+  /** Abitrary metdata to attach to errors for logging */
```
nborges-aws: super nit: arbitrary

nborges-aws (dismissed):

src/errors/errors.tsx
Hweinstock: good catch, let me fix that.

src/errors/classify.tsx
Hweinstock: Yeah good question. If the CLI sees a throttling exception its difficult to know if its an issue on the CLI making too many requests, or a user spamming through the CLI. 

We don't want any `internal` faults to be user-trigger-able, so I believe its safer to make this a user error and maintain the error name in telemetry. That way if we see a bunch of throttling across the board, we know there is likely a bug on the CLI.

src/index.ts
```diff
@@ -65,11 +66,9 @@ process.exit(
...
```
AlexanderRichey: I wonder if an alternative constructor would be more elegant here. Something like this?

```typescript
const error = AgentCoreCLIError.fromError(e)
```

src/errors/types.tsx
```diff
@@ -0,0 +1,10 @@
+export const ERROR_SOURCE = {
+  // note: this maps to the `client` error source in telemetry.
+  INTERNAL: "internal",
+  USER: "user",
+  SERVICE: "service",
+  UNKNOWN: "unknown",
```
AlexanderRichey: Can we get rid of `UNKNOWN`? It doesn't seem meaningfully different from `INTERNAL`, unless I'm missing something.

AlexanderRichey (dismissed):
Looks good! One thing I think would be a bit nicer would be not to have `classify` but to have an alternative constructor for this case that would do the same thing:

```typescript
const error = AgentCoreCLIError.fromError(e)
```

This would keep all of the public error handling code neatly exposed through one class.

Second thing is getting rid of `UNKNOWN` as an error type in favor of `INTERNAL`.

src/index.ts
Hweinstock: I like it! I'll swap it over.

src/errors/types.tsx
Hweinstock: discussed offline. distinction doesn't justify a separate field, aligning on INTERNAL.

AlexanderRichey (approved):

nborges-aws (approved):

PR #1823, opened 2026-07-23
fix(ci): provision review Python with uv on CodeBuild

notgitika:
## Summary

- replace `actions/setup-python` in the harness review workflow because it has no Amazon Linux 2023 Python builds
- use `setup-uv` to download Python 3.12 and activate a virtual environment
- install `boto3` into the activated environment

## Verification

- `npm run typecheck` (via pre-commit hook)
- Prettier and secretlint (via lint-staged)
- Full workflow dispatch on the CodeBuild runner: https://github.com/aws/agentcore-cli/actions/runs/30049510856
  - downloaded and activated CPython 3.12.13
  - installed boto3 successfully
  - completed the harness review and cleanup successfully

PR #1836, opened 2026-07-27
feat(memory): add read-only memory commands

notgitika:
This PR adds the first Memory surface to the refactored CLI with Read-only commands. 

Users can fetch a Memory with `agentcore memory get --id <memoryId>` and list Memories with service-side pagination through `--max-results` and `--next-token`. `get` also supports the `full` and `without_decryption` response views.

The commands are backed by a feature-scoped core `MemoryClient` using the AWS SDK. This also wires Memory into the root command tree and test client, adds fixture-backed command coverage, and updates the REAMDE.

Out of scope: data plane operations, memory mutations, TUI.

Next PR would be: dedicated TUI flows

Ran the bun CI checks and `bun build` successfully.

Manually tested all these commands E2E on my account and they were all successful and working as expected


notgitika:
AI review failing because of a bug. Fix: https://github.com/aws/agentcore-cli/pull/1823

src/handlers/memory/get/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const MEMORY_VIEWS = ["full", "without_decryption"] as const;
+
+export const createGetMemoryHandler = (core: Core) =>
+  createHandler({
+    name: "get",
+    description: "get an AgentCore Memory",
+    flags: [
+      flag("id", "the ID of the Memory", z.string().optional()),
+      flag("view", "response view", z.enum(MEMORY_VIEWS).default("full")),
```
Hweinstock: does providing a default at this level make still make it possible to open the TUI with no flags? Would the default cause the check in https://github.com/aws/agentcore-cli/blob/b57ed517eacd9397bf49c4aabcc6537de7c827b4/src/middleware/withTuiOnEmptyFlagsAndArgs.tsx#L20 to fail?

src/handlers/memory/list/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { JsonRendererKey } from "../../../tui";
+import type { Core } from "../../types";
+import { coreOptsFromCtx } from "../../utils";
+
+export const createListMemoriesHandler = (core: Core) =>
+  createHandler({
+    name: "list",
+    description: "list AgentCore Memories",
+    flags: [
+      flag("next-token", "pagination token returned by a previous request", z.string().optional()),
```
Hweinstock, line 12: OOS here, but curious to others thoughts. Do we think its worth defining a shared flag for stuff like this? We already have the same definitions for runtime, harness, and now memory.

Hweinstock (commented):
one question on TUI behavior, otherwise lgtm.

src/handlers/memory/get/index.tsx
notgitika: good catch! Memory doesn’t enable TUI routing in this PR, but the schema default would prevent empty-flag detection if we add it later. 

I removed the schema-level default and now `resolve flags.view ?? "full"` in the handler, preserving current behavior while keeping omitted flags empty.

src/handlers/memory/list/index.tsx
notgitika: Agreed, I’d prefer to keep that refactor out of this PR and follow up separately so it will be a follow-up once this is merged

notgitika:
Follow up PR: shared flags

src/handlers/memory/get/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+  createHandler({
+    name: "get",
+    description: "get an AgentCore Memory",
+    flags: [
+      flag("id", "the ID of the Memory", z.string().optional()),
+      flag("view", "response view", z.enum(MEMORY_VIEWS).optional()),
+    ],
+    handle: async (ctx, flags) => {
+      if (!flags.id) {
+        throw new TypeError("required option '--id <id>' not specified");
```
aidandaly24: This should probably throw `InputValidationError`, matching the current consensus error handling.

src/handlers/memory/memory.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { createRootHandler } from "../index";
+import { createGetMemoryHandler } from "./get";
+
+const REGION = "us-west-2";
+const FIXTURES = join(import.meta.dir, "__fixtures__");
+
+// These credential-free fixtures are synthetic but use the same SDK send-boundary
+// replay path as recorded fixtures, so command-to-SDK translation remains covered.
+const FIXTURE_MEMORY_ID = "agentcore_cli_memory_read_only_fixture-AbCd123456";
+const MISSING_MEMORY_ID = "missing_memory-0000000000";
```
aidandaly24: These fixtures cannot be regenerated using `fixtureFactories`' documented `RECORD=1` flow. The synthetic IDs do not exist, so recording replaces the successful get fixtures with not-found errors. I think Memory should use recorded responses from deterministic resources, matching Runtime and Identity.

aidandaly24 (changes requested):

src/handlers/memory/get/index.tsx
notgitika: Ah yes, my fault! Fixed now

src/handlers/memory/memory.test.tsx
notgitika: you're right that RECORD=1 would have overwritten the successful gets with not-found errors. fixed them now. The fixtures now hold real recorded responses

src/handlers/memory/list/index.tsx
Hweinstock: maybe we can raise as a parking lot. I do think there's an advantage to making them explicit so I'm curious to hear others' thoughts.

Hweinstock (dismissed):
nice!

aidandaly24 (dismissed):

Hweinstock (approved):

tejaskash (approved):

PR #1841, opened 2026-07-27
feat(eval): add eval online-eval CLI commands

jariy17:
## Summary
- Adds `agentcore eval online-eval create/get/list/update/pause/resume/delete`.
- `--agent <id>` accepts a plain AgentCore Runtime ID or a Harness ID and derives the CloudWatch log group + service name from it (a harness is itself backed by a runtime; resolution falls back from GetAgentRuntime to GetHarness). Verified against real resources in a live account.
- `--log-group-name`/`--service-name` are the escape hatch for agents that emit traces under a custom OTel service name rather than AgentCore's default runtime path.
- No execution role required at create time: Core auto-provisions and reuses a default one scoped to the resolved log group, mirroring the harness execution-role pattern.
- `update` merges over the existing config (the API replaces the whole `rule`/`dataSourceConfig`), so unset fields are preserved.

## Test plan
- [x] `bun x tsc --noEmit` clean
- [x] `bun test` — 391/391 pass
- [x] Smoke-tested `get`/`list` against real production online-eval configs in account 725476964917
- [x] Derivation formula verified against a real harness-backed runtime (create/get/delete round-trip)


jariy17:
Superseded by a rebase onto the latest `refactor` — reopening as a fresh PR with a single squashed commit.

PR #1845, opened 2026-07-27
fix: make data table columns terminal-aware

aidandaly24:
## Summary

Make `DataTable` allocate columns from terminal width so picker tables fill the screen without wrapping or welding adjacent fields.

## Changes

- add a pure shrink-then-drop width allocator with structural one-cell column gaps
- add explicit fixed/flex column sizing metadata to all six Runtime and Harness picker configurations
- require an explicit fixed width, default `minWidth` to it, and keep allocator internals private to `DataTable`
- represent dropped columns directly through omitted computed widths
- render full `version` headers and verify every visible production header fits its allocated width
- render compact UTC timestamps and the discriminating Runtime ID suffix
- preserve both rendered-value and raw-value filtering
- use display-cell-aware truncation for Unicode and ANSI-safe output
- remove unreachable interactive sorting inherited from the previous table implementation
- replace the unused `@inkui-cli/data-table` dependency with direct `cli-truncate` and `string-width` dependencies

## Validation

- `bun install --frozen-lockfile`
- `bun test` (403 passed)
- `bun test --coverage --coverage-reporter=lcov`
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- `bun run build`
- `bun audit`
- `bun pm pack --destination /tmp`

Live deploy-profile TUI validation in `us-east-1`:

- reproduced the base-branch failure at 80 columns: wrapped headers/rows, welded fields, and pagination text overlaid with table data
- verified Runtime list tables exactly fill 60, 80, and 120 columns with one physical line per row
- verified Harness list tables exactly fill 80 columns
- verified filtering by a rendered compact UTC timestamp returns the expected live Runtime row

### DIFF
--- a/src/components/RuntimeEndpointPicker.tsx
+++ b/src/components/RuntimeEndpointPicker.tsx
@@ -17,13 +17,12 @@ interface RuntimeEndpointRow extends Record<string, unknown> {
 export const runtimeEndpointColumns = [
   { key: "qualifier", header: "qualifier", flex: true },
   { key: "liveVersion", header: "live", width: 6, minWidth: 5 },
-  { key: "targetVersion", header: "target", width: 6, minWidth: 6 },
-  { key: "status", header: "status", width: 13, minWidth: 13 },
+  { key: "targetVersion", header: "target", width: 6 },
+  { key: "status", header: "status", width: 13 },
   {
     key: "lastUpdatedAt",
     header: "updated UTC",
     width: 16,
-    minWidth: 16,
     render: formatTimestamp,
   },
 ] satisfies DataTableColumn<RuntimeEndpointRow>[];
--- a/src/components/RuntimeEndpointPicker.tsx
+++ b/src/components/RuntimeEndpointPicker.tsx
@@ -2,8 +2,8 @@ import type { AgentRuntimeEndpoint } from "@aws-sdk/client-bedrock-agentcore-con
 import { useNavigate } from "react-router";
 import type { ScreenProps } from "../handlers/types";
 import { coreOptsFromCtx } from "../handlers/utils";
+import { formatTimestamp } from "./formatTimestamp";
 import { PaginatedTablePicker } from "./PaginatedTablePicker";
-import { formatTimestamp } from "./pickerFormatters";
 import type { DataTableColumn } from "./ui/data-table";
 
 interface RuntimeEndpointRow extends Record<string, unknown> {
--- a/src/components/RuntimeEndpointPicker.tsx
+++ b/src/components/RuntimeEndpointPicker.tsx
@@ -3,6 +3,8 @@ import { useNavigate } from "react-router";
 import type { ScreenProps } from "../handlers/types";
 import { coreOptsFromCtx } from "../handlers/utils";
 import { PaginatedTablePicker } from "./PaginatedTablePicker";
+import { formatTimestamp } from "./pickerFormatters";
+import type { DataTableColumn } from "./ui/data-table";
 
 interface RuntimeEndpointRow extends Record<string, unknown> {
   qualifier: string;
@@ -11,6 +13,20 @@ interface RuntimeEndpointRow extends Record<string, unknown> {
   status: string;
   lastUpdatedAt: string;
 }
+
+export const runtimeEndpointColumns = [
+  { key: "qualifier", header: "qualifier", flex: true },
+  { key: "liveVersion", header: "live", width: 6, minWidth: 5 },
+  { key: "targetVersion", header: "target", width: 6, minWidth: 6 },
+  { key: "status", header: "status", width: 13, minWidth: 13 },
+  {
+    key: "lastUpdatedAt",
+    header: "updated UTC",
+    width: 16,
+    minWidth: 16,
+    render: formatTimestamp,
+  },
+] satisfies DataTableColumn<RuntimeEndpointRow>[];
 
 function toRow(endpoint: AgentRuntimeEndpoint): RuntimeEndpointRow {
   return {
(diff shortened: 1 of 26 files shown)


src/components/ui/data-table/DataTable.tsx
```diff
@@ -44,6 +57,12 @@ export interface DataTableProps<T> {
   theme?: InkUITheme;
 }
 
+function headerWithSortIndicator(header: string, width: number, indicator: string): string {
```
aidandaly24: This should be removed this is accidental complexity.

src/components/ui/data-table/DataTable.tsx
```diff
@@ -1,19 +1,28 @@
...
+interface DataTableColumnBase<T> {
   key: keyof T & string;
   header: string;
   align?: "left" | "center" | "right";
-  sortable?: boolean;
   render?: (value: unknown, row: T) => string;
-  width?: number;
 }
 
+export type DataTableColumn<T> = DataTableColumnBase<T> & ColumnSizing;
```
Hweinstock: are we expecting other types to extend `DataTableColumnBase`? If not, I wonder if collapsing this into one type could simplify?

src/components/ui/data-table/columnWidths.ts
```diff
@@ -0,0 +1,94 @@
+export const COLUMN_GAP = 1;
+export const FLEX_MIN_WIDTH = 16;
+export const SELECTION_MARKER_WIDTH = 1;
+
+export type ColumnSizing =
+  | { flex: true; width?: never; minWidth?: never }
+  | { flex?: false; width: number; minWidth: number };
+
+export interface ComputedColumnWidths {
```
Hweinstock: should these two interfaces be types?

src/components/ui/data-table/columnWidths.ts
```diff
@@ -0,0 +1,94 @@
...
+  totalWidth: number;
+}
+
+interface FixedColumnWidth {
+  index: number;
+  width: number;
+  minWidth: number;
+}
+
+export function resolveBorderWidth(
```
Hweinstock: would bringing this out into an object make the optional parameters easier?
ex. 
```
resolveBorderWidth(border: { style?: string, left?: boolean, right?: boolean }) 
```

src/components/ui/data-table/columnWidths.ts
```diff
@@ -0,0 +1,94 @@
...
+  borderStyle: string | undefined,
+  borderLeft: boolean | undefined,
+  borderRight: boolean | undefined,
+): number {
+  if (borderStyle === undefined) return 0;
+  return Number(borderLeft !== false) + Number(borderRight !== false);
+}
+
+export function computeColumnWidths(
```
Hweinstock, line 33: should this take the header width as an input? When testing it looks like some of the headers are getting cut off 

Image

src/components/ui/data-table/columnWidths.ts
```diff
@@ -0,0 +1,94 @@
...
+    return [
+      {
+        index,
+        width: column.width,
+        minWidth: Math.min(column.minWidth, column.width),
+      },
+    ];
+  });
+
+  const gapCount = () => {
```
Hweinstock: im kind of confused what we're doing here. For example, `gapCount` is not an obvious concept to me. Do you see any way to simplify this logic or maybe add some concise comments to provide some intuition?

src/components/RuntimeEndpointPicker.tsx
```diff
@@ -12,6 +14,20 @@ interface RuntimeEndpointRow extends Record<string, unknown> {
   lastUpdatedAt: string;
 }
 
+export const runtimeEndpointColumns = [
+  { key: "qualifier", header: "qualifier", flex: true },
+  { key: "liveVersion", header: "live", width: 6, minWidth: 5 },
+  { key: "targetVersion", header: "target", width: 6, minWidth: 6 },
```
Hweinstock: As a dev how do I know what a reasonable `minWidth` should be? Does it make sense to default it to the width?

src/components/ui/data-table/DataTable.tsx
aidandaly24: Done in `1215f5d5`. `DataTableColumnBase` is gone; the common fields now live directly in `DataTableColumn<T>`, intersected only with the allocator sizing union.

src/components/ui/data-table/columnWidths.ts
aidandaly24: Done in `1215f5d5`. `ComputedColumnWidths` and `FixedColumnWidth` are now type aliases.

src/components/ui/data-table/columnWidths.ts
aidandaly24: Done in `1215f5d5`. `resolveBorderWidth` now takes one options object with optional `style`, `left`, and `right` fields.

src/components/ui/data-table/columnWidths.ts
aidandaly24: Good catch. The screenshot was showing the literal configured `ver` label rather than allocator truncation. In `1215f5d5` the Runtime and Harness list headers are now `version` with a seven-cell width and no smaller shrink floor. I kept the allocator header-agnostic because column sizing remains explicit, and added a regression that checks every visible production header across terminal widths 1 through 200.

src/components/ui/data-table/columnWidths.ts
aidandaly24: Done in `1215f5d5`. `gapCount` is now `gapWidth`, includes the separator multiplication itself, and is paired with `visibleFixedWidth` and `minimumTableWidth`. I also added one concise comment explaining that Ink inserts gaps between adjacent children and that the selection marker participates as a child.

src/components/RuntimeEndpointPicker.tsx
aidandaly24: Yes. Done in `1215f5d5`: `minWidth` is optional and defaults to `width`. Redundant equal declarations are removed, so an explicit smaller value now clearly means that the column is intentionally allowed to shrink.

src/components/formatTimestamp.ts
```diff
@@ -0,0 +1,5 @@
+export function formatTimestamp(value: unknown): string {
```
Hweinstock, line 1: i like how this cleans up the UI a bit!

Hweinstock (approved):

notgitika (approved):

PR #1846, opened 2026-07-27
feat(telemetry): setup client with logging sink

Hweinstock:
## Problem 
telemetry is not existent in this new cli. 

## Solution
We scaffold the telemetry implementation with three core pieces:
- `MetricSink`: the core abstraction for a destination of data. This could be logs, fs, or the collector. Allows us to decouple transport from orchestration logic. 
- `TelemetryClient`: the orchestration layer on top of metric sinks that validates metric shapes, and sends the data to the configured destinations. 
- `TelemetryAttributesRecorder`: a strongly typed object builder that allows us to set attributes later in the program execution. In the future we can pass this down in the context, and populate it in the handlers. We'll also want some middleware to populate common fields like command and global flags. 

## Testing 
- added tests on the client. 
- verified e2e by inspecting the logs on both executable and node build using the following:
run any command then
```
> cat ~/.agentcore/logs/output-2026-07-27.log | tail -n 1 | jq
{
  "level": "info",
  "msg": "recording telemetry",
  "time": 1785192842835,
  "cliSessionId": "d1b9e38c-2ec0-4062-9b04-a2034c6bebbf",
  "module": "loggingSink",
  "metricValue": 25,
  "metricAttributes": {
    "service.version": "0.0.0",
    "agentcore-cli.session_id": "d1b9e38c-2ec0-4062-9b04-a2034c6bebbf",
    "os.type": "Linux",
    "host.arch": "x64",
    "node.version": "v24.3.0",
    "exit_reason": "success"
  }
}
```
also failures shows up as failure under exit_reason. 

## Notes
- version requires some more effort to wire up, since we can no longer read the package.json when in an executable. Going to create a follow-up task for this.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -84,11 +84,14 @@ process.exit(
 
       throw error;
     } finally {
-      await telemetryClient.emit(
-        "cli.command_run",
-        Date.now() - startTime,
-        commandRunTelemetryRecorder.getAttributes(),
-      );
+      try {
+        const attributes = commandRunTelemetryRecorder.getAttributes();
+        await telemetryClient.emit("cli.command_run", Date.now() - startTime, attributes);
+      } catch (e) {
+        const error = AgentCoreCLIError.fromError(e);
+        rootLogger.child({ error: error.json() }).warn("failed to emit telemetry");
+        // telemetry is best-effort
+      }
       await telemetryClient.shutdown();
       await rootLogger.end();
     }
--- a/src/index.ts
+++ b/src/index.ts
@@ -14,11 +14,12 @@ import { FsReadWriteJson } from "./io";
 import { createFileLogger, LOG_LEVEL } from "./logging";
 import { runWithExitCode } from "./runnable";
 import { DefaultGlobalConfigAccessor } from "./globalConfig";
+import { DefaultTelemetryClient, TelemetryAttributesRecorder } from "./telemetry";
 
 process.exit(
   await runWithExitCode(async (argv: string[]) => {
+    const startTime = Date.now();
     // generate a unique identifier corresponding to this process of this CLI. (ex. one command invoke, one TUI session)
-    // TODO: wire this id into telemetry as well
     const cliSessionId = crypto.randomUUID();
 
     const rootLogger = createFileLogger({
@@ -33,15 +34,25 @@ process.exit(
       stderr: process.stderr,
     };
 
-    try {
-      const globalConfigAccessor = new DefaultGlobalConfigAccessor({
-        logger: rootLogger.child({ module: "globalConfigAccessor" }),
-        filePath: join(homedir(), ".agentcore", "config.json"),
-        json: new FsReadWriteJson({
-          logger: rootLogger.child({ module: "jsonDataSource" }),
-        }),
-      });
+    const globalConfigAccessor = new DefaultGlobalConfigAccessor({
+      logger: rootLogger.child({ module: "globalConfigAccessor" }),
+      filePath: join(homedir(), ".agentcore", "config.json"),
+      json: new FsReadWriteJson({
+        logger: rootLogger.child({ module: "jsonDataSource" }),
+      }),
+    });
 
+    const telemetryClient = new DefaultTelemetryClient({
+      logger: rootLogger.child({ module: "telemetry" }),
+      sessionId: cliSessionId,
+      globalConfigAccessor,
+    });
+
+    const commandRunTelemetryRecorder = new TelemetryAttributesRecorder("cli.command_run", {
+      exit_reason: "success",
+    });
+
+    try {
       rootLogger.info(`running CLI`);
 
       // factories (rather than instances) lets CoreClient build one client per
@@ -69,8 +80,16 @@ process.exit(
       rootLogger
         .child({ errorName: error.name, errorMessage: error.message, stack: error.stack ?? "" })
         .error();
+      // TODO: add error details to telemetry recorder;
+      commandRunTelemetryRecorder.record({ exit_reason: "failure" });
       throw e;
     } finally {
+      await telemetryClient.emit(
+        "cli.command_run",
+        Date.now() - startTime,
+        commandRunTelemetryRecorder.getAttributes(),
+      );
+      await telemetryClient.shutdown();
       await rootLogger.end();
     }
   }),
--- a/src/telemetry/client.tsx
+++ b/src/telemetry/client.tsx
@@ -1,0 +1,115 @@
+import type { Logger } from "../logging";
+import { LoggingSink } from "./loggingSink";
+import { resourceAttributesSchema, type ResourceAttributes } from "./shapes";
+import os from "os";
+import {
+  type AttributesOf,
+  type MetricSink,
+  type TelemetryClient,
+  type ValueOf,
+  METRICS,
+  type MetricName,
+} from "./types";
+import type { GlobalConfigAccessor } from "../globalConfig";
+
+export type DefaultTelemetryClientConfig = {
+  logger: Logger;
+  globalConfigAccessor: GlobalConfigAccessor;
+  sessionId: string;
+  metricSinks?: MetricSink[];
+};
+
+/**
+ * Implements {@link TelemetryClient} by validating and fanning out metrics to a list of {@link MetricSink} implementations.
+ */
+export class DefaultTelemetryClient implements TelemetryClient {
+  private logger: Logger;
+  private readonly sessionId: string;
+  private globalConfigAccessor: GlobalConfigAccessor;
+  private resourceAttributes: ResourceAttributes | undefined;
+  private metricSinks: MetricSink[] | undefined;
+
+  constructor(config: DefaultTelemetryClientConfig) {
+    this.logger = config.logger;
+    this.sessionId = config.sessionId;
+    this.globalConfigAccessor = config.globalConfigAccessor;
+    this.resourceAttributes = undefined;
+    this.metricSinks = config.metricSinks;
+  }
+
+  async emit<TMetricName extends MetricName>(
+    metricName: TMetricName,
+    metricValue: ValueOf<TMetricName>,
+    metricAttributes: Partial<AttributesOf<TMetricName>>,
+  ): Promise<void> {
+    try {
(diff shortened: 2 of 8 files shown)


src/telemetry/client.tsx
```diff
@@ -0,0 +1,115 @@
...
+          .child({ errorName: error.name, errorMessage: error.message })
+          .warn(`failed to shutdown metric sink with name '${sink.name}'`);
+      });
+    });
+    await Promise.all(promises);
+  }
+
+  private getMetricSinks(): MetricSink[] {
+    if (this.metricSinks !== undefined) return this.metricSinks;
+
```
Hweinstock, line 92: this will be async once we add fs sink for audit, so following lazy init pattern from attributes below.

src/telemetry/client.tsx
```diff
@@ -0,0 +1,115 @@
...
+    this.resourceAttributes = resourceAttributesSchema.parse({
+      "service.name": "agentcore-cli",
+      // TODO: wire up real package version.
+      "service.version": "0.0.0",
+      "agentcore-cli.installation_id": globalConfig.installationId,
+      "agentcore-cli.session_id": this.sessionId,
+      "os.type": os.type(),
+      "os.version": os.release(),
+      "host.arch": os.arch(),
+      "node.version": process.version,
```
Hweinstock, line 111: if I run this via bun, I get the wrong value here since bun hard-codes it for compatibility. However, when using the build for node, it does return the correct version.

src/index.ts
```diff
@@ -33,15 +34,25 @@ process.exit(
...
+      }),
+    });
+
+    const telemetryClient = new DefaultTelemetryClient({
+      logger: rootLogger.child({ module: "telemetry" }),
+      sessionId: cliSessionId,
+      globalConfigAccessor,
+    });
+
+    const commandRunTelemetryRecorder = new TelemetryAttributesRecorder("cli.command_run", {
```
Hweinstock, line 52: this is what we can pass down the context (will wire up in a future PR).

src/telemetry/client.tsx
```diff
@@ -0,0 +1,115 @@
...
+          .warn(`failed to shutdown metric sink with name '${sink.name}'`);
+      });
+    });
+    await Promise.all(promises);
+  }
+
+  private getMetricSinks(): MetricSink[] {
+    if (this.metricSinks !== undefined) return this.metricSinks;
+
+    this.metricSinks = [new LoggingSink({ logger: this.logger.child({ module: "loggingSink" }) })];
```
aidandaly24, line 93: It seems the default Logging sink is created and invoked still without resolving the telemetry preference. The persisted-config case also records again from a second CLI process, so this is not limited to the command that changes the setting.

```bash
CONFIG_HOME="$(mktemp -d /tmp/pr1846-config-test.XXXXXX)"
ENV_HOME="$(mktemp -d /tmp/pr1846-env-test.XXXXXX)"

# Persist the opt-out. This currently records one telemetry entry.
HOME="$CONFIG_HOME" bun run src/index.ts config telemetry.enabled false
CONFIG_LOG="$(find "$CONFIG_HOME/.agentcore/logs" -name 'output-*.log' -print -quit)"
rg -c '"msg":"recording telemetry"' "$CONFIG_LOG"
# 1

# Start another process with the persisted opt-out. It records another entry.
HOME="$CONFIG_HOME" bun run src/index.ts config telemetry.enabled
rg -c '"msg":"recording telemetry"' "$CONFIG_LOG"
# 2

# The environment override is also ignored.
HOME="$ENV_HOME" AGENTCORE_TELEMETRY_DISABLED=1 bun run src/index.ts config telemetry.enabled
ENV_LOG="$(find "$ENV_HOME/.agentcore/logs" -name 'output-*.log' -print -quit)"
rg -c '"msg":"recording telemetry"' "$ENV_LOG"
# 1
```

Shouldn't the resolved default sink list be empty when telemetry is disabled?

src/telemetry/types.tsx
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Static definition of all metrics the CLI emits.
+ */
+export const METRICS = {
+  "cli.command_run": {
+    attributeSchema: commandRunSchema,
+    // value describes duration (ms) of the command
+    valueSchema: z.number().min(0),
+  },
+} satisfies Record<string, { attributeSchema: z.ZodType; valueSchema: z.ZodType }>;
```
aidandaly24, line 36: From what I could tell the existing implementation defines `safeSchema` in `src/cli/telemetry/schemas/common-shapes.ts` to prevent command attributes from containing free-form strings. For example, `InvokeAttrs` records `has_session_id: z.boolean()` rather than the session ID itself. This accepts any `z.ZodType`. I think the `safeSchema` was probably helpfu so that the privacy constraint remained enforced by the type system, what do you think?

src/telemetry/types.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { commandRunSchema } from "./shapes";
+
+/**
+ * The primary interface for telemetry that orchestrates the emitting of metrics
+ */
+export interface TelemetryClient {
+  emit<TMetricName extends MetricName>(
+    metricName: TMetricName,
+    metricValue: ValueOf<TMetricName>,
+    attributes: Partial<AttributesOf<TMetricName>>,
```
aidandaly24: nit: `emit` accepts partial attributes even though it immediately validates them against a schema with required fields.  I tested this with `{}` it passes typechecking, but runtime validation rejects it, the sink receives nothing, and the error is reduced to an internal warning. The recorder can accumulate partial attributes internally, the final `emit` contract should probably require a complete `AttributesOf<TMetricName>` or a validated complete object. What do you think?

src/telemetry/loggingSink.tsx
```diff
@@ -0,0 +1,31 @@
+import type { Logger } from "../logging";
+import type { MetricSink } from "./types";
+
+type LoggingSinkConfig = {
```
jariy17, line 4: compliment: I love how you decided to make a config object instead of just passing logger. This allows to add more parameters if we need to.

aidandaly24 (commented):
Overall this looks good to me and clean. A few comments, let me know what you think. I wasn't sure on the exact decisions regarding the telemetry preference and if we always want the telemetry LoggingSink.

src/telemetry/types.tsx
```diff
@@ -0,0 +1,52 @@
...
+  ): Promise<void>;
+  shutdown(): Promise<void>;
+}
+
+/**
+ * A destination to send metric data.
+ */
+export interface MetricSink {
+  /** Send data to the given metric sink **/
+  send(metricName: string, value: number, attributes: Record<string, string | number>): void;
```
jariy17, line 21: Why isn't MetricSink.send() not returning a promise like shutdown?

src/telemetry/types.tsx
```diff
@@ -0,0 +1,52 @@
...
+ */
+export const METRICS = {
+  "cli.command_run": {
+    attributeSchema: commandRunSchema,
+    // value describes duration (ms) of the command
+    valueSchema: z.number().min(0),
+  },
+} satisfies Record<string, { attributeSchema: z.ZodType; valueSchema: z.ZodType }>;
+
+export type MetricName = keyof typeof METRICS & string;
```
jariy17: Nit: Redundant check—METRICS already enforces that the key is a string.

src/telemetry/types.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * A destination to send metric data.
+ */
+export interface MetricSink {
+  /** Send data to the given metric sink **/
+  send(metricName: string, value: number, attributes: Record<string, string | number>): void;
+  /** Flush and close the given metric sink **/
+  shutdown(): Promise<void>;
+  name: string;
```
jariy17: Since this is a interface, lets make this a getter function.

src/telemetry/recorder.tsx
```diff
@@ -0,0 +1,27 @@
+import type { MetricName, AttributesOf } from "./types";
+
+/**
+ * A strongly typed recorder for accumulating metric attributes, bound to a specific metric's schema.
+ */
+export class TelemetryAttributesRecorder<TMetricName extends MetricName> {
+  private attributes: Partial<AttributesOf<TMetricName>>;
+
+  constructor(
+    _metricName: TMetricName,
```
jariy17: We never store this. It would be nice to store this in a private variable for future middleware use.

src/telemetry/client.tsx
Hweinstock: I left it on because I think we want users to be able to audit telemetry without enabling it sent to the collector. I.e. maybe i disabled it, but I'm curious what would be sent. 

I do think we should disable logging when its not being sent or auditted (since its misleading), but I'll address that when I add the next sink.

src/telemetry/types.tsx
Hweinstock: yeah good question, I was planning to add that back once I get into defining the metric schemas.

src/telemetry/types.tsx
Hweinstock: yeah I went back and forth on this. The challenge is the recorder doesn't know when its attributes are no longer partial. So making this non-partial would require callers of `getData` on the recorder to be validating the shape, which feels like the job of the telemetry module. 

However, I agree the type signatures are awkward. I'm inclined to move this around so that we explicitly `end` the recorder lifespan (at which point we validate its accumulated attributes), and get the validated data. Then the telemetry client isn't responsible for any validation, and trusts the type signature. 

This is kind of similar to what the built-in span support looks like in `otel`.

src/telemetry/types.tsx
Hweinstock: This is inherited from how otel does it with their sdk client where emitting/sending data is non-blocking, and batched for us in the background. Ideally emit would be non-async as well, but we need to read from globalConfig (on first invoke)

src/telemetry/loggingSink.tsx
aidandaly24: Agreed I really like that system

jariy17 (approved):

aidandaly24 (approved):

PR #1848, opened 2026-07-28
refactor(errors): migrate input validation errors to use the modeled exception. 

Hweinstock:
### Problem 
Follow-up to https://github.com/aws/agentcore-cli/pull/1834. 

We defined the common errors, but we aren't using the modeled exceptions anywhere. 

### Solution 
Migrate input validation errors to use the shared input validation error. 

### Notes
- there are other errors we can migrate as well to new modeled exceptions, but keep the PR scope minimal and focused. 

### Testing 
Ran the CLI to verify this error is now classified correctly and logged at the root handler:

```
{"level":"error","msg":"","time":1785252586823,"cliSessionId":"ef345869-50c2-41ec-b282-f7d5d4044ece","error":{"name":"InputValidationError","message":"Invalid value for argu
ment 'key': Invalid option: expected one of \"telemetry\"|\"telemetry.enabled\"|\"telemetry.audit\"|\"telemetry.endpoint\"|\"installationId\"","stack":"{...}","exitCode":1,"meta":{},"source":"user"}}
```


src/router/flags.tsx
```diff
@@ -62,8 +63,9 @@ function attributeName(name: string): string {
 function validateFlag(flag: Flag, opts: Record<string, unknown>): unknown {
   const result = flag.schema.safeParse(coerce(flag.schema, opts[attributeName(flag.name)]));
   if (!result.success) {
-    throw new TypeError(
+    throw new InputValidationError(
```
aidandaly24, line 66: nit: Could we update the existing invalid-flag router test to assert `InputValidationError`, matching the new config assertions? It currently checks only the message and passed when this path threw `TypeError`, so the shared flag-classification change is otherwise untested.

src/handlers/identity/api-key-credential-provider/create/index.tsx
```diff
@@ -25,17 +26,21 @@ export const createCreateApiKeyCredentialProviderHandler = (core: Core, io: AppI
...
       }
+
       if (!hasApiKey && !hasSecretRef) {
-        throw new TypeError("either --api-key or --api-key-secret-reference is required");
+        throw new InputValidationError(
+          "either --api-key or --api-key-secret-reference is required",
+        );
       }
 
       const resolver = new SourceResolver({ stdin: io.stdin });
```
aidandaly24, line 46: One input-validation path appears to be missing from this migration the `SourceResolutionError` in `src/io/source.ts` still extends `TypeError`. Missing `file://` sources and invalid UTF-8 therefore continue to be classified as `internal`. Could it extend `InputValidationError`, with a regression test asserting `source: "user"`? I also get it if you think this is better as a separate PR. I think this may have been because I was working on the `SourceResolver` at the same time as you were creating the `AgentCore Error`.

aidandaly24 (changes requested):
Couple of comments, let me know what you think.

jariy17 (commented):
Pretty good but u missed: 
* accessor.tsx:80
* SourceResolutionError (io/source.ts:13)
* tui/index.tsx:46

src/router/flags.tsx
Hweinstock: yeah good call, I like asserting on type rather than message since it avoids coupling with the exact message.

Hweinstock:
> accessor.tsx:80
SourceResolutionError (io/source.ts:13)
tui/index.tsx:46

Migrated the source resolution, but the other two I think deserve their own errors. 

https://github.com/aws/agentcore-cli/blob/ccd402451b206fc0da61c1c4c575b62dc164f788/src/globalConfig/accessor.tsx#L76-L85 This error occurs when we write global config back out to the file and it fails zod validation. This should only happen if our zod schema is out of sync with the type in the signature, or we cast somewhere internally, so I don't think we want to mark it as a `user` error. 

I also think:
https://github.com/aws/agentcore-cli/blob/ccd402451b206fc0da61c1c4c575b62dc164f788/src/tui/index.tsx#L45-L47
might be better classified as an error to signify there is something wrong with the users environment rather than that they provided invalid input. However, I do think `user` error is right here. 

I can look into adding custom error types for these two as a follow-up.

jariy17 (approved):

aidandaly24 (approved):

PR #1856, opened 2026-07-29
feat: add OAuth2 provider support to identity

nborges-aws:
## Description
Adds command-line CRUDL for Identity OAuth2 credential providers:

- `identity oauth2-credential-provider create`
- `identity oauth2-credential-provider get`
- `identity oauth2-credential-provider list`
- `identity oauth2-credential-provider update`
- `identity oauth2-credential-provider delete`
- 
Supports two configuration paths:
- Guided CustomOauth2: `--client-id`, `--discovery-url`, and `--authorization-server-metadata` build the provider config from scalar flags. Only valid with `--vendor CustomOauth2`
- Provider configuration passthrough: `--provider-configuration <json>` accepts the complete SDK config union for any of the supported vendors. The CLI merges the resolved secret into the vendor config before sending

Secret handling follows the same pattern established with api-key-credential-provider:
- `--client-secret` is source-aware (inline, file://path, or - for stdin) and marked sensitive for log redaction
- `--client-secret-reference '{"secretId":"...","jsonKey":"..."}'` for external Secrets Manager secrets
- The two are mutually exclusive

This PR also generalizes `parseSecretReference` to a shared identity utility (src/handlers/identity/parser.tsx) parameterized by flag name.

## Related Issue

Closes #

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Manual tested cli by invoking all commands, both happy-path and negative validation. Expected results across the board. Additionally ran the full test suite, against latest HEAD after rebasing refactor branch:
- [x] bun test (402 pass, 0 fail)
- [x] bun run typecheck clean
- [x] bun run build

- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.


src/handlers/identity/oauth2-credential-provider/create/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+      flag("discovery-url", "OAuth2 discovery URL (guided Custom OAuth2)", z.string().optional()),
+      flag(
+        "authorization-server-metadata",
+        "authorization server metadata JSON (guided Custom OAuth2)",
+        z.string().optional(),
+      ),
+      flag(
+        "provider-configuration",
+        "complete OAuth2 provider configuration JSON (alternative to guided flags)",
+        z.string().optional(),
```
aidandaly24, line 47: I think `--provider-configuration` also needs to be marked sensitive in both create and update. This flag accepts the complete SDK config, which can include `clientSecret`. I tested that form and the plaintext value was written to `~/.agentcore/logs` while `--client-secret` was redacted. Marking both flag definitions sensitive should cover it.

src/handlers/identity/parser.tsx
```diff
@@ -0,0 +1,34 @@
+// Requires exactly { secretId, jsonKey } with non-empty string values.
+export function parseSecretReference(
+  flagName: string,
+  raw: string,
+): { secretId: string; jsonKey: string } {
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(raw);
+  } catch (error) {
+    throw new TypeError(
```
aidandaly24: This shared parser previously threw `InputValidationError`, but the move changes it to `TypeError`. The new OAuth handlers do the same. I reproduced a conflicting-secret invocation and the log classified it as `source: "internal"`. These user-input failures should remain `InputValidationError`.

src/handlers/identity/oauth2-credential-provider/create/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+          throw new TypeError(
+            "--provider-configuration must contain a single vendor config object",
+          );
+        }
+        const vendorConfig = config[configKey] as Record<string, unknown>;
+        vendorConfig.clientSecret = clientSecret;
+        vendorConfig.clientSecretConfig = clientSecretConfig;
+        vendorConfig.clientSecretSource = clientSecretSource;
+        oauth2ProviderConfigInput = config;
+      } else {
```
aidandaly24: The two configuration modes are not fully enforced here. Without `--provider-configuration`, this always builds `customOauth2ProviderConfig`, even for `--vendor GithubOauth2`, and guided Custom OAuth also allows the required discovery configuration to be omitted. I reproduced the resulting service validation error. We should require provider configuration for non-Custom vendors and one discovery form for the guided Custom path, including the equivalent update validation.

aidandaly24 (changes requested):

src/handlers/identity/oauth2-credential-provider/create/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+        "client-secret",
+        "the client secret (inline, file://path, or -)",
+        z.string().optional(),
+        {
+          sensitive: true,
+        },
+      ),
+      flag(
+        "client-secret-reference",
+        'external secret reference JSON: {"secretId":"<arn>","jsonKey":"<key>"}',
```
Hweinstock, line 34: does this need to be sensitive or is it strictly pointers to secrets?

src/handlers/identity/oauth2-credential-provider/create/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+      flag(
+        "provider-configuration",
+        "complete OAuth2 provider configuration JSON (alternative to guided flags)",
+        z.string().optional(),
+      ),
+      flag("tags", "tags as key=value (repeatable) or JSON object", z.array(z.string()).optional()),
+    ],
+    handle: async (ctx, flags) => {
+      if (!flags.name) {
+        throw new TypeError("required option '--name <name>' not specified");
```
Hweinstock: should we throw modeled input validation errors here? (same with below)

src/handlers/identity/oauth2-credential-provider/create/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+        throw new TypeError(
+          "guided flags (--client-id, --discovery-url, --authorization-server-metadata) are only valid with --vendor CustomOauth2; use --provider-configuration for other vendors",
+        );
+      }
+      if (hasProviderConfig && hasGuidedFlags) {
+        throw new TypeError(
+          "--provider-configuration and guided flags (--client-id, --discovery-url, --authorization-server-metadata) are mutually exclusive",
+        );
+      }
+      if (flags["discovery-url"] && flags["authorization-server-metadata"]) {
```
Hweinstock: same question for errors here.

src/handlers/identity/parser.tsx
```diff
@@ -0,0 +1,34 @@
...
+      `Invalid JSON for --${flagName}: ${error instanceof Error ? error.message : String(error)}`,
+    );
+  }
+
+  if (typeof parsed !== "object" || parsed === null || Array.isArray(parsed)) {
+    throw new TypeError(`--${flagName} must be a JSON object with "secretId" and "jsonKey"`);
+  }
+
+  const obj = parsed as Record<string, unknown>;
```
Hweinstock: could we simplify this with a strict zod schema? Something like:
```
 const secretReferenceSchema = z
    .object({
      secretId: z.string().min(1),
      jsonKey: z.string().min(1),
    })
    .strict(); 
```

src/handlers/identity/oauth2-credential-provider/create/index.tsx
nborges-aws: Yep, I went through and updated all instances of TypeError to the correct modeled error.

src/handlers/identity/oauth2-credential-provider/create/index.tsx
nborges-aws: Same as above handling of errors

src/handlers/identity/oauth2-credential-provider/create/index.tsx
nborges-aws: This is just a pointer to keys, so we're okay to leave it non-sensitive

src/handlers/identity/oauth2-credential-provider/create/index.tsx
nborges-aws: Good call. I updated our validations so we we require `--provider-configuration` for non-custom vendors

src/handlers/identity/oauth2-credential-provider/create/index.tsx
nborges-aws: Nice catch. I've updated to mark as sensitive in both spots

aidandaly24 (dismissed):
Thanks for fixing up the comments. LGTM!

src/handlers/identity/oauth2-credential-provider/update/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+        oauth2ProviderConfigInput = {
+          customOauth2ProviderConfig: {
+            clientId: flags["client-id"],
+            clientSecret,
+            clientSecretConfig,
+            clientSecretSource,
+            ...(Object.keys(oauthDiscovery).length > 0 && { oauthDiscovery }),
+          },
+        };
+      }
```
notgitika: it seems like this is dropping settings like OBO token exchange and auth method. it can also break providers using private endpoints. can we preserve the existing config somehow?

AlexanderRichey (commented):
This code looks good to me. Will hold off on approving until @notgitika's comment is addressed.

src/handlers/identity/oauth2-credential-provider/update/index.tsx
nborges-aws: Addressed in latest commit. Any guided update will overlay the updated settings atop the existing config

notgitika (approved):
Looks good to me!

aidandaly24 (approved):

PR #1858, opened 2026-07-29
feat(telemetry): enable audit mode for telemetry

Hweinstock:
### Problem 
The current CLI does not support `audit` mode. Audit mode allows customers to inspect the telemetry sent without enabling telemetry. 

### Solution
- implement a `MetricSink` that continually appends data to a file. 
- audit mode relies on a global config flag that defaults to false. 

### Tests and Verification
- updated unit tests to cover both sinks and verify that audit respects customer's settings. 

```
> rm -rf ~/.agentcore/telemetry
> agentcore config telemetry.audit true
true 
> cat ~/.agentcore/telemetry/* | jq
{
  "value": 29,
  "attrs": {
    "service.version": "0.0.0",
    "agentcore-cli.session_id": "61670d19-1359-46ab-98f9-226db0638353",
    "os.type": "Linux",
    "host.arch": "x64",
    "node.version": "v24.3.0",
    "exit_reason": "success"
  }
}

```


src/telemetry/client.tsx
```diff
@@ -87,10 +93,19 @@ export class DefaultTelemetryClient implements TelemetryClient {
     await Promise.all(promises);
   }
 
-  private getMetricSinks(): MetricSink[] {
+  private async getMetricSinks(): Promise<MetricSink[]> {
     if (this.metricSinks !== undefined) return this.metricSinks;
+    this.metricSinks = [];
```
notgitika: concurrent calls can see the empty sink list while config loads and drop metrics. I was able to reproduce this locally with 2 concurrent ones recording a single value. 

maybe we can cache and await for the sink initialization promise to prevent this?

notgitika (commented):
just one issue otherwise LGTM!

src/telemetry/client.tsx
Hweinstock: good catch! simplified the logic here to avoid this.

notgitika (approved):
Awesome!

aidandaly24 (approved):
This looks good to me. All makes sense and works well from my manual testing.

PR #1868, opened 2026-07-29
fix(errors): define new custom errors for failures. 

Hweinstock:
### Problem 
There are a few errors that are not yet migrated to the custom errors types. This PR does not aim to migrate all of them, but migrates a few that were raised in a previous PR. 

https://github.com/aws/agentcore-cli/pull/1848#issuecomment-5110918879

### Solution 
- define `InputValidationError` for when data passed internally fails validation in the config. 
- define `invalidEnvironmentError` for when tty is missing on stdin and stdout. 

### Testing
- updated tests to assert on error type instead of error message.

### DIFF
--- a/src/globalConfig/accessor.tsx
+++ b/src/globalConfig/accessor.tsx
@@ -8,7 +8,8 @@ import type { ReadWriteJson } from "../io";
 import type { Logger } from "../logging";
 import { globalConfigFileSchema } from "./types";
 import { DEFAULT_GLOBAL_CONFIG, applyOverrides } from "./config";
-import { AgentCoreCLIError, ERROR_SOURCE, type AgentCoreCLIErrorOptions } from "../errors";
+import z from "zod";
+import { InputValidationError } from "../errors";
 
 type DefaultGlobalConfigAccessorConfig = {
   logger: Logger;
@@ -76,7 +77,7 @@ export class DefaultGlobalConfigAccessor implements GlobalConfigAccessor {
   private async writeToConfigFile(data: GlobalConfigFileData): Promise<GlobalConfigFileData> {
     const dataParseResult = globalConfigFileSchema.safeParse(data);
     if (!dataParseResult.success) {
-      throw new InternalValidationError("failed to validate config data before writing to file", {
+      throw new InputValidationError(z.prettifyError(dataParseResult.error), {
         cause: dataParseResult.error,
       });
     }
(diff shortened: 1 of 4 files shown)


src/globalConfig/accessor.tsx
```diff
@@ -127,3 +128,10 @@ function diff<T extends Record<string, unknown>>(a: T, b: T): DeepPartial<T> {
 function isRecord(value: unknown): value is Record<string, unknown> {
   return typeof value === "object" && value !== null && !Array.isArray(value);
 }
+
+/** Error raised by invalid data coming from internal sources **/
+export class InternalValidationError extends AgentCoreCLIError {
```
Hweinstock: note: I intentionally defined these in the local files since they aren't used elsewhere. I think we can lift these up when we see an opportunity to generalize. I want to avoid lifting this up now, then trying to force these usages elsewhere.

aidandaly24: Makes sense to me, that's the only concern I could see from this PR.

aidandaly24 (dismissed):
Approved, I'd definitely extract the error to the common errors location when we see more use cases.

src/globalConfig/accessor.tsx
```diff
@@ -76,8 +76,9 @@ export class DefaultGlobalConfigAccessor implements GlobalConfigAccessor {
   private async writeToConfigFile(data: GlobalConfigFileData): Promise<GlobalConfigFileData> {
     const dataParseResult = globalConfigFileSchema.safeParse(data);
     if (!dataParseResult.success) {
-      // TODO: mark this as a client-source error.
-      throw new TypeError(z.prettifyError(dataParseResult.error));
+      throw new InternalValidationError("failed to validate config data before writing to file", {
+        cause: dataParseResult.error,
+      });
```
aidandaly24: One small nit comment, let me know what you think.

This path is reachable from CLI input, not only internal data. I ran `agentcore config telemetry '{"enabled":"nope"}'`. Then the config handler parses that object and casts it to `GlobalConfig`, then this logs `InternalValidationError` with `source: "internal"`. I feel like the handler should probably validate the updated config as user input and throw `InputValidationError`. Not a blocker or anything but just a thought.

jariy17 (dismissed):

src/globalConfig/accessor.tsx
Hweinstock: yeah thats a good point. I think this could be a user or internal error depending on how its reached, but we shouldn't make an internal error so easy to trigger. Also, I ran the same case you described and it produced a worse customer error message so let me actually swap this to InputValidationError. good catch!

aidandaly24 (approved):

jariy17 (approved):

PR #1872, opened 2026-07-30
feat(project): run npm install, uv sync, and git init after scaffolding

tejaskash:
## Summary

Stacked on #1809. After writing the project tree, `agentcore project create` now finishes the job so the scaffolded project is immediately usable:

1. `npm install` in `<project>/agentcore/cdk`
2. `uv sync` in `<project>/app/<agent>` (when the template ships a `pyproject.toml`)
3. `git init` at the project root

New flags: `--skip-install` (skips npm + uv) and `--skip-git`.

## Testing

- 516/516 `bun test` (7 new exec tests, 5 new manager tests, 2 new handler tests); manifest snapshot updated for the README; tsc + Prettier clean.
- Manual verification of all paths: full create (node_modules/.venv/.git present, npm/uv chatter in the log file, absent from terminal), each skip flag alone and combined, missing npm/git/uv (right error + exit 1, scaffold kept), failing `npm install` (captured output + retry hint), `--json` (stdout empty, progress on stderr), duplicate/reserved names, `--help`.

### DIFF
--- a/src/assets/templates/hello-world-python/README.md
+++ b/src/assets/templates/hello-world-python/README.md
@@ -27,6 +27,9 @@ curl -X POST http://localhost:8080/invocations \
   -d '{"prompt": "Hello!"}'
 ```
 
+<!-- TODO: replace the uv run + curl instructions with `agentcore dev` and
+     `agentcore invoke` once those commands are available. -->
+
 ## Build your agent
 
 Start in `main.py`:
--- a/src/assets/templates/hello-world-python/README.md
+++ b/src/assets/templates/hello-world-python/README.md
@@ -1,0 +1,57 @@
+# hello-world
+
+A minimal AgentCore Runtime agent built with the
+[Strands Agents SDK](https://strandsagents.com) — our recommended framework
+for building agents on AWS Bedrock AgentCore.
+
+## What's here
+
+- `main.py` — the agent. A `BedrockAgentCoreApp` wraps a Strands `Agent`;
+  the `@app.entrypoint` function receives each invocation payload and streams
+  the agent's response back to the caller.
+- `pyproject.toml` — Python dependencies, managed with
+  [uv](https://docs.astral.sh/uv/). `agentcore project create` has already run
+  `uv sync` for you (unless you passed `--skip-install`), so `.venv/` is ready.
+
+## Run it locally
+
+```bash
+uv run main.py
+```
+
+The app listens on http://localhost:8080. Invoke it:
+
+```bash
+curl -X POST http://localhost:8080/invocations \
+  -H "Content-Type: application/json" \
+  -d '{"prompt": "Hello!"}'
+```
+
+## Build your agent
+
+Start in `main.py`:
+
+- Change the `system_prompt` to shape your agent's behavior.
+- Give it tools — Strands ships ready-made ones and makes custom tools a
+  decorator away:
+
+  ```python
+  from strands import Agent, tool
+
+  @tool
+  def word_count(text: str) -> int:
+      """Count words in text."""
+      return len(text.split())
+
+  agent = Agent(system_prompt="You are a helpful assistant.", tools=[word_count])
+  ```
+
+- Add dependencies with `uv add <package>`.
+
+See the [Strands documentation](https://strandsagents.com/latest/documentation/docs/)
+for multi-agent patterns, MCP tools, and model configuration.
+
+## Deploy
+
+Deploy from the project root with the AgentCore CLI; the CDK app under
+`agentcore/cdk` provisions the Runtime that hosts this agent.
--- a/src/core/project/manager.test.ts
+++ b/src/core/project/manager.test.ts
@@ -35,7 +35,7 @@ function manager(): { manager: FsProjectManager; commands: { command: string[]; 
       runner: async (command, { cwd }) => {
         commands.push({ command, cwd });
       },
-      checkTool: () => {}, // CI hosts don't have uv installed
+      checkTool: async () => {}, // CI hosts don't have uv installed
     }),
     commands,
   };
@@ -132,7 +132,7 @@ describe("FsProjectManager.create", () => {
     await manager().manager.create({
       name: "example",
       template: PROJECT_TEMPLATES.HELLO_WORLD_PYTHON,
-      onProgress: (message) => messages.push(message),
+      onProgress: (event) => messages.push(event.message),
     });
 
     expect(messages).toEqual([
@@ -150,7 +150,7 @@ describe("FsProjectManager.create", () => {
       runner: async () => {
         throw new Error("npm exploded");
       },
-      checkTool: () => {},
+      checkTool: async () => {},
     });
 
     await expect(
@@ -163,11 +163,14 @@ describe("FsProjectManager.create", () => {
 
   test("refuses to create a project inside an existing project", async () => {
     const directory = await inTempDirectory();
-    await manager().create({ name: "root", template: PROJECT_TEMPLATES.HELLO_WORLD_PYTHON });
+    await manager().manager.create({
+      name: "root",
+      template: PROJECT_TEMPLATES.HELLO_WORLD_PYTHON,
+    });
 
     process.chdir(join(directory, "root"));
     await expect(
-      manager().create({ name: "child", template: PROJECT_TEMPLATES.HELLO_WORLD_PYTHON }),
+      manager().manager.create({ name: "child", template: PROJECT_TEMPLATES.HELLO_WORLD_PYTHON }),
     ).rejects.toBeInstanceOf(NestedProjectError);
   });
 });
--- a/src/core/project/manager.test.ts
+++ b/src/core/project/manager.test.ts
@@ -35,6 +35,7 @@ function manager(): { manager: FsProjectManager; commands: { command: string[]; 
       runner: async (command, { cwd }) => {
         commands.push({ command, cwd });
       },
+      checkTool: () => {}, // CI hosts don't have uv installed
     }),
     commands,
   };
@@ -149,6 +150,7 @@ describe("FsProjectManager.create", () => {
       runner: async () => {
         throw new Error("npm exploded");
       },
+      checkTool: () => {},
     });
 
     await expect(
--- a/src/core/project/manager.test.ts
+++ b/src/core/project/manager.test.ts
@@ -14,7 +14,9 @@ async function inTempDirectory(): Promise<string> {
   const directory = await mkdtemp(join(tmpdir(), "agentcore-manager-"));
   tempDirectories.push(directory);
   process.chdir(directory);
-  return directory;
+  // cwd is the realpath (macOS tmpdir lives behind a /var -> /private/var
+  // symlink), matching the paths the manager derives from process.cwd().
+  return process.cwd();
 }
 
 afterEach(async () => {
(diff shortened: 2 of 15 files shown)


src/assets/templates/hello-world-python/README.md
```diff
@@ -0,0 +1,57 @@
...
+uv run main.py
+```
+
+The app listens on http://localhost:8080. Invoke it:
+
+```bash
+curl -X POST http://localhost:8080/invocations \
+  -H "Content-Type: application/json" \
+  -d '{"prompt": "Hello!"}'
+```
```
tejaskash, line 28: We will update this when dev, and invoke are available

src/core/project/manager.tsx
```diff
@@ -6,13 +7,17 @@ import type {
...
+import { requireTool, runCommand, type CommandRunner } from "../../io";
 import { projectTree } from "./compose";
 import { defaultSource, type AssetSource } from "./source";
+import { TEMPLATES } from "./templates";
 import { writeTree } from "./tree";
 
 type ProjectManagerConfig = {
   logger: Logger;
   source?: AssetSource; // Bun executable or dist/assets depending on runtime
+  runner?: CommandRunner; // injectable so tests never spawn real processes
```
Hweinstock: could this get confusing with our own commands? I'm wondering if we should mirror the existing language of `child_process` to differentiate from our commands.

src/io/exec.ts
```diff
@@ -0,0 +1,86 @@
...
+      `'${rendered}' failed in ${cwd} (exit code ${exitCode ?? "unknown"}).\n\n` +
+        `${output.trim()}\n\n` +
+        `Fix the issue and run 'cd ${cwd} && ${rendered}' to retry.`,
+      { source: ERROR_SOURCE.USER, meta: { command, cwd, exitCode } },
+    );
+  }
+}
+
+/** Returns true if `tool` resolves to an executable on PATH. */
+export function toolOnPath(tool: string): boolean {
```
Hweinstock: would it be simpler and more reliable to try to run the tool? for example, something like: 
```
const gitIsAvailable = ableToRun('git', ['--version'])
```

src/handlers/project/create/index.ts
```diff
@@ -17,11 +19,22 @@ export const createCreateProjectHandler = (config: CreateProjectHandlerConfig) =
         "project template to scaffold from",
         z.enum(PROJECT_TEMPLATES).default(PROJECT_TEMPLATES.HELLO_WORLD_PYTHON),
       ),
+      flag(
+        "skip-install",
+        "skip installing dependencies (npm install, uv sync)",
+        z.boolean().default(false),
```
Hweinstock, line 25: Would putting defaults here make it impossible for the TUI to open on empty args?

src/handlers/project/create/index.ts
```diff
@@ -17,11 +19,22 @@ export const createCreateProjectHandler = (config: CreateProjectHandlerConfig) =
...
     ],
     handle: async (_ctx, flags) => {
-      await config.projectManager.create({
+      // Progress and success go to stderr, keeping stdout for machine output.
+      const project = await config.projectManager.create({
         name: flags["project-name"],
         template: flags["template"],
+        skipInstall: flags["skip-install"],
+        skipGit: flags["skip-git"],
+        onProgress: (message) => config.io.stderr.write(`${message}\n`),
```
Hweinstock: what do you think about using a structured event here instead of a string? I'm wondering if that could help us stay flexible to adding more info on these progress updates.

src/io/exec.ts
```diff
@@ -0,0 +1,86 @@
...
+  constructor(tool: string, installHint: string) {
+    super(`'${tool}' was not found on your PATH. ${installHint}`, {
+      source: ERROR_SOURCE.USER,
+      meta: { tool },
+    });
+  }
+}
+
+/** Error raised when a subprocess exits non-zero, carrying its captured output. */
+export class CommandFailedError extends AgentCoreCLIError {
```
Hweinstock: could this name be confusing with our own commands?

src/io/exec.ts
```diff
@@ -0,0 +1,86 @@
...
+  constructor(command: string[], cwd: string, exitCode: number | null, output: string) {
+    const rendered = command.join(" ");
+    super(
+      `'${rendered}' failed in ${cwd} (exit code ${exitCode ?? "unknown"}).\n\n` +
+        `${output.trim()}\n\n` +
+        `Fix the issue and run 'cd ${cwd} && ${rendered}' to retry.`,
+      { source: ERROR_SOURCE.USER, meta: { command, cwd, exitCode } },
```
Hweinstock, line 27: in what cases might we expect this to fail and why would it be a user fault?

src/core/project/manager.test.ts
```diff
@@ -66,7 +85,79 @@ describe("FsProjectManager.create", () => {
...
+
+    expect(messages).toEqual([
+      "Scaffolding project files...",
+      "Installing CDK dependencies (npm install)...",
+      "Syncing Python dependencies (uv sync)...",
+      "Initializing git repository...",
+    ]);
+  });
+
+  test("a failed step propagates and leaves the scaffolded files in place", async () => {
```
Hweinstock, line 146: is this the behavior we want? If I create a project and it fails to install, what is the customer supposed to do? not actually sure here.

tejaskash: This is the same behavior we have right now, if isntall fails we warn the user that they need to run npm i / uv sync on their own. We treat it as non-blocking

src/core/project/manager.tsx
tejaskash: Agreed — renamed across the board to mirror child_process vocabulary: `CommandRunner` → `ProcessRunner`, `runCommand` → `runProcess`, `CommandFailedError` → `ProcessFailedError`. (1516d075)

src/io/exec.ts
tejaskash: Done — replaced the PATH/PATHEXT probe with `toolAvailable(tool)`, which spawns `<tool> --version` and checks exit 0. Simpler and it verifies the tool actually runs rather than merely exists on disk. `requireTool` is now async accordingly. (1516d075)

src/io/exec.ts
tejaskash: Same rename covers this: `ProcessFailedError` now, no overlap with our command tree vocabulary. (1516d075)

src/handlers/project/create/index.ts
tejaskash: Good call — `onProgress` now takes a structured `CreateProgressEvent` (`{ message: string }` for now, defined next to `CreateProjectInput` in handlers/project/types.ts). Adding step ids / start-done-fail status later is additive rather than a signature break, and a TUI screen can consume the same events. (1516d075)

src/io/exec.ts
tejaskash: Two cases: 
1. the tool is missing mid-run which means the preflight passed and then the binary disappeared/PATH change (rare) 
2. The subprocess itself exits non-zero: npm install hitting a registry/proxy/auth issue, uv failing to resolve deps, git config problems.

src/handlers/project/create/index.ts
tejaskash: Fixed the middleware to count only cli-sourced flags via Commander's getOptionValueSource, so defaults no longer block TUI-on-empty.

src/io/exec.ts
Hweinstock: this is a tough one. I definitely see cases where it could be either, however its difficult to know without having more info. I think throwing modeled errors should help us here, and we can revisit depending on what we see.

src/core/project/manager.test.ts
```diff
@@ -24,14 +26,28 @@ afterEach(async () => {
   );
 });
 
-function manager(): FsProjectManager {
-  return new FsProjectManager({ logger: createSilentLogger() });
+// A manager whose runner records commands instead of spawning them.
```
Hweinstock, line 29: is there a way we can bring uv/npm into the runners to run real processes here? I feel like that could help make the guarantees stronger.

src/io/exec.ts
```diff
@@ -0,0 +1,85 @@
...
+    );
+  }
+}
+
+/** Returns true if `tool --version` runs successfully — more reliable than probing PATH. */
+export function toolAvailable(tool: string): Promise<boolean> {
+  return new Promise((resolve) => {
+    const child = spawn(tool, ["--version"], { stdio: "ignore", shell: useShell });
```
Hweinstock: should we explicitly pass the args (ex. `--version`) down to support tools that don't have that syntax? `ssh` is one that comes to mind.

src/middleware/withTuiOnEmptyFlagsAndArgs.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const root = new Router("agentcore", "test root")
+    .groupFlags(JsonKey)
+    .use(withTuiOnEmptyFlagsAndArgs(new TestCoreClient(), testIO().io))
+    .handler(leaf);
+
+  return { ran: () => handled, routed: root.route(["node", "agentcore", "leaf", ...args]) };
+}
+
+describe("withTuiOnEmptyFlagsAndArgs", () => {
+  test("opens the TUI on a bare invocation even when every flag has a default", async () => {
```
Hweinstock, line 34: nit: could we make these tests more concise with a `test.each` pattern?

src/middleware/withTuiOnEmptyFlagsAndArgs.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { withTuiOnEmptyFlagsAndArgs } from "./withTuiOnEmptyFlagsAndArgs";
+import { Router, createHandler, flag } from "../router";
+import { JsonKey } from "../handlers/keys";
+import { TestCoreClient, testIO } from "../testing";
+
+// The middleware decides between the TUI and the wrapped handler from what the
+// user actually typed. These tests route a real command tree so Commander's
+// option-source tracking (cli vs default) is exercised end to end; the TUI
+// branch is observable as the renderTui TTY error (testIO is not a TTY), the
+// headless branch as the leaf handler running.
```
Hweinstock: nit: i find this comment more confusing than helpful

Hweinstock (dismissed):

src/core/project/manager.test.ts
tejaskash: There definitely is, we use it for integration tests and End-to-ends. I'm not sure running actual commands is the best fit for unit tests

src/io/exec.ts
tejaskash: Done — toolAvailable/requireTool now take probeArgs defaulting to ["--version"], so ssh-style tools pass ["-V"]. Current call sites (npm/uv/git) are unchanged. (06154ddd)

src/middleware/withTuiOnEmptyFlagsAndArgs.test.tsx
tejaskash: Done — the three headless-branch cases are one test.each now; the bare-invocation/TUI case stays separate since it asserts the opposite branch. (06154ddd)

src/middleware/withTuiOnEmptyFlagsAndArgs.test.tsx
tejaskash: Agreed, trimmed it to two lines that just say what's observable: TUI attempt throws under non-TTY testIO, headless path runs the leaf handler. (06154ddd)

src/core/project/manager.test.ts
tejaskash: I looked at this and pulled back: the CI runners don't have uv installed (that's what broke the first Linux/Windows run and why checkTool/runner are injectable), so real npm/uv in the manager tests would make the suite host-dependent again — and a real npm install per test adds network + ~30s per case. The layering we have: exec.test.ts runs the real runProcess against real processes (node scripts asserting exit codes and captured output), manager tests assert command/cwd/ordering against the recorded fake, and the real npm/uv/git path is covered by the e2e CodeBuild job + manual verification (node_modules/.venv/.git all confirmed). If we want stronger guarantees I'd add it to the e2e suite where the toolchain is controlled, rather than the unit suite — happy to do that as a follow-up.

src/io/exec.ts
tejaskash: Agreed — ProcessFailedError carries command/cwd/exitCode in meta, so telemetry can slice by tool and exit code and we can reclassify once we see real distributions.

src/core/project/manager.test.ts
Hweinstock: makes sense, and I think I tend to agree we should keep a unit tests fast. Definitely think this is a high value e2e test when we get there.

Hweinstock (approved):

notgitika (approved):
This looks good to me thanks!

PR #1874, opened 2026-07-30
fix: resolve the gateway target for A/B test invocation URLs (#1854)

### LINKED ISSUE #1854: agentcore run ab-test outputs incorrect invocation URL (uses runtime name instead of gateway target name)
### Description

When running agentcore run ab-test, the CLI outputs an invocationUrl that uses the runtime name (PascalCase, e.g.  CustomerSupportAB) as the path segment, but the gateway target is actually registered with a different name (kebab-case, e.g. customer-support-ab). This causes all requests to the printed URL to fail with "No Target found for Target name:  CustomerSupportAB".

### Steps to Reproduce

1. Create an agent runtime named CustomerSupportAB with a gateway target
  2. Run agentcore run ab-test --mode config-bundle --name cs_prompt_abtest --gateway my-gateway-secure --runtime 
  CustomerSupportAB --control-bundle customerSupportControl --control-version LATEST --treatment-bundle 
  customerSupportTreatment --treatment-version LATEST --online-eval ABQualityMonitor --control-weight 80 --treatment-weight 20
  3. Observe the printed invocation URL
  4. Use the printed URL with curl to send traffic
  5. Also try: agentcore view ab-test <id> --json | jq -r '.invocationUrl'

### Expected Behavior

 The invocation URL should use the actual gateway target name. Based on agentcore status --json, the target is registered as  customer-support-ab, so the correct URL is:

  https://<gateway-id>.gateway.bedrock-agentcore.us-west-2.amazonaws.com/customer-support-ab/invocations

### Actual Behavior

The CLI prints and stores in .invocationUrl:

  https://<gateway-id>.gateway.bedrock-agentcore.us-west-2.amazonaws.com/CustomerSupportAB/invocations

  Requests to this URL fail:
  {"success":false,"error":"No Target found for Target name: CustomerSupportAB"}

  The correct target name (customer-support-ab) is visible in agentcore status --json under
  .deployedState.targets.default.resources.gateways["my-gateway-secure"].targets but the AB test command does not use it.

### CLI Version

_No response_

### Operating System

Linux

### Additional Context

Workaround: build the URL manually from agentcore status --json using the gateway URL + the actual target key name:

  GW_BASE=$(agentcore status --json | jq -r '.deployedState.targets.default.resources.gateways."my-gateway-secure".gatewayUrl')
  GATEWAY_URL="${GW_BASE}/customer-support-ab/invocations"

  This confirms the A/B test infrastructure itself works correctly — only the URL reported by the CLI is wrong. The mismatch
  appears to be that agentcore run ab-test substitutes the --runtime argument value directly into the URL path, rather than
  looking up the corresponding gateway target name.

jariy17:
## Problem

Closes #1854.

`agentcore run ab-test` in config-bundle mode printed an invocation URL built from the **runtime** name, but a gateway invocation path segment must be a **gateway target** name. Whenever the two differed, every request to the printed URL failed:

```json
{"success":false,"error":"No Target found for Target name: CustomerSupportAB"}
```

The same wrong URL was stored in `.invocationUrl` for `agentcore view ab-test <id> --json`, so scripts inherited it too.

The cause is one line in `getInvocationUrl`:

```ts
return record.agent ? `${baseUrl}/${record.agent}/invocations` : undefined;
```

`record.agent` comes from `opts.runtime ?? opts.name` (`handler.ts`), i.e. the runtime. It only happened to work when a gateway target was named identically to its runtime.

## Why not just look the target up

A config-bundle test has no target of its own to substitute:

- Its variants carry `configurationBundle` ARNs, and the schema forbids `target` on them (`target: z.never().optional()`).
- It attaches to the **whole gateway** — the service splits traffic with a gateway rule keyed on the bundles, which is why the execution role is granted `CreateGatewayRule`/`UpdateGatewayRule`.
- Confirmed against the service: every config-bundle test in a live account stores only `configurationBundle` per variant, and `e2e-tests/ab-test-config-bundle.test.ts` creates one on a gateway with **no targets at all** — it reaches RUNNING fine.

So the path segment is whichever gateway target the caller chooses to invoke. The CLI cannot know it, and a gateway may have several targets fronting one runtime (a canary beside prod), where each is equally valid.

## Change

Report the gateway base URL and name what to append, instead of guessing a path that 404s:

```
Gateway URL: https://<gatewayId>.gateway.bedrock-agentcore.<region>.amazonaws.com
  → append /<gateway-target>/invocations (see `agentcore status --json`)
```

**Target-based tests are unchanged** — their variants *are* targets, so `variants[0].targetName` is a real path and they keep a complete `Invocation URL`.

`getInvocationUrl` stays synchronous and pure, so all three call sites (`view --json`, `printABTestDetail`, the TUI detail view) pick this up with no plumbing changes.

Also corrects two comments that documented the bug as intended behaviour:

- `docs/ab-tests.md` — "config-bundle uses the agent name"
- `RunABTestFlow.tsx` — claimed targets deploy as `${project}-${target}`; the L3 CDK deploys each target under its spec name verbatim (`Gateway.ts`), which is what makes the spec name the URL path segment.

## `--json` shape change

For **config-bundle** tests only, `invocationUrl` is replaced by `gatewayUrl` + `invocationUrlHint`:

```jsonc
// before — a URL that 404s
{ "invocationUrl": "https://<gw>.../CustomerSupportAB/invocations" }

// after
{ "gatewayUrl": "https://<gw>...",
  "invocationUrlHint": "append /<gateway-target>/invocations (see `agentcore status --json`)" }
```

This is deliberate: a script reading `.invocationUrl` now gets nothing rather than a URL that silently fails. Target-based output is untouched. Both docs pages are updated.

## Testing

- 419 test files / 6005 tests pass; `tsc` clean; eslint/prettier/secretlint clean.
- 8 tests on the changed code, including a regression asserting the URL never contains the runtime name, and that the config-bundle detail view prints no `Invocation URL:` line.
- Verified end-to-end through a bundled + globally installed build (`npm run bundle` → `npm i -g`) for both modes, in `--json` and in the TUI (direct `view ab-test <id>` and via the interactive list). Config-bundle detail view:

```
Gateway: arn:aws:bedrock-agentcore:us-west-2:...:gateway/my-gateway-secure-abc123
Gateway filter: none
Gateway URL: https://my-gateway-secure-abc123.gateway.bedrock-agentcore.us-west-2.amazonaws.com
  → append /<gateway-target>/invocations (see `agentcore status --json`)
```

## Notes for reviewers

Two things I found while investigating but did **not** change, happy to file separately:

- `docs/ab-tests.md` documents `--max-duration-days` and `--traffic-header`, but neither flag is registered in `run/command.tsx` and `CreateABTestOptions` has no field for either.
- `--gateway-filter` scopes a config-bundle test to a target path (live tests do carry `{"targetPaths":["/orders/*"]}`). A future refinement could use a literal filter path as the printed URL's segment, but glob patterns make that ambiguous so I left it out.

### DIFF
--- a/src/cli/operations/jobs/ab-test/format.ts
+++ b/src/cli/operations/jobs/ab-test/format.ts
@@ -4,9 +4,16 @@ import { formatJobDate } from '../shared/format';
 import type { ABTestJobRecord } from '../shared/types';
 
 /**
- * Derive the gateway invocation URL from the stored gateway ARN.
- * Target-based: `https://{gateway}/{control-target-name}/invocations`.
- * Config-bundle: `https://{gateway}/{agent-name}/invocations`.
+ * Derive the URL to send test traffic to, from the stored gateway ARN.
+ *
+ * Target-based: a full invocation URL, `https://{gateway}/{control-target-name}/invocations`.
+ *
+ * Config-bundle: the gateway base URL only. These tests attach to the whole gateway — the variants
+ * are configuration bundles, and the service splits traffic with a gateway rule — so the path segment
+ * is whichever gateway target the caller invokes, which the CLI cannot know. The path segment must be
+ * a gateway TARGET name; substituting the runtime name (`record.agent`) produced URLs that failed with
+ * "No Target found for Target name: <runtime>" whenever a target was not named identically to its
+ * runtime (issue #1854). Callers append the target path themselves.
  */
 export function getInvocationUrl(record: ABTestJobRecord): string | undefined {
   const parts = record.gatewayArn.split(':');
@@ -18,8 +25,16 @@ export function getInvocationUrl(record: ABTestJobRecord): string | undefined {
     const targetName = record.variants[0]?.targetName;
     return targetName ? `${baseUrl}/${targetName}/invocations` : undefined;
   }
-  return record.agent ? `${baseUrl}/${record.agent}/invocations` : undefined;
+  return baseUrl;
 }
+
+/** True when `getInvocationUrl` yields a gateway base URL that still needs a target path appended. */
+export function isGatewayBaseUrl(record: ABTestJobRecord): boolean {
+  return record.mode !== 'target-based';
+}
+
+/** Names what the caller must append to a gateway base URL to reach a variant. */
+export const INVOCATION_PATH_HINT = 'append /<gateway-target>/invocations (see `agentcore status --json`)';
 
 export function printABTestHistory(records: ABTestJobRecord[]): void {
   if (records.length === 0) {
@@ -47,7 +62,14 @@ export function printABTestDetail(record: ABTestJobRecord): void {
   console.log(`Gateway: ${record.gatewayArn}`);
   console.log(`Gateway filter: ${record.gatewayFilter?.targetPaths?.[0] ?? 'none'}`);
   const invocationUrl = getInvocationUrl(record);
-  if (invocationUrl) console.log(`Invocation URL: ${invocationUrl}`);
+  if (invocationUrl) {
+    if (isGatewayBaseUrl(record)) {
+      console.log(`Gateway URL: ${invocationUrl}`);
+      console.log(`  → ${INVOCATION_PATH_HINT}`);
+    } else {
+      console.log(`Invocation URL: ${invocationUrl}`);
+    }
+  }
   console.log(`Started: ${formatJobDate(record.createdAt)}`);
   if (record.completedAt) console.log(`Stopped: ${formatJobDate(record.completedAt)}`);
   if (record.maxDurationExpiresAt) console.log(`Max duration expires: ${formatJobDate(record.maxDurationExpiresAt)}`);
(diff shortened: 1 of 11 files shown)


src/cli/operations/jobs/ab-test/format.ts
```diff
@@ -18,9 +25,17 @@ export function getInvocationUrl(record: ABTestJobRecord): string | undefined {
     const targetName = record.variants[0]?.targetName;
     return targetName ? `${baseUrl}/${targetName}/invocations` : undefined;
   }
-  return record.agent ? `${baseUrl}/${record.agent}/invocations` : undefined;
+  return baseUrl;
```
notgitika: This always returns only the gateway base URL. This breaks working same-name configurations and scripts consuming .invocationUrl, while leaving issue #1854’s workflow manual. Creation already loads gateway targets and runtime mappings, so it should emit a complete URL when one target uniquely matches, and expose candidates or a clear ambiguity when several match.

notgitika (dismissed):
left a comment otherwise LGTM

src/cli/operations/jobs/ab-test/format.ts
jariy17: I was thinking of this but I wanted to establish to the user that config bundles are applied to all gateway targets. However, I could go either with this.

notgitika (approved):

PR #1875, opened 2026-07-30
fix(project): refuse nested project creation and model embedded-asset error

tejaskash:
## Summary

Post-review follow-ups from #1809 

- **Modeled error for missing embedded assets** — `EmbeddedAssetSource.read` now throws `EmbeddedAssetNotFoundError` (extends `AgentCoreCLIError`, internal source, `assetPath` in meta) instead of a bare `Error`.
- **Refuse nested project creation** — `FsProjectManager.create` now walks up from the working directory looking for the `agentcore/agentcore.json` project marker and throws a modeled `NestedProjectError` (user source) if one is found, instead of scaffolding a project inside an existing one.

The "child lifted to root" behavior seen with `bun run start` is a `bun run` artifact — bun chdirs to the package root before running the script, so the child was created next to `root` rather than inside it. A compiled/linked binary would have nested the child inside `root`. Either way, nesting a project inside a project would confuse project resolution, so `create` now refuses it explicitly.

## Testing

- New test: creating a project from inside an existing project rejects with `NestedProjectError`.
- Updated test: missing embedded asset rejects with `EmbeddedAssetNotFoundError`.
- Full suite: 509 pass, 0 fail. `tsc --noEmit` and Prettier clean.
- Manual: `project create --project-name child` from inside a scaffolded `root` project prints `Error: cannot create a project inside an existing AgentCore project (found .../root/agentcore/agentcore.json)` and exits 1.

### DIFF
--- a/src/core/project/manager.test.ts
+++ b/src/core/project/manager.test.ts
@@ -2,7 +2,7 @@ import { afterEach, describe, expect, test } from "bun:test";
 import { mkdtemp, readdir, rm } from "node:fs/promises";
 import { join, relative } from "node:path";
 import { tmpdir } from "node:os";
-import { FsProjectManager } from "./manager";
+import { FsProjectManager, NestedProjectError } from "./manager";
 import { ProjectFileExistsError } from "./tree";
 import { PROJECT_TEMPLATES } from "../../handlers/project/types";
 import { createSilentLogger } from "../../testing";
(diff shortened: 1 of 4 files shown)


aidandaly24 (approved):
This looks good to me. I think it's a good choice to keep the Errors inside the files that they are used until we see more use cases for them across the package.

notgitika (approved):

PR #1876, opened 2026-07-30
fix(runtime): defer invoke required flag validation

aidandaly24:
## Summary

- declare Runtime invoke `--id` and `--payload` as optional at the Commander boundary
- validate omitted required values inside the handler while preserving usage exit codes
- verify bare `runtime invoke` reaches the existing TUI middleware and empty payloads remain valid

## Testing

- `bun test src/handlers/runtime/invoke` (54 pass)
- `bun test` (538 pass)
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- `bun run build`

### DIFF
--- a/src/handlers/runtime/invoke/index.tsx
+++ b/src/handlers/runtime/invoke/index.tsx
@@ -5,6 +5,7 @@ import type { AppIO } from "../../../io";
 import type { Core } from "../../types";
 import { coreOptsFromCtx } from "../../utils";
 import { JsonKey } from "../../keys";
+import { ExitCode } from "../../../runnable";
 import { RuntimeInvokeInterruptedError } from "./errors";
 import {
   normalizeRuntimeInvokeRequest,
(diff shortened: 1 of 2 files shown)


tejaskash (approved):
LGTM

src/handlers/runtime/invoke/index.tsx
```diff
@@ -49,6 +50,17 @@ export const createInvokeRuntimeHandler = (core: Core, io: AppIO) =>
       ),
     ],
     handle: async (ctx, flags) => {
+      if (flags.id === undefined) {
+        throw new InputValidationError("required option '--id <id>' not specified", {
+          exitCode: ExitCode.USAGE,
```
Hweinstock, line 55: I didn't realize that 2 was the standard for validation exceptions. Should we make this the default on the input validation error class? (could be a follow-up)

Hweinstock (approved):

PR #1877, opened 2026-07-30
feat(eval): add eval online-eval CLI commands

jariy17:
## What

Adds the imperative `agentcore eval online-eval` command surface (CLI only), following the conventions established in the `eval evaluator` review (#1822).

## Command structure

```
agentcore eval
└── online-eval                  # manage AgentCore online evaluation configs
    ├── create
    ├── get
    ├── list
    ├── update
    ├── pause                    # executionStatus -> DISABLED
    ├── resume                   # executionStatus -> ENABLED
    └── delete
```

### `eval online-eval create`
| Flag | Required | Notes |
|---|---|---|
| `--name` | ✓ | config name |
| `--agent` | ✓* | harness ID **or** runtime ID whose traffic to sample |
| `--endpoint` | | endpoint qualifier to scope to (default `DEFAULT`); only valid with `--agent` |
| `--data-source-config` | ✓* | JSON `DataSourceConfig` (source-aware); alternative to `--agent` |
| `--evaluator` | ✓ | evaluator ID(s) to apply |
| `--sampling-rate` | ✓ | percentage of sessions to sample (0.01–100) |
| `--session-timeout-minutes` | | minutes of inactivity before a session is complete (1–1440; service default 15) |
| `--filters` | | JSON `Filter[]` (source-aware) |
| `--role-arn` | | execution-role override; auto-provisioned when omitted |
| `--enable-on-create` | | `true \| false` (default `true`) |
| `--description` | | |

\* exactly one of `--agent` or `--data-source-config`.

### `eval online-eval update`
| Flag | Required | Notes |
|---|---|---|
| `--id` | ✓ | |
| `--sampling-rate` | | 0.01–100 |
| `--session-timeout-minutes` | | 1–1440 |
| `--filters` | | JSON `Filter[]` (source-aware) |
| `--evaluator` | | replaces the existing list |
| `--agent` | | repoint at a different harness/runtime |
| `--endpoint` / `--clear-endpoint` | | mutually exclusive; re-scope the agent the config was built from |
| `--data-source-config` | | JSON `DataSourceConfig` (source-aware); replaces the source outright |

The data source can be repointed three ways, in precedence order: `--data-source-config` replaces it outright, `--agent` re-derives it from a different agent, and `--endpoint`/`--clear-endpoint` alone re-scope the agent the config already used. `--agent`/`--data-source-config` and `--endpoint`/`--data-source-config` are mutually exclusive.

Fields left unset are preserved. `UpdateOnlineEvaluationConfig` does partial updates at the top level — an omitted field is untouched — but sending `rule` replaces the whole object, so the client merges the untouched sub-fields over the current config.

### `eval online-eval get` / `pause` / `resume` / `delete`
`--id` (required).

### `eval online-eval list`
`--max-results`, `--next-token` — server-side pagination only.

Source-aware values (`--data-source-config`, `--filters`) accept the value inline, `file://<path>`, or `-` for stdin; one flag per command may read stdin.

## Execution role

`CreateOnlineEvaluationConfig` requires a role the service can assume, and validates at create time that it can query the log groups it was pointed at. Core provisions `AgentCoreOnlineEval-<configName>` scoped to the resolved log groups unless `--role-arn` is passed, mirroring the harness execution-role pattern. The provisioning is idempotent: an existing role is reused and its inline policy refreshed. A name that would exceed IAM's 64-character cap is truncated with a hash suffix, since two configs sharing a truncated name would otherwise share one role.

Three parts of the policy are worth calling out, because each was derived from a live failure rather than the docs:

- spans live in `aws/spans`, not in the runtime's own log group, so both must be queryable;
- query access must be scoped to the runtime log-group **prefix** — a policy pinned to a single endpoint's log group is rejected as insufficient;
- evaluators encrypted with a customer managed key need `kms:Decrypt` on that key, which the service also validates at create time. Core resolves each referenced evaluator's key and scopes the statement to exactly those keys, omitting it entirely when nothing is encrypted.

The first two surface only as `ValidationException: The provided execution role does not have permissions to access the specified log groups`.

Moving the data source invalidates the role's scope, so `update` re-scopes a CLI-provisioned role when `--agent`/`--data-source-config`/`--endpoint` changes it. `--update-role false` opts out. A role named via `--role-arn` is never edited; both skip paths emit a warning naming the role and the log groups it needs access to.

That re-scope is sequenced so a failed update cannot strand a live config. The role is widened to the **union** of the old and new log groups *before* `UpdateOnlineEvaluationConfig`, then narrowed to the new set only after it succeeds. A superset role is valid for either config state, so if the update fails the config still points at the old runtime with a role that still covers it. A rollback-on-failure ordering was considered and rejected: IAM is eventually consistent, so the compensating call can itself fail and none of these take effect instantly, whereas a superset is never wrong for either state. If the final narrow fails, the role is left broader than necessary and a `narrow-failed` warning says so — a later successful update re-scopes it.

A freshly created role is not immediately assumable (IAM is eventually consistent) and the service rejects the create rather than retrying, so the create is retried on that specific error with a bounded backoff — only when the CLI provisioned the role.

### Known service limitation

KMS resolution reads `GetEvaluator.kmsKeyArn`, which the service currently reports only for roughly two minutes after an evaluator is created, then omits. Until that is fixed, creating a config that references an older encrypted evaluator needs an explicit `--role-arn`. The encrypted-evaluator fixture is hand-authored to represent the documented behavior; a `RECORD=1` run overwrites it with the live response and it must be restored afterwards.

## Testing

- `bun test` — full suite green (566 tests).
- Handler tests follow the fixture/golden pattern (real `CoreClient` + recorded SDK seam), not `TestCoreClient`. Recorded against a live account with `RECORD=1`; the recording provisions the execution role, creates one config, exercises get/list/update/pause/resume, then deletes the config, so it leaves no config residue. Both record and replay modes pass.
- Policy assertions live in `src/core/onlineEvalExecutionRole.test.ts` rather than the fixture layer: recorded IAM responses are empty, so the policy body is not observable through a golden. Those tests fail if the KMS statement is dropped, if query access is scoped to an endpoint log group instead of the runtime prefix, or if role names are blindly truncated.
- The transactional ordering was verified against the live API: a forced update failure leaves the config on the old runtime with the role still covering it (union), and a successful `--agent` repoint narrows the role to the new runtime only.
- Exercised by hand against the live API beyond the recorded flow: `file://` and stdin sources, the one-stdin-per-command guard, all three `update` data-source paths (`--agent`, `--data-source-config`, `--endpoint`) applied in sequence to one config, and the client-side rejection of `--endpoint` on a config built from custom log groups.

---

TUI flows for these commands will come in a later PR.


src/core/onlineEvalExecutionRole.tsx
```diff
@@ -0,0 +1,172 @@
...
+// group(s) being sampled. Idempotent: an existing role is reused and its
+// inline policy refreshed.
+
+const POLICY_NAME = "AgentCoreOnlineEvalExecutionPolicy";
+
+// executionRoleName derives the default role's name from the online eval config
+// name. IAM role names cap at 64 characters.
+export function onlineEvalExecutionRoleName(configName: string): string {
+  return `AgentCoreOnlineEval-${configName}`.slice(0, 64);
+}
```
notgitika, line 45: when you slice 64 chars out of this, it also includes the chars in `AgentCoreOnlineEval-` therefore keeps only the first 44 config-name characters.

can we just use a hash suffix to preserve uniqueness?

src/core/eval.tsx
```diff
@@ -153,4 +176,281 @@ export class EvalClient implements CoreEvalClient {
...
+    let dataSourceConfig = current.dataSourceConfig;
+    if (update.dataSourceConfig !== undefined) {
+      dataSourceConfig = update.dataSourceConfig;
+    } else if (update.agent !== undefined) {
+      dataSourceConfig = await agentDataSource(
+        update.agent,
+        update.clearEndpoint ? DEFAULT_ENDPOINT_QUALIFIER : update.endpoint,
+        this.clients,
+        options,
+      );
```
notgitika, line 281: this updates the log groups but leaves the role scoped to the old runtime. can we refresh the managed policy, and support replacing custom roles here?

src/core/onlineEvalExecutionRole.tsx
```diff
@@ -0,0 +1,172 @@
...
+        // online evaluation must be able to invoke them.
+        Sid: "InvokeCodeBasedEvaluators",
+        Effect: "Allow",
+        Action: ["lambda:GetFunction", "lambda:InvokeFunction"],
+        Resource: `arn:aws:lambda:${region}:${accountId}:function:*`,
+      },
+    ],
+  });
+}
+
```
notgitika, line 170: no KMS permissions?  `kms:Decrypt`
is this going to be a follow up? I would suggest that we resolve evaluator KMS keys and grant access, or require `--role-arn` for that case.

src/core/eval.tsx
```diff
@@ -153,4 +176,281 @@ export class EvalClient implements CoreEvalClient {
...
+        : input.dataSourceConfig;
+    const control = this.clients.control(toClientConfig(options));
+
+    // The service validates at create time that the role can query the log groups
+    // it was pointed at, and the required policy is not obvious, so provision a
+    // default role scoped to them unless the caller brought their own.
+    const evaluationExecutionRoleArn =
+      input.evaluationExecutionRoleArn ??
+      (await ensureDefaultOnlineEvalExecutionRole(
+        this.clients.iam(toClientConfig(options)),
```
notgitika: https://github.com/aws/agentcore-cli/blob/refactor/src/core/harness.tsx#L123

instead of `toClientConfig` can we do what harness does here to prevent the endpoint override leak? just pass `{ region: options.region }`

notgitika (changes requested):

src/core/eval.tsx
jariy17: I went back and forth on this — I didn't want to modify role permissions without the customer's knowledge, but I'll add a flag to disable updating for auto-provisioned role permissions.

notgitika: Honestly yea that makes sense to me

src/core/onlineEvalExecutionRole.tsx
jariy17: Sure thing, I'll update with a hash suffix

src/core/onlineEvalExecutionRole.tsx
jariy17: Fixing that now.

src/core/eval.tsx
jariy17: yep, i'll fix it now

notgitika (dismissed):

notgitika (dismissed):

src/core/eval.tsx
```diff
@@ -153,4 +181,354 @@ export class EvalClient implements CoreEvalClient {
...
+      } else if (roleArn) {
+        roleScopeWarning = {
+          reason: managed ? "update-declined" : "custom-role",
+          roleArn,
+          logGroupNames,
+        };
+      }
+    }
+
+    const response = await control.send(
```
nborges-aws, line 399: If this update command fails, could the role-policy update above leave the config in a broken state? I'm thinking of a case where the role is scoped to the new runtime's logs, and then the eval config update fails. Meaning the config still points to the old runtime. Do you think we need to make this transactional in some way, maybe only add to the scope before update, and then restrict it down after update succeeds?

jariy17: Ah yes, nice catch!. I'll fix it now by making it transactional.

notgitika (approved):

nborges-aws (approved):

PR #1878, opened 2026-07-30
feat(memory): add read-only Memory TUI

notgitika:
This PR asds read-only TUI screens for listing and inspecting AgentCore Memories.

Manually tested all flows. 

This reuses the shared DataTable through PaginatedTablePicker, with https://github.com/aws/agentcore-cli/pull/1845 ’s column configuration

### DIFF
--- a/src/handlers/memory/get/screen.tsx
+++ b/src/handlers/memory/get/screen.tsx
@@ -1,0 +1,103 @@
+import { useQuery } from "@tanstack/react-query";
+import { Box, Text, useInput } from "ink";
+import { useNavigate, useParams } from "react-router";
+import { JsonDetail } from "../../../components/JsonDetail";
+import { KeyValueTable } from "../../../components/KeyValueTable.js";
+import { Layout } from "../../../components/Layout";
+import { darkTheme } from "../../../components/ui/_core.js";
+import { Divider } from "../../../components/ui/divider/Divider.js";
+import { Spinner } from "../../../components/ui/spinner";
+import type { ScreenProps } from "../../types";
+import { coreOptsFromCtx } from "../../utils";
+
+function useMemoryDetail({ ctx, core }: ScreenProps, memoryId: string | undefined) {
+  const opts = coreOptsFromCtx(ctx);
+  return useQuery({
+    queryKey: ["memory", opts.region, memoryId, "full"],
+    queryFn: () => core.memory.getMemory(memoryId!, "full", opts),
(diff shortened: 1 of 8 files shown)


notgitika:
Unrelated but I will look into this failure for AI reviewer separately. It now throws a new error. https://github.com/aws/agentcore-cli/actions/runs/30582321369/job/91005460337?pr=1878 

>ERROR: Failed to invoke harness: An error occurred (AccessDeniedException) when calling the InvokeHarness operation: User: arn:aws:sts::685197708687:assumed-role/GitHubActions-HarnessReview/GitHubActions is not authorized to perform: bedrock-agentcore:InvokeHarness on resource: *** because no identity-based policy allows the bedrock-agentcore:InvokeHarness action

notgitika:
Fixed the reviewer ^

aidandaly24 (approved):
I have no comments, approved.

src/handlers/memory/get/screen.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const { memoryId } = useParams();
+  const detail = useMemoryDetail(props, memoryId);
+  const memory = detail.data?.memory;
+
+  useInput((input, key) => {
+    if (key.escape) {
+      navigate(-1);
+      return;
+    }
+    if (input === "r" && detail.isError) {
```
Hweinstock, line 33: q: do we want to allow the user to refresh even if there isn't an error? (ex. maybe something changed and they want to check). 

This looks consistent with https://github.com/aws/agentcore-cli/blob/refactor/src/handlers/runtime/get/screen.tsx#L48 so just curious what behavior we want.

src/handlers/memory/get/screen.tsx
```diff
@@ -0,0 +1,103 @@
+import { useQuery } from "@tanstack/react-query";
+import { Box, Text, useInput } from "ink";
+import { useNavigate, useParams } from "react-router";
+import { JsonDetail } from "../../../components/JsonDetail";
```
Hweinstock, line 4: love to see us using so many common components!

Hweinstock (approved):

src/handlers/memory/get/screen.tsx
notgitika: nice catch I think we would want that right? maybe I created a new resource while the TUI screen was open and I just want to refresh it to see if it shows up. 

Let me add that quickly in a follow up in both places

PR #1881, opened 2026-07-31
feat: wire up a global constants file with package version

Hweinstock:
### Problem 
The package version is not wired up into telemetry or logging. 

### Solution 
- define a constants.ts at the top level, import the package.json and extract the version. 

### Notes
originally thought this wouldn't work on bun executable, but it does! It appears to pull the file in to the executable. 

### Testing 
Ran 3 commands, one from a node build `bun run build`, one from dev `bun run start`, and one from exec `./dist/bin/agentcore-linux-x64 `. All three produced logs with the correct version number.


tejaskash (dismissed):
LGTM

notgitika (dismissed):

notgitika (approved):

tejaskash (approved):

PR #1883, opened 2026-07-31
feat(dev): CodeZip dev runner and process lifecycle core

tejaskash:
## Summary
- stream long-running subprocess stdout and stderr as typed async-generator events from the `io` boundary
- cancel process groups through `AbortSignal`, with graceful termination, hard-kill escalation, and descendant cleanup
- simplify `CodeZipDevRunner` to yield status and process events without callbacks or lifecycle handles
- run Python environments through `uv run`; use uvicorn for HTTP and direct Python execution for MCP, A2A, and AGUI
- run TypeScript projects through `npm exec -- tsx watch`, installing dependencies only when `node_modules` is absent
- remove unused port discovery and public implementation-helper exports from this PR

## Verification
- `bun test` (981 tests)
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- `bun run build`


src/io/exec.ts
```diff
@@ -87,3 +105,151 @@ export const runProcess: ProcessRunner = ([executable, ...args], { cwd, onOutput
...
+    }
+  } finally {
+    options.signal?.removeEventListener("abort", terminate);
+    stdout.close();
+    stderr.close();
+    if (!closed) {
+      terminate();
+      await closedPromise;
+    }
+    if (terminating && processTreeAlive(child)) killTree(child, "SIGKILL");
```
aidandaly24: I was wondering if we should check the process group even when `terminating` is false. I reproduced a parent that spawned a long-running child, unreferenced it, and exited 0: `streamProcess()` returned successfully while the descendant remained alive. A wrapper or reloader exiting before its worker could therefore leave the development server holding its port. Would it make sense to clean up any surviving process group whenever the direct child closes and add a parent-exits-first regression test?

tejaskash: Fixed

aidandaly24 (approved):

jariy17 (approved):

PR #1886, opened 2026-07-31
feat: Implement TUI for identity

nborges-aws:
**Note: This PR is built upon and depends on #1856. Should not be merged until previous PR is, at which point diff will correctly render.**

## Description
Adds read-only TUI screens for inspecting Identity credential providers

- `identity api-key-credential-provider` — interactive menu, list picker, and get detail view
- `identity oauth2-credential-provider`— interactive menu, list picker, and get detail view
- Read operations feature interactive view in TUI
- Write operations (create/update/delete) are CLI only. TUI features `CliOnlyScreen`, which points user at the equivalent CLI command (e.g. This operation is available in the CLI: `agentcore identity api-key-credential-provider create --help`)
- Adds optional `maxPageSize` to the `PaginatedTablePicker`. Identity list API caps maxResults at 20, which caused issues when calling list/get from a tall terminal. The calculation was based on terminal height and could easily exceed 20, resulting in service exceptions. Optional cap allows identity to pass max value of 20, while other primitives are unchanged.
- ReadMe updates: adds Identity to the interactive-TUI notes, documents the read-only + CLI-only-mutations behavior, and backfills the OAuth2 credential provider command tree entries I missed in #1856 

## Related Issue

Closes #

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Manually tested the TUI with `bun run start` and interacting with the TUI screen. Also ran the full suite against latest HEAD after rebasing the refactor branch:

- [x] bun test src/ (590 pass, 0 fail)
- [x] bun run typecheck clean
- [x] bun run lint:check clean
- [x] bun run format:check clean
- [x] bun run build

- [x] I ran `npm run lint`

## Checklist

- [ ] Any dependent changes have been merged and published -- **Depends on #1856 which should be merged first**

---

terms of your choice.


src/handlers/identity/oauth2-credential-provider/get/screen.tsx
```diff
@@ -0,0 +1,111 @@
...
+      ]}
+    >
+      {detail.isPending ? (
+        <Spinner label="Loading OAuth2 credential provider…" />
+      ) : detail.isError ? (
+        <Text color="red">Error: {(detail.error as Error).message}</Text>
+      ) : (
+        <Box flexDirection="column">
+          <Box flexDirection="column" paddingLeft={1}>
+            <KeyValueTable
```
notgitika: non-blocking for this PR but can we show status and failureReason here? a failed provider otherwise might look healthy unless the user opens the raw JSON.

notgitika (dismissed):

src/components/CliOnlyScreen.tsx
```diff
@@ -0,0 +1,39 @@
...
+      breadcrumb={breadcrumb}
+      keyHints={[
+        { key: "esc", label: "back" },
+        { key: "ctl+c", label: "quit" },
+      ]}
+    >
+      <Box flexDirection="column" paddingLeft={1}>
+        <Text color={darkTheme.colors.text}>This operation is available via the CLI:</Text>
+        <Box paddingTop={1}>
+          <Text color={darkTheme.colors.focus}>{command}</Text>
```
AlexanderRichey: Can we read the actual help text from the Commander command? This should be available through the context object that's passed down. If that's required as a prop here, you can then get the help from the commander command. We do this in `createHelpDefault`.

src/handlers/identity/oauth2-credential-provider/get/screen.tsx
```diff
@@ -0,0 +1,111 @@
...
+                vendor: detail.data.credentialProviderVendor ?? "-",
+                secretSource: detail.data.clientSecretSource ?? "-",
+                clientSecretArn: detail.data.clientSecretArn?.secretArn ?? "-",
+                ...(detail.data.callbackUrl ? { callbackUrl: detail.data.callbackUrl } : {}),
+                createdAt: detail.data.createdTime?.toISOString() ?? "-",
+                updatedAt: detail.data.lastUpdatedTime?.toISOString() ?? "-",
+                arn: detail.data.credentialProviderArn ?? "",
+              }}
+            />
+          </Box>
```
AlexanderRichey: Any reason why we're not using the ResourceDetailScreen component here?

AlexanderRichey (changes requested):

src/handlers/identity/oauth2-credential-provider/get/screen.tsx
nborges-aws: Just a miss by me. Updated to use the right component

src/components/CliOnlyScreen.tsx
nborges-aws: This CliOnlyScreen I created is actually obsolete now, as a solution for the same problem was merged in a few days ago. So in my latest commit this entire file has been deleted

src/handlers/identity/oauth2-credential-provider/get/screen.tsx
nborges-aws: Good call. Both fields have been added to the details

AlexanderRichey (approved):

aidandaly24 (approved):

notgitika (approved):
Thanks for addressing the comments this LGTM now!

PR #1895, opened 2026-08-03
feat(memory): add read-only event and record commands

notgitika:
This PR adds `get` and `list` commands for memory events and records and tests for the same.
Core + handler implementation only, no-TUI so far.

bun run typecheck, format, lint and test passed (there was one un-related failure, not introduced by this commit. ran `bun test src/handlers/memory/memory.test.tsx` to confirm)
Smoke tested with my account.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,13 @@ agentcore                          # interactive TUI
 │       └── list                   # list a Runtime's endpoints
 ├── memory                         # inspect AgentCore Memories
 │   ├── get                        # fetch a Memory by id
-│   └── list                       # list Memories (server-side paginated)
+│   ├── list                       # list Memories (server-side paginated)
+│   ├── event
+│   │   ├── get                    # get an Event from a Memory session
+│   │   └── list                   # list Events from a Memory session
+│   └── record
+│       ├── get                    # get a long-term Memory record
+│       └── list                   # list long-term Memory records
 ├── gateway                        # inspect AgentCore Gateways
 │   ├── get                        # get a Gateway by id
 │   ├── list                       # list Gateways (server-side paginated)
--- a/README.md
+++ b/README.md
@@ -136,6 +136,10 @@ agentcore runtime endpoint list --id <runtimeId> --max-results 20
 agentcore memory get --id <memoryId>
 agentcore memory get --id <memoryId> --view without_decryption
 agentcore memory list --max-results 20
+agentcore memory event get --memory <memoryId> --actor-id <actorId> --session-id <sessionId> --event-id <eventId>
+agentcore memory event list --memory <memoryId> --actor-id <actorId> --session-id <sessionId> --max-results 20
+agentcore memory record get --memory <memoryId> --record-id <recordId>
+agentcore memory record list --memory <memoryId> --namespace <namespace> --max-results 20
 
 # Inspect Gateway resources without project configuration or deployment
 agentcore gateway get --id <gatewayId>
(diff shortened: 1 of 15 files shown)


src/core/memory.tsx
```diff
@@ -1,3 +1,17 @@
+import {
```
tejaskash, line 1: NIT - can this be a TS file since it doesn't contain any JSX

tejaskash (dismissed):
LGTM

src/core/memory.tsx
notgitika: ah yes, I was just following the convention in the other files. I can get a follow up PR that fixes it for everything

src/handlers/memory/event/get/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+import type { Core } from "../../../types";
+import { coreOptsFromCtx } from "../../../utils";
+import { JsonRendererKey } from "../../../../tui";
+
+export const createGetMemoryEventHandler = (core: Core) =>
+  createHandler({
+    name: "get",
+    description: "get an AgentCore Memory Event",
+    flags: [
+      flag("memory", "the ID of the Memory", z.string()),
```
aidandaly24: If you check out other primitives right now we use `.optional()` on everything. The general standard we've set in the CLI is to make all commands optional in Commander and then check for required flags separately and throw InputValidationError. The execution order we are looking for from my understanding is:
1. Commander parses and validates required options.
2. The router action runs.
3. Middleware runs.
4. The Runtime handler runs.
This is not meant to be forever, but until we figure out a solution for this. For example check out this PR: https://github.com/aws/agentcore-cli/pull/1876

src/handlers/memory/event/list/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+      flag("max-results", "maximum number of events to return; default 20", z.number().optional()),
+      flag("next-token", "pagination token returned by a previous request", z.string().optional()),
+    ],
+
+    handle: async (ctx, flags) => {
+      if (flags["include-parent-branches"] && !flags.branch) {
+        throw new InputValidationError("'--include-parent-branches' requires '--branch'");
+      }
+
+      const eventMetadata = parseJsonFlag<EventMetadataFilterExpression[]>(
```
aidandaly24: `parseJsonFlag<EventMetadataFilterExpression[]>` only casts the parsed JSON. I tested `--metadata-filters '{}'`, and it reached the SDK as `{"filter":{"eventMetadata":{}}}` instead of failing locally. The command contract says structured JSON should reject incorrect top-level types and invalid members, so event and record filters should use runtime schemas and test a non-array and malformed expression.

README.md
```diff
@@ -136,6 +136,10 @@ agentcore runtime endpoint list --id <runtimeId> --max-results 20
 agentcore memory get --id <memoryId>
 agentcore memory get --id <memoryId> --view without_decryption
 agentcore memory list --max-results 20
+agentcore memory event get --memory <memoryId> --actor-id <actorId> --session-id <sessionId> --event-id <eventId>
```
aidandaly24, line 145: The examples include these commands, but the command tree still shows Memory with only get and list. We should add the event and record groups there too.

aidandaly24 (changes requested):

src/handlers/memory/event/list/index.tsx
notgitika: nice catch, lemme fix it

src/handlers/memory/event/get/index.tsx
notgitika: completely forgot about this one! fixing it now

README.md
notgitika: interesting, I thought it did, let me check

aidandaly24: Would be lines 68-70 i believe

aidandaly24 (dismissed):
Thank you for making the updates this LGTM

aidandaly24 (dismissed):

src/handlers/memory/metadataFilters.ts
```diff
@@ -0,0 +1,75 @@
...
+        message: "right is required for EQUALS_TO",
+      });
+    }
+  });
+
+const memoryRecordMetadataValueSchema = z.union([
+  z.object({ stringValue: z.string() }).strict(),
+  z.object({ stringListValue: z.array(z.string()) }).strict(),
+  z.object({ numberValue: z.number() }).strict(),
+  z.object({ dateTimeValue: z.coerce.date() }).strict(),
```
nborges-aws: Can we replace this with something like:
```
z.object({
  dateTimeValue: z.iso.datetime().transform((value) => new Date(value)),
}).strict()
```

Otherwise, `z.coerce.date()` will accept values like null and zero and resolve them to dates instead of rejecting the input

notgitika: Fixed in 04738677. The schema now validates an ISO datetime with an explicit UTC designator or offset before transforming it to the SDK-required Date. Added coverage for valid conversion and local rejection of null, numeric, and malformed datetime values.

nborges-aws (approved):

aidandaly24 (approved):

PR #1896, opened 2026-08-03
fix: make paginated tables fit terminal dimensions

aidandaly24:
## Summary

- render the DataTable filter prompt in the existing divider row so filtering does not change table height
- increase the terminal-derived page size by one so full paginated tables reach the footer
- make Runtime Version and Harness Version tables fill the available terminal width
- left-truncate overlong filter queries so the filter remains one row while preserving the full value for matching
- cover all production table width configurations and the paginated `60x24` layout with regressions

## Testing

- focused DataTable, pagination, width, and key-hint tests (`30 pass`, `0 fail`)
- CI unit tests on Linux, macOS, and Windows
- source-scoped TypeScript check excluding ignored generated `dist` assets
- `bun run lint:check`
- `bun run format:check`
- `bun run build`
- live Runtime Version and Harness Version validation at `120x32`
- live paginated Runtime List validation at `60x24`
- live long-filter validation at `60x24`

### DIFF
--- a/src/components/ui/data-table/DataTable.tsx
+++ b/src/components/ui/data-table/DataTable.tsx
@@ -143,7 +143,7 @@ export function DataTable<T extends Record<string, unknown>>({
             onNextPage();
           }
         } else setCurrentPage((p) => Math.min(totalPages - 1, p + 1));
-      } else if (input === "/") {
+      } else if (searchable && input === "/") {
         setSelectedRow(0);
         setCurrentPage(0);
         setSearchMode(true);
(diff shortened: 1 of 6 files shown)


src/components/ui/data-table/DataTable.tsx
```diff
@@ -191,26 +191,18 @@ export function DataTable<T extends Record<string, unknown>>({
     const width = computedWidths.widths[index];
     return width === undefined ? [] : [{ column, width }];
   });
+  const filterLine = searchMode ? (
```
nborges-aws, line 206: nit: if the filter is very long, it pushes down the table and impede rendering. Maybe as follow up we can left-truncate long inputs?

nborges-aws (dismissed):

src/components/ui/data-table/DataTable.tsx
aidandaly24: Addressed in aa072c55. The table now preserves the full query for matching while left-truncating only the visible value to the computed table width. Added a regression with 33 matching rows and a long filter, and verified the one-line filter live at 60x24.

nborges-aws (approved):

src/components/ui/data-table/DataTable.tsx
```diff
@@ -191,26 +191,33 @@ export function DataTable<T extends Record<string, unknown>>({
...
   });
+  const filterPrefix = "/ Filter: ";
+  const filterCursor = searchMode ? "█" : "";
+  const availableQueryWidth = Math.max(
+    0,
+    computedWidths.totalWidth - stringWidth(filterPrefix) - stringWidth(filterCursor),
+  );
+  const visibleSearchQuery =
+    availableQueryWidth > 0
+      ? cliTruncate(searchQuery, availableQueryWidth, {
```
AlexanderRichey, line 202: What is `cliTruncate`?

AlexanderRichey (approved):

PR #1899, opened 2026-08-04
feat(gateway): add create commands

aidandaly24:
## Summary

- add create commands for Gateways, Targets, connector-backed Targets, and Rules
- expose complete service-model inputs through JSON sources plus focused Target and Connector shortcuts
- omit `protocolType` by default for unrestricted Gateways and map `--protocol mcp` to an MCP-only Gateway
- require the service-modeled `--role-arn` when creating a Gateway
- default curated Connector shortcuts to `GATEWAY_IAM_ROLE`; require exactly one explicit credential configuration for exact `--connector-configuration`

## IAM scope

- this PR does not create, tag, mutate, or delete IAM roles or permission policies
- Gateway and Target execution-role permissions remain customer-managed
- automatic role provisioning and permission reconciliation are deferred to the shared IAM management plane

## Stack

- base: `refactor`, including merged connector read PR #1922
- this PR: #1899
- Update and Delete remain separate upstack PRs

## Testing

- focused Gateway validation, handler mapping, source-aware help, and connector shortcut tests
- fixture-backed Gateway, HTTP Target, Web Search Connector, and Rule creation through the real router and Core
- live deploy-account probes proving every tested Connector rejects omitted credentials, explicit `[]` is rejected by the service, explicit `GATEWAY_IAM_ROLE` is accepted, and the CLI shortcut default reaches `READY`
- the fixture provisions its prerequisite role and Web Search permission outside production code, then passes `--role-arn` through the CLI
- full lint, formatting, typecheck, Gateway stack tests, bundle, and AgentCore E2E coverage
- every disposable probe Target, Gateway, inline policy, and role was deleted and confirmed absent


src/core/gatewayExecutionRole.ts
```diff
@@ -0,0 +1,296 @@
...
+}
+
+function gatewayArnPattern(
+  partition: string,
+  region: string,
+  accountId: string,
+  gatewayName: string,
+): string {
+  return `arn:${partition}:bedrock-agentcore:${region}:${accountId}:gateway/${gatewayName}-*`;
+}
```
AlexanderRichey: I understand that gateway is itself complicated, but the complexity that we're introducing in this code seems excessive. I think it's likely that we can achieve the same functionality with much simpler code. Let's start small and simple and layer on complexity, where justified, as we go. To do that in this case, I think we should omit all of the IAM role handling business here and develop the simplest possible solution—a default, uneditable role. After we get that code where it should be, we can add this functionality.

AlexanderRichey (dismissed):

src/core/gatewayExecutionRole.ts
aidandaly24: This makes sense to me, I definitely agree. I'll update to fit that simpler model and we can build as we go. That should hopefully cut down on a lot of accidental complexity.

src/core/gateway.tsx
```diff
@@ -63,6 +100,15 @@ export class GatewayClient implements CoreGatewayClient {
     );
   }
 
+  async createGatewayTarget(
+    input: CreateGatewayTargetInput,
+    options: CoreOptions,
+  ): Promise<CreateGatewayTargetResponse> {
+    return this.clients
+      .control(toClientConfig(options))
+      .send(new CreateGatewayTargetCommand(input));
```
nborges-aws, line 98: I know we're going for minimal complexity, but we need to append target-specific permissions to the execution role before we send the create command, otherwise the connector target won't work. 

The default execution role created by CLI below only has a trust policy. So if I then create a web search gateway target here, the default role will be missing `InvokeWebSearch` and fail.

src/handlers/gateway/gateway.create.test.tsx
```diff
@@ -0,0 +1,543 @@
...
+          (response) => response.status === "READY",
+        );
+
+        // TODO: Remove this fixture-only grant when managed Gateway roles reconcile Target permissions.
+        if (isRecording()) {
+          await createIamClient({ region: REGION }).send(
+            new PutRolePolicyCommand({
+              RoleName: GatewayExecutionRole.roleName(GATEWAY_NAME, REGION),
+              PolicyName: WEB_SEARCH_POLICY_NAME,
+              PolicyDocument: JSON.stringify(webSearchPolicy(gateway.gatewayArn)),
```
nborges-aws, line 543: Test is actually appending the missing permission that I mentioned in my comment above

src/core/gateway.tsx
aidandaly24: I think its best to leave it out of this PR, unless there is a consensus that we should include it here. I am going to follow up after delete with all of the IAM role permissions management. I do update the role in the test like you note below to have the test succeed, but I left a TODO to remove it later because I think that it makes sense to keep the IAM role management in one PR. I also think it will be easier to do IAM management after the `Update` PR is in as well since the management side will have overlap between create and update.

src/handlers/gateway/connector/create/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+        "metadata-configuration",
+        "metadata propagation (JSON; inline, file://<path>, or - for stdin)",
+        z.string().optional(),
+      ),
+      flag(
+        "private-endpoint",
+        "private endpoint (JSON; inline, file://<path>, or - for stdin)",
+        z.string().optional(),
+      ),
+      flag("client-token", "idempotency token", z.string().optional()),
```
jariy17: no client-token please.

src/handlers/gateway/rule/create/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+    name: "create",
+    description: "create a Gateway Rule",
+    flags: [
+      flag("gateway-id", "the parent Gateway ID", z.string().optional()),
+      flag(
+        "priority",
+        "Rule priority from 1 to 1000000",
+        z.number().int().min(1).max(1_000_000).optional(),
+      ),
+      flag("conditions", "Rule conditions (JSON Condition[])", z.string().optional()),
```
jariy17: should this use the Source Resolver here?

src/handlers/gateway/rule/create/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+      flag("actions", "Rule actions (JSON Action[])", z.string().optional()),
```
jariy17: Source Resolver?

src/handlers/gateway/rule/create/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+      flag("description", "Rule description", z.string().optional()),
+      flag("client-token", "idempotency token", z.string().optional()),
```
jariy17: No need for client token here.

src/handlers/utils.tsx
```diff
@@ -51,6 +51,27 @@ export function parseJsonFlagWithSchema<T>(
   return result.data;
 }
 
+export function parseJsonObjectFlag<T extends object>(
```
jariy17, line 54: should this go into the io library? I'm fine with it here.

jariy17 (changes requested):
Pretty good, just need some nits.

nborges-aws:
LGTM, will approve once TJ's callouts are addressed

src/handlers/utils.tsx
aidandaly24: I think it belongs here, this is where the flag parsing is. I had mentioned somewhere that we could make a flag parsing class.

src/handlers/gateway/connector/create/index.tsx
aidandaly24: It's exposed in harness and eval so I'm just wondering why not? I am fine to remove it and I actually agree that it is a valid UX simplification.

src/handlers/gateway/rule/create/index.tsx
aidandaly24: It does use it below

src/handlers/gateway/connector/create/index.tsx
aidandaly24: Fixed in e81eb0a6: removed the connector create --client-token flag and no longer forwards clientToken in the CreateGatewayTarget request.

src/handlers/gateway/rule/create/index.tsx
aidandaly24: Already implemented: --actions is resolved through SourceResolver before parseJsonArrayFlag, so it supports inline JSON, file:// paths, and stdin. No functional change needed.

src/handlers/gateway/rule/create/index.tsx
aidandaly24: Fixed in e81eb0a6: removed the rule create --client-token flag and no longer forwards clientToken in the CreateGatewayRule request.

src/handlers/gateway/connector/create/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+      const privateEndpoint = parseJsonObjectFlag<PrivateEndpoint>(
+        "private-endpoint",
+        await source.resolveText("private-endpoint", flags["private-endpoint"]),
+      );
+      const input: CreateGatewayTargetInput = {
+        gatewayIdentifier: flags["gateway-id"],
+        name: flags.name,
+        targetConfiguration,
+        ...(flags.description ? { description: flags.description } : {}),
+        ...(credentialProviderConfigurations ? { credentialProviderConfigurations } : {}),
```
nborges-aws: nit: should `--credential-provider-configurations` be required? If its omitted and passes through nothing, I would expect the command to fail.

nborges-aws (dismissed):

src/handlers/gateway/connector/create/index.tsx
aidandaly24: Checking this out thanks for the call out

aidandaly24: Good callout. I live-tested this in the deploy account before changing the CLI. Web Search, Bedrock Mantle, OpenAI, and Bedrock Knowledge Bases all reject omission with `ValidationException: Credential provider configurations is not defined`; explicit `[]` is also rejected by the modeled minimum length. The implementation now defaults curated `--connector` shortcuts to one `GATEWAY_IAM_ROLE` configuration, requires exactly one explicit configuration with `--connector-configuration`, and rejects empty/multiple arrays before Core. A live `gateway connector create --connector web-search` without the credential flag reached `READY`, and all disposable resources were confirmed deleted.

nborges-aws (dismissed):

jariy17 (dismissed):
LGTM

jariy17 (approved):

nborges-aws (approved):

PR #1900, opened 2026-08-04
feat(gateway): add update commands

aidandaly24:
## Summary

- add Gateway, Target, connector-backed Target, and Rule update commands
- support explicit set, clear, and no-op validation semantics
- GET current Gateway and Target state, merge requested changes, and send the complete replacement requests required by the service
- keep Rule updates patch-shaped
- preserve complete API access through JSON sources plus focused MCP endpoint and Connector shortcuts
- keep Connector shortcut construction handler-owned and Connector classification local to each layer

## IAM scope

- this PR does not create, mutate, or reconcile IAM roles or permission policies
- `--role-arn` updates only the Gateway service configuration
- execution-role permissions remain customer-managed until the shared IAM management plane lands

## Stack

- base: #1899
- this PR: #1900
- top: #1901

## Testing

- focused Gateway validation, flag mapping, request merge, connector classification, and fixture replay tests
- fixture-backed Gateway, Target, Connector, and Rule updates through the real router and Core
- full build, lint, formatting, typecheck, cross-platform unit-test, compiled-binary smoke-test, and AgentCore E2E CI coverage


src/core/gateway.tsx
```diff
@@ -132,4 +229,100 @@ export class GatewayClient implements CoreGatewayClient {
...
+    if (replacement === undefined) return current;
+    return replacement === null ? undefined : replacement;
+  }
+
+  private static isConnectorTarget(configuration: TargetConfiguration | undefined): boolean {
+    return (
+      configuration?.mcp?.connector !== undefined ||
+      configuration?.inference?.connector !== undefined
+    );
+  }
```
aidandaly24, line 320: This function is also here, but I didn't want to import from the handlers into the core. This avoids circular dependencies.

https://github.com/aws/agentcore-cli/blob/c44a4e572b73d5890d2e0e88adab6882de728f65/src/handlers/gateway/connector/gatewayConnectorTarget.ts#L6-L11

src/core/gateway.test.ts
```diff
@@ -0,0 +1,306 @@
...
+  expect(commands[0]).toBeInstanceOf(GetGatewayTargetCommand);
+  expect((commands[0] as GetGatewayTargetCommand).input).toEqual({
+    gatewayIdentifier: patch.gatewayId,
+    targetId: patch.targetId,
+  });
+  return (commands[1] as UpdateGatewayTargetCommand).input;
+}
+
+describe("GatewayClient updateGateway", () => {
+  test("preserves required and omitted fields while replacing selected configuration", async () => {
```
Hweinstock: what coverage or guarantees do we get from these tests that we don't get from the top level handler tests that use fixtures? I'm wondering if testing on the client is focusing too heavily on implementation details, rather than behavior.

src/core/gateway.tsx
```diff
@@ -51,6 +64,76 @@ export class GatewayClient implements CoreGatewayClient {
       .send(new GetGatewayCommand({ gatewayIdentifier: id }));
   }
 
+  async updateGateway(
+    patch: GatewayUpdatePatch,
+    options: CoreOptions,
+  ): Promise<UpdateGatewayResponse> {
+    const control = this.clients.control(toClientConfig(options));
+    const current = await control.send(new GetGatewayCommand({ gatewayIdentifier: patch.id }));
+    const name = GatewayClient.required(current.name, patch.id, "name");
```
Hweinstock: would a get gateway response ever return a response without these fields? If we did get a response without these, should that be classified as a `service` error instead of `client`?

src/core/gateway.tsx
```diff
@@ -51,6 +64,76 @@ export class GatewayClient implements CoreGatewayClient {
...
+    const roleArn = GatewayClient.required(current.roleArn, patch.id, "role ARN");
```
Hweinstock: would it be simpler to deep merge current and patch then let the API reject? I'm wondering if this could get difficult to maintain as the fields change/grow. 

Something like 
```
const current = await.control.send(new GetGatewayCommand({ gatewayIdentifier: patch.id }));
const request = deepReplace(current, patch);
return control.send(new UpdateGatewayCommand(request));
```
where deepReplace walks the object tree calling the existing replace method. 

The downside is we lose the custom error messages, but I wonder if we could rely on the service validation for this?

src/core/gateway.tsx
```diff
@@ -132,4 +229,100 @@ export class GatewayClient implements CoreGatewayClient {
...
   }
+
+  async updateGatewayRule(
+    input: GatewayRuleUpdateInput,
+    options: CoreOptions,
+  ): Promise<UpdateGatewayRuleResponse> {
+    return this.clients.control(toClientConfig(options)).send(new UpdateGatewayRuleCommand(input));
+  }
+
+  private async updateTarget(
```
Hweinstock, line 300: similar question here. I wonder if we can avoid maintaining all this client side validation.

src/handlers/gateway/connector/update/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+        "replacement outbound credentials (JSON array; inline, file://<path>, or - for stdin)",
+        z.string().optional(),
+      ),
+      flag(
+        "metadata-configuration",
+        "replacement metadata propagation (JSON; inline, file://<path>, or - for stdin)",
+        z.string().optional(),
+      ),
+      flag(
+        "private-endpoint",
```
Hweinstock, line 57: one day I would love to generate these from the api spec, but that day is not today lol.

src/handlers/gateway/target/update/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+    ],
+    handle: async (ctx, flags) => {
+      if (!flags["gateway-id"]) {
+        throw new InputValidationError("required option '--gateway-id <gateway-id>' not specified");
+      }
+      if (!flags["target-id"]) {
+        throw new InputValidationError("required option '--target-id <target-id>' not specified");
+      }
+
+      for (const [name, value, clear] of [
```
Hweinstock: i like this pattern for handling mutually exclusive. Do you see a natural way to generalize it a bit to re-use across these few handlers that use it in this PR?

src/handlers/gateway/types.tsx
```diff
@@ -24,8 +38,38 @@ export type CreateGatewayTargetInput = CreateGatewayTargetRequest;
 
 export type CreateGatewayRuleInput = CreateGatewayRuleRequest;
 
+export type GatewayUpdatePatch = {
```
Hweinstock, line 33: is there a strong motivation to define our own types here, instead of derive it from the SDK types?

src/core/gateway.tsx
aidandaly24: I considered this, but deep merging would treat every nested object as a patch. Our JSON flags are full replacements, so it could silently preserve fields the customer intended to remove. Policy Engine is the one intentional partial merge because we expose Policy Engine as separate `--policy-engine-arn` and `--policy-engine-mode flags`. An explicit mapper keeps those preserve, replace, and clear semantics unambiguous.

src/core/gateway.test.ts
aidandaly24: That’s fair. The fixture-backed flow already covers the normal `GET-to-Update` path, so the simple preservation tests are redundant. I’ll trim this file to the clear, merge, shorthand, and classification behaviors the fixture flow does not cover.

src/handlers/gateway/types.tsx
aidandaly24: We need a distinct patch type because it includes CLI-only semantics such as clear sentinels and endpoint shorthand. However, fields that map directly to the service should derive their types from `UpdateGatewayRequest` and `UpdateGatewayTargetRequest`. I’ll update those while keeping only the CLI-specific additions explicit.

src/core/gateway.tsx
aidandaly24: Good catch. `name` and `authorizerType` are modeled as required, and `roleArn` is required by Create and Update. If GET omits them, that is a service-response issue rather than invalid user input. I changed this guard, and the equivalent required Target configuration guard, to report `source: service`.

src/core/gateway.tsx
aidandaly24: I kept the checks that enforce CLI-only semantics: `connector update` must remain connector-backed, and `--endpoint` only applies to an existing MCP server Target. The service cannot infer those command and shorthand contracts. Missing required `targetConfiguration` from GET now reports a service error instead of a user validation error.

src/handlers/gateway/target/update/index.tsx
aidandaly24: Implemented as `validateSetClearConflicts` in the shared handler utilities and reused it across Gateway, Target, Connector, and Rule updates. Resource-specific validation remains in each handler.

src/handlers/utils.tsx
Hweinstock, line 30: OOS here, but I feel like we should rename this to avoid it becoming a dumping ground of a bunch of functionality? Looks mostly parsing related? maybe `src/handlers/parsing.tsx`?

src/handlers/utils.tsx
```diff
@@ -72,6 +72,16 @@ export function parseJsonArrayFlag<T>(name: string, raw: string | undefined): T[
   return parsed as T[];
 }
 
+export function validateSetClearConflicts(
```
Hweinstock: nit: my understanding is that the existing logic holds for any mutually exclusive input (not just set and clear). Does it make sense to name it something to capture that? Ex. `assertMutuallyExclusiveInputs` or something.

Hweinstock (dismissed):

src/handlers/utils.tsx
aidandaly24: Agreed that `utils.tsx` is broad and could become a dumping ground. I kept the file rename out of this PR since this change only adds one handler-validation helper. Splitting the existing parsing utilities can be a focused follow-up.

src/handlers/utils.tsx
aidandaly24: Updated this to `assertMutuallyExclusiveInputs`. It now accepts both input names and values, and is used for ordinary pairs such as `--endpoint`/`--target-configuration` and `--connector`/`--connector-configuration`, as well as set/clear pairs.

tejaskash (approved):

src/handlers/utils.tsx
aidandaly24: I also think that the parsing stuff should be moved into a class. There is only one function here that is not parsing.

src/core/gateway.tsx
```diff
@@ -58,6 +74,77 @@ export class GatewayClient implements CoreGatewayClient {
...
   }
 
+  async updateGateway(
+    patch: GatewayUpdatePatch,
+    options: CoreOptions,
+  ): Promise<UpdateGatewayResponse> {
+    const control = this.clients.control(toClientConfig(options));
+    const current = await control.send(new GetGatewayCommand({ gatewayIdentifier: patch.id }));
+    const resource = `Gateway "${patch.id}"`;
+    const name = GatewayClient.required(current.name, resource, "name");
```
jariy17, line 84: Please check the smithy model, I'm pretty sure the name is required in the response. AWS SDK v3's response parameters are all | undefined for some reason.

src/core/gateway.tsx
```diff
@@ -58,6 +74,77 @@ export class GatewayClient implements CoreGatewayClient {
...
+    if (patch.authorizerConfiguration !== undefined && authorizerType !== "CUSTOM_JWT") {
+      throw new InputValidationError(
+        "Authorizer configuration can only be updated for a CUSTOM_JWT Gateway",
+      );
+    }
+
+    let policyEngineConfiguration = current.policyEngineConfiguration;
+    if (patch.policyEngineConfiguration === null) {
+      policyEngineConfiguration = undefined;
+    } else if (patch.policyEngineConfiguration !== undefined) {
```
jariy17, line 100: Lets choose one null or undefined.

src/core/gateway.tsx
```diff
@@ -189,6 +290,97 @@ export class GatewayClient implements CoreGatewayClient {
...
+    };
+    if (connectorOnly && !GatewayClient.isConnectorTarget(request.targetConfiguration)) {
+      throw new InputValidationError(
+        "Connector updates require an MCP or inference connector Target configuration",
+      );
+    }
+    return control.send(new UpdateGatewayTargetCommand(request));
+  }
+
+  private static replace<T>(
```
jariy17, line 367: Is there a way we can use this.replace instead of GatewayClient.replace?

Hweinstock (approved):
LGTM!

jariy17 (commented):
nit: These are nits so make follow up prs for this.

PR #1906, opened 2026-08-04
feat: add Memory event TUI

notgitika:
This PR adds TUI implementation to the memory event list and get commands implemented in PR #1895 

The flow: 
Memory -> Actor -> Session -> Event selection 

It lists events with branch and occurrence time and opens up selected event as an entire JSON.  Pagination, filtering, retries, empty states, and back navigation is implemented.

I had to add internal read-only Actor and Session listing APIs, which we could as future work make into their own commands.

I tested myself + using the TUI harness skill.

Typcheck, lint, build and test passes


AlexanderRichey (approved):

aidandaly24 (approved):

notgitika:
Updating readme as follow up

PR #1911, opened 2026-08-04
feat(eval): add datasets CLI commands 

nborges-aws:
## Description
Adds command-line CRUDL plus versioning for AgentCore evaluation datasets:
* eval dataset create
* eval dataset get
* eval dataset list
* eval dataset delete
* eval dataset publish

**`eval dataset update` _will land in follow up PR_**

* `DataSourceType` union on `CreateDataset` allowing users to pass in s3 uri or file path for data set creation.
* `source.tsx` - Responsible for translating file path in dataSourceType union to the service API expected `inlineExample` format. `inlineExamples` is a dict, with list of JSONL items inside. For user convenience and ability to pass their JSONL draft file as a whole -- we accept filepath, and parse the file into `inlineExamples` list before call to service
* `--file-path` flag conditional logic on delete: `eval datasets get` without file path flag will fetch metadata and print to command line (getDataset API). When file path flag is included, CLI will pull content from pre-signed downloadUrl provided in metadata, and writes content to provided file location (streamed into temp file and renamed into place)
* New error types in `src/errors`: `DatasetDownloadError` (SERVICE, covers a missing or unfetchable presigned URL) and `DatasetWriteError` (USER, covers local write issues)
* EvalClient now receives the injected CoreFetch from CoreClient, since dataset content is served from S3
* create, publish, delete are all async operation which return CREATING / UPDATING / DELETING

## Related Issue

Closes #

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing
Manually tested commands using `bun run ...` against personal account, happy path and negative validation.

Full suite against latest HEAD after rebasing onto refactor:

- bun test (729 pass, 0 fail)
- bun run typecheck clean
- bun run lint:check clean
- bun run format:check clean
- bun run build clean

- [x] I ran `npm run lint`

## Checklist

---


src/handlers/eval/dataset/publish/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const createPublishDatasetHandler = (core: Core) =>
+  createHandler({
+    name: "publish",
+    description: "publish the current DRAFT as a new immutable version",
+    flags: [flag("id", "the ID of the dataset to publish", z.string().optional())],
+    handle: async (ctx, flags) => {
+      if (!flags["id"]) throw new InputValidationError("required option '--id <id>' not specified");
+
+      // Publishing an unmodified DRAFT creates a version identical to the last
```
nborges-aws, line 16: current implementation will allow user to publish a version that is unmodified from the last. If silently publishing unmodified version feels wrong here, we can check `draftStatus` for `UNMODIFIED` and use that to prompt/warn/reject publishing.

src/core/eval.tsx
```diff
@@ -447,6 +468,147 @@ export class EvalClient implements CoreEvalClient {
...
+      callback(null, lastByte === undefined || lastByte === NEWLINE ? undefined : "\n");
+    },
+  });
+}
+
+// streamToFile writes via a temporary file in the same directory as `filePath`,
+// renamed into place once the transfer completes. Streams so a large dataset
+// is never held in memory. A failed or aborted transfer removes the temp file and
+// leaves any existing file at `filePath` untouched.
+async function streamToFile(
```
AlexanderRichey: This seems like it should be in the io package.

src/core/eval.tsx
```diff
@@ -447,6 +468,147 @@ export class EvalClient implements CoreEvalClient {
...
+    options: CoreOptions,
+    signal?: AbortSignal,
+  ): Promise<GetDatasetResponse> {
+    const dataset = await this.getDataset(id, version, options);
+
+    // The consolidated file is written asynchronously, so a dataset that is still
+    // ingesting has no URL to offer yet. Report the status, which is what tells
+    // the caller whether to retry.
+    if (!dataset.downloadUrl) {
+      throw new DatasetDownloadError(
```
jariy17: This isn't really a user fault, more of service issue. Should we add a retry mechanism here?

src/core/eval.tsx
```diff
@@ -447,6 +468,147 @@ export class EvalClient implements CoreEvalClient {
...
+          `retry once it reports ACTIVE`,
+        { meta: { datasetId: id, datasetVersion: dataset.datasetVersion, status: dataset.status } },
+      );
+    }
+
+    let body: ReadableStream<Uint8Array>;
+    try {
+      const response = await this.fetch(dataset.downloadUrl, { signal });
+      if (!response.ok) {
+        throw new DatasetDownloadError(
```
jariy17: Again this could be a service issue or an Internal issue.

src/errors/errors.tsx
```diff
@@ -142,3 +142,22 @@ export class RuntimeInvokeResponseError extends AgentCoreCLIError {
...
+  constructor(message: string, options?: AgentCoreCLIErrorOptions) {
+    super(message, { source: ERROR_SOURCE.SERVICE, ...options });
+  }
+}
+
+/**
+ * A downloaded dataset could not be written to its destination.
+ * Local and user-fixable: a missing directory, a permission denial, etc.
+ */
+export class DatasetWriteError extends AgentCoreCLIError {
```
AlexanderRichey: Can these exceptions be made more general? Maybe `NetworkingError` and `FileWriteError`?

AlexanderRichey (changes requested):

src/core/eval.tsx
```diff
@@ -447,6 +468,147 @@ export class EvalClient implements CoreEvalClient {
...
+          { meta: { datasetId: id, status: response.status } },
+        );
+      }
+      if (!response.body) {
+        throw new DatasetDownloadError(`Dataset "${id}" download returned an empty response`, {
+          meta: { datasetId: id },
+        });
+      }
+      body = response.body;
+    } catch (error) {
```
jariy17, line 510: Concern about the catch-all wrap here. Blanket-tagging every unknown fetch failure as SERVICE is a guess, and it can mislabel things:
    1. A bug in our own request construction is really INTERNAL, not SERVICE.
    2. A runtime or polyfill issue is also INTERNAL.
  
'd only wrap the cases we can positively attribute to the service, like the non-ok HTTP status above. Let anything unrecognized fall through to [fromError](https://github.com/aws/agentcore-cli/blob/180627e0e50890e5d3ab612cd786f5eedc3426e3/src/errors/errors.tsx#L41) so it categorizes based on what it actually observes.

src/core/eval.tsx
jariy17: There should be a library package for this

src/handlers/eval/dataset/get/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+          );
+        return;
+      }
+
+      // --file-path downloads the contents via the presigned download URL in metadata
+      const controller = new AbortController();
+      const interrupt = () => controller.abort();
+      process.once("SIGINT", interrupt);
+      try {
+        const response = await core.eval.downloadDataset(
```
jariy17, line 43: We should add filePath to the response here. Right now the output is just metadata, so nothing in it confirms where the file was written. Returning { ...dataset, filePath } gives the customer clear confirmation of the download and its location.

jariy17 (changes requested):
Pretty Good! Just look at the error handling for the handlers

src/core/eval.tsx
nborges-aws: I've classified this as a SERVICE error. I considered adding a retry mechanism, but it was plausible that after a short-polling loop it still wouldn't be active. I lead towards classifying this as a service-side retryable exception, and informing the user to retry once ACTIVE as the right call. Updated code does exactly that

src/core/eval.tsx
nborges-aws: I think for this case we're good to leave as service. If we hit `!response.ok` then we successfully made the http call and got a non-2xx

jariy17 (approved):
LGTM, though this is more of a general question for the team. What does --json do here? I thought we already returned JSON output for every CLI command.

AlexanderRichey (approved):

PR #1922, opened 2026-08-05
feat(gateway): add connector read commands

aidandaly24:
## Summary

- add `gateway connector get` for connector-backed Gateway Targets
- add `gateway connector list` by filtering Target summaries with `targetType: CONNECTOR`
- preserve existing Target read APIs and service pagination tokens

## Testing

- `bun test src/handlers/gateway/gateway.test.tsx`
- `bun run build`
- focused Prettier and oxlint checks

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a></sub>


src/handlers/gateway/connector/isConnectorTarget.ts
```diff
@@ -0,0 +1,7 @@
+import type { TargetConfiguration } from "@aws-sdk/client-bedrock-agentcore-control";
+
+export function isConnectorTarget(configuration: TargetConfiguration | undefined): boolean {
```
jariy17: nit: it seems like this is only useful for get handler. Let's put this in the index.tsx file.

src/handlers/gateway/gateway.test.tsx
```diff
@@ -210,6 +214,67 @@ describe("gateway reads", () => {
     ]);
   });
 
+  test.each([
```
jariy17, line 217: is there a reason why we aren't doing any golden tests here?

jariy17 (changes requested):
Looks good! Left nit comment and unit test question (TDD)

src/handlers/gateway/gateway.test.tsx
aidandaly24: Added fixture-backed golden coverage for connector `get` and filtered `list` in afcaac52, recorded against the e2e-test account. The focused unit cases remain for MCP/inference classification, token-preserving filtering, and non-connector rejection.

src/handlers/gateway/connector/isConnectorTarget.ts
aidandaly24: Moved `isConnectorTarget` into `connector/get/index.tsx` and removed the standalone module in afcaac52.

src/handlers/gateway/gateway.test.tsx
aidandaly24: Follow-up: replaced the initial failed Target recording with a genuinely READY OpenAI connector in 8aca33f3. It uses the existing DONOTDELETEe2eOAI provider in the e2e account; no key value is present in the fixtures.

jariy17 (dismissed):

jariy17 (approved):

nborges-aws (approved):

PR #1925, opened 2026-08-05
feat: add project build command for CDK

notgitika:
This PR adds the first implementation of `agentcore project build` (no TUI).

Here are some design considerations I took:

1. Build is separate from deploy. build creates the artifacts and saves a manifest; deploy will use that manifest to deploy the artifact. The `deploy` work will be in a follow up PR.
2. Only CDK is supported now, but the manifest is not CDK-only. CDK details live in a CDK-specific artifact so Terraform or no-IAC etc can be added later.
3. The build flow decides what to validate and create, while filesystem access and command execution live in the IO layer.
4. The manifest makes builds reusable. It stores the project inputs, CLI version, targets, and generated artifacts so deploy can tell whether it needs to rebuild

The manifest is the contract between the commands: build produces a validated, versioned artifact description, and deploy will consume it only when its project fingerprint and CLI version are still current.

---
What this includes:
1. Resolves the current project from the working directory
2. Validates the project config and deployment targets
3. builds and synthesizes the CDK application and verifies that every configured target produced a stack.
4. Writes the build manifest to `agentcore/.build/manifest.json`.
5. prints progress to stderr; the final build result is a JSON

Build, format, lint and typecheck pass. Running more tests locally while this is in draft PR mode.


AlexanderRichey (changes requested):
There's a lot of good stuff here, but I feel that this PR is too big to merge in all at once. Let's break this down into smaller pieces so that we can have a more productive review cycle. Let's start with the higher level interfaces, align on those, and then move on to the implementations.

I'm also seeing a lot of code duplicating functionality we already have. We should be using ReadWriteJson, for example, instead of writing that same functionality over again. I think it would be helpful to take some time to check out the current codebase to see what's available and to direct your coding agent to use these existing utilities rather than rewriting.

notgitika:
Closing and Dividing this PR into 3-4 PRs. Will link all of them here. 

Starting with:
 
https://github.com/aws/agentcore-cli/pull/1935

PR #1930, opened 2026-08-06
fix(runtime): polish invoke TUI errors and sessions

aidandaly24:
## Summary

- surface available error names and messages for every non-abort Runtime lookup, invocation, and response-stream failure
- render modeled AWS service failures as a structured type, HTTP status, message, and request ID while preserving redaction for arbitrary transport causes
- generate and display a Runtime session ID before the first TUI send, while honoring `--session-id`
- remove the redundant `JSON payload` heading and trailing response whitespace before settled metadata

## Testing

- `bun test src/core/core.test.ts src/handlers/runtime/invoke/invoke.screen.test.tsx` (`60` pass)
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- TUI Harness: successful live invocation without `--session-id`; generated ID persisted in request and response
- TUI Harness: intentionally short session ID rendered the AWS `ValidationException` type, HTTP status, complete message, and request ID on separate lines
- TUI Harness: final SSE event now renders directly above session metadata while preserving spacing between events
- `bun test src`: the only failures were two unchanged `src/io/exec.test.ts` subprocess-output assertions that reproduce when run alone in this environment

### DIFF
--- a/src/core/runtime.tsx
+++ b/src/core/runtime.tsx
@@ -1,4 +1,5 @@
 import { randomUUID } from "node:crypto";
+import { ServiceException } from "@smithy/core/client";
 import {
   GetAgentRuntimeCommand,
   GetAgentRuntimeEndpointCommand,
@@ -172,6 +173,10 @@ export class RuntimeClient implements CoreRuntimeClient {
         name?: unknown;
         $metadata?: { httpStatusCode?: unknown; requestId?: unknown };
       };
+      const serviceMessage =
+        ServiceException.isInstance(error) && error.message.trim()
+          ? error.message.trim()
+          : undefined;
       const diagnostics: string[] = [];
       if (typeof sdkError?.name === "string" && sdkError.name !== "Error") {
         diagnostics.push(sdkError.name);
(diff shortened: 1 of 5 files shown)


notgitika (approved):

src/core/runtime.tsx
```diff
@@ -194,7 +199,10 @@ export class RuntimeClient implements CoreRuntimeClient {
         })
         .debug("Runtime invocation SDK request failed");
       throw new Error(
```
AlexanderRichey, line 201: Should this be an AgentCoreError?

src/core/runtime.tsx
AlexanderRichey, line 184: Is all of this complexity necessary? Why not simply throw the error? All of these details should already be handled in the error message, no? If they're not handled, then extracting this complexity into a reusable helper would help to simplify the main flow at this level.

nborges-aws (approved):

AlexanderRichey (approved):
Approved with a couple comments we might want to address in a follow up.

PR #1908, opened 2026-08-04
feat: add Memory record TUI

notgitika:
This PR adds TUI implementation to the memory record list and get commands implemented in PR https://github.com/aws/agentcore-cli/pull/1895

PR #1906 adds the event TUI. Stacks on top of the both of them.

The flow:
Adds Memory selection followed by namespace scope entry. Supports namespace and namespace-path.

It lists records with content, strategy, and creation timeand opens up selected event as an entire JSON. Pagination, filtering, retries, empty states, and back navigation is implemented.

I tested this myself + using the TUI harness skill.

Typcheck, lint, build and test passes

Additional TUI entry points:
- Events and records are available from the top-level Memory menu and beside detail in a selected Memory's action list.
- Record-list columns preserve IDs, content, strategy IDs, and timestamps at wide terminal widths.

PR #1931, opened 2026-08-06
fix: move events and records under get memory instead of memory top level

notgitika:
This PR moves the memory picker from where it was before (adjacent to memory get and list) to now being under a specific memory get. The rest of the logic and flow remains the same. I have verified e2e with both that the navigation and back spaces work well.

Where it stays now:

image


notgitika:
added into #1908

PR #1932, opened 2026-08-06
feat: add read-only datasets TUI

nborges-aws:
Adds an interactive, read-only TUI for the `eval dataset` command tree. Follows conventional menu → list → detail → JSON flow. Bare commands `eval dataset`, `eval dataset list`, `eval dataset get` now open TUI flows instead of missing flag error messaging.

## Screens

| Path | Screen |
|---|---|
| `eval dataset` | `RouterScreen` menu |
| `eval dataset list` / `eval dataset get` | `DatasetPicker` using `PaginatedTablePicker` |
| `eval dataset get/:datasetId` | `ResourceDetailScreen` |
| `eval dataset get/:datasetId/json` | `JsonDetail` |

## Testing
* Added `dataset.screen.test.tsx` covering navigation, menu filtering, metadata rendering, etc.
* `bun run test` (892 pass, 0 fail)

image
image

---

### DIFF
--- a/src/handlers/eval/dataset/get/screen.tsx
+++ b/src/handlers/eval/dataset/get/screen.tsx
@@ -1,0 +1,74 @@
+import { useQuery } from "@tanstack/react-query";
+import { useNavigate, useParams } from "react-router";
+import { JsonDetail } from "../../../../components/JsonDetail";
+import { ResourceDetailScreen } from "../../../../components/ResourceDetailScreen";
+import type { ScreenProps } from "../../../types";
+import { coreOptsFromCtx } from "../../../utils";
+
+function useDatasetDetail({ ctx, core }: ScreenProps, datasetId: string | undefined) {
+  const opts = coreOptsFromCtx(ctx);
+  return useQuery({
+    queryKey: ["dataset", opts.region, datasetId],
+    queryFn: () => core.eval.getDataset(datasetId!, undefined, opts),
+    enabled: datasetId !== undefined,
+  });
+}
+
+export function DatasetGetScreen(props: ScreenProps) {
+  const navigate = useNavigate();
+  const { datasetId } = useParams();
+  const detail = useDatasetDetail(props, datasetId);
+  const dataset = detail.data;
+
+  return (
+    <ResourceDetailScreen
+      breadcrumb={["agentcore", "eval", "dataset", "get", datasetId ?? ""]}
+      isPending={detail.isPending}
+      error={detail.isError ? (detail.error as Error) : null}
+      items={{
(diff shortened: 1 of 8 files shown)


jariy17 (dismissed):

src/handlers/eval/dataset/update/index.tsx
```diff
@@ -1,12 +1,11 @@
 import z from "zod";
 import { createHandler, flag } from "../../../../router";
 import { InputValidationError } from "../../../../errors";
-import type { AppIO } from "../../../../io";
 import { JsonRendererKey } from "../../../../tui";
 import type { Core } from "../../../types";
 import { coreOptsFromCtx } from "../../../utils";
 
-export const createUpdateDatasetHandler = (core: Core, io: AppIO) =>
+export const createUpdateDatasetHandler = (core: Core) =>
```
jariy17: Seperate the write only dataset handlers into their own pr.

jariy17 (changes requested):
Seperate the update commands in their own PR

src/handlers/eval/dataset/update/index.tsx
nborges-aws: This PR was incorrectly pulling in commits from the `dataset update` CLI command PR. That's why this file was showing up. The PR is now showing the correct commits/diff now that I've fixed conflicts and rebased

jariy17 (approved):

src/handlers/eval/dataset/get/screen.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const navigate = useNavigate();
+  const { datasetId } = useParams();
+  const detail = useDatasetDetail(props, datasetId);
+  const dataset = detail.data;
+
+  return (
+    <ResourceDetailScreen
+      breadcrumb={["agentcore", "eval", "dataset", "get", datasetId ?? ""]}
+      isPending={detail.isPending}
+      error={detail.isError ? (detail.error as Error) : null}
```
Hweinstock, line 27: should we use `AgentCoreCLIError.from` here?

Hweinstock (approved):

src/handlers/eval/dataset/get/screen.tsx
nborges-aws: Hmm yeah looking into it now, this is a good call out. It would probably make the most sense to use `AgentCoreCLIError.from` inside ResourceDetailScreen and the other shared resources. That way the its standardized across all screens. The call site would then just do: 
`detail.isError ? detail.error : null`

PR #1935, opened 2026-08-06
feat: define build manifest contract

notgitika:
This PR defines the versioned manifest contract to be shared between `build` and `deploy` commands.
It includes no command implementation yet.

You can look into the first draft PR I posted: https://github.com/aws/agentcore-cli/pull/1925

design considerations I took: 

- build writes its manifest to this path `agentcore/.build/manifest.json`
- The manifest records its format version, project name, CLI version, input fingerprint, build time, deployment targets, and artifact
- Artifacts are a discriminated union. CDK is the first supported one. Terraform and no-IaC modes can add their own typed variants later
- CDK manifests require exactly one synthesized stack for every recorded deployment target. Unit tests added for this.
- The CLI build response is a wrapper containing both `manifestPath` and the persisted `manifest`.
- Deploy will rebuild when the manifest is missing, invalid, unsupported, or has a stale input fingerprint.

ran typecheck, lint, format and test successfully.

### DIFF
--- a/src/core/project/buildManifest.ts
+++ b/src/core/project/buildManifest.ts
@@ -1,0 +1,120 @@
+import z from "zod";
+
+/** Relative to the project root; build writes this file and deploy reads it. */
+export const BUILD_MANIFEST_PATH = "agentcore/.build/manifest.json";
+
+export const BUILD_MANIFEST_VERSION = 1;
+
+export const DeploymentTargetSchema = z
+  .object({
+    name: z
+      .string()
+      .min(1)
+      .max(64)
+      .regex(
+        /^[a-zA-Z][a-zA-Z0-9-]*$/,
+        "must start with a letter and contain only letters, numbers, and hyphens",
+      ),
+    description: z.string().max(256).optional(),
+    account: z.string().regex(/^[0-9]{12}$/, "must be a 12-digit AWS account ID"),
+    region: z.string().min(1),
+  })
+  .strict();
+
+export type DeploymentTarget = z.infer<typeof DeploymentTargetSchema>;
+
+const ProjectRelativePathSchema = z
+  .string()
+  .min(1)
+  .refine(
+    (path) =>
+      !path.startsWith("/") &&
+      !path.startsWith("\\") &&
+      !/^[a-zA-Z]:[\\/]/.test(path) &&
+      !path.includes("\\") &&
+      path.split("/").every((segment) => segment !== "" && segment !== "." && segment !== ".."),
+    "must be a non-empty path relative to the project root",
+  );
+
+const CdkCloudAssemblyArtifactSchema = z
+  .object({
+    kind: z.literal("cdk-cloud-assembly"),
+    path: ProjectRelativePathSchema,
+    stacksByTarget: z.record(z.string().min(1), z.string().min(1)),
+  })
+  .strict();
+
+/**
+ * The artifact is the deploy dispatch point. Add a new discriminated variant
+ * before implementing another deployment mechanism.
+ */
+export const BuildArtifactSchema = z.discriminatedUnion("kind", [CdkCloudAssemblyArtifactSchema]);
+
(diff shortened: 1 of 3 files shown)


src/core/project/buildManifest.ts
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * The artifact is the deploy dispatch point. Add a new discriminated variant
+ * before implementing another deployment mechanism.
+ */
+export const BuildArtifactSchema = z.discriminatedUnion("kind", [CdkCloudAssemblyArtifactSchema]);
+
+export type BuildArtifact = z.infer<typeof BuildArtifactSchema>;
+
+/**
+ * Persisted handofffrom build to deploy
```
tejaskash: NIT - typo

src/core/project/buildManifest.ts
```diff
@@ -0,0 +1,120 @@
...
+ * Deploy requires a new build when the manifest is missing, unreadable, not valid, has unsupported manifestVerison,
+ * or if its inputFinderprint differs from current project inputs. Otherwise, deploy would use the
+ * typed artifact to select its backend
+ */
+export const BuildManifestSchema = z
+  .object({
+    manifestVersion: z.literal(BUILD_MANIFEST_VERSION),
+    projectName: z.string().min(1),
+    cliVersion: z.string().min(1),
+    inputFingerprint: z.string().regex(/^[a-f0-9]{64}$/),
```
tejaskash, line 68: What does this field help us do ?

src/core/project/buildManifest.ts
```diff
@@ -0,0 +1,120 @@
...
+    builtAt: z.iso.datetime(),
+    targets: z.array(DeploymentTargetSchema),
```
tejaskash: Based on the logic below, wouldn't this accept an `[]` with an empty `stacksByTarget` ? It would mean that this manifest is valid but undeployable by CDK. This would fail later at the deploy step. Can we decide on what direction we want to take - add validation here vs let it fail at deploy?

tejaskash (dismissed):
Approving with questions

src/core/project/buildManifest.ts
notgitika: fixing

src/core/project/buildManifest.ts
notgitika: This field identifies the exact build-relevant project inputs used to produce the artifact.`deploy` recomputes it, sees that when it differs, the saved artifact is stale and deploy rebuilds instead of deploying old output.

Let me know if that is not the required behavior

notgitika: adding a comment here to explain the field

src/core/project/buildManifest.ts
notgitika: I think validating it here makes sense. Fixing it now in this PR

src/core/project/buildManifest.ts
```diff
@@ -0,0 +1,121 @@
...
+
+/** The JSON response emitted by `agentcore project build`. */
+export const BuildResultSchema = z
+  .object({
+    manifestPath: z.literal(BUILD_MANIFEST_PATH),
+    manifest: BuildManifestSchema,
+  })
+  .strict();
+
+export type BuildResult = z.infer<typeof BuildResultSchema>;
```
AlexanderRichey, line 121: Which of these types does the existing CLI have today?

I tend to think we just need an interface that has one method in it called `build()`. Where does all of this other stuff come into play?

notgitika: Comparing to the existing CLI, these types don't exist. There is just a `LocalCdkProject.build(): Promise<void> ` +  the CDK wrapper’s in-memory synth() result.

I introduced them here to create a new persisted handoff between build and deploy commands. It does make things complicated. 

I thought more about this and I think the easiest design could be that we just keep a small internal cache record beside that cdk output, rather than introducing a broader public artifact contract now

tejaskash (approved):
LGTM

PR #1937, opened 2026-08-06
refactor(runtime): propagate invoke SDK errors

aidandaly24:
## Summary

- rethrow IAM `InvokeAgentRuntime` SDK failures without wrapping or rebuilding their diagnostics
- let the existing root error boundary classify headless failures
- preserve direct `ServiceException` metadata for structured TUI rendering
- remove Runtime-specific logger plumbing and speculative redaction tests

Follow-up to the error-boundary review comments on #1930.

## Testing

- `bun test src/core/core.test.ts src/handlers/runtime/invoke/invoke.screen.test.tsx src/handlers/runtime/invoke/invoke.test.tsx` (`85` pass)
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- TUI Harness: headless invalid session surfaced the original AWS validation message
- TUI Harness: interactive invalid session retained structured type, HTTP status, message, and request ID
- `bun test src` reached `816` pass; the only two failures were unchanged `src/io/exec.test.ts` subprocess-output assertions that reproduce in this environment

### DIFF
--- a/src/core/runtime.tsx
+++ b/src/core/runtime.tsx
@@ -1,5 +1,4 @@
 import { randomUUID } from "node:crypto";
-import { ServiceException } from "@smithy/core/client";
 import {
   GetAgentRuntimeCommand,
   GetAgentRuntimeEndpointCommand,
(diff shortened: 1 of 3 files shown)


nborges-aws (approved):

src/core/runtime.tsx
```diff
@@ -169,41 +167,7 @@ export class RuntimeClient implements CoreRuntimeClient {
...
-          }),
-        })
-        .debug("Runtime invocation SDK request failed");
-      throw new Error(
-        `Runtime invocation failed${serviceMessage ? `: ${serviceMessage}` : ""}${
-          diagnostics.length ? ` (${diagnostics.join(", ")})` : ""
-        }`,
-        { cause: error },
-      );
+      throw error;
```
AlexanderRichey, line 170: Do these errors need to be wrapped with `AgentCoreCLIError.fromError(error)` with the relevant settings so we know the error was a service error?

aidandaly24: Good question. Headless failures are already converted through `AgentCoreCLIError.fromError` in the root catch, which recognizes `ServiceException` and assigns the appropriate error source. The TUI consumes the original exception directly for structured rendering, so I think wrapping it inside Core would duplicate that boundary.

jariy17 (approved):

PR #1938, opened 2026-08-06
feat(eval): read-only batch-evaluation TUI (list + get)

jariy17:
## What

Adds the read-only **TUI** for `agentcore eval batch-evaluation`, wiring it into the eval TUI alongside `evaluator` and `online-eval`:

- **`list`** — a paginated, searchable table of batch evaluations (`BatchEvaluationPicker` over `PaginatedTablePicker`). Enter on a row opens that job.
- **`get`** — a raw-JSON detail screen (`JsonDetail`). No metadata hub: the value is the full response (job metadata + per-session `results`), which the JSON shows cleanly. Unwraps `getBatchEvaluation`'s `{ detail }` (metadata + merged CloudWatch results).
- **menu** — `BatchEvaluationScreen` (`RouterScreen`). This route is **mandatory**: the eval menu already lists `batch-evaluation`, and without a route the entry falls through to the `HelpScreen` catch-all and exits the app.

A bare `agentcore eval batch-evaluation` now opens the TUI (`renderTui` + `withTuiOnEmptyFlagsAndArgs`) instead of printing help.

## Demo

batch-evaluation-tui-demo

## Files

- `src/components/BatchEvaluationPicker.tsx` — list body (columns/`toRow`/`loadPage`), mirrors `OnlineEvalPicker`.
- `src/handlers/eval/batch-evaluation/list/screen.tsx` — `BatchEvaluationListScreen`; enter → get.
- `src/handlers/eval/batch-evaluation/get/screen.tsx` — `BatchEvaluationGetJsonScreen` + `useBatchEvaluationDetail`.
- `src/handlers/eval/batch-evaluation/screen.tsx` — `BatchEvaluationScreen` menu.
- `index.tsx` re-wired to `renderTui`; `Root.tsx` gets the menu + list + bare-get-redirect + `get/:id` routes.

## Testing

- `batch-evaluation.screen.test.tsx` — menu (get/list), picker (rows, Core-options, bare-get redirect, selection → JSON, empty), detail (full JSON incl. merged results, CloudWatch-failure-degrades, retry).
- Full suite: **752 pass**, typecheck + lint clean.

### DIFF
--- a/src/handlers/eval/batch-evaluation/get/screen.tsx
+++ b/src/handlers/eval/batch-evaluation/get/screen.tsx
@@ -1,0 +1,37 @@
+import { useQuery } from "@tanstack/react-query";
+import { useParams } from "react-router";
+import { JsonDetail } from "../../../../components/JsonDetail";
+import type { ScreenProps } from "../../../types";
+import { coreOptsFromCtx } from "../../../utils";
+
+// getBatchEvaluation already fetches the job and merges the per-session CloudWatch
+// results, returning `{ detail, resultsError? }`. The screen unwraps `detail`;
+// `resultsError` is ignored here — a CloudWatch read failure just omits `results`
+// (the CLI is the surface that warns on stderr).
+function useBatchEvaluationDetail({ ctx, core }: ScreenProps, id: string | undefined) {
+  const opts = coreOptsFromCtx(ctx);
+  return useQuery({
+    queryKey: ["batch-evaluation", opts.region, id],
+    queryFn: async () => (await core.eval.getBatchEvaluation(id!, opts)).detail,
+    enabled: id !== undefined,
+  });
+}
+
+// Batch-evaluation get is raw JSON only — no metadata hub. The value is the full
+// response (job metadata + per-session results), which the JSON already shows
+// cleanly; a curated field subset would just hide data.
+export function BatchEvaluationGetJsonScreen(props: ScreenProps) {
+  const { batchEvaluationId } = useParams();
+  const detail = useBatchEvaluationDetail(props, batchEvaluationId);
+
+  return (
+    <JsonDetail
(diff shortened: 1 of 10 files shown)


src/handlers/eval/batch-evaluation/get/screen.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+// getBatchEvaluation already fetches the job and merges the per-session CloudWatch
+// results, returning `{ detail, resultsError? }`. The screen unwraps `detail`;
+// `resultsError` is ignored here — a CloudWatch read failure just omits `results`
+// (the CLI is the surface that warns on stderr).
+function useBatchEvaluationDetail({ ctx, core }: ScreenProps, id: string | undefined) {
+  const opts = coreOptsFromCtx(ctx);
+  return useQuery({
+    queryKey: ["batch-evaluation", opts.region, id],
+    queryFn: async () => (await core.eval.getBatchEvaluation(id!, opts)).detail,
```
aidandaly24: I may be missing an intended TUI-specific behavior, but should we preserve `resultsError` here instead of dropping it? A CloudWatch access or read failure currently renders normal job JSON without `results`, so the user cannot distinguish incomplete data from a job that simply has no results yet. The headless command warns for this reason. Would it make sense to return both values from the query and show a persistent warning above the JSON while still preserving the metadata?

aidandaly24 (dismissed):
One comment though it can be a follow up

src/handlers/eval/batch-evaluation/get/screen.tsx
jariy17: I'll update now to fix it.

aidandaly24 (approved):

nborges-aws (approved):

PR #1940, opened 2026-08-06
feat: centralize support TUI command routing logic

nborges-aws:
## Problem
Some command groups have read-only TUIs while supporting both read and write operations through the CLI.

Previously, write commands were hidden directly in the `RouterScreen`. This filtering happens only after the TUI has already launched. Because `withTuiOnEmptyFlagsAndArgs` runs earlier, invoking an omitted write command without flags still launched Ink instead of reaching its normal CLI validation or help behavior.

## Solution

Add `Router.supportedTuiCommands()` as the source of truth for TUI menu visibility and bare-command dispatch. Also, migrated the read-only TUIs which were filtering commands to use this new structure. 

The allowlist is propagated through the Commander command tree and used by both:
- `withTuiOnEmptyFlagsAndArgs`, to determine whether a bare command should launch the TUI.
- `RouterScreen`, to determine which commands should appear in TUI menus.

Commands excluded from the TUI remain registered in Commander and fully available through the CLI. Additionally, any Router that omits .supportedTuiCommands() will include all its command keys. 

## Description
- Add `Router.supportedTuiCommands()`.
- Add `isTuiCommandSupported()` for consumers of compiled commands.
- Update `withTuiOnEmptyFlagsAndArgs` to delegate unsupported commands to their normal CLI handlers.
- Update `RouterScreen` to filter using router metadata.
- Remove the `RouterScreen` `omit` prop.
- Migrated read-only Eval and Identity TUIs to the new API.
- Add test coverage for menu filtering, nested propagation, backward compatibility, and bare CLI-only commands.

## Related Issue

Closes #

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

`bun run test` (825 pass, 0 fail)

- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.

### DIFF
--- a/src/router/router.tsx
+++ b/src/router/router.tsx
@@ -24,13 +24,36 @@ export const GlobalConfigAccessorKey: ContextKey<GlobalConfigAccessor> =
   contextKey<GlobalConfigAccessor>("globalConfigAccessor");
 export const ProjectKey = contextKey<Project>("project");
 
-// Allowlist of commands supported in TUI. Ensures read-only TUIs do not
-// render write commands, and ensures bare write commands do not launch TUI
-const supportedTuiCommandNames = new WeakMap<Handler, ReadonlySet<string>>();
-const compiledTuiSupport = new WeakMap<Command, boolean>();
+// RoutedCommand keeps the compiled handler and Commander command tree together.
+// TUI consumers can therefore read handler metadata without module-level state.
+class RoutedCommand extends Command {
+  constructor(readonly handler: Handler) {
+    super(handler.name());
+  }
+}
 
 export function isTuiCommandSupported(command: Command): boolean {
-  return compiledTuiSupport.get(command) ?? true;
+  return command instanceof RoutedCommand ? command.handler.doesSupportTui() : true;
+}
+
+interface TuiChildSupportProvider {
+  supportsTuiCommand(commandName: string): boolean;
+}
+
+function isTuiChildSupportProvider(h: Handler): h is Handler & TuiChildSupportProvider {
+  return typeof (h as Partial<TuiChildSupportProvider>).supportsTuiCommand === "function";
+}
+
+function withEffectiveTuiSupport(handler: Handler, supported: boolean): Handler {
+  return {
+    name: () => handler.name(),
+    description: () => handler.description(),
+    flags: () => handler.flags(),
+    arguments: () => handler.arguments(),
+    doesSupportTui: () => supported,
+    handle: (ctx, flags, args) => handler.handle(ctx, flags, args),
+    children: () => handler.children(),
+  };
 }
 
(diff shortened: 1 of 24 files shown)


src/router/router.tsx
```diff
@@ -24,6 +24,15 @@ export const GlobalConfigAccessorKey: ContextKey<GlobalConfigAccessor> =
   contextKey<GlobalConfigAccessor>("globalConfigAccessor");
 export const ProjectKey = contextKey<Project>("project");
 
+// Allowlist of commands supported in TUI. Ensures read-only TUIs do not
+// render write commands, and ensures bare write commands do not launch TUI
+const supportedTuiCommandNames = new WeakMap<Handler, ReadonlySet<string>>();
+const compiledTuiSupport = new WeakMap<Command, boolean>();
```
AlexanderRichey: Is this a global? 🧐

src/router/router.tsx
```diff
@@ -203,6 +217,14 @@ export class Router implements Handler, MiddlewareProvider, DefaultHandlerProvid
     return this;
   }
 
+  // supportedTuiCommands limits this router's interactive menu and bare-command
+  // TUI dispatch to the named children. Without this call, every child keeps the
+  // existing TUI behavior.
+  supportedTuiCommands(...commands: string[]): this {
+    supportedTuiCommandNames.set(this, new Set(commands));
```
AlexanderRichey: No globals please! 😅 We can achieve the same thing, if we need it, by adding a new method to the `Handler` interface `doesSupportTui()`.

AlexanderRichey (dismissed):
Another idea we might pursue is to automatically show help text for unsupported write commands in the TUI. I left a comment on a different CR about this. See the code in `createHelpDefault` for how this can work.

nborges-aws:
> Another idea we might pursue is to automatically show help text for unsupported write commands in the TUI. I left a comment on a different CR about this. See the code in `createHelpDefault` for how this can work.

@AlexanderRichey I think this is a good idea and worth a follow up. I had proposed something similar to this with a generic screen to show the corresponding CLI command. I do think we still need the changes in this PR (pending me refactoring globals to handler), as in my mind they're separate flows. 

1. A user already browsing the TUI could be shown a generic CLI-help view for unsupported commands
2. A user invoking a CLI-only command in a bare state should remain in CLI, and receive help via CLI output

Otherwise, a bare CLI-only command would launch a TUI screen telling the user to return to using the CLI. Which feels wrong to me.

src/router/router.tsx
nborges-aws: Updated PR with this logic. Removed globals and added handler method for `doesSupportTui()`

aidandaly24 (approved):
This looks good to me no comments

jariy17 (approved):

PR #1941, opened 2026-08-06
fix(memory): expand record TUI columns

notgitika:
Rebalances record-list columns so record IDs take remaining space while content, strategy IDs, and timestamps retain useful widths. Adds a wide-terminal regression test covering the reported record layout.


notgitika:
added into #1908

PR #1942, opened 2026-08-06
feat(project): port complete project schema

tejaskash:
## Summary

- Port the complete v1 project schema from the existing CLI into resource-owned Zod modules.
- Expose the schema through `src/core/project/schema.ts` and derive project handler types from it.
- Preserve custom cross-resource validation while sharing repeated rules such as unique resource names.
- Add focused tests only for owned custom validation and helper behavior.

## Validation

- `bun test` (965 pass)
- `bun test src/core/project/schema` (77 pass)
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- `bun run build`

### DIFF
--- a/src/core/project/schema/ab-test.ts
+++ b/src/core/project/schema/ab-test.ts
@@ -1,0 +1,111 @@
+import { z } from "zod";
+export const ABTestNameSchema = z
+  .string()
+  .min(1, "Name is required")
+  .max(48)
+  .regex(
+    /^[a-zA-Z][a-zA-Z0-9_]{0,47}$/,
+    "Must begin with a letter and contain only alphanumeric characters and underscores (max 48 chars)",
+  );
+export const ABTestDescriptionSchema = z.string().min(1).max(200).optional();
+export const ABTestModeSchema = z
+  .enum(["config-bundle", "target-based"])
+  .optional()
+  .default("config-bundle");
+export type ABTestMode = z.infer<typeof ABTestModeSchema>;
+export const VariantNameSchema = z.enum(["C", "T1"]);
+export const VariantWeightSchema = z.number().int().min(1).max(100);
+export const ConfigurationBundleRefSchema = z.object({
+  bundleArn: z.string().min(1),
+  bundleVersion: z.string().min(1),
+});
+export type ConfigurationBundleRef = z.infer<typeof ConfigurationBundleRefSchema>;
+export const TargetRefSchema = z.object({
+  targetName: z.string().min(1).max(100),
+});
+export type TargetRef = z.infer<typeof TargetRefSchema>;
+const ConfigBundleVariantConfigSchema = z.object({
+  configurationBundle: ConfigurationBundleRefSchema,
+  target: z.never().optional(),
+});
+const TargetVariantConfigSchema = z.object({
+  configurationBundle: z.never().optional(),
+  target: TargetRefSchema,
+});
+export const VariantConfigurationSchema = z.union([
+  ConfigBundleVariantConfigSchema,
+  TargetVariantConfigSchema,
+]);
+export type VariantConfiguration = z.infer<typeof VariantConfigurationSchema>;
+export const ABTestVariantSchema = z.object({
+  name: VariantNameSchema,
+  weight: VariantWeightSchema,
+  variantConfiguration: VariantConfigurationSchema,
+});
+export type ABTestVariant = z.infer<typeof ABTestVariantSchema>;
+export const PerVariantOnlineEvaluationConfigSchema = z.object({
+  treatmentName: VariantNameSchema,
+  onlineEvaluationConfigArn: z.string().min(1),
+});
+export type PerVariantOnlineEvaluationConfig = z.infer<
+  typeof PerVariantOnlineEvaluationConfigSchema
+>;
+export const ABTestEvaluationConfigSchema = z.union([
+  z.object({ onlineEvaluationConfigArn: z.string().min(1) }),
+  z.object({
+    perVariantOnlineEvaluationConfig: z.array(PerVariantOnlineEvaluationConfigSchema).length(2),
+  }),
+]);
+export type ABTestEvaluationConfig = z.infer<typeof ABTestEvaluationConfigSchema>;
+export const GatewayFilterSchema = z.object({
+  targetPaths: z.array(z.string().min(1).max(500)).max(1),
+});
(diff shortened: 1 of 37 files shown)


src/core/project/schema/gateway.ts
```diff
@@ -0,0 +1,786 @@
...
+export const AgentCoreMcpRuntimeToolSchema = z
+  .object({
+    name: z.string().min(1),
+    toolDefinition: ToolDefinitionSchema,
+    compute: AgentCoreRuntimeComputeConfigSchema,
+    bindings: z.array(McpRuntimeBindingSchema).optional(),
+  })
+  .strict();
+export type AgentCoreMcpRuntimeTool = z.infer<typeof AgentCoreMcpRuntimeToolSchema>;
+export interface AgentCoreMcpSpec {
```
aidandaly24, line 782: I couldn't find a current consumer for `AgentCoreMcpSpec`, and these fields also appear in `AgentCoreProjectSpecSchema`. Would it make sense to remove this for now to avoid maintaining both shapes? If not one other thing is this should probably be a type instead of an interface, would fit since this represents concrete data.

src/core/project/schema/mcp-defs.ts
```diff
@@ -0,0 +1,45 @@
...
+export const SchemaPrimitiveTypeSchema = z.enum([
+  "string",
+  "number",
+  "object",
+  "array",
+  "boolean",
+  "integer",
+]);
+export type SchemaPrimitiveType = z.infer<typeof SchemaPrimitiveTypeSchema>;
+export interface SchemaDefinition {
```
aidandaly24, line 11: I think this should also probably be a type not an interface.

src/core/project/schema/gateway.ts
```diff
@@ -0,0 +1,786 @@
...
+              }),
+            )
+            .optional(),
+        }),
+      )
+      .optional(),
+    passthrough: PassthroughConfigSchema.optional(),
+  })
+  .strict()
+  .superRefine((data, ctx) => {
```
aidandaly24, line 310: I noticed that the non-applicable-field checks do not cover every configuration field. For example, a `lambda` target with valid `compute` and `toolDefinitions` still passes when it also includes an `apiGateway` block. Similarly, an `apiGateway` target accepts `schemaSource`. It would probably make sense to define the allowed configuration fields for each target type and reject the remaining fields, what do you think?

This behavior is inherited from the legacy schema, so it may be better treated as a follow-up if this PR wants to remain a strict port.

src/core/project/schema/harness.ts
```diff
@@ -0,0 +1,567 @@
...
+      if (TOOL_TYPES_REQUIRING_CONFIG.has(tool.type)) {
+        ctx.addIssue({
+          code: "custom",
+          message: `Tool type "${tool.type}" requires a "${expectedKey}" config`,
+          path: ["config"],
+        });
+      }
+      return;
+    }
+    const configKeys = Object.keys(tool.config);
```
aidandaly24, line 245: I think this check may run after the union has already removed additional configuration keys. For example, I tried a config containing both `remoteMcp` and `agentCoreBrowser` it succeeds and the parsed result silently drops `agentCoreBrowser`, so `configKeys.length` is still one here. Since this check appears intended to enforce exactly one configuration type, would making the union branches strict help ensure the extra configuration is rejected instead?

This behavior is consistent with the old CLI, but is probably still a bug.

src/core/project/schema/ab-test.ts
```diff
@@ -0,0 +1,111 @@
...
+});
+export type ABTestVariant = z.infer<typeof ABTestVariantSchema>;
+export const PerVariantOnlineEvaluationConfigSchema = z.object({
+  treatmentName: VariantNameSchema,
+  onlineEvaluationConfigArn: z.string().min(1),
+});
+export type PerVariantOnlineEvaluationConfig = z.infer<
+  typeof PerVariantOnlineEvaluationConfigSchema
+>;
+export const ABTestEvaluationConfigSchema = z.union([
```
aidandaly24, line 53: I think this is also a bug from the old CLI. Supplying both `onlineEvaluationConfigArn` and `perVariantOnlineEvaluationConfig` currently succeeds and silently drops the per-variant configuration because the first union branch strips it. `VariantConfigurationSchema` above uses `z.never().optional()` to prevent this. Would it make sense to use the same pattern here, or make both branches strict?

jariy17 (approved):
Looks good to me. I validated with the smithy models and the cdk package and they're are no issues introduced by this PR.

src/core/project/schema/harness.ts
```diff
@@ -0,0 +1,567 @@
...
+  })
+  .superRefine((data, ctx) => {
+    if (data.containerUri !== undefined && data.dockerfile !== undefined) {
+      ctx.addIssue({
+        code: "custom",
+        message: "containerUri and dockerfile are mutually exclusive",
+        path: ["containerUri"],
+      });
+    }
+    if (data.networkMode === "VPC" && !data.networkConfig) {
```
aidandaly24, line 496: This is a bug from the old CLI as well and appears to be out of sync with the CDK contract. `HarnessSpecSchema` accepts a Dockerfile build in VPC mode without `networkConfig.vpcId`, while the generated CDK app's L3 CDK version in the schema rejects it, and `toBuildNetwork()` throws because CodeBuild cannot infer the VPC from subnet IDs. We could probably just do this refinement here so this schema does not accept a configuration that later fails during CDK synthesis.

aidandaly24 (changes requested):
There's a few bugs here that are consistent behavior with the old CLI. I feel like we may as well fix them now since they're pretty simple. Let me know what you think also open to a follow up. Others are just mainly the interfaces that should be types for consistency with the code rules that we have been adopting.

AlexanderRichey (approved):
Let's merge this in to unblock dev. @tejaskash you may want to check out @aidandaly24 's comments cut a follow up PR if necessary.

src/core/project/schema/ab-test.ts
jariy17: This is a leftover from when AB-tests were a deployable resource. Now that AB-tests are jobs rather than part of agentcore.json, this fix isn't needed. We're leaving it in for now to avoid introducing a breaking schema change.

aidandaly24 (approved):

src/core/project/schema/ab-test.ts
aidandaly24: Ahhh that makes sense yeah disregard this one.

PR #1945, opened 2026-08-07
feat: add config-bundle CLI commands

nborges-aws:
## Description
Adds imperative CRUDL and version listing commands for AgentCore evaluation configuration bundles:

* `eval config-bundle create`
* `eval config-bundle get`
* `eval config-bundle list`
* `eval config-bundle update`
* `eval config-bundle delete`
* `eval config-bundle version list`

## Summary of changes
* `--components` accepts inline JSON, `file://<path>`, or stdin (`-`) via `SourceResolver`.
* Component input parsed and validated via `parseJsonFlagWithSchema` utility.
* `get` calls the latest-bundle API by default and the version API when optional `--version` is provided.
* `list` returns configuration bundles, while `version list` returns all versions belonging to one bundle.
* `update` requires `--id`, `--components`, and `--commit-message`; `--kms-key-arn` remains optional. This counteracts the service API docs and smithy models, but is the confirmed validation structure in the service itself
* Updates fetches the latest bundle version and pass it as the sole `parentVersionIds` entry, as required by service lineage validation.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Manually smoke tested all commands against a configuration bundle in my personal account:
* Creating a bundle from a components file
* Listing bundles
* Getting the latest bundle
* Getting a specific version
* Updating components and creating a new version
* Listing bundle versions
* Deleting the bundle

- [x] `bun run test` (993 pass, 0 fail)

- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.

### DIFF
--- a/src/core/configBundle.test.ts
+++ b/src/core/configBundle.test.ts
@@ -67,11 +67,14 @@ describe("EvalClient configuration bundles", () => {
       return {};
     });
 
-    await client.getConfigurationBundle("b-1", undefined, OPTIONS);
-    await client.getConfigurationBundle("b-1", "v-2", OPTIONS);
+    await client.getConfigurationBundle("b-1", undefined, "review-branch", OPTIONS);
+    await client.getConfigurationBundle("b-1", "v-2", "mainline", OPTIONS);
 
     expect(sent[0]).toBeInstanceOf(GetConfigurationBundleCommand);
-    expect((sent[0] as GetConfigurationBundleCommand).input).toEqual({ bundleId: "b-1" });
+    expect((sent[0] as GetConfigurationBundleCommand).input).toEqual({
+      bundleId: "b-1",
+      branchName: "review-branch",
+    });
     expect(sent[1]).toBeInstanceOf(GetConfigurationBundleVersionCommand);
     expect((sent[1] as GetConfigurationBundleVersionCommand).input).toEqual({
       bundleId: "b-1",
@@ -118,6 +121,7 @@ describe("EvalClient configuration bundles", () => {
       await client.updateConfigurationBundle(
         "b-1",
         {
+          branchName: "review-branch",
           components: COMPONENTS,
           commitMessage: "Replace order support configuration",
           kmsKeyArn: "arn:aws:kms:us-west-2:123456789012:key/new",
@@ -128,10 +132,14 @@ describe("EvalClient configuration bundles", () => {
 
     expect(sent).toHaveLength(2);
     expect(sent[0]).toBeInstanceOf(GetConfigurationBundleCommand);
-    expect((sent[0] as GetConfigurationBundleCommand).input).toEqual({ bundleId: "b-1" });
+    expect((sent[0] as GetConfigurationBundleCommand).input).toEqual({
+      bundleId: "b-1",
+      branchName: "review-branch",
+    });
     expect(sent[1]).toBeInstanceOf(UpdateConfigurationBundleCommand);
     expect((sent[1] as UpdateConfigurationBundleCommand).input).toEqual({
       bundleId: "b-1",
+      branchName: "review-branch",
       components: COMPONENTS,
       commitMessage: "Replace order support configuration",
       kmsKeyArn: "arn:aws:kms:us-west-2:123456789012:key/new",
@@ -149,6 +157,7 @@ describe("EvalClient configuration bundles", () => {
     const promise = client.updateConfigurationBundle(
       "b-1",
       {
+        branchName: "mainline",
         components: COMPONENTS,
         commitMessage: "Replace order support configuration",
         kmsKeyArn: "arn:aws:kms:us-west-2:123456789012:key/new",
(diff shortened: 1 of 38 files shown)


src/core/eval.tsx
```diff
@@ -465,6 +481,83 @@ export class EvalClient implements CoreEvalClient {
...
+      .send(new ListConfigurationBundlesCommand({ nextToken, maxResults }));
+  }
+
+  async updateConfigurationBundle(
+    id: string,
+    update: UpdateConfigurationBundleInput,
+    options: CoreOptions,
+  ): Promise<UpdateConfigurationBundleResponse> {
+    const control = this.clients.control(toClientConfig(options));
+    const current = await control.send(new GetConfigurationBundleCommand({ bundleId: id }));
```
aidandaly24, line 522: I may be missing the intended branch behavior, but can we select the branch explicitly before choosing the parent version? I live-tested a mainline v1 followed by a `review-branch` v2. `GetConfigurationBundle` without `branchName` returned v2, and this CLI update then created v3 on `review-branch`. Passing `branchName: "mainline"` correctly returned v1. Would it make sense to expose `--branch-name`, default it to `mainline`, and pass it to both Get and Update so an ordinary update cannot silently continue whichever branch was modified most recently?

src/handlers/eval/types.tsx
```diff
@@ -102,6 +111,14 @@ export type RoleScopeWarning = {
 };
 
 export type CreateDatasetInput = CreateDatasetRequest;
+export type CreateConfigurationBundleInput = Pick<
+  CreateConfigurationBundleRequest,
+  "bundleName" | "components" | "kmsKeyArn"
+>;
```
aidandaly24, line 117: Are these field exclusions intentional? The live service accepted `description`, `branchName`, `commitMessage`, `createdBy`, and `tags` during create, and accepted `description` plus `branchName` during update when the required components, commit message, and parent were supplied. The SDK also models `bundleName` and explicit `parentVersionIds`. Would it make sense to expose the common scalar fields and typed JSON for the structured ones? Also definitely get if you have reasoning behind why these are not included.

src/core/eval.tsx
nborges-aws: Agreed it makes sense to have branch name here. Good catch. I've added the flag and updated tests around it

src/handlers/eval/types.tsx
nborges-aws: These exclusions were intentional for first pass implementation. I think we want to avoid adding a bespoke flag for each of these options, but could be supported in typed JSON perhaps. I will discuss with TJ and see if this something we want to pursue in a follow up PR.

src/handlers/eval/config-bundle/create/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { resolveConfigurationBundleComponents } from "../components";
+
+export const createCreateConfigBundleHandler = (core: Core, io: AppIO) =>
+  createHandler({
+    name: "create",
+    description: "create a configuration bundle and its initial immutable version",
+    flags: [
+      flag("name", "the name of the configuration bundle", z.string().optional()),
+      flag(
+        "components",
```
aidandaly24, line 17: Should this be sensitive?

aidandaly24 (dismissed):

src/core/configBundle.test.ts
```diff
@@ -0,0 +1,207 @@
+import { describe, expect, test } from "bun:test";
```
jariy17, line 1: Shouldn't the handler's golden tests cover these core functions?

src/core/eval.tsx
```diff
@@ -601,6 +617,87 @@ export class EvalClient implements CoreEvalClient {
...
+  async updateConfigurationBundle(
+    id: string,
+    update: UpdateConfigurationBundleInput,
+    options: CoreOptions,
+  ): Promise<UpdateConfigurationBundleResponse> {
+    const control = this.clients.control(toClientConfig(options));
+    const current = await control.send(
+      new GetConfigurationBundleCommand({ bundleId: id, branchName: update.branchName }),
+    );
+    if (!current.versionId) {
```
jariy17, line 662: According to the internal service model, this is required. I don't know why aws sdk v3 has all required response parameters as | undefined.

src/handlers/eval/config-bundle/config-bundle.test.tsx
```diff
@@ -0,0 +1,473 @@
+import { afterEach, describe, expect, test } from "bun:test";
```
jariy17, line 1: SHould this be a golden test?

src/handlers/eval/types.tsx
jariy17: I don't see a user setting a `createdBy` by I can see customers wanting to set branchName, commitMessage. We can do this as Follow up PR.

jariy17 (changes requested):
Looks good but needs golden tests.

src/core/eval.tsx
nborges-aws: Yeah I saw the field was required, but the SDK type has `string | undefined`. I added this check to follow the service model

jariy17 (approved):
Unit tests could changed as a follow up

aidandaly24 (approved):
Thanks for making the updates!

PR #1955, opened 2026-08-07
feat(memory): add actor and session lists

notgitika:
This PR adds first-class actor and session list commands for AgentCore Memory.

Adds `agentcore memory actor list` and `agentcore memory session list`, with required selectors and pagination

Also adds corresponding TUI flows from the top-level Memory menu and a selected Memory's detail actions.

Reuses the actor/session picker components for the existing event flow.

Covers command behavior, direct TUI navigation, pagination, empty/retry states, and the existing event/record flows.

I did targeted memory tests command line + TUI. typecheck, lint, formatting, and bundle all pass.


aidandaly24 (approved):

jariy17 (approved):

jariy17:
Nice work! My worry is that adding actor and session to the top-level memory menu crowds it — customers still need to pick a memory first for those to be meaningful. Should it live under the selected memory rather than alongside it?

PR #1956, opened 2026-08-07
feat: add read-only Gateway TUI

aidandaly24:
## Summary

- launch bare Gateway branches and read leaves into the TUI through the shared Handler tree while preserving flag-driven headless JSON behavior
- align TUI routes with the CLI command hierarchy for Gateway, Target, Connector, and Rule `get`/`list`
- use staged Gateway selection for nested resources, with direct bare `get` routes redirecting through the corresponding list flow
- present Connector as a separate resource experience while continuing to use Gateway Target APIs internally
- keep parent selection and scoped resource listing in dedicated Gateway, Target, Connector, and Rule picker components
- centralize Connector get/list filtering and validation in `GatewayClient`, with CLI and TUI consumers delegating through `CoreGatewayClient`
- fill Connector pages to `maxResults`, cap discovery at 101 Target pages, and use lookahead so `nextToken` is returned only when another Connector exists
- use router-level `supportedTuiCommands` so read commands enter the TUI while bare CLI-only create commands run normal validation
- reuse the established router menu, paginated table, resource detail, JSON, loading, retry, empty, keyboard, and responsive layout components

## Testing

- `bun test src/handlers/gateway/gateway.test.tsx src/handlers/gateway/gateway.screen.test.tsx src/handlers/gateway/gateway.fixture.test.tsx src/core/gateway.test.ts src/components/ui/data-table/columnWidths.test.ts` (`88` pass)
- `bun test --coverage --coverage-reporter=text` (`1081` pass; `93.73%` functions, `97.88%` lines)
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- `bun run build`
- `npm pack --ignore-scripts`
- help smoke for Gateway root plus all eight `get`/`list` leaves (`12/12`)
- TUI harness at `100x30`, confirming Gateway and Connector menus expose read-only commands only
- deploy-account pagination probe: reused a `ListGatewayTargets` token created with `maxResults=2` in a request with `maxResults=1`, returning the next distinct Target
- `RECORD=1 AWS_PROFILE=e2e-test bun test src/handlers/gateway/gateway.fixture.test.tsx -t 'lists Gateway Connectors'`

The full source run retains seven local `src/io/exec.test.ts` failures plus three associated errors on files byte-identical to `origin/refactor`.


src/core/gateway.tsx
```diff
@@ -98,6 +102,41 @@ export class GatewayClient implements CoreGatewayClient {
...
+  async listGatewayConnectors(
+    gatewayId: string,
+    nextToken: string | undefined,
+    maxResults: number | undefined,
+    options: CoreOptions,
+  ): Promise<ListGatewayTargetsResponse> {
+    const items: TargetSummary[] = [];
+    let token = nextToken;
+
+    while (true) {
```
jariy17: This is unbounded scan, we should throw a truncationError like [this](https://github.com/aws/agentcore-cli/pull/1964).  Should be more than gateway_limit [(100)](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/bedrock-agentcore-limits.html#gateway-endpoints-quotas)

aidandaly24:
### Gateway read-only TUI demo

Gateway read-only TUI demo

[Download the MP4](https://gist.githubusercontent.com/aidandaly24/fb44643df48f63b3d0408b070e4bcc76/raw/fb1d8c37aa70f449f50af7b9aa1ed12c384e4a01/agentcore-gateway-read-tui-demo.mp4)

Recorded with the real TUI components and synthetic Gateway/Connector data. No AWS account or credential metadata is included.

src/core/gateway.tsx
jariy17: ▎ Flagging a second issue in this block, separate from the unbounded scan.
  ▎
  ▎ Early stop. const full = maxResults === undefined ? items.length > 0 : .... On the no-maxResults path, items.length > 0 returns after the first Target page that yields any connector. So a Gateway with 30 connectors can return just the 2 that
  ▎ landed in page 1. Every other list returns a full service page. This one doesn't.
  ▎
  ▎ Wrong token. return { ...response, items } returns response.nextToken, which is a ListGatewayTargets token. It means more Targets remain, not more Connectors. The TUI surfaces a "more →" affordance whenever a token is present, so a user can
  ▎ page forward into an empty Connector screen while connectors still exist deeper in the Target stream.
  ▎
  ▎ Fix: treat the undefined case like the numeric one. Fill to a default page size instead of stopping at the first hit, so the page and its token stay connector-aligned. At minimum, add a comment that the token paginates Targets, not
  ▎ Connectors.

jariy17 (dismissed):
Could be a follow up pr? It's pretty god

src/core/gateway.tsx
aidandaly24: Fixed in 24f49f82. Connector discovery is now capped at 101 Target pages (one above the documented default 100-target Gateway quota) and throws the merged ResultTruncationError instead of scanning indefinitely or returning partial results. I added an ever-advancing-token regression test that verifies the cap.

aidandaly24: Fixed in 24f49f82. An omitted maxResults now fills a 100-row logical Connector page instead of stopping at the first hit. After a page fills, Core looks ahead and returns the service token used to fetch the first later Target page known to contain a Connector; if none exists, nextToken is omitted. This keeps AWS tokens, avoids buffered/custom token state, and prevents a false "more" page. Added focused tests for default filling, aligned lookahead, and exhausted lookahead.

src/components/GatewayRulePicker.tsx
```diff
@@ -0,0 +1,77 @@
+import type { GatewayRuleDetail } from "@aws-sdk/client-bedrock-agentcore-control";
+import { useNavigate } from "react-router";
+import type { ScreenProps } from "../handlers/types";
+import { coreOptsFromCtx } from "../handlers/utils";
+import { PaginatedTablePicker } from "./PaginatedTablePicker";
+import type { DataTableColumn } from "./ui/data-table";
+
+interface GatewayRuleRow extends Record<string, unknown> {
```
Hweinstock, line 8: wow, all these common components make this so easy to read!

src/core/gateway.tsx
```diff
@@ -98,6 +105,55 @@ export class GatewayClient implements CoreGatewayClient {
...
+    nextToken: string | undefined,
+    maxResults: number | undefined,
+    options: CoreOptions,
+  ): Promise<ListGatewayTargetsResponse> {
+    const pageSize = maxResults ?? DEFAULT_CONNECTOR_PAGE_SIZE;
+    const items: TargetSummary[] = [];
+    let token = nextToken;
+    let filling = true;
+
+    for (let page = 0; page < MAX_CONNECTOR_TARGET_PAGES; page++) {
```
Hweinstock, line 131: does the api not support server side filtering? if so that feels like a major gap, especially if this is a common customer use-case.

src/handlers/gateway/gateway.test.tsx
```diff
@@ -84,273 +53,47 @@ describe("gateway command hierarchy", () => {
...
+    ["Target list", ["gateway", "target", "list"]],
+    ["Connector", ["gateway", "connector"]],
+    ["Connector get", ["gateway", "connector", "get"]],
+    ["Connector list", ["gateway", "connector", "list"]],
+    ["Rule", ["gateway", "rule"]],
+    ["Rule get", ["gateway", "rule", "get"]],
+    ["Rule list", ["gateway", "rule", "list"]],
+  ] as const)("opens the TUI for a bare %s command", async (_label, args) => {
+    await expect(run([...args])).rejects.toThrow(
+      "interactive mode requires a TTY on stdin and stdout",
```
Hweinstock: is there a way to make this test more explicit? Right now we're relying on the behavior that tty is required for this test to work.

src/core/gateway.tsx
aidandaly24: It doesn't support it at the moment. We could probably request Gateway do it. But I wonder if since they haven't, maybe it really isn't that useful to the customer...

This could be removed for just gateway target list/get. I think that would make sense. But we wanted to separate connector for CUD because it is helpful for containing flags to specific use cases.

src/handlers/gateway/gateway.test.tsx
aidandaly24: Good call. Fixed in 9d530a5e. The Gateway test now compiles the real command tree and asserts isTuiCommandSupported directly for every read path, plus false for each CLI-only create path. The create cases still assert normal handler validation, while the generic TTY dispatch behavior remains owned by the middleware unit test.

jariy17 (approved):

src/core/gateway.tsx
```diff
@@ -27,6 +31,9 @@ import type {
 import type { AwsClients, CoreOptions } from "./types";
 import { toClientConfig } from "./utils";
 
+const DEFAULT_CONNECTOR_PAGE_SIZE = 100;
+const MAX_CONNECTOR_TARGET_PAGES = 101;
```
notgitika, line 35: it's unlikely but just to test the bounds: the code assumes a GW can never have more than 100 Targets, but customers can increase that quota. connectors after the first 101 pages will not be found then, right?

src/core/gateway.tsx
```diff
@@ -98,6 +105,55 @@ export class GatewayClient implements CoreGatewayClient {
...
+    const pageSize = maxResults ?? DEFAULT_CONNECTOR_PAGE_SIZE;
+    const items: TargetSummary[] = [];
+    let token = nextToken;
+    let filling = true;
+
+    for (let page = 0; page < MAX_CONNECTOR_TARGET_PAGES; page++) {
+      const requestToken = token;
+      const requestSize = filling ? pageSize - items.length : DEFAULT_CONNECTOR_PAGE_SIZE;
+      const response = await this.listGatewayTargets(gatewayId, token, requestSize, options);
+      const connectors = (response.items ?? []).filter(
```
notgitika, line 135: with `--max-results 1`, the cli may make one API call per target which means that a GW with 100 non-connector targets could cause 100 calls, right? this would be too slow/throttled, right?

notgitika (approved):
added 2 minor edge case related comments but otherwise this PR looks good to me, nice!

src/core/gateway.tsx
aidandaly24: Addressed in follow-up PR #1979. Connector discovery now requests up to 1000 Targets per scan, so the 101-request safety cap is no longer based on the adjustable default 100-Target quota. It remains explicitly bounded and still throws ResultTruncationError rather than returning partial results.

src/core/gateway.tsx
aidandaly24: Addressed in follow-up PR #1979. Connector discovery now scans up to 1000 Targets per request even when `--max-results 1`; if that scan contains extra Connectors, Core replays only the exact prefix needed to produce an aligned AWS nextToken. This avoids one call per non-Connector Target without adding custom cursor state.

PR #1962, opened 2026-08-10
feat(dev): add container dev runner

tejaskash:
## Summary
- add a streamed `ContainerDevRunner` for Docker, Podman, and Finch builds
- validate build inputs, protect widened contexts with `.dockerignore`, and map protocol-specific loopback ports
- keep environment and build-argument values out of argv while explicitly cleaning up project-scoped containers
- cover runtime selection, command construction, validation, cancellation, cleanup, and failure sequencing with fake process tests

## Stack
- depends on #1883
- intentionally targets `feat/dev-server-core`; retarget to `refactor` after #1883 merges
- handler wiring and AWS credential materialization remain out of scope

## Verification
- `bun test` (1013 tests)
- `bun test src/core/dev/` (30 tests)
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- `bun run build`


src/core/dev/container.ts
```diff
@@ -0,0 +1,202 @@
...
+}
+
+function ensureBuildContextDockerignore(context: string): string | undefined {
+  const dockerignore = join(context, ".dockerignore");
+  if (existsSync(dockerignore)) return undefined;
+  writeFileSync(dockerignore, BUILD_CONTEXT_DOCKERIGNORE);
+  return dockerignore;
+}
+
+function projectIdentifier(projectRoot: string): string {
```
Hweinstock: nit: should this be `hashString`? I don't see any project specific logic here.

src/core/dev/container.ts
```diff
@@ -0,0 +1,202 @@
...
+      if (dockerignore) {
+        yield { type: "status", message: `Created protective ${dockerignore}` };
+      }
+    }
+
+    const runtimeName = input.runtime.name.toLowerCase();
+    const projectId = projectIdentifier(input.projectRoot);
+    const imageTag = `agentcore-dev/${runtimeName}-${projectId}`;
+    const containerName = `agentcore-dev-${runtimeName}-${projectId}`;
+    input.signal.throwIfAborted();
```
Hweinstock: should this be after the await? My understanding is that we only need to re-check when our code is suspended in a promise.

src/core/dev/container.ts
```diff
@@ -0,0 +1,202 @@
...
+          env: { ...process.env, ...forwardedEnv },
+          signal: input.signal,
+        },
+      );
+    } finally {
+      await this.removeContainer(tool, containerName, context);
+    }
+  }
+
+  private async resolveContainerTool(signal: AbortSignal): Promise<ContainerTool> {
```
Hweinstock, line 171: do you think its worth wiring up the logger here? I think it might be helpful to get some logs on how this tool resolution executes.

src/core/dev/container.ts
```diff
@@ -0,0 +1,202 @@
...
+    for (const tool of CONTAINER_TOOLS) {
```
Hweinstock, line 172: im not familiar with podman or finch, but wondering if its safe to silently default to use them? I wonder if the user should have some sort of control over which tool is run.

src/core/dev/container.ts
```diff
@@ -0,0 +1,202 @@
...
+  ): Promise<void> {
+    try {
+      for await (const _event of this.streamProcess([tool, "rm", "-f", containerName], {
+        cwd,
+        signal: AbortSignal.timeout(CLEANUP_TIMEOUT_MS),
+      })) {
+        // Best-effort cleanup intentionally discards command output.
+      }
+    } catch {
+      // A missing container and an unavailable daemon are both safe to ignore here.
```
Hweinstock, line 202: would a warning log be helpful here?

src/core/dev/container.test.ts
```diff
@@ -0,0 +1,524 @@
...
+    expect(build.command.join(" ")).not.toContain("hello-world");
+    expect(build.command.join(" ")).not.toContain("development");
+
+    const dockerignore = await readFile(join(root, ".dockerignore"), "utf8");
+    for (const pattern of [".env", "**/.env", "**/node_modules", "agentcore/"]) {
+      expect(dockerignore).toContain(pattern);
+    }
+  });
+
+  test.each([
```
Hweinstock, line 169: love this pattern!

src/core/dev/container.ts
tejaskash: Renamed this to `hashString` and moved `resolve(projectRoot)` to the call site, so the helper now has a genuinely generic contract.

src/core/dev/container.ts
tejaskash: Yes. Moved `throwIfAborted()` immediately after the awaited stale-container cleanup. I also added a regression that aborts while cleanup is suspended and proves build/run never starts.

src/core/dev/container.ts
tejaskash: I kept logger wiring out of this runner. Selection is already surfaced through the `Building image with ${tool}` status event, while `toolAvailable` exposes only a boolean, so probe logs would add mechanics without diagnostic detail. The handler owns the app logger and can log the consumed status stream when this runner is wired into `project dev`.

src/core/dev/container.ts
tejaskash: This is intentional and preserves the existing CLI contract: auto-detect Docker, then Podman, then Finch, using the first CLI that supports both version and build probes. The runner uses only their shared build/run/rm subset and names the selected tool in the status stream. A user override would belong in the future command-handler options rather than this runner.

src/core/dev/container.ts
tejaskash: I kept this silent intentionally. The pre-run `rm -f` normally fails when no stale container exists, so warning on the shared cleanup path would warn on every healthy first run. The process API also does not distinguish that expected case from daemon unavailability reliably; actual build/run failures still surface with command output.

src/core/dev/container.test.ts
tejaskash: Thanks. Kept the pattern and added another lifecycle regression for aborting during stale-container cleanup.

src/core/dev/container.ts
```diff
@@ -0,0 +1,202 @@
...
+    if (input.runtime.buildContextPath) {
+      const dockerignore = ensureBuildContextDockerignore(context);
+      if (dockerignore) {
+        yield { type: "status", message: `Created protective ${dockerignore}` };
+      }
+    }
+
+    const runtimeName = input.runtime.name.toLowerCase();
+    const projectId = hashString(resolve(input.projectRoot));
+    const imageTag = `agentcore-dev/${runtimeName}-${projectId}`;
```
notgitika: can we sanitize underscores when creating the image name, docker has rules here like only 2 consecutive underscore (runtimename can have many).

src/core/dev/container.ts
```diff
@@ -0,0 +1,202 @@
...
+    const containerName = `agentcore-dev-${runtimeName}-${projectId}`;
+    await this.removeContainer(tool, containerName, context);
+    input.signal.throwIfAborted();
+
+    const buildArgs = input.runtime.customDockerBuildArgs ?? {};
+    const buildArgFlags = Object.keys(buildArgs).flatMap((key) => ["--build-arg", key]);
```
notgitika: finch does not forward arbitrary env vars into its VM, so `--build-arg KEY` loses the value.

should probably be `--build-arg KEY=value`

src/core/dev/container.ts
```diff
@@ -0,0 +1,202 @@
...
+          "--name",
+          containerName,
+          "-p",
+          `127.0.0.1:${input.port}:${containerPort}`,
+          ...envNameFlags,
+          imageTag,
+        ],
+        {
+          cwd: context,
+          env: { ...process.env, ...forwardedEnv },
```
notgitika: for docker specifically, build args and app env vars are passed into Docker’s own environment. names like `DOCKER_HOST` can change which Docker daemon is used so can we keep these values separate?

src/core/dev/container.ts
```diff
@@ -0,0 +1,202 @@
...
+    }
+  }
+
+  private async resolveContainerTool(signal: AbortSignal): Promise<ContainerTool> {
+    for (const tool of CONTAINER_TOOLS) {
+      const hasVersion = await this.toolAvailable(tool);
+      signal.throwIfAborted();
+      if (!hasVersion) continue;
+
+      const canBuild = await this.toolAvailable(tool, ["build", "--help"]);
```
notgitika, line 177: while I was trying on my own system, it seems like this treats a failed `finch build --help` as an unavailable binary. it was installed but not initialized. It should instead suggest running `finch vm init`. this is also likely a bug in the current CLI but I would like to have it fixed here if possible!

notgitika (changes requested):

src/core/dev/container.ts
tejaskash: Build args are now passed as explicit `--build-arg KEY=value` arguments, including for Finch. I added a Finch-specific regression and a redacted error-command path so build failures do not print the values.

src/core/dev/container.ts
tejaskash: Separated both build args and app env from the container CLI process environment. The CLI now receives the unchanged `process.env`, while app values are passed explicitly to the container and redacted from errors. Added a `DOCKER_HOST` regression to prove the app value cannot select the Docker daemon.

src/core/dev/container.ts
tejaskash: When Finch responds to `--version` but fails the build-capability probe, the runner now raises a user-facing environment error that says to run `finch vm init`. Added a focused regression for this state.

src/core/dev/container.ts
tejaskash: Normalized image-name components by collapsing runs of 3+ underscores to the supported 2-underscore separator. Added a regression using `Hello___World`.

src/core/dev/container.ts
```diff
@@ -0,0 +1,242 @@
...
+  private readonly toolAvailable: ToolAvailable;
+
+  constructor(config: ContainerDevRunnerConfig = {}) {
+    this.streamProcess = config.streamProcess ?? streamProcess;
+    this.toolAvailable = config.toolAvailable ?? toolAvailable;
+  }
+
+  public async *run(input: DevServerInput): AsyncGenerator<DevEvent, void> {
+    input.signal.throwIfAborted();
+    const context = join(
```
aidandaly24, line 60: nit: `DirectoryPathSchema` currently accepts any nonempty string. I used `buildContextPath: ".."` and the runner built the parent directory and wrote its generated `.dockerignore` outside the project. Could we resolve the real path and require the context to remain under `projectRoot` before checking or mutating it?

aidandaly24 (approved):
One small nit that could be fixed another time otherwise this lgtm:

src/io/exec.ts
```diff
@@ -66,6 +66,8 @@ export type StreamProcessOptions = {
   cwd: string;
   env?: NodeJS.ProcessEnv;
   signal?: AbortSignal;
+  /** Command rendered in errors when the actual arguments contain sensitive values. */
+  redactedCommand?: string[];
```
aidandaly24, line 70: I like this feature a lot!

src/core/dev/container.ts
tejaskash: Addressed in #1966. Container build contexts are now canonicalized for containment checks and rejected if they resolve outside the project root, before Dockerfile checks, tool probes, or .dockerignore writes. Added regressions for both `..` traversal and an in-project symlink pointing outside.

notgitika (approved):

PR #1924, opened 2026-08-05
feat(eval): add read-only batch-evaluation CLI (get, list)

jariy17:
## What

Adds the **read-only** slice of the imperative `batch-evaluation` commands.
```
agentcore eval batch-evaluation
├── get      # retrieve one job + CloudWatch-backed per-session results
└── list     # list jobs (service-side pagination)
```

## Commands

**`eval batch-evaluation get`**
- `--id <id>` (required)
- `--disable-cw-results` — skip CloudWatch retrieval, return service-side metadata only
- Default: on a terminal job (`COMPLETED`/`COMPLETED_WITH_ERRORS`/`FAILED`/`STOPPED`) it reads per-session results from the CloudWatch output log stream. **A CloudWatch failure warns on stderr and never hides or replaces the job status** (stdout stays clean JSON).

**`eval batch-evaluation list`**
- `--max-results` / `--next-token` — service-side pagination. Returns the service page verbatim (no client-side filtering).

## How

- **New CloudWatch Logs client in `AwsClients`** — `logs()` accessor + `createLogsClient` factory, injected at the app edge and in the test fixture seam. Region-only config, mirroring the IAM-client pattern (the AgentCore endpoint override must not leak onto CloudWatch).
- **`EvalClient` (data plane)** — `getBatchEvaluation`, `listBatchEvaluations`, and `getFullBatchEvaluationResults(batchEvaluationId, logGroupName, options)` (derives the per-job result stream `run-<batchEvaluationId>`). Batch evaluation lives on `@aws-sdk/client-bedrock-agentcore`, not the control plane.
- **Result-record parser** keeps each score's `level` (`aws.bedrock_agentcore.evaluation_level`) and scope ids (`session.id`, `traceId`), read verbatim from the recorded stream shape; a malformed log line is warned and skipped, never fatal.

## Example: `get` default vs `--disable-cw-results`

Real output from a freshly-run batch evaluation (`GTProbe2`, COMPLETED, 9 sessions, 2 evaluators — one GT-requiring, one reference-free).

**Default** — fetches per-session results from the CloudWatch output stream (showing first 4 of 18 results):

```json
{
  "batchEvaluationId": "GTProbe2_1786034545579-8ffefc851e",
  "batchEvaluationName": "GTProbe2_1786034545579",
  "status": "COMPLETED",
  "createdAt": "2026-08-06T16:42:26.468Z",
  "evaluators": [
    { "evaluatorId": "Builtin.Correctness" },
[...]

PR #1964, opened 2026-08-10
fix(eval): surface batch-evaluation result truncation as a warning

jariy17:
## What

Follow-up to #1924. Addresses @aidandaly24's non-blocking review comment on `src/core/batchEvaluationResults.tsx`:

> Hitting `MAX_RESULT_PAGES` should flow back as `resultsError` instead of returning the accumulated array as complete. `GetLogEvents` is limited by 1 MB as well as 10,000 events, so large explanations can make 100 pages much less than roughly one million results. Right now the cap only writes to the file logger, while `get --json` succeeds with partial `results` and no customer-visible warning.

## Change

`readEvaluationResults` now **throws** when it exhausts `MAX_RESULT_PAGES` with the forward token still advancing (i.e. the stream has more pages than we read), instead of logging to the file logger and returning the partial list.

`getBatchEvaluation` already wraps the call in a `try/catch` that routes any throw into `resultsError`, and the `get` handler already turns `resultsError` into a stderr warning via `warnCloudWatchFailure` while leaving stdout metadata clean. So truncation now travels the **same customer-visible path as every other CloudWatch read failure** — no new plumbing.

Before: `get --json` on a very large job returned partial `results` that read as complete, with only a file-logger line.
After: the job metadata still prints on stdout; a stderr warning states the results are incomplete.

## Test

Added a unit test with an ever-advancing token so the loop runs into the cap; asserts `readEvaluationResults` rejects rather than returning a partial list. Existing pagination/parse tests unchanged and green (`bun test src/core/batchEvaluationResults.test.ts` → 6 pass; handler tests → 12 pass).

### DIFF
--- a/src/core/batchEvaluationResults.test.ts
+++ b/src/core/batchEvaluationResults.test.ts
@@ -1,5 +1,6 @@
 import { test, expect } from "bun:test";
 import type { CloudWatchLogsClient, OutputLogEvent } from "@aws-sdk/client-cloudwatch-logs";
+import { ResultTruncationError } from "../errors";
 import { createSilentLogger } from "../testing";
 import {
   isTerminalStatus,
@@ -174,9 +175,14 @@ test("readEvaluationResults throws (not silently truncates) when it hits the pag
     }),
   } as unknown as CloudWatchLogsClient;
 
-  await expect(
-    readEvaluationResults(everAdvancing, "lg", "ls", createSilentLogger()),
-  ).rejects.toThrow(/incomplete/);
+  const err = await readEvaluationResults(everAdvancing, "lg", "ls", createSilentLogger()).then(
+    () => undefined,
+    (e) => e as ResultTruncationError,
+  );
+  expect(err).toBeInstanceOf(ResultTruncationError);
+  expect(err?.message).toMatch(/incomplete/);
+  expect(err?.source).toBe("internal"); // our page cap, not a user or service fault
+  expect(err?.meta).toMatchObject({ logGroupName: "lg", logStreamName: "ls" });
 });
 
 test("parseEvaluationLogEvent warns on and skips an unparseable line", () => {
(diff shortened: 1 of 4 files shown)


jariy17:
## Live demo against the exploratory account

Verified the new warning path end-to-end. To force truncation on a real job I temporarily set `MAX_RESULT_PAGES = 1` (reverted before commit — the merged code keeps `100`), then ran `eval batch-evaluation get` against a `COMPLETED` job in account `725476964917` whose result stream spans more than one `GetLogEvents` page.

**1. Truncation now warns (stderr) while the job metadata still prints (stdout):**

```console
$ bun run src/index.ts eval batch-evaluation get --id BugBash3P_BugBash3P_1785443436425-b1c354f3b9 --region us-east-1 | head -12
warning: could not retrieve CloudWatch results (batch-evaluation results exceed 1 CloudWatch pages; retrieved 5 results are incomplete). Job status is unaffected. See CloudWatch: region us-east-1, log group /aws/bedrock-agentcore/evaluations/batch-evaluations/results/default, stream run-BugBash3P_BugBash3P_1785443436425-b1c354f3b9.
{
  "batchEvaluationId": "BugBash3P_BugBash3P_1785443436425-b1c354f3b9",
  "batchEvaluationArn": "arn:aws:bedrock-agentcore:us-east-1:725476964917:batch-evaluate/BugBash3P_BugBash3P_1785443436425-b1c354f3b9",
  "batchEvaluationName": "BugBash3P_BugBash3P_1785443436425",
  "status": "COMPLETED",
  ...
```

**2. The warning goes to stderr only — stdout stays clean, machine-readable JSON, and `results` is absent rather than present-but-partial:**

```console
$ bun run src/index.ts eval batch-evaluation get --id ... --region us-east-1 2>/dev/null | python3 -c 'import sys,json; d=json.load(sys.stdin); print("stdout is valid JSON. status:", d["status"], "| results key present:", "results" in d)'
stdout is valid JSON. status: COMPLETED | results key present: False
```

This is the intended behavior: truncation travels the same customer-visible `resultsError` → stderr-warning path as any other CloudWatch read failure, and a script piping stdout still gets parseable JSON with no silently-partial `results` array.

aidandaly24 (approved):

nborges-aws (approved):

PR #1965, opened 2026-08-10
feat(eval): add batch-evaluation evaluate

jariy17:
## Summary

Adds `agentcore eval batch-evaluation evaluate` — the write path that starts an evaluation over existing sessions. `batch-evaluation` previously shipped read-only (`get` + `list`); this makes it runnable.

The command submits an async `StartBatchEvaluation` job (the service gathers the sessions) and returns a durable job id. Retrieve results with the existing `get`.

## Usage

```
agentcore eval batch-evaluation evaluate \
  --agent <harness-id|runtime-id> \
  --evaluator Builtin.Helpfulness \
  --name my-eval
```

### Session source — exactly one

| Flag | Source |
|---|---|
| `--agent <harness-id\|runtime-id>` | resolved to the runtime's CloudWatch log group (reuses `resolveAgentToRuntime`) |
| `--online-eval <id>` | sessions an online-eval config already sampled |
| `--data-source-config <json>` | raw SDK `DataSourceConfig` escape hatch — inline / `file://` / stdin |

### Filters & options

- `--start-time` / `--end-time` (ISO-8601, together) — optional time window.
- `--session-ids` — specific sessions (agent arm only).
- `--ground-truth <json>` — inline / `file://` / stdin → `evaluationMetadata`.
- `--name` (required), `--description`, `--kms-key-arn`.

JSON-valued flags resolve via `SourceResolver` + `parseJsonFlag`, matching `online-eval/create`.

## Design notes

- **Core:** `startBatchEvaluation` on `EvalClient` maps the resolved source to the `dataSourceConfig` union (`cloudWatchLogs` / `onlineEvaluationConfigSource` / raw passthrough) and sends via the data-plane client. Reuses the existing agent→runtime→log-group resolution.
- **Source resolution is inlined in the handler** (one consumer today). It will be extracted to a shared util when `ondemand evaluate` lands as a follow-up and becomes the second consumer.
- **Long flags only** — this framework's `flag()` has no short aliases; matches every existing handler.

## Scope

- `batch-evaluation evaluate` only. `ondemand evaluate` (sync, client-side) and `simulate` (dataset replay) are deferred to follow-up phases.
- `--lookback-days` is intentionally **not** included yet; only explicit `--start-time`/`--end-time` windows for now.
- `--insights` and `--tags` (supported by the SDK request) are deferred.

## Verification

Verified end-to-end against a live account (`us-west-2`, real runtime + `Builtin.Helpfulness`):

batch evaluate demo

Submitted (HTTP 202, `PENDING`) → `get` polled to **`COMPLETED`**, 5/5 sessions, `averageScore: 0.9`, with the resolved `dataSourceConfig` echoed back.

## Tests

Handler command-flow tests (real root handler + `TestCoreClient`): required `--name`/`--evaluator`, source-arm selection (agent + online-eval resolved shapes), zero/multiple arms rejected, `--session-ids` rejected on the online-eval arm.

`bun run typecheck`, `bun test src/` (1001 pass), `oxlint` — all clean.


src/handlers/eval/batch-evaluation/__fixtures__/StartBatchEvaluationCommand.4ab72d4356f13abe.json
```diff
@@ -0,0 +1,20 @@
+{
+  "batchEvaluationId": "golden_batch_evaluate-0c95e0cf8b",
+  "batchEvaluationArn": "arn:aws:bedrock-agentcore:us-west-2:725476964917:batch-evaluate/golden_batch_evaluate-0c95e0cf8b",
```
aidandaly24: these tests are using the wrong account 685197708687

aidandaly24 (dismissed):

.github/workflows/slack-pr-review-notification.yml
```diff
@@ -17,5 +17,8 @@ permissions:
 
 jobs:
   call:
+    # Fork PRs never receive repo secrets/OIDC from GitHub, so the reusable's
```
Hweinstock: is this intentionally part of this PR? feels unrelated to batch evals?

src/core/eval.tsx
```diff
@@ -280,6 +286,66 @@ export class EvalClient implements CoreEvalClient {
       .send(new ListBatchEvaluationsCommand({ nextToken, maxResults }));
   }
 
+  // startBatchEvaluation submits the async, service-side job. Core translates the
+  // resolved SessionSourceValue into the dataSourceConfig union: the agent arm
+  // resolves the harness/runtime id to a log group (reusing agentDataSource), the
+  // online-eval arm points at a config ARN, and the raw arm passes JSON through.
```
Hweinstock: I'm not sure I follow this comment, is it describing the function below? If so, should we localize that info to avoid drift?

src/core/eval.tsx
```diff
@@ -280,6 +286,66 @@ export class EvalClient implements CoreEvalClient {
...
+  // dataSourceConfig union. The agent arm reuses the same runtime resolution +
+  // log-group derivation the control-plane agentDataSource uses, then attaches the
+  // session-id / time-range filters; the raw arm is returned verbatim.
+  private async dataSourceConfigForSource(
+    source: SessionSourceValue,
+    options: CoreOptions,
+  ): Promise<DataPlaneDataSourceConfig> {
+    if (source.origin === "raw") return source.dataSourceConfig;
+
+    // SessionWindow is already { startTime, endTime } — the SDK's SessionFilterConfig shape.
```
Hweinstock: I feel like the code expresses this more concisely than the comment by the fact that we didn't need to transform it before passing to SDK. 

I tend to think comments should be reserved for pieces of information that the code cannot or fails to express.

src/core/eval.tsx
```diff
@@ -280,6 +286,66 @@ export class EvalClient implements CoreEvalClient {
...
+      return {
+        onlineEvaluationConfigSource: {
+          onlineEvaluationConfigArn: source.onlineEvaluationConfigId,
+          timeRange,
+        },
+      };
+    }
+
+    const qualifier = source.endpoint ?? DEFAULT_ENDPOINT_QUALIFIER;
+    const { runtimeId, runtimeName } = await resolveAgentToRuntime(
```
Hweinstock: couldn't this be a runtime or a harness? I thought we could pass harnessId to eval apis?

src/handlers/eval/batch-evaluation/evaluate/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+import { createHandler, flag } from "../../../../router";
+import { InputValidationError } from "../../../../errors";
+import { JsonRendererKey } from "../../../../tui";
+import { SourceResolver, type AppIO } from "../../../../io";
+import type { Core } from "../../../types";
+import type { SessionMetadataShape, DataSourceConfig } from "@aws-sdk/client-bedrock-agentcore";
+import { coreOptsFromCtx, parseJsonFlag } from "../../../utils";
+import type { SessionSourceValue, SessionWindow } from "../../types";
+
+// createEvaluateBatchEvaluationHandler wires `batch-evaluation evaluate`: an
```
Hweinstock: Do you feel that this comment expresses something that the code is unable to?

src/handlers/eval/batch-evaluation/evaluate/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+          groundTruth,
+          kmsKeyArn: flags["kms-key-arn"],
+        },
+        coreOptsFromCtx(ctx),
+      );
+      ctx.require(JsonRendererKey).renderJson(response);
+    },
+  });
+
+type SourceFlags = {
```
Hweinstock: is there a nice way we can derive this from the flags to avoid drift?

src/handlers/eval/batch-evaluation/evaluate/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+  "start-time"?: string;
+  "end-time"?: string;
+  "session-ids"?: string[];
+};
+
+// resolveSource picks exactly one source arm and validates the filters legal for
+// it. `rawDataSourceConfig` is the already-parsed --data-source-config (JSON
+// resolution is async and happens in the handler; this stays sync/pure). Kept
+// local for now; extract to a shared util when on-demand evaluate reuses it.
+function resolveSource(
```
Hweinstock: what exactly is the source of? Source of data for the evaluations? `resolveSource` feels somewhat ambiguous.

src/handlers/eval/batch-evaluation/evaluate/index.tsx
```diff
@@ -0,0 +1,186 @@
...
+  endpoint?: string;
+  "online-eval"?: string;
+  "start-time"?: string;
+  "end-time"?: string;
+  "session-ids"?: string[];
+};
+
+// resolveSource picks exactly one source arm and validates the filters legal for
+// it. `rawDataSourceConfig` is the already-parsed --data-source-config (JSON
+// resolution is async and happens in the handler; this stays sync/pure). Kept
```
Hweinstock: Isn't this already communicated by the function definition?

Hweinstock (commented):
mostly some questions about how we're using comments in the code. I am of the opinion we should reserve them for details the code can't express, and make the code as readable as possible. 

I found most of the comments distracting/confusing, and wonder if it would have the same effect on agents (especially if the information becomes stale).

.github/workflows/slack-pr-review-notification.yml
jariy17: No this was a artifact for another PR

src/core/eval.tsx
jariy17:
```
async function resolveAgentToRuntime(
  agent: string,
  clients: AwsClients,
  options: CoreOptions,
): Promise<{ runtimeId: string; runtimeName: string }> {
  const control = clients.control(toClientConfig(options));
  try {
    const runtime = await control.send(new GetAgentRuntimeCommand({ agentRuntimeId: agent }));
    if (runtime.agentRuntimeName) {
      return { runtimeId: agent, runtimeName: runtime.agentRuntimeName };
    }
  } catch (error) {
    if ((error as Error).name !== "ResourceNotFoundException") throw error;
  }

  const harness = await control.send(new GetHarnessCommand({ harnessId: agent }));
  const environment = harness.harness?.environment;
  const runtimeEnv =
    environment && "agentCoreRuntimeEnvironment" in environment
      ? environment.agentCoreRuntimeEnvironment
      : undefined;
  if (!runtimeEnv?.agentRuntimeId || !runtimeEnv?.agentRuntimeName) {
    throw new InputValidationError(`"${agent}" does not exist as a runtime or a harness`, {
      meta: { agent },
    });
  }
  return { runtimeId: runtimeEnv.agentRuntimeId, runtimeName: runtimeEnv.agentRuntimeName };
}
```
As shown above, the function can return both an harness id/name or a runtime id/name. I'll rename this reflect this.

src/core/eval.tsx
jariy17: I'll remove this commnet

src/handlers/eval/batch-evaluation/evaluate/index.tsx
jariy17: I'll remove this comment

src/core/eval.tsx
jariy17: I'm just going to remove it because the code explains itself.

src/handlers/eval/batch-evaluation/evaluate/index.tsx
jariy17: Removing this code comment too.

src/handlers/eval/batch-evaluation/evaluate/index.tsx
jariy17: I'll rename it to resolveDataSource

src/handlers/eval/batch-evaluation/__fixtures__/StartBatchEvaluationCommand.4ab72d4356f13abe.json
jariy17: I'll use 685197708687 to record these.

src/handlers/eval/batch-evaluation/evaluate/index.tsx
Hweinstock: Discussed in person this is actually meant to match this API shape, not the flag shape: https://docs.aws.amazon.com/boto3/latest/reference/services/bedrock-agentcore/client/start_batch_evaluation.html#. 

We might want to pull out this functionality for insights later, but thats TBD.

Hweinstock (approved):

aidandaly24 (approved):

PR #1966, opened 2026-08-10
feat(project): wire dev handler

tejaskash:
## Summary
- implement `agentcore project dev` runtime selection, port resolution, runner dispatch, structured output, and signal handling
- reuse `withProject` for project discovery and actionable missing-project guidance
- load runtime and `.env.local` values directly; containers forward host AWS variables, mount `~/.aws` read-only, and fail early when neither credential source exists
- reuse shared JSON rendering for NDJSON events and generalize command interruption handling

Stacks on #1962.

## Verification
- `bun test` (1107 pass)
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- `bun run build`
- executable smoke: scaffolded a real Python project, ran `project dev`, received healthy `/ping`, interrupted with exit 130, and verified child cleanup and port release
- executable no-project probe verified the actionable `project create` guidance

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,9 @@
     "agentcore": "./dist/index.js"
   },
   "main": "./dist/index.js",
+  "engines": {
+    "node": ">=20.12.0"
+  },
   "files": [
     "dist"
   ],
--- a/src/core/dev/container.test.ts
+++ b/src/core/dev/container.test.ts
@@ -1,8 +1,9 @@
 import { afterEach, describe, expect, test } from "bun:test";
 import { createHash } from "node:crypto";
-import { mkdir, mkdtemp, readFile, rm, symlink, writeFile } from "node:fs/promises";
+import { mkdir, mkdtemp, readFile, rm, stat, symlink, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join, resolve } from "node:path";
+import { parseEnv } from "node:util";
 import { InputValidationError, InvalidEnvironmentError } from "../../errors";
 import type { DevEvent, DevServerInput } from "../../handlers/project/dev/types";
 import {
@@ -17,6 +18,7 @@ import { ContainerDevRunner } from "./container";
 type ProcessCall = {
   command: string[];
   options: StreamProcessOptions;
+  envFile?: { path: string; contents: string; mode: number };
 };
 
 type StreamBehavior = (
@@ -67,7 +69,14 @@ function harness(
 ) {
   const calls: ProcessCall[] = [];
   const fakeStreamProcess: ProcessStreamer = async function* (command, options) {
-    calls.push({ command, options });
+    const call: ProcessCall = { command, options };
+    calls.push(call);
+    const envFileFlag = command.indexOf("--env-file");
+    if (envFileFlag >= 0) {
+      const path = command[envFileFlag + 1]!;
+      const [contents, metadata] = await Promise.all([readFile(path, "utf8"), stat(path)]);
+      call.envFile = { path, contents, mode: metadata.mode & 0o777 };
+    }
     if (config.stream) yield* config.stream(command, options);
   };
   return {
@@ -161,7 +170,10 @@ describe("ContainerDevRunner", () => {
       ".",
     ]);
     expect(build.options.cwd).toBe(root);
-    expect(build.options.env).toBe(process.env);
+    expect(build.options.env).toEqual({
+      AWS_ACCESS_KEY_ID: "test-access-key",
+      AWS_SECRET_ACCESS_KEY: "test-secret-key",
+    });
     expect(build.options.redactedCommand).toContain("AGENT_NAME=<redacted>");
     expect(build.options.redactedCommand).toContain("TARGET=<redacted>");
     expect(build.options.redactedCommand?.join(" ")).not.toContain("hello-world");
@@ -195,20 +207,15 @@ describe("ContainerDevRunner", () => {
       containerName(root),
       "-p",
       `127.0.0.1:3000:${containerPort}`,
-      "-e",
-      "AWS_ACCESS_KEY_ID",
-      "-e",
-      "AWS_SECRET_ACCESS_KEY",
-      "-e",
-      "API_KEY",
-      "-e",
-      "PORT",
-      "-e",
-      "LOCAL_DEV",
-      ...(protocol === "MCP" ? ["-e", "FASTMCP_PORT"] : []),
+      "--env-file",
+      run.envFile!.path,
       imageTag(root),
     ]);
-    expect(run.options.env).toMatchObject({
+    expect(run.options.env).toEqual({
+      AWS_ACCESS_KEY_ID: "test-access-key",
+      AWS_SECRET_ACCESS_KEY: "test-secret-key",
+    });
+    expect(parseEnv(run.envFile!.contents)).toEqual({
       AWS_ACCESS_KEY_ID: "test-access-key",
       AWS_SECRET_ACCESS_KEY: "test-secret-key",
       API_KEY: "super-secret",
@@ -216,6 +223,8 @@ describe("ContainerDevRunner", () => {
       LOCAL_DEV: "1",
       ...(protocol === "MCP" ? { FASTMCP_PORT: "8000" } : {}),
     });
+    if (process.platform !== "win32") expect(run.envFile!.mode).toBe(0o600);
+    await expect(readFile(run.envFile!.path, "utf8")).rejects.toThrow();
     expect(run.command.join(" ")).not.toContain("super-secret");
     expect(run.command.join(" ")).not.toContain("test-secret-key");
   });
@@ -235,9 +244,9 @@ describe("ContainerDevRunner", () => {
 
     const run = commandCall(calls, "run");
     expect(run.command).toContain(`${awsDirectory}:/aws-config:ro`);
-    expect(run.command).toContain("AWS_PROFILE");
-    expect(run.command).toContain("AWS_CONFIG_FILE");
-    expect(run.options.env).toMatchObject({
+    expect(run.command).not.toContain("AWS_PROFILE");
+    expect(run.command).not.toContain("AWS_CONFIG_FILE");
+    expect(parseEnv(run.envFile!.contents)).toMatchObject({
       AWS_PROFILE: "sandbox",
       AWS_REGION: "us-east-1",
       AWS_CONFIG_FILE: "/aws-config/config",
@@ -299,7 +308,7 @@ describe("ContainerDevRunner", () => {
     );
   });
 
-  test("supplies app variable values through the container CLI environment", async () => {
+  test("keeps app variables out of the container CLI control environment", async () => {
     const projectRuntime = runtime();
     const root = await projectRoot(projectRuntime);
     const { calls, runner } = harness();
@@ -309,9 +318,10 @@ describe("ContainerDevRunner", () => {
     await collect(runner.run(runInput));
 
     const run = commandCall(calls, "run");
-    expect(run.command).toContain("DOCKER_HOST");
+    expect(run.command).not.toContain("DOCKER_HOST");
     expect(run.command.join(" ")).not.toContain("tcp://application-value");
-    expect(run.options.env?.DOCKER_HOST).toBe("tcp://application-value");
+    expect(run.options.env?.DOCKER_HOST).toBeUndefined();
+    expect(parseEnv(run.envFile!.contents).DOCKER_HOST).toBe("tcp://application-value");
   });
 
   test("selects the first tool that supports container builds", async () => {
--- a/src/core/dev/container.test.ts
+++ b/src/core/dev/container.test.ts
@@ -1,6 +1,6 @@
 import { afterEach, describe, expect, test } from "bun:test";
 import { createHash } from "node:crypto";
-import { mkdir, mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
+import { mkdir, mkdtemp, readFile, rm, symlink, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join, resolve } from "node:path";
 import { InputValidationError, InvalidEnvironmentError } from "../../errors";
@@ -61,6 +61,8 @@ function harness(
   config: {
     available?: (tool: string, probeArgs?: string[]) => Promise<boolean>;
     stream?: StreamBehavior;
+    awsDirectory?: string;
+    processEnv?: NodeJS.ProcessEnv;
   } = {},
 ) {
   const calls: ProcessCall[] = [];
@@ -77,6 +79,11 @@ function harness(
         (async (tool) => {
           return tool === "docker";
         }),
+      awsDirectory: config.awsDirectory ?? join(tmpdir(), "agentcore-container-no-aws"),
+      processEnv: config.processEnv ?? {
+        AWS_ACCESS_KEY_ID: "test-access-key",
+        AWS_SECRET_ACCESS_KEY: "test-secret-key",
+      },
     }),
   };
 }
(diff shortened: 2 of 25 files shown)


src/core/dev/container.test.ts
```diff
@@ -409,6 +449,31 @@ describe("ContainerDevRunner", () => {
...
+    for (const buildContextPath of ["..", "linked"]) {
+      const { calls, runner } = harness({
+        available: async (tool) => {
+          probes.push(tool);
+          return true;
+        },
+      });
+
+      await expect(collect(runner.run(input(root, runtime({ buildContextPath }))))).rejects.toThrow(
+        "container build context must be within the project root",
```
Hweinstock, line 495: should we validate the error types in these tests?

src/core/dev/container.ts
```diff
@@ -65,12 +81,36 @@ export class ContainerDevRunner implements DevRunner {
       throw new InputValidationError(`container build context directory not found: ${context}`);
     }
 
+    const canonicalContext = realpathSync(context);
+    const relativeContext = relative(realpathSync(input.projectRoot), canonicalContext);
+    if (
+      relativeContext === ".." ||
+      relativeContext.startsWith(`..${sep}`) ||
+      isAbsolute(relativeContext)
```
Hweinstock: in what cases does `relative` return an absolute path?

src/core/dev/container.ts
```diff
@@ -65,12 +81,36 @@ export class ContainerDevRunner implements DevRunner {
...
       throw new InputValidationError(`container Dockerfile not found: ${dockerfilePath}`);
     }
 
+    const hasAwsCredentials = Boolean(
+      (input.env?.AWS_ACCESS_KEY_ID ?? this.processEnv.AWS_ACCESS_KEY_ID) &&
+      (input.env?.AWS_SECRET_ACCESS_KEY ?? this.processEnv.AWS_SECRET_ACCESS_KEY),
+    );
+    const hasAwsConfig = existsSync(this.awsDirectory);
+    if (!hasAwsCredentials && !hasAwsConfig) {
+      throw new InputValidationError(
```
Hweinstock: do you think it would be useful to distinguish the missing credentials case from the invalid inputs case in telemetry with a separate error type here?

src/core/dev/port.ts
```diff
@@ -0,0 +1,45 @@
...
+const MAX_PORT_ATTEMPTS = 100;
+export const DEV_PORTS = { HTTP: 8080, AGUI: 8080, MCP: 8000, A2A: 9000 } as const;
+
+export type DevPort = {
+  port: number;
+  requestedPort: number;
+};
+
+function portInUse(port: number, suffix = ""): InputValidationError {
+  return new InputValidationError(
```
Hweinstock: would it be simpler to make this function its own error type?

src/errors/errors.tsx
```diff
@@ -135,7 +135,7 @@ export class EmbeddedAssetNotFoundError extends AgentCoreCLIError {
   }
 }
 
-export class RuntimeInvokeInterruptedError extends AgentCoreCLIError {
+export class CommandInterruptedError extends AgentCoreCLIError {
```
Hweinstock, line 141: I think this is doing something similar to https://github.com/aws/agentcore-cli/pull/1986/changes (haven't fully reviewed that one yet)

src/handlers/project/dev/index.ts
```diff
@@ -1,11 +1,119 @@
...
+    throw new InputValidationError(
+      "This project has no runtimes. Add a runtime to agentcore/agentcore.json and retry.",
+    );
+  }
+  const available = project.runtimes.map(({ name }) => name).join(", ");
+
+  if (name) {
+    const runtime = project.runtimes.find((candidate) => candidate.name === name);
+    if (runtime) return runtime;
+    throw new InputValidationError(
```
Hweinstock: should this be a resourceNotFoundError or something?

src/handlers/project/dev/index.ts
```diff
@@ -1,11 +1,119 @@
...
+}
+
+export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
   createHandler({
     name: "dev",
     description: "run the project locally for development",
-    handle: async () => {
-      throw new NotImplementedError("agentcore project dev is not implemented yet");
+    flags: [
+      flag("agent", "runtime to run", z.string().optional()),
```
Hweinstock, line 61: would using the flag `agent` imply it works for runtime and harness?

src/handlers/project/index.ts
```diff
@@ -23,7 +25,20 @@ export function createProjectHandler(config: ProjectHandlerConfig): Router {
...
+    withProject({ projectManager: config.projectManager })(
+      createDevProjectHandler({
+        io: config.io,
+        runners: {
+          CodeZip: new CodeZipDevRunner(),
+          Container: new ContainerDevRunner(),
+        },
+        loadDevEnvironment,
+        checkPort,
+        forceExit: () => process.exit(130),
```
Hweinstock: I think we want to avoid killing the process directly so that we can get telemetry here.

package.json
```diff
@@ -7,6 +7,9 @@
     "agentcore": "./dist/index.js"
   },
   "main": "./dist/index.js",
+  "engines": {
+    "node": ">=20.12.0"
```
Hweinstock, line 11: why exactly is this needed? is there a component bun is missing that we need here?

Hweinstock (changes requested):
only blocking comment is the direct process kill.

src/handlers/project/dev/index.ts
tejaskash: Yes

src/core/dev/container.test.ts
tejaskash: Addressed in 32e0283c. The escaped-context regression now asserts `InputValidationError` in addition to the actionable message. The missing-credentials path also asserts its concrete environment error type.

src/core/dev/container.ts
tejaskash: On Windows, `path.relative` returns an absolute path when the canonical paths are on different drives. For example, `win32.relative("C:\\project", "D:\\outside")` returns `D:\\outside`.

src/core/dev/container.ts
tejaskash: Agreed. Addressed in 32e0283c: missing container credentials now throw `InvalidEnvironmentError`, so telemetry distinguishes them from invalid project input while retaining user attribution.

src/core/dev/port.ts
tejaskash: I kept this as a small message helper. An occupied explicit port and an exhausted default-port search have the same recovery and telemetry category (`InputValidationError`), so a dedicated class would add structure without distinct behavior.

src/errors/errors.tsx
tejaskash: Yes, #1986 is the broader cancellation model and should supersede this rename when it lands. It is still open and currently conflicts with `refactor`, so #1966 cannot depend on it yet. This branch keeps the minimal generalization needed by project dev and preserves the newer Runtime response-streaming implementation from current `refactor`.

src/handlers/project/dev/index.ts
tejaskash: This is a failed selection against local `agentcore.json`, not a missing remote service resource. `InputValidationError` keeps it user-attributed and lets the message list the valid configured runtimes; a resource-not-found error would imply a service lookup.

src/handlers/project/index.ts
tejaskash: Addressed in 32e0283c. The direct `process.exit`/`forceExit` path is removed. Repeated signals are idempotent, cleanup completes, and the handler throws `CommandInterruptedError` through the root path so telemetry is emitted.

package.json
tejaskash: The published `dist/index.js` runs under Node (the bin has a Node shebang), and this change uses `node:util.parseEnv`, which was added in Node 20.12. Bun is the development/build runner, but the npm artifact still needs the Node engine floor declared.

src/core/dev/port.ts
Hweinstock: wouldn't a dedicated class change the telemetry behavior?

tejaskash: Hmm, I see telemetry records `error_name: error.name`, so a dedicated subclass does change the telemetry behavior. Fixed this

src/handlers/project/dev/index.ts
Hweinstock: would we want to distinguish valid user input that reference non-existent resources from invalid input in telemetry?

tejaskash: Yes, that distinction is useful. Addressed in e6c8aa10: an explicit `--agent` name that is not configured now throws `ResourceNotFoundError`, so telemetry records that separately from malformed/ambiguous input. Zero runtimes and multiple runtimes without a selector remain `InputValidationError`.

src/handlers/project/dev/index.ts
Hweinstock: I see, forgot we're still deploying on dev for harness...

package.json
Hweinstock: node 20 is pretty old, we'll probably want to bump this at some point.

src/handlers/project/dev/index.ts
```diff
@@ -1,11 +1,122 @@
...
+          runtime,
+          projectRoot: project.rootPath,
+          port: devPort.port,
+          env,
+          signal: controller.signal,
+        })) {
+          renderEvent(config.io, event, json);
+        }
+      } catch (error) {
+        if (!controller.signal.aborted) throw error;
```
Hweinstock, line 117: this will likely conflict with https://github.com/aws/agentcore-cli/pull/1986, but I think we can fix this as a follow-up to unblock.

src/handlers/project/dev/environment.ts
```diff
@@ -0,0 +1,65 @@
...
+
+type DevEnvironmentLoaderConfig = {
+  readFile?: (path: string, encoding: BufferEncoding) => Promise<string>;
+};
+
+async function localEnvironment(
+  projectRoot: string,
+  read: (path: string, encoding: BufferEncoding) => Promise<string>,
+): Promise<Record<string, string>> {
+  const path = join(projectRoot, "agentcore", ".env.local");
```
Hweinstock, line 29: this feels like project specific code, should we move it into the project module?

tejaskash: Addressed in 85d64984. The environment loader and its tests now live under `src/handlers/project/dev/`, project wiring imports it directly, and `src/io` no longer exports project-specific types or behavior.

src/handlers/project/dev/index.ts
tejaskash: Agreed. We will let #1986 supersede this interruption type as a follow-up; keeping the current implementation in #1966 avoids blocking project dev while that PR is reconciled with `refactor`.

Hweinstock (dismissed):

Hweinstock (dismissed):

src/core/dev/container.ts
```diff
@@ -116,15 +156,23 @@ export class ContainerDevRunner implements DevRunner {
...
       PORT: String(containerPort),
       LOCAL_DEV: "1",
-    };
+    });
     if (input.runtime.protocol === "MCP") {
       forwardedEnv.FASTMCP_PORT = String(containerPort);
     }
+    const awsMount = hasAwsConfig ? ["-v", `${this.awsDirectory}:/aws-config:ro`] : [];
+    if (awsMount.length) {
+      forwardedEnv.AWS_CONFIG_FILE = "/aws-config/config";
```
notgitika, line 165: won't these be fully visible on command line once the below code executes? it does ``${key}=${value}`,`

is there a safer way to do this? we can also try supplying this through the Docker CLI process environment

notgitika: we could pass only variable names as arguments

```
docker run \
    -e AWS_ACCESS_KEY_ID \
    -e AWS_SECRET_ACCESS_KEY \
    -e AWS_SESSION_TOKEN \
    image

```

and then pass through this

```
const dockerEnv = {
    ...process.env,
    ...forwardedEnv,
  };
```

what do you think?

src/core/dev/container.ts
```diff
@@ -65,12 +81,36 @@ export class ContainerDevRunner implements DevRunner {
...
+      );
+    }
+
     const dockerfile = input.runtime.dockerfile ?? DOCKERFILE_NAME;
     const dockerfilePath = join(context, dockerfile);
     if (!isFile(dockerfilePath)) {
       throw new InputValidationError(`container Dockerfile not found: ${dockerfilePath}`);
     }
 
+    const hasAwsCredentials = Boolean(
```
notgitika, line 94: maybe OOS and could be a follow up but could we use the AWS SDK default credential provider here? The current check/mount may not support sso, web identity, or custom aws config paths.

notgitika (changes requested):

src/core/dev/container.ts
tejaskash: Addressed in `90abf1fd`. Container environment values are now supplied only through the container CLI subprocess environment; `docker`/`podman`/`finch run` argv contains only `-e NAME`, so credentials and `.env.local` values are no longer visible in process listings.

src/core/dev/container.ts
tejaskash: Agreed this is worth a follow-up. Using the SDK default provider changes credential resolution and refresh semantics, especially for SSO, web identity, custom config paths, and long-running dev sessions. This PR keeps the current shared-config mount and preflight scope while removing argv exposure.

Hweinstock (dismissed):

Hweinstock (dismissed):

src/handlers/runtime/invoke/index.tsx
```diff
@@ -91,12 +95,22 @@ export const createInvokeRuntimeHandler = (core: Core, io: AppIO) =>
...
+            core,
+            io,
+          );
+        } catch (error) {
+          if (error instanceof InvalidEnvironmentError) {
+            throw new InputValidationError(error.message, {
+              cause: error,
+              exitCode: ExitCode.USAGE,
+            });
+          }
```
notgitika: this is a bit weird. runtime invoke reports InputValidationError, while gw and other TUI paths report InvalidEnvironmentError for the same condition. 

Also, I'm not sure how this is related to this PR so can we just remove it?

src/core/dev/container.ts
```diff
@@ -149,18 +191,7 @@ export class ContainerDevRunner implements DevRunner {
...
-          "run",
-          "--rm",
-          "--name",
-          containerName,
-          "-p",
-          `127.0.0.1:${input.port}:${containerPort}`,
-          ...redactedEnvFlags,
-          imageTag,
-        ],
+        env: { ...this.processEnv, ...forwardedEnv },
```
notgitika: thanks for addressing my earlier comment. this approach keeps secrets out of argv, but it also puts application variables into the Docker CLI’s control environment. `docker run -e NAME ` reads NAME from that environment, so an app variable such as `DOCKER_HOST` redirects `docker run` to another daemon, while removeContainer still targets the default daemon. 

passing the same env to cleanup would fix the mismatch but would still let app configuration control Docker itself. could we keep the CLI environment stable and pass app variables separately, for eg through a temporary mode-0600 --env-file removed in finally? 

Let's also add a regression test covering DOCKER_HOST.

src/handlers/project/index.ts
```diff
@@ -23,7 +26,19 @@ export function createProjectHandler(config: ProjectHandlerConfig): Router {
...
   project.handler(createAddProjectResourceHandler(config));
   project.handler(createRemoveProjectHandler());
-  project.handler(createDevProjectHandler());
+  project.handler(
+    withProject({ projectManager: config.projectManager })(
+      createDevProjectHandler({
+        io: config.io,
+        runners: {
+          CodeZip: new CodeZipDevRunner(),
+          Container: new ContainerDevRunner(),
```
notgitika, line 35: this wiring makes the existing CodeZip runner reachable from project dev. Its codeLocation and entrypoint paths are not confined to the project root, while the schema accepts traversal such as ../outside. 

This can execute or run npm install outside the project, including through symlinks. Could we add canonical project-root confinement for CodeZip paths before enabling it here, matching the Container runner’s protection? could be a follow up

src/handlers/runtime/invoke/index.tsx
tejaskash: Removed in `f0b65c99`. This PR no longer remaps `InvalidEnvironmentError` in Runtime invoke; the existing `refactor` behavior is restored unchanged.

src/core/dev/container.ts
tejaskash: Addressed in `f0b65c99`. Container values now go through a temporary mode-0600 `--env-file`, removed in `finally`. Build, run, and cleanup keep the same container CLI environment. The regression verifies `DOCKER_HOST` is present only in the env file and cannot redirect the CLI.

src/handlers/project/index.ts
tejaskash: Addressed in `f0b65c99`. CodeZip now applies the same canonical project-root guard as the Container runner to both `codeLocation` and the resolved entrypoint before npm or the server process can run. Tests cover traversal and symlink/junction escapes for both paths.

notgitika (approved):

aidandaly24 (approved):
This LGTM, I will replace the user cancellation error with mine in #1986

aidandaly24:
Saw that Harrison had also approved earlier so merged

PR #1967, opened 2026-08-10
feat(gateway): add headless invoke command

aidandaly24:
## Summary

- add `agentcore gateway invoke` as a headless, project-independent command with `GET`, `POST`, and `DELETE` support
- resolve the data-plane URL and authorizer from `GetGateway`, including SigV4, bearer-token, and unsigned requests
- support relative paths with query strings, source-aware payloads and tokens, application/session headers, and raw, file, or JSON response output
- redact path, payload, application-header, and bearer-token values from command debug logs
- preserve 3xx/4xx/5xx response bodies in the selected output mode, then return a failure exit for non-2xx responses
- sign `AUTHENTICATE_ONLY` requests with SigV4, matching live service behavior
- share the streaming response writer with Runtime invoke while preserving Runtime behavior
- keep the V1 surface raw: no TUI, `--request-type`, structured `--query`, or discovery commands

## Validation

- `bun test`: 1,087 passed, 0 failed on current `refactor`
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- `bun run build`

### Live Gateway validation

Using `AWS_PROFILE=deploy` in account `603141041947`:

- Ingress `NONE`: HTTP 200 with file/stdin payloads and raw/JSON/file output
- Ingress `AWS_IAM`: HTTP 200 with SigV4 and repeated query keys
- Ingress `CUSTOM_JWT`: HTTP 200 with inline/file/stdin bearer-token sources
- Ingress `AUTHENTICATE_ONLY`: HTTP 200 with SigV4 through the CLI; `--bearer-token` is rejected locally
- Outbound API key: successful API Gateway MCP target invocation
- Outbound OAuth 2LO: successful `CLIENT_CREDENTIALS` MCP target invocation
- Outbound OAuth 3LO: Cognito authorization-code consent, `CompleteResourceTokenAuth` session binding, and successful per-user tool call with downstream user-token evidence
- Outbound OBO: RFC 8693 token exchange and successful OpenAPI tool call with a distinct opaque `obo.*` downstream token
- JWT passthrough: backend token hash exactly matched the scoped Cognito caller-token hash
- Gateway IAM role: successful AgentCore Runtime HTTP target invocation with Runtime session and SSE body
- Inference: 47 models discovered through `v1/models`; OpenAI-compatible chat completion returned `inference gateway ready`
- HTTP 405 response: preserved `Method Not Allowed` in raw and JSON output, then exited 1 without duplicate stderr
- query-secret redaction probe: the secret had zero log matches and the execution record stored `path: [REDACTED]`

The reusable matrix and retry commands are documented at `gateway-invoke/auth-matrix/README.md` in the task workspace.

### DIFF
--- a/src/io/streamingResponse.ts
+++ b/src/io/streamingResponse.ts
@@ -1,0 +1,179 @@
+import { createWriteStream } from "node:fs";
+import { pipeline } from "node:stream/promises";
+
+export type StreamingResponse = {
+  contentType: string;
+  body: AsyncIterable<Uint8Array>;
+};
+
+export type StreamingResponseOutput = {
+  stdout: NodeJS.WriteStream;
+  stderr: NodeJS.WriteStream;
+  outputFile?: string;
+  json?: boolean;
+  signal?: AbortSignal;
+};
+
+export type StreamingResponseWriter<T extends StreamingResponse> = {
+  metadata: (response: T) => Record<string, unknown>;
+  summary: (response: T, byteCount: number, complete: boolean, error?: string) => string;
+  fail: (error: unknown) => never;
+  binaryTtyError: string;
+};
+
+const TEXTUAL_SEQUENCE_MEDIA_TYPES = new Set([
+  "application/x-ndjson",
+  "application/ndjson",
+  "application/json-seq",
+]);
+
+function mediaType(contentType: string): string {
+  return contentType.split(";", 1)[0]!.trim().toLowerCase();
+}
+
+export function classifyStreamingResponse(contentType: string): "json" | "text" | "binary" {
+  const type = mediaType(contentType);
+  if (type === "application/json" || /^application\/[^/]+\+json$/.test(type)) {
+    return "json";
+  }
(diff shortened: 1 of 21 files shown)


tejaskash (dismissed):

src/io/streamingResponse.ts
```diff
@@ -0,0 +1,179 @@
...
+  });
+  await writeChunk(output.stdout, envelope, output.signal, writer.fail);
+}
+
+export async function writeStreamingResponse<T extends StreamingResponse>(
+  response: T,
+  output: StreamingResponseOutput,
+  writer: StreamingResponseWriter<T>,
+): Promise<void> {
+  if (
```
nborges-aws, line 127: would a successful DELETE operation pass this check? If it has no content type this would throw?

aidandaly24: Good call out thanks. I've updated now with your comment.

tejaskash (dismissed):

nborges-aws (dismissed):
LGTM thanks for fixes!

tejaskash (approved):
👍

nborges-aws (approved):

PR #1968, opened 2026-08-11
fix(runtime): update shell session wire protocol

priyasingh2111:
## Description
  
  Remove the 0x03 metadata frame dependency and 0xFF close frame send from the shell connection flow to align with the K8s wire protocol.
  
  **Changes:**
  1. **Removed `encodeClose()` method** from `ShellFramer` — CLI already uses `ws.close()` for detach, this was dead code
  2. **Removed confirmation frame blocking wait** — connection is now ready immediately after WebSocket opens + shellId read from the 101 upgrade
  header (eliminates up to 10s wait)
  3. **Removed `reconnected` / `bytesDropped` from `ShellConnection`** — these were populated by the confirmation frame we no longer wait for
  4. **Confirmation frames silently swallowed during iteration** — forward-compatible if server still sends them
  
  Exit codes, keepalive/reconnect, and Ctrl+] detach all continue to work unchanged.
  
  ## Related Issue
  
  
  
  Closes #
  
  ## Documentation PR
  
  
  
  N/A — no user-facing documentation changes needed.
  
  ## Type of Change
  
  
  
  - [x] Bug fix
  - [ ] New feature
  - [x] Breaking change
  - [ ] Documentation update
  
  ## Testing
  
  - [x] I ran `npm run lint`
  
  **Results:**
  - 6,068 unit tests pass (422 test files)
  - 0 lint errors (33 pre-existing warnings unchanged)
  - Format check clean
  - Live integration tests: 10/10 against prod (connection in ~3s, no 10s wait)
  
  **Breaking changes:**
  - `ShellConnection.reconnected` and `ShellConnection.bytesDropped` no longer exist
  - `ShellReconnectOptions.onBytesDropped` and `onNewSession` callbacks removed
  - `ShellFramer.encodeClose()` method removed
  - Old clients will hard-fail (10s timeout then error) when the server stops sending the 0x03 confirmation frame — this is an accepted breaking change
  per team decision (no protocol negotiation)
  
  ## Checklist
  
  ---
  
  terms of your choice.

### DIFF
--- a/src/cli/aws/connect-shell.ts
+++ b/src/cli/aws/connect-shell.ts
@@ -1,6 +1,5 @@
 import { ShellKickedError } from '../../lib/errors/types';
 import { getCredentialProvider } from './account';
-import { ShellChannel, ShellFramer, parseStatusFrame } from './shell-framer';
 import { dataPlaneEndpoint } from './stage-endpoint';
 import { Sha256 } from '@aws-crypto/sha256-js';
 import { HttpRequest } from '@smithy/protocol-http';
@@ -22,10 +21,6 @@ export interface ShellReconnectOptions {
   onAttempt?: (attempt: number, reason: string) => void;
   /** Called when close code 4000 is received — another client took the session. */
   onKicked?: () => void;
-  /** Called when reconnect yields a fresh shell (previous session expired). */
-  onNewSession?: (shellId: string) => void;
-  /** Called when the confirmation frame reports bytes lost during disconnect. */
-  onBytesDropped?: (n: number) => void;
 }
 
 export interface ConnectShellOptions {
@@ -41,18 +36,13 @@ export interface ConnectShellOptions {
   reconnect?: ShellReconnectOptions;
   /** Bearer token for CUSTOM_JWT auth. When set, authenticates via WebSocket subprotocol instead of SigV4. */
   bearerToken?: string;
-  /** Milliseconds to wait for the STATUS confirmation frame before failing. Default: 10_000 */
-  confirmationTimeoutMs?: number;
 }
 
 export interface ShellConnection {
   ws: WebSocket;
-  /** The server-assigned shell identifier (from wire `shellId`). */
+  /** The server-assigned shell identifier (from the 101 upgrade response header). */
   shellId: string;
   sessionId?: string;
-  reconnected: boolean;
-  /** Bytes of output lost during a disconnect, reported in the confirmation frame. */
-  bytesDropped?: number;
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


src/cli/aws/connect-shell.ts
```diff
@@ -202,42 +183,20 @@ async function openWebSocket(options: ConnectShellOptions): Promise<ShellConnect
...
         } else {
-          fail(new Error(`WebSocket closed before confirmation (code ${code})`));
+          fail(new Error(`WebSocket closed before open (code ${code})`));
         }
       }
     });
 
-    ws.on('message', (data: Buffer) => {
+    // Connection is ready immediately after WebSocket opens — no confirmation frame wait.
+    ws.on('open', () => {
```
aidandaly24, line 192: Could we update the close handling so only code 1000 is treated as a clean exit? Now that the connection resolves when the WebSocket opens, an abnormal close such as 1006 can happen before the shell is usable. Treating that as exit 0 reports a failed connection as successful.

src/cli/aws/connect-shell.ts
```diff
@@ -202,42 +183,20 @@ async function openWebSocket(options: ConnectShellOptions): Promise<ShellConnect
...
-        resolve(conn);
-      }
-      // termination before confirmation — treat as error
-      if (parsed.type === 'termination') {
-        fail(new Error('Shell terminated before confirmation frame'));
-      }
+      settled = true;
+      resolve({
+        ws,
+        shellId: shellIdFromHeader ?? shellId ?? '',
```
aidandaly24: Could we return an error when a new connection does not include the shell ID header? Returning an empty shell ID means the user cannot reconnect to the session.

aidandaly24 (changes requested):

src/cli/aws/connect-shell.ts
priyasingh2111: Addressed in the next commit!

src/cli/aws/connect-shell.ts
priyasingh2111: Addressed in the next commit!

src/cli/commands/exec/action.ts
```diff
@@ -472,12 +459,25 @@ export async function handleShellSession(ctx: ExecContext, options: ExecOptions)
...
+      // prints the reconnect hint instead of a spurious exit line.
+      let resolvedExitCode: number | null;
+      if (exitCode !== null) {
+        resolvedExitCode = exitCode;
+      } else if (code === 1000) {
+        resolvedExitCode = 0;
+      } else if (code === 4000) {
+        resolvedExitCode = null;
+      } else {
+        resolvedExitCode = 1;
```
aidandaly24, line 479: Could we keep abnormal WebSocket disconnects separate from shell exit code 1? A close such as 1006 is a transport failure, and the remote shell may still be available to reconnect. Mapping it to exit code 1 prints “session closed” and suppresses the reconnect command. Could this return a failure while keeping `exitCode` unset and showing the reconnect hint?

PR #1969, opened 2026-08-11
refactor(project): tighten the ProjectEvent and create() contracts

notgitika:
Groundwork for `agentcore project build`, split out so the contract change is reviewable on its own.

## What

`ProjectEvent` was a three-field bag with every field optional:

```ts
export type ProjectEvent = {
  message?: string;
  subprocessOutput?: string;
  project?: Project;
};
```

`subprocessOutput` and `project` were never written and never read — I checked every producer and consumer. Only `message` was ever used, and because it was optional the `create` handler had to guard it (`if (event.message)`) on every iteration. So the type is now just:

```ts
export type ProjectEvent = { message: string };
```

and the guard in `handlers/project/create/index.ts` is gone.

`create` also declared `AsyncGenerator<ProjectEvent>`, leaving `TReturn` as the default `any`. It is now `AsyncGenerator<ProjectEvent, Project>`.

## The type hole that tightening exposed

Naming the return type surfaced a real mismatch: `create` was building its return value out of `TemplateSpec.runtimes`, which is `unknown[]`, while `Project.runtimes` is `ProjectRuntime[]`. `any` had been hiding it.

Rather than cast, `create` now reads the project back through `resolve()`:

```ts
const project = await this.resolve({ filePath: destination });
```

That honours a comment already sitting in the code ("Return the same shape resolve() would: a created project is a resolvable one"), and means the returned runtimes are schema-validated instead of being the template's loosely-typed spec sections passed through unchecked.

The matching `as Project` cast in `manager.test.ts` is dropped too, so the test now type-checks the real thing.

## Verification

- `bun test` — 1061 pass, 0 fail
- `tsc --noEmit` clean, `oxlint` clean

No behaviour change; the only user-visible difference is that progress messages are no longer conditionally skipped, and they were never absent in practice.

### DIFF
--- a/src/core/project/manager.tsx
+++ b/src/core/project/manager.tsx
@@ -66,7 +66,7 @@ export class FsProjectManager implements ProjectManager {
     }
   }
 
-  public async *create(input: CreateProjectInput): AsyncGenerator<ProjectEvent> {
+  public async *create(input: CreateProjectInput): AsyncGenerator<ProjectEvent, Project> {
     // Scaffold into a fresh directory, refusing to nest inside an existing project.
     const enclosing = enclosingProjectRoot(process.cwd());
     if (enclosing) {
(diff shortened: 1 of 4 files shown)


tejaskash (approved):

src/core/project/manager.tsx
```diff
@@ -106,12 +106,16 @@ export class FsProjectManager implements ProjectManager {
...
-      name: input.name,
-      rootPath: destination,
-      runtimes: TEMPLATES[input.template].spec.runtimes ?? [],
-    };
+    // A created project is a resolvable one, so read it back rather than
+    // duplicating the template's shape: the returned runtimes are then
+    // schema-validated instead of the template's loosely-typed spec sections.
+    const project = await this.resolve({ filePath: destination });
+    if (!project) {
+      throw new ProjectStateError(
```
Hweinstock, line 114: OOS here but `ProjectStateError` feels pretty vague to me. What do you think of  replacing it with something like `ProjectValidationError` for when reading a project fails?

Hweinstock (approved):
LGTM!

src/core/project/manager.tsx
notgitika: Ahh yes that makes sense, I will make a note and have it out as a followup!

PR #1970, opened 2026-08-11
feat(project): implement project build

notgitika:
`agentcore project build` compiles the project's CDK app and synthesizes its CloudFormation templates, so the deployable artifacts exist before deploy.

The generated `package.json` defines `cdk` as `"npm run build && cdk"`, so that one command covers both compile and synthesis — no need to construct a `node_modules/.bin/cdk` path or branch on `win32`.

## Synthesis is credential-free, by design

An earlier draft of this called STS `GetCallerIdentity` and wrote `agentcore/aws-targets.json` as a side effect of `build`. That isn't needed: each stack takes an explicit `env: { account, region }` from `aws-targets.json` and nothing in the app calls `fromLookup`, so synth never talks to AWS. The account is only load-bearing at deploy.

So `build` stays offline and does not write config. When `aws-targets.json` is still `[]`, the CDK app raises its own error, which already says what to do:

```
AgentCore CDK synthesis failed: No deployment targets configured.
Please define targets in agentcore/aws-targets.json
```

wrapped by `ProcessFailedError` with `Fix the issue and run 'cd .../agentcore/cdk && npm run cdk -- synth --quiet' to retry.`

Deliberately not guessing the account also avoids a trap the earlier draft had: it wrote `region` unvalidated, but `AgentCoreRegionSchema` is an enum of 9 regions. `--region eu-west-2` would have written a region synth then rejects, and because the CLI's own re-read used `z.array(z.unknown())` it would see a non-empty list and never repair the file.

## withProject goes into service

`src/middleware/withProject.tsx` already existed but was never exported and never used. It now resolves the enclosing project and hands it to the handler via `ProjectKey`, so the handler body is just:

```ts
for await (const event of config.projectManager.build(project)) {
  config.io.stderr.write(`${event.message}\n`);
}
```

Two changes to it:

- `cwd` is optional and resolved per invocation (`config.cwd ?? process.cwd()`) rather than captured at wiring time, so the directory the user actually ran in is the one searched.
- Failure is a `ProjectStateError` naming the path searched and the file looked for, and pointing at `agentcore project create`.

It wraps **only** `build`, not the whole router — `create` refuses to nest inside an existing project, so requiring one would break it.

## Second commit: `runtimeVersion` on the CodeZip template

Included here rather than as a follow-up, so `build` is never merged in a state where it fails on the config the CLI itself just scaffolded.

The CDK construct library's schema refines `build !== 'Container' && !runtimeVersion` into `runtimeVersion is required for CodeZip builds`, and our `hello-world-python` template didn't set it. Five lines: `runtimeVersion: "PYTHON_3_14"` on the template runtime plus the matching test assertion. Container builds take their version from the image, so the container template is untouched.

A follow-up will mirror that refinement into our own `AgentEnvSpecSchema`, which currently marks `runtimeVersion` plain `.optional()` — that way the CLI catches this class of error itself instead of surfacing CDK's zod path error.

## Verification

Beyond unit tests, I ran this against a real scaffolded project with a real `npm install`:

1. Empty `aws-targets.json` → the CDK app's "No deployment targets configured" error, wrapped with the retry hint.
2. One target filled in, with `AWS_ACCESS_KEY_ID`/`SECRET`/`SESSION_TOKEN` blanked and `AWS_PROFILE` pointed at a nonexistent profile → `Built project 'Demo'`, emitting `cdk.out/AgentCore-Demo-dev.template.json`. Confirms synth needs no credentials.
3. Stripping `runtimeVersion` back out of that same project → reproduces the CodeZip failure, which is what motivated the second commit.

Unit coverage: the exact synth command and cwd, the missing-`node_modules` error, subprocess failure propagation, the progress event, resolution from a nested directory, and failing outside a project.

- `bun test` — 1061 pass, 0 fail
- `tsc --noEmit` clean, `oxlint` clean

### DIFF
--- a/src/core/project/manager.tsx
+++ b/src/core/project/manager.tsx
@@ -54,7 +54,12 @@ export class FsProjectManager implements ProjectManager {
     const configPath = join(rootPath, "agentcore", "agentcore.json");
     try {
       const spec = await this.json.read(configPath, ProjectSpecSchema);
-      return { name: spec.name, rootPath, runtimes: spec.runtimes };
+      return {
+        name: spec.name,
+        rootPath,
+        managedBy: spec.managedBy,
+        runtimes: spec.runtimes,
+      };
     } catch (error) {
       // A malformed agentcore.json is a user-correctable problem, not a crash.
       if (error instanceof DeserializationError) {
@@ -119,6 +124,25 @@ export class FsProjectManager implements ProjectManager {
   }
 
   public async *build(project: Project): AsyncGenerator<ProjectEvent, void> {
+    // agentcore.json records which backend owns the project's artifacts. CDK is the
+    // only one today; a terraform or no-IaC backend adds an arm here rather than
+    // editing the CDK path.
+    switch (project.managedBy) {
+      case "CDK":
+        yield* this.buildWithCdk(project);
+        break;
+      default: {
+        // Exhaustiveness: a new ManagedBy member fails to compile until it is handled.
+        const unsupported: never = project.managedBy;
+        throw new ProjectStateError(
+          `project '${project.name}' declares an unsupported backend: ${String(unsupported)}`,
+        );
+      }
+    }
+  }
+
+  // Compiles the generated CDK app and synthesizes its CloudFormation templates.
+  private async *buildWithCdk(project: Project): AsyncGenerator<ProjectEvent, void> {
     const cdkDir = join(project.rootPath, "agentcore", "cdk");
 
     // The generated CDK app is built from its own node_modules; without them the
--- a/src/core/project/manager.tsx
+++ b/src/core/project/manager.tsx
@@ -118,6 +118,26 @@ export class FsProjectManager implements ProjectManager {
     return project;
   }
 
+  public async *build(project: Project): AsyncGenerator<ProjectEvent, void> {
+    const cdkDir = join(project.rootPath, "agentcore", "cdk");
+
+    // The generated CDK app is built from its own node_modules; without them the
+    // failure would otherwise surface as an opaque "cdk: not found".
+    if (!existsSync(join(cdkDir, "node_modules"))) {
+      throw new ProjectStateError(
+        `CDK dependencies are missing for project '${project.name}'. ` +
+          `Run 'cd ${cdkDir} && npm install'.`,
+      );
+    }
+    await this.checkTool("npm", "Install Node.js: https://nodejs.org/");
+
+    // The generated package.json defines `cdk` as "npm run build && cdk", so this
+    // single command compiles the app and then synthesizes it. Synthesis needs no
+    // AWS credentials: each stack's environment comes from aws-targets.json.
+    yield { message: "Synthesizing CloudFormation templates" };
+    await this.run(["npm", "run", "cdk", "--", "synth", "--quiet"], cdkDir);
+  }
+
   // Runs a command with its output streamed to the file logger.
   private run(command: string[], cwd: string): Promise<void> {
     return this.runner(command, { cwd, onOutput: (chunk) => this.logger.debug(chunk) });
--- a/src/handlers/project/build/index.ts
+++ b/src/handlers/project/build/index.ts
@@ -1,11 +1,26 @@
-import { createHandler } from "../../../router";
-import { NotImplementedError } from "../../../errors";
+import { createHandler, ProjectKey } from "../../../router";
+import type { AppIO } from "../../../io";
+import type { ProjectManager } from "../types";
 
-export const createBuildProjectHandler = () =>
+type BuildProjectHandlerConfig = {
+  projectManager: ProjectManager;
+  io: AppIO;
+};
+
+export const createBuildProjectHandler = (config: BuildProjectHandlerConfig) =>
   createHandler({
     name: "build",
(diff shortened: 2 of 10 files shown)


src/core/project/manager.tsx
```diff
@@ -118,6 +118,26 @@ export class FsProjectManager implements ProjectManager {
...
+          `Run 'cd ${cdkDir} && npm install'.`,
+      );
+    }
+    await this.checkTool("npm", "Install Node.js: https://nodejs.org/");
+
+    // The generated package.json defines `cdk` as "npm run build && cdk", so this
+    // single command compiles the app and then synthesizes it. Synthesis needs no
+    // AWS credentials: each stack's environment comes from aws-targets.json.
+    yield { message: "Synthesizing CloudFormation templates" };
+    await this.run(["npm", "run", "cdk", "--", "synth", "--quiet"], cdkDir);
```
tejaskash, line 162: Are we not using `@aws-cdk/toolkit-lib` anymore?

src/core/project/manager.tsx
```diff
@@ -118,6 +118,26 @@ export class FsProjectManager implements ProjectManager {
     return project;
   }
 
+  public async *build(project: Project): AsyncGenerator<ProjectEvent, void> {
```
tejaskash, line 126: How does this build impl handle alternative project backends like terraform and SDK if we chose to implement those in the future?

src/handlers/project/build/index.ts
```diff
@@ -1,11 +1,26 @@
...
+      // withProject has already resolved the enclosing project.
+      const project = ctx.require(ProjectKey);
+
+      // Progress goes to stderr, keeping stdout for machine output. Subprocess
+      // output goes to the debug log; on failure ProcessFailedError carries it.
+      for await (const event of config.projectManager.build(project)) {
+        config.io.stderr.write(`${event.message}\n`);
+      }
+
+      config.io.stderr.write(`Built project '${project.name}'\n`);
```
tejaskash, line 24: I like that this is kept clean. It doesn't need to be aware of the build backend or any of the other steps.

src/core/project/manager.tsx
tejaskash, line 162: I think we had discussed that build would: 
- Validate Schemas
- Generate ZIP artifacts for CodeZIP Agents
- Run CDK Synth

I'm not sure if Generate ZIP artifacts for CodeZIP Agents is happening here or if we decided to move that to a different PR.

tejaskash (changes requested):

src/core/project/manager.tsx
notgitika: For now, I went with a subprocess to keep the PR small. using toolkit-lib means porting the wrapper and the schema pinning first, and deploy is what actually needs those. So I will introduce it when I introduce deploy. It should be easy to switch later, build() already yields events so it's a 2-way door decision

src/core/project/manager.tsx
notgitika: In the commit I just pushed, we use the `managedBy` field in agentcore.json

src/core/project/manager.tsx
notgitika: All three are happening. schema validation is up front, `withProject` resolves the project through ProjectSpecSchema, so build can't run on an invalid agentcore.json (#1972  makes that catch the CodeZip runtimeVersion rule and name the field). 

the ZIP is generated by synth: the construct library's packager runs `uv pip install` and stages the asset, so cdk.out gets the zip and the template's `CodeConfiguration.Code.S3` points at it. 

I validated and checked a real synth to be sure. so there's no separate ZIP step to write, doing our own would duplicate the packager deploy relies on. LMK what you think

tejaskash (dismissed):

src/core/project/manager.tsx
```diff
@@ -118,6 +123,45 @@ export class FsProjectManager implements ProjectManager {
...
+      throw new ProjectStateError(
+        `CDK dependencies are missing for project '${project.name}'. ` +
+          `Run 'cd ${cdkDir} && npm install'.`,
+      );
+    }
+    await this.checkTool("npm", "Install Node.js: https://nodejs.org/");
+
+    // The generated package.json defines `cdk` as "npm run build && cdk", so this
+    // single command compiles the app and then synthesizes it. Synthesis needs no
+    // AWS credentials: each stack's environment comes from aws-targets.json.
```
aidandaly24, line 160: Non-blocking and fine as a follow-up: synthesis succeeds without credentials, but `AWS_PROFILE` currently causes the pinned `ConfigIO.readAWSDeploymentTargets()` to call STS even when every target already has an account. I confirmed this by redirecting STS locally. The build still succeeded, but made six `GetCallerIdentity` attempts and inherited the retry latency. Could we avoid that fallback when account values are already present so build is fully offline?

aidandaly24 (dismissed):
This looks good to me one really minor comment:

src/core/project/manager.tsx
notgitika: creating a follow up issue for this. thanks for catching this!

aidandaly24 (approved):

tejaskash (approved):
LGTM

PR #1977, opened 2026-08-11
fix(tui): preserve picker selection after detail views

### LINKED ISSUE #1976: Preserve TUI picker selection after detail views
## Problem

`PaginatedTablePicker` keeps its cursor only in component-local state. Opening a selected resource unmounts the picker, so returning from its detail screen resets the cursor to the first row. This affects every resource flow built on the shared picker.

## Expected experience

- preserve the selected row by stable value when a picker opens a detail screen
- restore that row when the user returns with Escape
- apply the behavior once in `PaginatedTablePicker` for every resource family
- keep existing pagination, resize, transition, and filtering reset behavior
- avoid resource-specific cursor state in individual screens

## Verification

- shared Harness picker regression
- Memory Event picker regression
- full repository coverage suite
- TUI-harness validation across both resource families

aidandaly24:
## Summary

- preserve the selected row when any `PaginatedTablePicker` opens a detail screen and the user returns with Escape
- store the selected row's stable value in that list's history entry and restore it when the picker remounts
- keep the behavior inside the shared picker instead of adding resource-specific navigation state
- retain existing pagination, resize, transition, and filtering reset behavior

Fixes #1976

## Testing

- shared Harness list/detail regression for restoring the third selected row
- Memory Event list/JSON regression for restoring the third selected Event
- `bun test src/components/PaginatedTablePicker.test.tsx src/components/ui/data-table/DataTable.test.tsx src/handlers/memory` (`92` pass)
- `bun test --coverage --coverage-reporter=lcov`
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- TUI harness at `100x28`, independently confirming third-row restoration in both Memory Event and Harness list flows

### DIFF
--- a/src/components/PaginatedTablePicker.test.tsx
+++ b/src/components/PaginatedTablePicker.test.tsx
@@ -180,6 +180,29 @@ describe("paginated table picker contract", () => {
     ).toBe(false);
   });
 
+  test("restores the selected row after returning from its detail screen", async () => {
+    const selected = harness({ harnessName: "third-harness", harnessId: "third-harness" });
+    const core = coreWith([
+      harness({ harnessName: "first-harness", harnessId: "first-harness" }),
+      harness({ harnessName: "second-harness", harnessId: "second-harness" }),
+      selected,
+    ]);
+    core.harness.setGetResponse(getResponse(selected));
+    const r = renderScreen("/agentcore/harness/list", { core });
+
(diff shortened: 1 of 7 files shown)


jariy17 (dismissed):
This is fine but how is having events the first item a better customer experience.

PR #1978, opened 2026-08-11
feat(gateway): add interactive invoke console

aidandaly24:
## Summary

- add a persistent POST JSON console for `agentcore gateway invoke`
- add Gateway switching, raw path editing, generated/continued sessions, and live response streaming
- preserve the headless Gateway invoke contract while allowing CLI flags to seed interactive context
- share the multiline editor with Runtime and keep long payloads within stable terminal rows

## UX

- `Ctrl+T` switches Gateway and clears Gateway-specific state
- `Ctrl+P` edits the Gateway-relative path; blank uses the exact Gateway URL
- `Ctrl+V` toggles completed JSON between raw and pretty forms
- CUSTOM_JWT Gateways proactively require a CLI-provided bearer token
- Runtime and MCP session metadata continues across requests

## Verification

- `206 pass, 0 fail` focused Gateway/Runtime invoke tests
- `1197 pass, 0 fail` full `bun test src`
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- `bun run build`
- live TUI validation: NONE, AWS_IAM, CUSTOM_JWT, AUTHENTICATE_ONLY, JWT passthrough, OAuth 3LO, OAuth OBO, inference, Runtime HTTP, MCP aggregation, Lambda, and real SSE streaming

## Stack

Depends on #1967. Review this PR against `feat/gateway-invoke`.

### DIFF
--- a/src/handlers/gateway/invoke/index.tsx
+++ b/src/handlers/gateway/invoke/index.tsx
@@ -3,9 +3,8 @@ import {
   GatewayInvokeInterruptedError,
   GatewayInvokeResponseError,
   InputValidationError,
-  InvalidEnvironmentError,
 } from "../../../errors";
-import type { AppIO } from "../../../io";
+import { SourceResolver, type AppIO } from "../../../io";
 import { ExitCode } from "../../../runnable";
 import { createHandler, flag, PathKey } from "../../../router";
 import { renderTuiAt } from "../../../tui";
@@ -80,7 +79,7 @@ export const createInvokeGatewayHandler = (core: Core, io: AppIO) =>
           const applicationHeaders = parseGatewayInvokeHeaders(flags.header);
           const bearerToken = await resolveGatewayInvokeTuiBearerToken(
             flags["bearer-token"],
-            io.stdin,
+            new SourceResolver({ stdin: io.stdin }),
           );
           const launchContext = {
             gatewayId: flags.id,
@@ -91,22 +90,12 @@ export const createInvokeGatewayHandler = (core: Core, io: AppIO) =>
             applicationHeaders,
             bearerToken,
           };
-          try {
-            await renderTuiAt(
-              `${ctx.require(PathKey)}/${encodeURIComponent(flags.id)}`,
-              ctx.withValue(GatewayInvokeLaunchContextKey, launchContext),
-              core,
-              io,
-            );
-          } catch (error) {
-            if (error instanceof InvalidEnvironmentError) {
-              throw new InputValidationError(error.message, {
-                cause: error,
-                exitCode: ExitCode.USAGE,
-              });
-            }
-            throw error;
-          }
+          await renderTuiAt(
+            `${ctx.require(PathKey)}/${encodeURIComponent(flags.id)}`,
+            ctx.withValue(GatewayInvokeLaunchContextKey, launchContext),
+            core,
+            io,
+          );
           return;
         }
       }
--- a/src/handlers/gateway/invoke/index.tsx
+++ b/src/handlers/gateway/invoke/index.tsx
@@ -22,7 +22,11 @@ import {
 import { writeGatewayInvokeResponse } from "./response";
 import { GatewayInvokeLaunchContextKey } from "./launchContext";
 
-export const createInvokeGatewayHandler = (core: Core, io: AppIO) =>
+export const createInvokeGatewayHandler = (
+  core: Core,
+  io: AppIO,
+  renderInvokeTui: typeof renderTuiAt = renderTuiAt,
+) =>
   createHandler({
     name: "invoke",
     description: "invoke an AgentCore Gateway",
@@ -90,7 +94,7 @@ export const createInvokeGatewayHandler = (core: Core, io: AppIO) =>
             applicationHeaders,
             bearerToken,
           };
-          await renderTuiAt(
+          await renderInvokeTui(
             `${ctx.require(PathKey)}/${encodeURIComponent(flags.id)}`,
             ctx.withValue(GatewayInvokeLaunchContextKey, launchContext),
             core,
--- a/src/handlers/gateway/invoke/index.tsx
+++ b/src/handlers/gateway/invoke/index.tsx
@@ -3,10 +3,12 @@ import {
   GatewayInvokeInterruptedError,
   GatewayInvokeResponseError,
   InputValidationError,
+  InvalidEnvironmentError,
 } from "../../../errors";
 import type { AppIO } from "../../../io";
 import { ExitCode } from "../../../runnable";
-import { createHandler, flag } from "../../../router";
+import { createHandler, flag, PathKey } from "../../../router";
+import { renderTuiAt } from "../../../tui";
 import { JsonKey } from "../../keys";
 import type { Core } from "../../types";
 import { coreOptsFromCtx } from "../../utils";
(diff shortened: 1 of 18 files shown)


src/handlers/gateway/invoke/screen.tsx
```diff
@@ -0,0 +1,555 @@
...
+    abortRef.current = controller;
+
+    try {
+      const response = await core.gateway.invokeGateway(request, opts, controller.signal);
+      updateExchange({
+        heading: `Response · ${response.statusCode} · ${response.contentType || "-"}`,
+        metadata: metadata(response),
+        state: "streaming",
+      });
+      const kind = classifyStreamingResponse(response.contentType);
```
nborges-aws, line 258: I think we need the same `statusCode !== 204 && statusCode !== 205` clause here that we did in the headless invoke PR

aidandaly24: Good catch. Fixed in `e14e8848` by exempting HTTP 204 and 205 from the TUI binary/unknown-content guard, matching the headless invoke behavior. I also added screen coverage for both statuses with an empty content type and body; the focused Gateway invoke screen suite passes 31/31.

nborges-aws (dismissed):
LGTM

src/handlers/gateway/invoke/index.tsx
```diff
@@ -58,6 +62,55 @@ export const createInvokeGatewayHandler = (core: Core, io: AppIO) =>
       }
 
       const jsonOutput = ctx.require(JsonKey);
+      if (flags.payload === undefined) {
```
Hweinstock, line 68: does this break the pattern of tui only no flags? If so, could this be confusing for customers?

src/handlers/gateway/invoke/index.tsx
```diff
@@ -58,6 +62,55 @@ export const createInvokeGatewayHandler = (core: Core, io: AppIO) =>
...
+          };
+          try {
+            await renderTuiAt(
+              `${ctx.require(PathKey)}/${encodeURIComponent(flags.id)}`,
+              ctx.withValue(GatewayInvokeLaunchContextKey, launchContext),
+              core,
+              io,
+            );
+          } catch (error) {
+            if (error instanceof InvalidEnvironmentError) {
```
Hweinstock: why do we need to re-wrap this? Is there a specific `InvalidEnvironmentError` that should be a `InputValidationError`?

src/handlers/gateway/invoke/invoke.test.tsx
```diff
@@ -377,6 +382,87 @@ describe("gateway invoke", () => {
...
+    await expect(runCommand(core, output.io, ["gateway", "invoke"])).rejects.toThrow(
+      "interactive mode requires a TTY on stdin and stdout",
+    );
+    expect(core.gateway.calls).toEqual([]);
+  });
+
+  test("deep-links an id-only invoke and seeds interactive request context", async () => {
+    const core = configuredCore();
+    const output = captureIO();
+    const render = spyOn(tui, "renderTuiAt").mockResolvedValue(undefined);
```
Hweinstock: is there a way to avoid spying by leveraging the existing dependency injection?

src/handlers/gateway/invoke/request.test.ts
```diff
@@ -65,6 +68,26 @@ describe("Gateway invoke sources", () => {
...
+  const error = await resolveGatewayInvokeTuiBearerToken(missing, stdin()).catch((error) => error);
+
+  expect(error).toBeInstanceOf(InputValidationError);
+  expect(error.message).toContain("could not read '--bearer-token' from file");
+  expect(error.cause).toBeInstanceOf(SourceResolutionError);
+});
+
+test("preserves unexpected TUI bearer-token source failures", async () => {
+  const failure = new TypeError("source failed");
+  const resolve = spyOn(SourceResolver.prototype, "resolveText").mockRejectedValue(failure);
```
Hweinstock: same question as above. Is there way we can inject real implementations?

src/handlers/gateway/invoke/request.ts
```diff
@@ -62,6 +62,25 @@ export async function resolveGatewayInvokeSources(
   }
 }
 
+export async function resolveGatewayInvokeTuiBearerToken(
+  source: string | undefined,
+  stdin: NodeJS.ReadStream,
+): Promise<string | undefined> {
+  if (source === "-") {
+    throw new InputValidationError(
+      "stdin bearer tokens are not available when launching the interactive console",
```
Hweinstock, line 71: don't we need to support stdin for sensitive fields?

src/handlers/gateway/invoke/request.ts
```diff
@@ -62,6 +62,25 @@ export async function resolveGatewayInvokeSources(
...
+    );
+  }
+  try {
+    return await new SourceResolver({ stdin }).resolveText("bearer-token", source);
+  } catch (error) {
+    if (error instanceof SourceResolutionError) {
```
Hweinstock: whats wrong with the sourceResolutionError bubbling up?

src/handlers/gateway/invoke/index.tsx
aidandaly24: Yes, but Runtime invoke already follows this pattern where a missing `--payload` launches the TUI. I think it technically could be a little bit confusing, but I personally think there's are a couple good reasons for why we should keep it this way. For one, we wanted to keep the TUI simple to start. When building runtime invoke I built an advanced options menu, which was accessible in the invoke TUI via a ctrl + key command. This added complexity to the code and we thought it would be better to start simple and layer this complexity on top in the future. So to keep these configurations accessible, I opted to not open the TUI on no `--payload`. Second, I think for things like `--session-id` it is more intuitive to pass it in via the CLI before opening the agent than after opening up the agent.

aidandaly24: The harness invoke TUI is also not strictly no-flags only so I think it's okay to break the pattern on invoke for now.

I think it'd be worthwhile though to have a conversation about this with the team though.

src/handlers/gateway/invoke/index.tsx
aidandaly24: Tbh this only exists to change the exit code. I agree with what I think you may be getting at which is that this is unnecessary complexity. I'll remove this.

Runtime invoke also uses this, and I'll update it there as well. I think what I could do is have this here throw a `new InvalidEnvironmentError(message, { exitCode: ExitCode.USAGE })` instead.

https://github.com/aws/agentcore-cli/blob/refactor/src/tui/index.tsx#L49-L51

Then remove the catch/rewrite here and in Runtime.

src/handlers/gateway/invoke/request.ts
aidandaly24: This is unfortunately a trade off of using the TUI. Headless invoke supports `--bearer-token -`. Interactive invoke deliberately does not because Ink needs the same stdin for keyboard input. Consuming it for the token would drain or block the TUI. Interactive mode supports inline and `file://` tokens. So you still can pass bearer token securely through a file for interactive. Runtime has the same restriction.

Hweinstock (approved):
lgtm! thanks for answering my earlier questions, those decisions make sense.

nborges-aws (approved):

PR #1979, opened 2026-08-12
fix: batch Gateway Connector target scans

aidandaly24:
## Summary

- scan Gateway Targets in service-max 1000-item batches when building logical Connector pages
- replay only the exact Target prefix needed when a scan contains Connector overflow, preserving AWS pagination tokens without custom cursor state
- treat the 101-scan limit as a request safety cap rather than tying it to the adjustable default Target quota

Follow-up to #1956, addressing the late review comments about raised Target quotas and one-request-per-Target behavior with `--max-results 1`.

## Testing

- `bun test src/core/gateway.test.ts src/handlers/gateway/gateway.fixture.test.tsx` (22 pass)
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- `bun run build`
- `bun test src` (1100 pass; 7 unchanged local `src/io/exec.test.ts` process-runner failures on files identical to `refactor`)


src/core/gateway.tsx
```diff
@@ -123,34 +124,50 @@ export class GatewayClient implements CoreGatewayClient {
     maxResults: number | undefined,
     options: CoreOptions,
   ): Promise<ListGatewayTargetsResponse> {
-    const pageSize = maxResults ?? DEFAULT_CONNECTOR_PAGE_SIZE;
+    const connectorPageSize = maxResults ?? DEFAULT_CONNECTOR_PAGE_SIZE;
```
notgitika, line 127: nit: maxResults should be validated as an integer from 1–1000. currently 0, -1 or 1.25 is being accepted

notgitika (approved):
thanks for addressing it!

src/core/gateway.tsx
Hweinstock: would the api reject this for us?

Hweinstock (approved):
i do think we should see if gateway is planning to support server side filters here, because this is fairly complex.

PR #1980, opened 2026-08-12
feat(dev): collect local OTEL traces in project dev

tejaskash:
## Review after #2043

Until #2043 merges, the diff shows both PRs' commits (GitHub's stack tooling wouldn't retarget the base); the work to review here is the single collector commit on top. It collapses to just that once #2043 merges.

## What this does

`agentcore project dev` gives a local agent tracing with no extra config: an in-process OTLP/HTTP receiver starts on a loopback port, the agent's OpenTelemetry SDK is pointed at it via env vars, and traces persist through #2043's TraceStore to `agentcore/.cli/traces/`. Nothing leaves the machine; `--no-traces` or `instrumentation.enableOtel: false` opts out.

- **`io/httpServer.ts`** — loopback-only `node:http` primitive (body caps, abort-signal close).
- **`otel/collector.ts`** — `POST /v1/traces` + `/v1/logs`, protobuf or JSON. Pinned `@opentelemetry/otlp-transformer@0.213.0`, the last version shipping request decoders — do not bump.
- **Handler wiring** — collector lifecycle on the command's abort signal; signal-specific OTEL env vars so stray shell values can't redirect traces; `host.docker.internal` rewrite for containers; sitecustomize-on-PYTHONPATH so `uvicorn --reload` workers stay instrumented (a wrapper would only instrument the reloader parent).

## Verification

Per-layer tests (receiver over real ephemeral-port HTTP with protobuf + JSON fixtures; handler with fake runners/collector) plus end to end: built bundle → real `project create` → `project dev` → real Bedrock invocation → traces on disk. Protobuf decode verified under bun, the Node bundle, and a compiled binary.

### DIFF
--- a/src/core/dev/codezip.ts
+++ b/src/core/dev/codezip.ts
@@ -1,19 +1,28 @@
 import { existsSync } from "node:fs";
-import { join, resolve } from "node:path";
+import { delimiter, join, resolve } from "node:path";
 import { InputValidationError } from "../../errors";
 import type { DevEvent, DevRunner, DevServerInput } from "../../handlers/project/dev/types";
-import { streamProcess, type ProcessStreamer, type StreamProcessOptions } from "../../io";
+import {
+  runProcess,
+  streamProcess,
+  type ProcessRunner,
+  type ProcessStreamer,
+  type StreamProcessOptions,
+} from "../../io";
 import { isDirectory, isFile, resolvePathWithinProject } from "./path";
 
 type CodeZipDevRunnerConfig = {
   streamProcess?: ProcessStreamer;
+  runProcess?: ProcessRunner;
 };
 
 export class CodeZipDevRunner implements DevRunner {
   private readonly streamProcess: ProcessStreamer;
+  private readonly runProcess: ProcessRunner;
 
   constructor(config: CodeZipDevRunnerConfig = {}) {
     this.streamProcess = config.streamProcess ?? streamProcess;
+    this.runProcess = config.runProcess ?? runProcess;
   }
 
   public async *run(input: DevServerInput): AsyncGenerator<DevEvent, void> {
@@ -41,7 +50,47 @@ export class CodeZipDevRunner implements DevRunner {
 
     yield { type: "status", message: "Starting development server" };
     const serverProcess = commandForRuntime(entrypoint!, directory, input);
+    if (entrypoint!.endsWith(".py") && input.env?.OTEL_EXPORTER_OTLP_ENDPOINT) {
+      const sitecustomizeDir = await this.findOtelSitecustomizeDir(directory);
+      if (sitecustomizeDir) {
+        const existing = serverProcess.options.env?.PYTHONPATH;
+        serverProcess.options.env = {
+          ...serverProcess.options.env,
+          PYTHONPATH: existing ? `${sitecustomizeDir}${delimiter}${existing}` : sitecustomizeDir,
+        };
+      } else {
+        yield {
+          type: "status",
+          message:
+            "OTEL auto-instrumentation is not installed in the agent environment; traces will not be collected. Add aws-opentelemetry-distro to the agent's dependencies to enable them.",
+        };
+      }
+    }
     yield* this.streamProcess(serverProcess.command, serverProcess.options);
+  }
+
+  /**
+   * Locate the auto-instrumentation sitecustomize.py directory inside the agent's
+   * uv environment. Prepending it to PYTHONPATH instruments every Python process —
+   * an `opentelemetry-instrument` wrapper would only instrument uvicorn's reloader
+   * parent, leaving the re-spawned worker processes untraced.
+   */
+  private async findOtelSitecustomizeDir(directory: string): Promise<string | undefined> {
+    const output: string[] = [];
+    const probe =
+      "import opentelemetry.instrumentation.auto_instrumentation as m; import os; print(os.path.dirname(m.__file__))";
+    try {
+      await this.runProcess(["uv", "run", "python", "-c", probe], {
+        cwd: directory,
+        onOutput: (chunk) => output.push(chunk),
+      });
+    } catch {
+      return undefined;
+    }
+    const sitecustomizeDir = output.join("").trim().split("\n").at(-1)?.trim();
+    if (!sitecustomizeDir || !existsSync(join(sitecustomizeDir, "sitecustomize.py")))
+      return undefined;
+    return sitecustomizeDir;
   }
 }
 
--- a/src/core/dev/codezip.ts
+++ b/src/core/dev/codezip.ts
@@ -16,6 +16,8 @@ type CodeZipDevRunnerConfig = {
   runProcess?: ProcessRunner;
 };
 
+const SITECUSTOMIZE_MARKER = "AGENTCORE_OTEL_SITECUSTOMIZE=";
+
 export class CodeZipDevRunner implements DevRunner {
   private readonly streamProcess: ProcessStreamer;
   private readonly runProcess: ProcessRunner;
@@ -51,7 +53,7 @@ export class CodeZipDevRunner implements DevRunner {
     yield { type: "status", message: "Starting development server" };
     const serverProcess = commandForRuntime(entrypoint!, directory, input);
     if (entrypoint!.endsWith(".py") && input.env?.OTEL_EXPORTER_OTLP_ENDPOINT) {
-      const sitecustomizeDir = await this.findOtelSitecustomizeDir(directory);
+      const sitecustomizeDir = await this.findOtelSitecustomizeDir(directory, input.signal);
       if (sitecustomizeDir) {
         const existing = serverProcess.options.env?.PYTHONPATH;
         serverProcess.options.env = {
@@ -75,19 +77,29 @@ export class CodeZipDevRunner implements DevRunner {
    * an `opentelemetry-instrument` wrapper would only instrument uvicorn's reloader
    * parent, leaving the re-spawned worker processes untraced.
    */
-  private async findOtelSitecustomizeDir(directory: string): Promise<string | undefined> {
+  private async findOtelSitecustomizeDir(
+    directory: string,
+    signal: AbortSignal,
+  ): Promise<string | undefined> {
     const output: string[] = [];
-    const probe =
-      "import opentelemetry.instrumentation.auto_instrumentation as m; import os; print(os.path.dirname(m.__file__))";
+    // uv writes sync progress to stderr, which merges into onOutput, so the path
+    // is printed behind a marker and read from that line rather than the last one.
+    const script = `import opentelemetry.instrumentation.auto_instrumentation as m, os; print("${SITECUSTOMIZE_MARKER}" + os.path.dirname(m.__file__))`;
     try {
-      await this.runProcess(["uv", "run", "python", "-c", probe], {
+      await this.runProcess(["uv", "run", "python", "-c", script], {
         cwd: directory,
         onOutput: (chunk) => output.push(chunk),
+        signal,
       });
     } catch {
       return undefined;
     }
-    const sitecustomizeDir = output.join("").trim().split("\n").at(-1)?.trim();
+    const marked = output
+      .join("")
+      .split("\n")
+      .map((line) => line.trim())
+      .find((line) => line.startsWith(SITECUSTOMIZE_MARKER));
+    const sitecustomizeDir = marked?.slice(SITECUSTOMIZE_MARKER.length);
     if (!sitecustomizeDir || !existsSync(join(sitecustomizeDir, "sitecustomize.py")))
       return undefined;
     return sitecustomizeDir;
--- a/src/core/dev/container.ts
+++ b/src/core/dev/container.ts
@@ -183,6 +183,11 @@ export class ContainerDevRunner implements DevRunner {
       containerName,
       "-p",
       `127.0.0.1:${input.port}:${containerPort}`,
+      // Docker Engine on Linux does not define host.docker.internal (Desktop,
+      // Podman, and Finch do); the mapping makes the OTLP endpoint rewrite
+      // resolve everywhere and is harmless where the name already exists.
+      "--add-host",
+      "host.docker.internal:host-gateway",
       ...awsMount,
       "--env-file",
       envFile,
--- a/src/core/dev/otel/collector.ts
+++ b/src/core/dev/otel/collector.ts
@@ -1,0 +1,152 @@
+// Decodes OTLP/HTTP protobuf payloads (the only protocol Python and Node OTEL
+// SDKs export over HTTP) with the generated types from @opentelemetry/otlp-transformer.
+// The version is pinned: newer releases dropped the generated request decoders.
+import root from "@opentelemetry/otlp-transformer/build/src/generated/root";
+import { type HttpRequest, type HttpResponse, startHttpServer } from "../../../io";
+import { TraceStore } from "./store";
+import type { OtlpPayload } from "./types";
+
+/** The slice of a generated protobufjs message type the collector (and its tests) use. */
+export interface OtlpMessageType {
+  decode(data: Uint8Array): unknown;
+  fromObject(object: object): unknown;
+  encode(message: unknown): { finish(): Uint8Array };
+}
+
+// The generated root's declaration file types it as an opaque protobufjs Root,
+// so the real static-message shape is asserted once, here.
+const { trace, logs } = (
+  root as unknown as {
+    opentelemetry: {
+      proto: {
+        collector: {
+          trace: { v1: { ExportTraceServiceRequest: OtlpMessageType } };
+          logs: { v1: { ExportLogsServiceRequest: OtlpMessageType } };
+        };
+      };
+    };
+  }
+).opentelemetry.proto.collector;
+
+export const ExportTraceServiceRequest = trace.v1.ExportTraceServiceRequest;
+export const ExportLogsServiceRequest = logs.v1.ExportLogsServiceRequest;
+type OtlpDecoder = Pick<OtlpMessageType, "decode">;
+
+export interface OtelCollector {
+  /** The loopback port the OTLP/HTTP receiver listens on. */
+  port: number;
+  /** Reads the traces this collector persists. */
+  store: TraceStore;
+  /** Environment variables that point an agent's OTEL SDK at this collector. */
+  envVars: Record<string, string>;
+  /** Stops the receiver. Also invoked by the start signal, if one was given. */
+  close(): Promise<void>;
+}
+
+export interface StartOtelCollectorOptions {
+  /** Directory to persist OTLP JSON Lines files into. */
+  tracesDirectory: string;
+  /** Closes the collector when aborted. */
+  signal?: AbortSignal;
+}
+
+/**
+ * Starts an in-process OTLP/HTTP receiver for dev mode on an OS-assigned
+ * loopback port. Accepts `POST /v1/traces` and `POST /v1/logs` in protobuf or
+ * JSON encoding and appends the raw payloads to a TraceStore.
+ */
+export async function startOtelCollector(
+  options: StartOtelCollectorOptions,
+): Promise<OtelCollector> {
+  const store = new TraceStore(options.tracesDirectory);
+  const server = await startHttpServer((request) => route(request, store), {
+    signal: options.signal,
+  });
+
+  return { port: server.port, store, envVars: otelEnvVars(server.port), close: server.close };
+}
+
+async function route(request: HttpRequest, store: TraceStore): Promise<HttpResponse> {
+  if (request.method === "POST" && request.url === "/v1/traces") {
+    return ingest(request, store, ExportTraceServiceRequest);
+  }
+  if (request.method === "POST" && request.url === "/v1/logs") {
+    return ingest(request, store, ExportLogsServiceRequest);
+  }
+  if (request.method === "GET" && request.url === "/") {
+    return json(200, { status: "ok" });
+  }
+  return { status: 404 };
+}
+
+async function ingest(
+  request: HttpRequest,
+  store: TraceStore,
+  decoder: OtlpDecoder,
+): Promise<HttpResponse> {
+  let payload: OtlpPayload;
+  try {
+    payload = decodePayload(request.body, String(request.headers["content-type"] ?? ""), decoder);
+  } catch {
+    return json(400, { error: "Invalid OTLP payload" });
+  }
+  await store.append(payload);
+  return json(200, {});
+}
+
+/**
+ * Decode an OTLP payload. The JSON round-trip on the protobuf path converts the
+ * message to plain objects (protobufjs renders Long as string and bytes as base64).
+ */
+function decodePayload(body: Buffer, contentType: string, decoder: OtlpDecoder): OtlpPayload {
+  if (contentType.includes("application/json")) {
+    return JSON.parse(body.toString()) as OtlpPayload;
+  }
+  return JSON.parse(JSON.stringify(decoder.decode(new Uint8Array(body)))) as OtlpPayload;
+}
+
+/**
+ * Environment for a spawned agent so its OTEL SDK exports to the collector at
+ * `port`. Signal-specific variables are set alongside the generic ones because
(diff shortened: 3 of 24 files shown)


Hweinstock:
based on description, it sounds like https://github.com/aws/agentcore-cli/pull/2039 needs to land first so looking at that one now.

src/assets/cdk/bin/cdk.ts
```diff
@@ -110,12 +110,9 @@ async function main() {
...
   const targets = await configIO.readAWSDeploymentTargets();
-
-  // `project build` runs before a project has anywhere to deploy, so an empty target
-  // list is not an error: it synthesizes a single environment-agnostic stack, which is
-  // enough to typecheck the app and produce a template. Only a stack synthesized for a
-  // real target is a deploy candidate; the target tag below is what marks one.
-  const stackTargets: (AwsDeploymentTarget | undefined)[] = targets.length > 0 ? targets : [undefined];
+  if (targets.length === 0) {
+    throw new Error('No deployment targets configured. Please define targets in agentcore/aws-targets.json');
+  }
```
notgitika: we don't need an aws-targets.json to be populated in order to go ahead with build. build is AWS account agnostic. that is why we had that implementation earlier.  that was done in this PR https://github.com/aws/agentcore-cli/pull/2011/changes

is there a reason we added it back? if not, can we revert it?

tejaskash:
All four addressed:
- Unrelated reversions — the commit was a whole-tree snapshot taken off a stale base, silently reverting everything merged since. Rebuilt from explicit paths; diff is OTEL/dev only now.
- Linux host-gateway — container runs pass `--add-host=host.docker.internal:host-gateway`; harmless on Desktop/Podman/Finch.
- 413 before close — oversized bodies get the 413 (+ `Connection: close`) before the socket closes. Tested with a real 51 MiB request.

src/assets/cdk/bin/cdk.ts
tejaskash: Stale rebase — the branch no longer touches cdk.ts (`git diff origin/refactor...HEAD` is empty). Nothing to revert.

src/core/dev/otel/collector.ts
```diff
@@ -0,0 +1,166 @@
...
+  onError?: (error: unknown) => void;
+}
+
+/**
+ * Starts an in-process OTLP/HTTP receiver for dev mode on an OS-assigned
+ * loopback port. Accepts `POST /v1/traces` and `POST /v1/logs` in protobuf or
+ * JSON encoding and appends the raw payloads to a TraceStore.
+ */
+export async function startOtelCollector(
+  options: StartOtelCollectorOptions,
```
Hweinstock, line 63: when I see options I think optional, but I think `tracesDirectory` is required. Do you think moving the required params into `input` would communicate this better?

src/core/dev/otel/collector.ts
```diff
@@ -0,0 +1,166 @@
...
+  if (contentType.includes("application/json")) {
+    return JSON.parse(body.toString()) as OtlpPayload;
+  }
+  return JSON.parse(JSON.stringify(decoder.decode(new Uint8Array(body)))) as OtlpPayload;
+}
+
+/**
+ * Environment for a spawned agent so its OTEL SDK exports to the collector at
+ * `port`. Signal-specific variables are set alongside the generic ones because
+ * they take precedence in the SDK — a stray OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
```
Hweinstock: would it be a valid use case for a customer to bring their own collector? I could imagine a case where a customer is developing an agent, and already has some collector they use in production that they want to send local traces to. 

I agree with the choice to not silently pick up on this, but wondering if theres a way we can offer an explicit opt-out of our collector.

src/io/httpServer.ts
```diff
@@ -0,0 +1,119 @@
+// Uses node:http rather than Bun.serve because the npm bundle targets Node,
```
Hweinstock, line 1: i feel like `io` is kind of becoming a utils dumping ground. does it make sense to keep this in `otel/` until we are ready to re-use? The logic looks mostly generic, but I see an OTLP reference in line 73, so unsure if we want this as a common abstraction yet. Unless we already see a place we can re-use this?

src/router/flags.tsx
```diff
@@ -5,15 +5,17 @@ import type { Flag, GlobalFlag } from "./handler";
 import { coerce, formatZodError, inspect } from "./schema";
 
 // toOption builds a Commander Option from a flag's schema. Booleans become value-less
-// toggles; everything else takes a value (`<name>` / variadic `<name...>`). A
-// required, non-boolean flag is made mandatory; defaults are forwarded.
+// toggles — declared as `--no-<name>` when they default to true, so the flag turns
```
Hweinstock: this comment is hard to understand. why exactly do we want this behavior?

src/core/dev/otel/collector.ts
```diff
@@ -0,0 +1,166 @@
+// Decodes OTLP/HTTP protobuf payloads (the only protocol Python and Node OTEL
```
Hweinstock, line 1: what do you think about moving this to the overrideComments in the package.json like we did in the old CLI? https://github.com/aws/agentcore-cli/blob/aa855edf1da47181f9035666bb81d7a564c378c9/package.json#L168. 

I kind of liked that since it kept the comment close to the source.

src/core/dev/otel/collector.ts
```diff
@@ -0,0 +1,166 @@
...
+// SDKs export over HTTP) with the generated types from @opentelemetry/otlp-transformer.
+// The version is pinned: newer releases dropped the generated request decoders.
+import root from "@opentelemetry/otlp-transformer/build/src/generated/root";
```
Hweinstock, line 4: i'm kind of confused why we need to reach in to the internals of this package like this. Wouldn't anyone implementing a collector need these shapes, where are they getting it from?

src/core/dev/otel/collector.ts
```diff
@@ -0,0 +1,166 @@
...
+    OTEL_INSTRUMENTATION_GENAI_CAPTURE_MESSAGE_CONTENT: "true",
+    OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED: "true",
+  };
+}
+
+/**
+ * Rewrite loopback OTLP endpoints so a containerized agent can reach the
+ * collector on the host. host.docker.internal resolves on Docker Desktop,
+ * Finch, and Podman; bare-metal Linux Docker would additionally need
+ * `--add-host=host.docker.internal:host-gateway` (matches the reference CLI).
```
Hweinstock: is the old CLI the reference CLI? Could this comment become outdated?

src/core/dev/otel/collector.ts
```diff
@@ -0,0 +1,166 @@
...
+ */
+export function rewriteOtelEndpointForContainer(
```
Hweinstock, line 185: is there a way we can avoid overrides and instead return the correct endpoints the first time in `otelEnvVars` above?

It feels simpler to get once, rather than get than overwrite but maybe I'm missing something that makes this harder.

tejaskash:
#2039 was superseded by #2043, which just merged — this is unblocked and rebased on it.

src/core/dev/otel/collector.ts
```diff
@@ -0,0 +1,166 @@
...
+  let payload: OtlpPayload;
+  try {
+    payload = decodePayload(request.body, String(request.headers["content-type"] ?? ""), decoder);
+  } catch {
+    return json(400, { error: "Invalid OTLP payload" });
+  }
+  try {
+    await store.append(payload);
+  } catch (error) {
+    // A persistence failure (disk full, permissions) is the collector's problem,
```
Hweinstock, line 110: is this comment describing a specific implementation of `onError`?

src/core/dev/otel/collector.ts
```diff
@@ -0,0 +1,166 @@
...
+    // not the agent's: ack the export so the SDK exporter stops retrying the batch
+    // forever, and report it once so the user isn't left with silently missing traces.
+    onError?.(error);
+  }
+  return json(200, {});
```
notgitika: nit: seems like only decode failures return 400. A structurally incorrect but valid json like ({"resourceSpans": 5}) throws inside `store.append` and gets reported as a persistence error, so the user sees "failed to persist traces to" for what is actually a bad payload

src/core/dev/otel/collector.ts
```diff
@@ -0,0 +1,166 @@
...
+
+/**
+ * Decode an OTLP payload. The JSON round-trip on the protobuf path converts the
+ * message to plain objects (protobufjs renders Long as string and bytes as base64).
+ */
+function decodePayload(body: Buffer, contentType: string, decoder: OtlpDecoder): OtlpPayload {
+  if (contentType.includes("application/json")) {
+    return JSON.parse(body.toString()) as OtlpPayload;
+  }
+  return JSON.parse(JSON.stringify(decoder.decode(new Uint8Array(body)))) as OtlpPayload;
```
notgitika: nit: this seems too complicated with the parse, stringify and then decode. can we simplify this?

src/core/dev/codezip.ts
```diff
@@ -41,8 +50,48 @@ export class CodeZipDevRunner implements DevRunner {
 
     yield { type: "status", message: "Starting development server" };
     const serverProcess = commandForRuntime(entrypoint!, directory, input);
+    if (entrypoint!.endsWith(".py") && input.env?.OTEL_EXPORTER_OTLP_ENDPOINT) {
```
Hweinstock, line 55: is there a way to avoid hardcoding this to python? My understanding is that this is the code that causes the agent code to emit traces, which we're going to need for TS but will require some different logic (or other languages we support going forward). 

Is there a way to provide an interface or establish a clear extension path to other languages?

src/core/dev/codezip.ts
```diff
@@ -41,8 +50,48 @@ export class CodeZipDevRunner implements DevRunner {
...
+   * uv environment. Prepending it to PYTHONPATH instruments every Python process —
+   * an `opentelemetry-instrument` wrapper would only instrument uvicorn's reloader
+   * parent, leaving the re-spawned worker processes untraced.
+   */
+  private async findOtelSitecustomizeDir(directory: string): Promise<string | undefined> {
+    const output: string[] = [];
+    const probe =
+      "import opentelemetry.instrumentation.auto_instrumentation as m; import os; print(os.path.dirname(m.__file__))";
+    try {
+      await this.runProcess(["uv", "run", "python", "-c", probe], {
```
Hweinstock: kind of related to above, but I wonder if language specific functionality should live elsewhere so avoid special casing. happy to sync on this since I don't see an obvious way to accomplish this.

src/core/dev/otel/collector.ts
tejaskash: TypeScript already makes `tracesDirectory` required, so you can't omit it. A single options object is also how we do it elsewhere (`startHttpServer`, `DevServerInput`). I'd keep it.

src/core/dev/otel/collector.ts
tejaskash: `--no-traces` does exactly this. It skips our collector and injects nothing, so the agent keeps your own `OTEL_*` env and sends traces to your collector.

src/io/httpServer.ts
tejaskash: It's already reused I have a PR in flight where Inspector server imports it. The only OTLP-specific thing was that one comment, now generalized in 507dfd64.

src/router/flags.tsx
tejaskash: Rewrote it in 507dfd64. A boolean that's on by default is shown as `--no-<name>` so you can turn it off, and Commander stores that under the positive name, so `--no-traces` sets `traces=false`.

src/core/dev/otel/collector.ts
tejaskash: This is a normal dependency, not an override, so `overrideComments` doesn't apply. package.json can't hold comments anyway. The reason is tied to the import, so it lives there.

src/core/dev/otel/collector.ts
tejaskash: Agents export protobuf, not JSON, so we have to decode protobuf. The decoder is only reachable at `build/src/generated/root` (it isn't in the package's public exports), and newer versions removed it, so we pin `0.213.0`. The old CLI does the same.

src/core/dev/otel/collector.ts
tejaskash: Fixed in 507dfd64. Dropped that line.

src/core/dev/otel/collector.ts
tejaskash: The collector always uses loopback. Only the handler knows when it's a container and needs `host.docker.internal`, so the rewrite happens there. I can instead pass a `host` into `otelEnvVars` and drop the rewrite. Want that here or later?

src/core/dev/otel/collector.ts
tejaskash: You're right. Fixed in 507dfd64 so the comment only says what the collector does: ack the export and call `onError`. The caller decides how to show it.

src/core/dev/codezip.ts
tejaskash: Agreed, it's Python-only. I don't want to add an abstraction for one language yet. When we add TS traces (#1893), I'll pull this into a per-language step then. Let's sync.

src/core/dev/codezip.ts
tejaskash: Same as above. When a second language needs it, I'll move this out of the `.py` check into a per-language step. Let's sync.

src/core/dev/codezip.ts
```diff
@@ -41,8 +50,48 @@ export class CodeZipDevRunner implements DevRunner {
...
+        onOutput: (chunk) => output.push(chunk),
+      });
+    } catch {
+      return undefined;
+    }
+    const sitecustomizeDir = output.join("").trim().split("\n").at(-1)?.trim();
+    if (!sitecustomizeDir || !existsSync(join(sitecustomizeDir, "sitecustomize.py")))
+      return undefined;
+    return sitecustomizeDir;
+  }
```
notgitika: `onOutput` is stdout and stderr merged, and uv writes progress to stderr so `.at(-1)` isn't always the python print. when it picks a uv line we tell the user to install aws-opentelemetry-distro on a project that already has it, and tracing just quietly doesn't happen.

can we print a marker and match on that instead?

src/core/dev/container.ts
```diff
@@ -183,6 +183,11 @@ export class ContainerDevRunner implements DevRunner {
       containerName,
       "-p",
       `127.0.0.1:${input.port}:${containerPort}`,
+      // Docker Engine on Linux does not define host.docker.internal (Desktop,
+      // Podman, and Finch do); the mapping makes the OTLP endpoint rewrite
+      // resolve everywhere and is harmless where the name already exists.
+      "--add-host",
```
notgitika, line 189: I don’t think host-gateway makes the collector reachable on Linux. It resolves to the host bridge address, while the collector only listens on 127.0.0.1. Could we use a container-reachable listener or proxy and add a Linux container test?

src/io/httpServer.ts
```diff
@@ -0,0 +1,119 @@
...
+ * the given port (0 lets the OS assign one). Handler errors become plain 500s;
+ * oversized bodies become 413s. Aborting the signal closes the server.
+ */
+export async function startHttpServer(
+  handler: HttpRequestHandler,
+  options: { port?: number; signal?: AbortSignal } = {},
+): Promise<HttpServerHandle> {
+  const server = createServer((request, response) => {
+    void respond(handler, request, response);
+  });
```
notgitika, line 50: there is no `.catch()` here, and the 500 fallback below calls `writeHead` which throws if we already started writing. exporters time out and disconnect mid-response, and nothing in `src/index.ts` catches unhandled rejections, so that kills `project dev`.

can we add `.catch(() => {})` here and guard the catch with `if (!response.headersSent)`? the existing test throws before headers go out, so it takes the happy path through that catch.

src/core/dev/container.ts
```diff
@@ -183,6 +183,11 @@ export class ContainerDevRunner implements DevRunner {
...
+      "host.docker.internal:host-gateway",
```
aidandaly24, line 190: I think this fixes hostname resolution on Linux, but the container still cannot reach the collector because it only listens on `127.0.0.1`. I tested the current head with native Linux Docker 25: `host.docker.internal` resolved, but the request failed with `ECONNREFUSED`. The same container reached a control server bound to `0.0.0.0`.

src/handlers/project/dev/index.ts
```diff
@@ -103,12 +117,47 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
...
+                config.io,
+                {
+                  type: "status",
+                  message: `Warning: failed to persist traces to ${tracesDirectory} (${detail}); collected traces may be incomplete.`,
+                },
+                json,
+              );
+            },
+          });
+          otelEnv = otelEnvForRuntime(collector, runtime);
```
aidandaly24, line 145: I may be missing where this happens for container builds, but I think this only points an existing OTEL SDK at the collector and it does not start instrumentation. I generated and built the current `hello-world-python-container` template with these environment variables. After importing the complete app, it still had only a `ProxyTracerProvider`. The same image under `opentelemetry-instrument` had a real `TracerProvider`. Would it make sense to make sure the container path actually starts instrumentation, or update the scaffolded Dockerfile so the default container template can produce traces?

src/handlers/project/dev/index.ts
```diff
@@ -103,12 +117,47 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
         });
         controller.signal.throwIfAborted();
 
+        let otelEnv: Record<string, string> = {};
+        if (flags.traces && (runtime.instrumentation?.enableOtel ?? true)) {
+          const tracesDirectory = join(project.rootPath, "agentcore", ".cli", "traces", "otlp");
+          let tracePersistErrorReported = false;
+          collector = await config.startTraceCollector({
+            tracesDirectory,
+            signal: controller.signal,
```
aidandaly24: I think sharing this signal with the collector can lose the agent's final spans. `streamProcess` sends the child `SIGTERM` and gives it up to two seconds to exit, but the collector closes immediately on the same abort. I tested a child exporting from its `SIGTERM` handler and it received `ECONNREFUSED`. Since the `finally` block already closes the collector after the runner exits, should the collector stay alive until then?

src/handlers/project/dev/index.ts
```diff
@@ -103,12 +117,47 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
...
+        }
+        controller.signal.throwIfAborted();
+
         const runner = config.runners[runtime.build];
         for await (const event of runner.run({
           runtime,
           projectRoot: project.rootPath,
           port: devPort.port,
-          env,
+          env: { ...env, ...otelEnv },
```
aidandaly24, line 162: I think loaded OTEL settings can still break the local setup because only variables present in `otelEnv` are overridden here. I tested a real Python exporter with `OTEL_EXPORTER_OTLP_COMPRESSION=gzip`: the process exited successfully, but the collector returned `400` and persisted no trace and `none` persisted it. Settings such as `OTEL_SDK_DISABLED=true` or `OTEL_TRACES_EXPORTER=none` would also survive this merge. Since the environment template says the CLI owns OTEL settings while collection is enabled, would it make sense to clear or override settings that are incompatible with the local collector?

src/core/dev/otel/collector.ts
```diff
@@ -0,0 +1,166 @@
...
+  return json(200, {});
+}
+
+/**
+ * Decode an OTLP payload. The JSON round-trip on the protobuf path converts the
+ * message to plain objects (protobufjs renders Long as string and bytes as base64).
+ */
+function decodePayload(body: Buffer, contentType: string, decoder: OtlpDecoder): OtlpPayload {
+  if (contentType.includes("application/json")) {
+    return JSON.parse(body.toString()) as OtlpPayload;
```
aidandaly24: nit: I think this cast lets syntactically valid but invalid OTLP JSON bypass the `400` path. I tried `null`, `[]`, and `42` and they all returned `200`, and `null` also went through `onError`, which can consume the command's one persistence-error warning. Would it make sense to validate the top-level OTLP shape before calling `store.append()`?

src/core/dev/codezip.ts
aidandaly24: nit: Would it make sense to make this probe abortable as well? It runs before the server process and `runProcess` has no signal option, so `Ctrl+C` while `uv run` is still resolving or syncing the environment leaves `project dev` waiting for that child even though the command has already been aborted.

src/core/dev/otel/collector.ts
tejaskash: Fixed in adf371b8. The top-level shape is now checked before `append`, so `{"resourceSpans": 5}` returns 400 instead of surfacing as a persistence error.

src/core/dev/otel/collector.ts
tejaskash: Pulled the round-trip into a named `decodeProtobufToPlainObject` in adf371b8. The stringify then parse is what flattens protobufjs Long to string and bytes to base64, which is the shape the rest of the code reads, so it stays, just clearer now.

src/core/dev/codezip.ts
tejaskash: Fixed in adf371b8. The path is now printed behind a marker and read from that line, so uv's stderr progress can no longer be mistaken for it.

src/core/dev/container.ts
tejaskash: Fixed in adf371b8. The collector now binds 0.0.0.0 for container runtimes, so a Linux container reaches it over the bridge. The host-gateway mapping stays for name resolution.

src/io/httpServer.ts
tejaskash: Fixed in adf371b8. Added the `.catch(() => {})` and an `if (!response.headersSent)` guard, so a client that disconnects mid-response can no longer take down `project dev`.

src/core/dev/container.ts
tejaskash: Thanks for the ECONNREFUSED repro, that was the missing piece. Fixed in adf371b8: the collector binds 0.0.0.0 for container runtimes so the container reaches it over the bridge.

src/handlers/project/dev/index.ts
tejaskash: Good catch. Fixed in adf371b8. The container template now runs under `opentelemetry-instrument`, which starts a real TracerProvider instead of leaving the proxy one.

src/handlers/project/dev/index.ts
tejaskash: Right, the shared signal closed the collector too early. Fixed in adf371b8: the collector no longer takes the signal and closes only in `finally`, after the child's shutdown grace, so the final spans land.

src/handlers/project/dev/index.ts
tejaskash: Fixed in adf371b8. While tracing is on the CLI now forces the settings that would break local collection: OTEL_SDK_DISABLED=false, traces and logs exporter=otlp, and compression=none. We merge these over the loaded env, so a gzip or a none from .env.local no longer wins.

src/core/dev/otel/collector.ts
tejaskash: Fixed in adf371b8, same change as Gitika's note. The top-level shape is validated before `append`, so `null`, `[]`, and `42` return 400 and never reach onError.

src/core/dev/codezip.ts
tejaskash: Fixed in adf371b8. runProcess now takes a signal and the discovery step passes `input.signal`, so Ctrl+C during `uv run` cancels it instead of leaving dev waiting.

src/assets/cdk/bin/cdk.ts
tejaskash: This branch doesn't touch `cdk.ts`, and there's no aws-targets requirement in it on the current head, so there's nothing to revert here. If you're still seeing one, point me at the line and I'll take it.

notgitika (approved):

aidandaly24 (approved):
Thanks for making the fixes. This looks good to me!

PR #1983, opened 2026-08-12
feat(eval): add ondemand evaluate (synchronous, client-side)

jariy17:
## Summary

Adds `agentcore eval ondemand evaluate` — a **client-side** evaluation of existing sessions. On-demand gathers the sessions' traces from CloudWatch **on the client**, calls the `Evaluate` data-plane API directly, and prints scores. 

|  | `batch-evaluation evaluate` | `ondemand evaluate` (this PR) |
|---|---|---|
| SDK call | `StartBatchEvaluation` | `Evaluate` (data plane) |
| Trace gathering | service-side | **client-side** (CloudWatch Logs Insights) |
| Returns | job id (poll with `get`) | scores, synchronously |
| Source arms | `--agent` / `--online-eval` / `--data-source-config` | `--agent` only |

## Usage

```
agentcore eval ondemand evaluate \
  --agent <harness-id|runtime-id> \
  --evaluator Builtin.Helpfulness \
  --session-ids <id...>            # or --lookback-days N, or --start-time/--end-time
```

### Flags
- `--agent` (required), `--endpoint`, `--evaluator <ids...>` (required)
- time filter: `--lookback-days N` **or** `--start-time`/`--end-time` (ISO-8601, together)
- `--session-ids <ids...>`, `--trace-id <id>` — independent, AND-ed fetch filters
- `--ground-truth <json>` — inline / `file://` / `-` → SDK-native `EvaluationReferenceInput[]`

## Tests

- Golden fixture suite (`ondemand.fixture.test.tsx`) — recorded `GetAgentRuntime` + Insights `StartQuery`/`GetQueryResults` (both log groups) + `Evaluate` fixtures, driven through the real root handler against a pinned window, diffed against `evaluate.golden.json`.
- Command-flow suite (`ondemand.test.tsx`, `TestCoreClient`) — source-arm validation, `getTracesForAgent → evaluate` orchestration/order, `--lookback-days` window math, `--trace-id`, ground-truth passthrough.

`tsc --noEmit`, `bun test src/` (1067 pass), `oxlint`, `prettier` — all clean.


src/core/eval.tsx
```diff
@@ -1261,6 +1364,212 @@ async function agentDataSource(
...
+
+// extractToolCallSpanIds pulls span ids of tool-call spans — those carrying a tool
+// name attribute (kind=CLIENT alone is too broad, per the old CLI).
+function extractToolCallSpanIds(spans: SpanRecord[]): string[] {
+  const spanIds: string[] = [];
+  for (const span of spans) {
+    const spanId = span.spanId;
+    if (typeof spanId !== "string" || spanId.length === 0) continue;
+    const attrs = span.attributes as Record<string, unknown> | undefined;
+    if (attrs?.["gen_ai.tool.name"] ?? attrs?.["tool.name"]) spanIds.push(spanId);
```
aidandaly24: Should tool-call selection use the standard operation/kind markers rather than requiring a tool-name attribute? Current telemetry identifies tool spans using `gen_ai.operation.name === "execute_tool"`, `openinference.span.kind === "TOOL"`, or `traceloop.span.kind === "tool"`. The name fields are not always present. I reproduced a valid LangGraph-style tool span producing an empty `toolCallSpanIds`, so a `TOOL_CALL` evaluator makes no `Evaluate` request. Could we use the same marker logic as the evaluation SDK?

src/core/eval.tsx
```diff
@@ -1261,6 +1364,212 @@ async function agentDataSource(
...
+  evaluatorIds: string[],
+  control: BedrockAgentCoreControlClient,
+  _options: CoreOptions,
+): Promise<Map<string, EvaluatorLevel>> {
+  const levels = new Map<string, EvaluatorLevel>();
+  for (const id of new Set(evaluatorIds)) {
+    try {
+      const evaluator = await control.send(new GetEvaluatorCommand({ evaluatorId: id }));
+      levels.set(id, evaluator.level ?? "SESSION");
+    } catch {
```
aidandaly24: Could evaluator lookup failures propagate instead of silently defaulting to `SESSION`? The level determines whether `Evaluate` receives `traceIds`, `spanIds`, or no target. I reproduced an `AccessDeniedException` here causing a trace evaluator to be submitted without an evaluation target, which can either evaluate the wrong scope or hide the actual permissions error. If a fallback is needed for compatibility, could it be limited to a missing level rather than every exception?

src/core/eval.tsx
```diff
@@ -384,6 +415,78 @@ export class EvalClient implements CoreEvalClient {
...
+              evaluationInput: { sessionSpans: trace.spans } as unknown as EvaluationInput,
+              evaluationTarget: target,
+              evaluationReferenceInputs: refsBySession.get(trace.sessionId),
+            }),
+          );
+          results.push(...(response.evaluationResults ?? []));
+        }
+      }
+    }
+    return { sessionsEvaluated: input.traces.length, results };
```
aidandaly24: Could `sessionsEvaluated` count sessions for which an `Evaluate` request was actually sent? TRACE and TOOL_CALL sessions with no IDs are skipped above but still included here. I reproduced `sessionsEvaluated: 1` with zero API calls and zero results. Tracking submitted sessions, or naming this `sessionsDiscovered`, would make the output less misleading.

src/handlers/eval/ondemand/evaluate/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+      const traces = await core.eval.getTracesForAgent(
+        { agent: flags["agent"], endpoint: flags["endpoint"], window, sessionIds, traceId },
+        opts,
+      );
+
+      // Ground truth is a typed SDK-shape passthrough (identical to batch's handler):
+      // resolve inline / file:// / -, then hand the array to core verbatim — core
+      // groups it by session.
+      const resolver = new SourceResolver({ stdin: io.stdin });
+      const groundTruth = parseJsonFlag<EvaluationReferenceInput[]>(
```
nborges-aws: Can we change this to use `parseJsonArrayFlag` helper instead? Since `groundTruth` has to be an array.

nborges-aws:
Agree with Aidans findings + one additional comment

src/core/eval.tsx
jariy17: I'll add support to detect these "gen_ai.operation.name === "execute_tool", openinference.span.kind === "TOOL", or traceloop.span.kind === "tool"

src/core/eval.tsx
jariy17: AWS SDK v3 responses have | undefined even if they are required. I'm just going to do ! because every evaluator must have a level.

src/core/eval.tsx
```diff
@@ -384,6 +415,98 @@ export class EvalClient implements CoreEvalClient {
...
+    const data = this.clients.data(toClientConfig(options));
+    const control = this.clients.control(toClientConfig(options));
+    const levels = await resolveEvaluatorLevels(input.evaluatorIds, control, options);
+    const refsBySession = groupRefsBySession(input.groundTruth);
+
+    const results: EvaluationResultContent[] = [];
+    // Sessions that actually produced Evaluate results — distinct from the sessions
+    // handed in, since a TRACE/TOOL_CALL session with no matching ids makes no call.
+    const evaluatedSessions = new Set<string>();
+    for (const evaluatorId of input.evaluatorIds) {
```
nborges-aws, line 481: nit: should we use `levels.keys()` here instead of the raw id array? Consistent with rest of code + leverages the dedup logic in `resolveEvaluatorLevels()`. A follow up item if you think its worth it

nborges-aws (approved):

aidandaly24 (approved):

src/core/eval.tsx
```diff
@@ -384,6 +415,98 @@ export class EvalClient implements CoreEvalClient {
...
+
+    const { runtimeId, runtimeName } = await resolveAgentToNameAndId(
+      input.agent,
+      this.clients,
+      options,
+    );
+    const logGroupName = runtimeLogGroup(runtimeId, qualifier);
+    const serviceName = runtimeServiceName(runtimeName, qualifier);
+
+    // CloudWatch Insights takes epoch seconds. Discovery defaults to now-7d when
```
Hweinstock, line 429: nit: i feel like the code explains this already.

src/core/eval.tsx
```diff
@@ -384,6 +415,98 @@ export class EvalClient implements CoreEvalClient {
...
+
+    const logs = this.clients.logs(toClientConfig(options));
+    const queryString = buildSpanQuery(serviceName, input.sessionIds, input.traceId);
+
+    // Runtime group required (missing = agent has no traces); aws/spans optional now
+    // https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/observability-configure.html#observability-configure-unified-traces
+    const [runtimeRows, sharedRows] = await Promise.all([
+      runInsightsQuery(logs, [logGroupName], queryString, startSec, endSec).catch((error) => {
+        if (error instanceof ResourceNotFoundException) {
+          throw new InputValidationError(
```
Hweinstock, line 444: should this be a different error type for telemetry? I wonder if it would be useful to distinguish invalid inputs from valid inputs without results.

src/core/eval.tsx
```diff
@@ -384,6 +415,98 @@ export class EvalClient implements CoreEvalClient {
...
+    // handed in, since a TRACE/TOOL_CALL session with no matching ids makes no call.
+    const evaluatedSessions = new Set<string>();
+    for (const evaluatorId of input.evaluatorIds) {
+      const level = levels.get(evaluatorId) ?? "SESSION";
+      for (const trace of input.traces) {
+        // TRACE/TOOL_CALL sessions with no ids at that level contribute no calls
+        // (empty batch list); SESSION always makes one call with no target.
+        for (const target of targetBatches(level, trace)) {
+          const response = await data.send(
+            new EvaluateCommand({
```
Hweinstock, line 488: q: if a single evaluate fails, do we want to fail the entire run?

src/core/eval.tsx
```diff
@@ -384,6 +415,98 @@ export class EvalClient implements CoreEvalClient {
...
+          );
+          const evaluationResults = response.evaluationResults ?? [];
+          if (evaluationResults.length > 0) evaluatedSessions.add(trace.sessionId);
+          results.push(...evaluationResults);
+        }
+      }
+    }
+    return {
+      sessionsRequested: input.traces.length,
+      sessionsEvaluated: evaluatedSessions.size,
```
Hweinstock, line 505: q: I see we return the sessions evaluated and the results separately. Is there a use case for getting the results for a certain session or is it more useful in aggregate?

src/core/eval.tsx
```diff
@@ -1261,6 +1384,221 @@ async function agentDataSource(
   };
 }
 
+// sanitizeQueryValue strips single quotes so an id can't break out of the quoted
+// Insights filter literal it is interpolated into (matches the old CLI).
+function sanitizeQueryValue(value: string): string {
+  return value.replace(/'/g, "");
+}
+
+// buildSpanQuery is the single-phase Insights query: scope to one runtime by its
```
Hweinstock, line 1393: nit: is there info in this comment not expressed by the code?

src/core/eval.tsx
```diff
@@ -1261,6 +1384,221 @@ async function agentDataSource(
...
+  const queryId = started.queryId;
+
+  // Phase 1: wait for completion. A large scan can take minutes, so the deadline is
+  // generous; each poll costs one cheap GetQueryResults call.
+  let status = "Running";
+  for (let i = 0; i < 300 && status !== "Complete"; i++) {
+    const result = await logs.send(new GetQueryResultsCommand({ queryId }));
+    status = result.status ?? "Unknown";
+    if (status === "Failed" || status === "Cancelled" || status === "Timeout") {
+      throw new NetworkingError(`CloudWatch Logs Insights query ${status.toLowerCase()}`, {
```
Hweinstock, line 1436: is networking the right error here? Feels more like a timeout/cancellation depending on status.

src/core/eval.tsx
```diff
@@ -1261,6 +1384,221 @@ async function agentDataSource(
...
+  }
+  query += `\n     | sort @timestamp asc\n     | limit ${INSIGHTS_MAX_ROWS}`;
+  return query;
+}
+
+// runInsightsQuery starts a CloudWatch Logs Insights query, waits for it to finish,
+// then drains all result pages. GetQueryResults returns <=10k rows per call, so a
+// long session's spans span multiple pages (nextToken); dropping any would score a
+// partial conversation. Fails fast if the ceiling is hit — that belongs in batch.
+async function runInsightsQuery(
```
Hweinstock, line 1417: oos here, but I wonder if a generic `waitFor`/`pollUntil` could be useful. Some extension of https://github.com/aws/agentcore-cli/blob/refactor/src/testing/timing.tsx

src/core/eval.tsx
```diff
@@ -1261,6 +1384,221 @@ async function agentDataSource(
...
+    const message = row.find((f) => f.field === "@message")?.value;
+    const sessionId = row.find((f) => f.field === "sessionId")?.value;
+    // Drop orphan records with no session id (e.g. system logs) — an
+    // unidentifiable session can't be evaluated.
+    if (!message || !sessionId) continue;
+    let doc: SpanRecord;
+    try {
+      doc = JSON.parse(message) as SpanRecord;
+    } catch {
+      continue;
```
Hweinstock, line 1482: is it worth logging a warning here or would this be noisy?

src/handlers/eval/types.tsx
```diff
@@ -203,6 +205,54 @@ export type StartBatchEvaluationInput = {
...
+// EvaluateInput is the CLI-facing shape for the synchronous Evaluate path.
+// `groundTruth` is the SDK-native array, passed verbatim; EvalClient groups it by
+// session (context.spanContext.sessionId) and attaches per session.
+export type EvaluateInput = {
+  traces: SessionTrace[];
+  evaluatorIds: string[];
+  groundTruth?: EvaluationReferenceInput[];
+};
+
+// EvaluateResult returns the raw Evaluate API results across all evaluators and
```
Hweinstock, line 244: nit: these comments describe usages of the type and feel like they could drift.

src/handlers/eval/ondemand/evaluate/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+      ctx.require(JsonRendererKey).renderJson(result);
+    },
+  });
+
+// resolveWindow validates on-demand's time filter: --lookback-days maps to
+// [now - n days, now]; the explicit --start-time/--end-time pair must come together
+// with start before end. On-demand owns this rather than reusing batch's resolver:
+// batch has no --lookback-days and its window feeds a service-side data source, not
+// a client-side Insights query.
+function resolveWindow(flags: {
```
Hweinstock, line 99: how does this differ from https://github.com/aws/agentcore-cli/blob/ea3fa84922d82870a2bc7daadf863dc0a88c7763/src/handlers/eval/batch-evaluation/evaluate/index.tsx#L164?

src/handlers/eval/ondemand/ondemand.fixture.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+  testIO,
+} from "../../../testing";
+import { createRootHandler } from "../../index";
+
+const REGION = "us-west-2";
+const FIXTURES = join(import.meta.dir, "__fixtures__");
+
+// Record with: RECORD=1 bun test src/handlers/eval/ondemand/ondemand.fixture.test.tsx
+//
+// This exercises the real seam end to end: parsing → handler → CoreClient →
```
Hweinstock, line 18: nit: this feels overly verbose.

src/handlers/eval/ondemand/ondemand.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+        "evaluate",
+        "--evaluator",
+        "Builtin.Helpfulness",
+        "--session-ids",
+        "s1",
+      ]),
+    ).rejects.toThrow(/--agent/);
+  });
+
+  test("requires --evaluator", async () => {
```
Hweinstock, line 86: would these make sense as a test.each pattern for the different set of flags that reject?

src/core/eval.tsx
jariy17: I was thinking we create a new Exception called TracesNotFound once we add observability to the cli.

src/core/eval.tsx
jariy17: On demand evaluate should be use for small evaluations (1-2 sessions with 2-3 evaluators) so this situation is unlikely to happen and also you could multiple evaluate calls per session if one of those fails, you have an incomplete evaluations

src/core/eval.tsx
jariy17: Removed in follow-up PR #2007: https://github.com/aws/agentcore-cli/pull/2007

src/core/eval.tsx
jariy17: Addressed in follow-up PR #2007: missing runtime telemetry now throws ResourceNotFoundError instead of InputValidationError, while preserving the CloudWatch exception as its cause. https://github.com/aws/agentcore-cli/pull/2007

src/core/eval.tsx
jariy17: Yes, fail-fast is intentional for now. One session can require multiple Evaluate API calls across evaluators and target batches. Continuing after one fails could return incomplete results for that session while appearing successful. On-demand evaluation targets small synchronous runs; larger workloads should use batch evaluation. We can add explicit partial-failure handling later if customers need it.

src/core/eval.tsx
jariy17: Removed in follow-up PR #2007: https://github.com/aws/agentcore-cli/pull/2007

src/core/eval.tsx
jariy17: Results are already associated with sessions through context.spanContext.sessionId. Keeping them flat preserves the SDK response shape while still supporting per-session filtering and aggregate analysis. I also confirmed the session ID is present in a live on-demand evaluation response.

src/core/eval.tsx
jariy17: Agreed this could be useful, but I will keep it out of scope and extract it when another production caller needs the same polling behavior.

src/core/eval.tsx
jariy17: Addressed in follow-up PR #2007 by passing the existing logger into the grouping helper and warning at most once when malformed telemetry records are skipped. https://github.com/aws/agentcore-cli/pull/2007

src/core/eval.tsx
jariy17: Addressed in follow-up PR #2007 with a CloudWatchQueryError that carries caller-defined messages plus query status and ID metadata. https://github.com/aws/agentcore-cli/pull/2007

src/handlers/eval/ondemand/evaluate/index.tsx
jariy17: Batch's resolveWindow feeds the batch evaluation API's DataSourceConfig shape, which supports CloudWatch log groups and onlineEvaluationConfigArn.

src/handlers/eval/types.tsx
jariy17: Removed the usage-oriented type comments in follow-up PR #2007: https://github.com/aws/agentcore-cli/pull/2007

src/handlers/eval/ondemand/ondemand.fixture.test.tsx
jariy17: Removed the verbose fixture comment block in follow-up PR #2007: https://github.com/aws/agentcore-cli/pull/2007

src/handlers/eval/ondemand/ondemand.test.tsx
jariy17: Converted the repeated validation cases to test.each in follow-up PR #2007: https://github.com/aws/agentcore-cli/pull/2007

PR #1986, opened 2026-08-12
feat(errors): add shared user cancellation error

aidandaly24:
### Problem

Commander exits and user cancellations bypass the shared CLI error model. Runtime and Gateway invoke define resource-specific interruption errors, while Project dev defines a separate command interruption type for the same user cancellation outcome.

### Solution

- classify `CommanderError` in `AgentCoreCLIError.fromError`, preserving help as exit `0` and mapping parse failures to usage exit `2`
- add an opt-in `SilentCLIError` category so only intentionally silent errors skip generic root stderr output
- add a silent, user-sourced `UserCancellationError` with exit code `130`
- centralize process SIGINT listener lifecycle in `withUserCancellation` for Runtime invoke, Gateway invoke, dataset get, and dataset update
- use the shared cancellation error as each operation's `AbortSignal.reason`, including Project dev
- preserve Project dev's SIGINT/SIGTERM handling, single `Shutting down…` message, repeated-signal behavior, and listener cleanup
- preserve Runtime and Gateway partial-response interruption summaries while propagating the original typed cancellation reason
- leave TUI-local cancellation and low-level platform `AbortError` handling unchanged

### Verification

- focused cancellation suites across errors, root handling, Runtime, Gateway, Project dev, and datasets (`146 pass, 0 fail`)
- full local source suite (`1544 pass, 0 fail`)
- `bun run typecheck`
- `bun run lint:check`
- Prettier check for every changed file
- `bun run build`
- `git diff --check`
- GitHub full unit suites:
  - Linux: `1544 pass, 0 fail`
  - Windows: `1544 pass, 0 fail`
  - macOS: `1544 pass, 0 fail`
- Linux, Windows, and macOS builds
- AgentCore E2E CodeBuild
- bundled CLI smoke checks:
  - nested help: exit `0`, help on stdout, no stderr or error log
  - nested unknown option: exit `2`, one Commander error line, classified as a user error

### DIFF
--- a/src/errors/errors.tsx
+++ b/src/errors/errors.tsx
@@ -12,8 +12,6 @@ export interface AgentCoreCLIErrorOptions extends ErrorOptions {
   exitCode?: number;
   /** Describes the name of the underlying error, defaults to AgentCoreCLIError */
   name?: string;
-  /** Whether the root handler should display this error's message */
-  displayMessage?: boolean;
 }
 
 /** Base error for CLI failures, including their source, metadata, and process exit code. */
@@ -21,7 +19,6 @@ export class AgentCoreCLIError extends Error {
   readonly source: ErrorSource;
   readonly meta: Record<string, unknown>;
   readonly exitCode: number;
-  readonly displayMessage: boolean;
 
   constructor(message?: string, options?: AgentCoreCLIErrorOptions) {
     super(message, options);
@@ -29,7 +26,6 @@ export class AgentCoreCLIError extends Error {
     this.source = options?.source ?? ERROR_SOURCE.INTERNAL;
     this.meta = options?.meta ?? {};
     this.exitCode = options?.exitCode ?? 1;
-    this.displayMessage = options?.displayMessage ?? true;
   }
   /** Convert the error into an object with its attributes enumerated as keys **/
   json(): Record<string, unknown> {
@@ -38,7 +34,6 @@ export class AgentCoreCLIError extends Error {
       message: this.message,
       stack: this.stack,
       exitCode: this.exitCode,
-      displayMessage: this.displayMessage,
       meta: this.meta,
       source: this.source,
     };
@@ -48,13 +43,12 @@ export class AgentCoreCLIError extends Error {
     if (error instanceof AgentCoreCLIError) return error;
 
     if (error instanceof CommanderError) {
-      return new AgentCoreCLIError(error.message, {
+      return new SilentCLIError(error.message, {
         cause: error,
         source: ERROR_SOURCE.USER,
         name: error.name,
         meta: { code: error.code },
         exitCode: error.exitCode === 0 ? 0 : 2,
-        displayMessage: false,
       });
     }
 
@@ -78,6 +72,9 @@ export class AgentCoreCLIError extends Error {
     return new AgentCoreCLIError(String(error), { cause: error });
   }
 }
+
+/** Base for CLI errors intentionally omitted from root stderr output. */
+export class SilentCLIError extends AgentCoreCLIError {}
 
 /** Error raised for invalid user input. */
 export class InputValidationError extends AgentCoreCLIError {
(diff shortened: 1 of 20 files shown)


src/errors/errors.tsx
```diff
@@ -135,19 +149,17 @@ export class EmbeddedAssetNotFoundError extends AgentCoreCLIError {
...
+      source: ERROR_SOURCE.USER,
+      exitCode: 130,
+    });
   }
 }
 
-export class RuntimeInvokeResponseError extends AgentCoreCLIError {
-  readonly reported = true;
-
+export class RuntimeInvokeResponseError extends SilentCLIError {
```
Hweinstock, line 169: why are runtime invoke responses silent? I thought this was the error we get when the stream parsing fails.

src/runnable/index.tsx
```diff
@@ -48,20 +32,8 @@ export async function runWithExitCode(
...
-      const name = reported instanceof AgentCoreCLIError ? "Error" : reported.name;
-      console.error(`${name}: ${reported.message}`);
-    }
-    if (error instanceof CommanderError) {
-      return error.exitCode === 0 ? ExitCode.SUCCESS : ExitCode.USAGE;
-    }
-    if ((error as Error)?.name === "AbortError") return ExitCode.INTERRUPTED;
-    if (caught instanceof AgentCoreCLIError) return caught.exitCode;
-    return ExitCode.FAILURE;
+    const error = AgentCoreCLIError.fromError(caught);
```
Hweinstock, line 53: nice, really like how simple this is now!

src/index.ts
```diff
@@ -91,12 +91,14 @@ process.exit(
       await rootHandler.route(argv, context);
     } catch (e) {
       const error = AgentCoreCLIError.fromError(e);
-      rootLogger.child({ error: error.json() }).error();
-      commandRunMetricEvent.setAttributes({
-        exit_reason: "failure",
-        error_name: error.name,
-        error_source: error.source,
-      });
+      if (error.exitCode !== 0) {
```
Hweinstock, line 94: i wonder if it makes sense to expand the `exit_reason` attribute to accept a `cancelled` value. That way we still get telemetry for these cancellations. 

could be a follow-up since we'll need to adjust the backend schema to accommodate.

src/handlers/eval/dataset/get/index.tsx
Hweinstock: do we need the same `controller.signal.throwIfAborted();` check here?

Also wondering if it makes sense to build an abstraction for this since it seems there's already a few consumers?

Something like:

```
 async function withCancellation<T>(fn: (signal: AbortSignal) => Promise<T>): Promise<T> {
    const controller = new AbortController();
    const interrupt = () => controller.abort(new UserCancellationError());
    process.once("SIGINT", interrupt);
    try {
      return await fn(controller.signal);
    } catch (error) {
      throw error;
    } finally {
      controller.abort();
      process.off("SIGINT", interrupt);
    }
  }
```

Hweinstock (commented):
i like the approach! few comments, but I think some could be follow-ups I could help pick up.

src/errors/errors.tsx
aidandaly24: Yeah, this is the error we get when response streaming fails. By the time it reaches the root, `writeStreamingResponse` has already written the sanitized incomplete response summary to `stderr`. Making this error silent just prevents a second generic `Error: response stream failed` line. It still goes through structured logging and telemetry.

src/index.ts
aidandaly24: Yeah, I think cancelled would be clearer. Right now cancellation still emits telemetry as failure with `error_source: user`. I kept the new exit reason out of this PR since it also requires a backend schema change, so I think that should be a follow-up.

src/handlers/eval/dataset/get/index.tsx
aidandaly24: Agreed. After rebasing, Gateway invoke and dataset update introduced two more consumers, so the abstraction makes sense now. I added `withUserCancellation` under `src/runnable` and moved Runtime invoke, Gateway invoke, dataset get, and dataset update onto it. It owns the SIGINT listener, cleanup, and `throwIfAborted()` normalization, while TUI-local cancellation stays separate. I'll do the same with dev when its merged.

src/index.ts
Hweinstock: agreed. Once we start getting data we can decide if separating into a separate reason makes sense.

Hweinstock (dismissed):
awesome!

nborges-aws (approved):

Hweinstock (approved):

PR #1992, opened 2026-08-13
fix(schemas): lift schemas to avoid circular dep

Hweinstock:
### Problem 
The project schemas break the convention of dependency inversion between the handlers and core. Handlers define interfaces, while core implements the interfaces, causing core to depend on handlers. However, the schemas reverse that pattern by causing the handlers to also depend on core. 

The challenge is that core needs these schemas for reading/writing to the agentcore.json, and handlers needs these schemas to derive the interfaces for the Project (that core then implements). 

### Solution 
- lift the schemas into a top level module to avoid the circular import. 
- alternatively considered decoupling the interfaces for working with the project from the schemas, but this provides no clear benefit and significant duplication. The schemas themselves also naturally describe the interface since the config is ultimately user facing. 
- Remove renaming in the top-level export, and instead rename at the source to simplify. This effects AgentEnvSpec --> ProjectRuntimeSpec, and AgentCoreProjectRuntime --> ProjectRuntime. 

### Testing / Verification
This is a pure moving/renaming change so relying on CI checks for validation.


tejaskash (approved):
Great find and clean fix!

jariy17 (approved):

PR #1996, opened 2026-08-13
fix(exec): derive region from the ARN so --runtime needs no project

### LINKED ISSUE #1995: exec: --runtime <arn> requires --region, and otherwise demands a project + aws-targets.json + a deploy
### What happened

`agentcore exec --runtime <full-arn>` fails outside an AgentCore project unless `--region` is also passed:

```
$ agentcore exec --it --runtime arn:aws:bedrock-agentcore:us-east-1:111122223333:runtime/MyRuntime-abc123 --session-id <id>
AWS Targets config file not found at: /path/to/cwd/agentcore/aws-targets.json
```

Adding `--region us-east-1` to the identical command in the identical directory clears the error entirely, so the region is the only thing missing — and it is already field 3 of the ARN that was just supplied.

Working around it by creating the config produces a chain of further walls, none of which `exec` actually needs:

1. `AWS Targets config file not found at: .../agentcore/aws-targets.json`
2. after creating that → `State config file not found at: .../agentcore/.cli/deployed-state.json`
3. with an empty state file → `No deployed targets found. Run 'agentcore deploy' first.`

Net effect: to `exec` against a runtime you already have the ARN for, the CLI demands you create a project, hand-write `aws-targets.json`, and complete a deploy through this CLI. That rules out `exec` for anyone whose runtimes are deployed by CDK, a pipeline, or a personal stack — i.e. the CLI cannot be adopted for `exec` alone.

### Root cause

`loadExecContext` in `src/cli/commands/exec/action.ts`. The "explicit ARN, skip deployed state" short-circuit is gated on **both** flags:

```ts
if (options.runtimeArn?.startsWith('arn:') && options.region) {
  return assertInteractiveHarnessUnsupported(options, { region: options.region, runtimeArn: options.runtimeArn });
}
```

Without `--region` it falls through to `readAWSDeploymentTargets()` / `readDeployedState()`, which throw before anything else runs. The `--harness <arn>` short-circuit a few lines below has the identical gate.

There is already a branch labelled "`--runtime <arn>` with no `--region`" further down, but it sits *after* the config reads and only uses config for `options.region ?? targetConfig.region` — a value the ARN already carries. So the config reads exist purely to recover something the caller supplied.

`exec --help` gives no hint either; it lists `--region <region>  AWS region` as though optional.

### Expected

A full ARN should resolve with no project and no config on disk. Only a *name* should need deployed state.

### Fix

Derive the region from ARN field 3 and drop the `&& options.region` conjunct from both short-circuits, so config is read only when a name has to be resolved. A region-less or malformed ARN should still fall through to config as a last resort.

PR: #1996

notgitika:
## Description

`agentcore exec --runtime <full-arn>` fails outside an AgentCore project unless `--region` is also passed:

```
$ agentcore exec --it --runtime arn:aws:bedrock-agentcore:us-east-1:111122223333:runtime/MyRuntime-abc123 --session-id <id>
AWS Targets config file not found at: /path/to/cwd/agentcore/aws-targets.json
```

`loadExecContext` gated its "explicit ARN, skip deployed state" short-circuit on **both** flags (`options.runtimeArn?.startsWith('arn:') && options.region`), so an ARN alone fell through to `readAWSDeploymentTargets()` / `readDeployedState()`, which throw first. The branch further down labelled "`--runtime <arn>` with no `--region`" sat *after* those reads and used config only for `options.region ?? targetConfig.region` — a value already present as field 3 of the ARN that was just supplied. So the CLI demanded a project, a hand-written `aws-targets.json`, and a completed deploy purely to recover something the caller had handed it, which rules out `exec` for anyone whose runtimes are deployed by CDK, a pipeline, or a personal stack.

This derives the region from the ARN so `--region` is optional, and the short-circuit fires on the ARN alone. Config is now read only when a *name* has to be resolved. A region-less or malformed ARN still falls through to config as a last resort, so nothing that worked before stops working. `--harness <arn>` had the identical gate and gets the identical treatment.

`regionFromArn` already existed in `src/cli/operations/jobs/shared/region.ts`, so it is reused rather than rewritten. It moves to a new `src/cli/aws/arn.ts` and is re-exported from its old location to keep the jobs-facing import path stable. Two test files factory-mock the whole `aws/region` module, so it could not live there — and the split is cleaner regardless: `aws/region.ts` detects the ambient region from env and shared config files, while this is a pure string function over an ARN the caller already holds. Splitting on `:` rather than matching a partition keeps it correct for GovCloud and China ARNs.

## Related Issue

Closes #1995

## Documentation PR

N/A

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] I ran `npm run lint`

Unit: 423/423 files, 6095/6095 tests. Typecheck, lint, prettier and secretlint clean. `test:integ` did not run to completion — I stopped it rather than block this PR, so it never printed a summary. Every integ test that did run passed and no failures appeared in the output, but I am not claiming a green integ run. No integ test touches `exec`, `cli/aws`, or the jobs region path; please let CI be the authority there.

Also verified end-to-end against a built CLI rather than by unit test alone, in an empty directory with no project:

- installed 0.25.0 → `AWS Targets config file not found at: /private/tmp/exec-fix-check/agentcore/aws-targets.json`
- this branch → reaches the real `InvokeAgentRuntime` call in `us-east-1` and fails only on cross-account authorization for a runtime in an account I have no access to

That is the correct terminal state: region resolved from the ARN, config wall gone, no project and no `aws-targets.json` on disk.

New unit tests cover region-from-ARN with config readers mocked to reject and asserted `not.toHaveBeenCalled()` (standing in for "no project at all"), an explicit `--region` overriding the ARN, `aws-us-gov` and `aws-cn` partitions, and the malformed-ARN fall-back to config.

## Checklist

---

terms of your choice.

### DIFF
--- a/src/cli/aws/arn.ts
+++ b/src/cli/aws/arn.ts
@@ -1,0 +1,17 @@
+/**
+ * ARN parsing helpers.
+ *
+ * Kept separate from region.ts: that module *detects* the ambient region from the environment and
+ * shared config files, while these are pure string functions over an ARN the caller already holds.
+ */
+
+/**
+ * Parse the region out of a service ARN.
+ * ARN format: arn:{partition}:{service}:{region}:{account}:{resource} → field index 3 is the region.
+ * Splitting on ':' rather than matching a partition keeps this correct for GovCloud and China ARNs.
+ * Returns undefined for a malformed or region-less ARN so callers can fall back.
+ */
+export function regionFromArn(arn: string): string | undefined {
+  const region = arn.split(':')[3];
+  return region && region.length > 0 ? region : undefined;
+}
--- a/src/cli/commands/exec/__tests__/action.test.ts
+++ b/src/cli/commands/exec/__tests__/action.test.ts
@@ -767,13 +767,55 @@ describe('loadExecContext --runtime as ARN or name', () => {
     ).not.toHaveBeenCalled();
   });
 
-  it('resolves region from config when --runtime is a full ARN but --region is omitted', async () => {
+  // A full ARN already carries its region, so exec must not need a project to recover it. Reading
+  // config here used to fail outright with "AWS Targets config file not found" for anyone deploying
+  // runtimes outside this CLI (CDK, pipelines, personal stacks) — the ARN was enough all along.
+  it('takes the region from the ARN when --region is omitted, without reading config', async () => {
+    // Config reads throw, standing in for "no project / no aws-targets.json in cwd".
+    const noProject = {
+      readAWSDeploymentTargets: vi.fn().mockRejectedValue(new Error('AWS Targets config file not found')),
+      readDeployedState: vi.fn().mockRejectedValue(new Error('State config file not found')),
+    } as unknown as ConfigIO;
+
+    const ctx = await loadExecContext({ runtimeArn: 'arn:aws:bedrock-agentcore:eu-west-2:123:runtime/X' }, noProject);
+
+    expect(ctx.region).toBe('eu-west-2'); // from the ARN, not config
+    expect(ctx.runtimeArn).toBe('arn:aws:bedrock-agentcore:eu-west-2:123:runtime/X');
+    expect(
+      (noProject as unknown as { readAWSDeploymentTargets: ReturnType<typeof vi.fn> }).readAWSDeploymentTargets
+    ).not.toHaveBeenCalled();
+    expect(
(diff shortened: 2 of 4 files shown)


nborges-aws (approved):
LGTM, thanks for fixing!

PR #1998, opened 2026-08-13
feat(project): scaffold add harness handler

Hweinstock:
### Problem 
We're missing functionality to add resources to a project. This PR sets up the handler for `agentcore project add harness` as an example, and scaffolds the remaining work for add. 

### Solution 
- setup a handler that takes the same flags at the imperative harness create (with a extra few project specific ones). Alternatively considered using different flags from imperative, but this creates an inconsistent UX where project command inputs are similar to api shapes, but slightly different (ex. containerUri is top level, vs nested in json). 
- convert those flags into the spec format expected by the schemas. 
- scaffold an add function for the shared functionality. 

### Testing / Verification 
- created a project, attempted to add a harness and got the not implemented error. 
- add unit tests for the flag parsing. 

## Next Steps
- I want to move the scaffolding of runtime into the add function, then call that from the handler when we're scaffolding. This will make it easier to re-use the logic, and handle harness in the same place. 
- implement the rest of add via json source.


src/core/project/manager.tsx
```diff
@@ -123,6 +130,15 @@ export class FsProjectManager implements ProjectManager {
     return project;
   }
 
+  // eslint-disable-next-line require-yield
+  public async *add<TResource extends ProjectResource>(
```
Hweinstock: i'm imagining minimal special casing here. We'll want to scaffold files for runtime/harness, then the rest should be simple shared logic to write the schema changes. We will also need some special casing for harness since it links to the config in the scaffolded files.

src/handlers/project/add/harness/index.ts
```diff
@@ -0,0 +1,400 @@
...
+      for await (const event of config.projectManager.add(project, "harness", harnessConfig)) {
+        config.io.stderr.write(`${event.message}\n`);
+      }
+
+      config.io.stderr.write(`added harness '${flags["name"]}' to '${project.name}'`);
+    },
+  });
+
+/** Converts the SDK's tagged-union model config into the flat project-schema shape. */
+function toModelConfig(modelConfig: HarnessModelConfiguration): HarnessModel {
```
Hweinstock, line 162: we handle all the api shape --> project schema conversion in the handler to keep the shared logic clean.

jariy17 (approved):

src/handlers/project/project.test.ts
```diff
@@ -98,21 +108,261 @@ describe("project create", () => {
...
+    ["tags", ["--name", "x", "--tags", '{"team":"ml"}']],
+    ["allowed-tools", ["--name", "x", "--allowed-tools", "*", "@builtin"]],
+    [
+      "max-iterations, max-tokens, timeout-seconds",
+      ["--name", "x", "--max-iterations", "10", "--max-tokens", "4096", "--timeout-seconds", "60"],
+    ],
+  ])("%s", async (_label, flags) => {
+    await inProject();
+    // TODO: update to verify that the project updates.
+    await expect(run(["add", "harness", ...flags])).rejects.toThrow("not yet implemented");
```
jariy17, line 306: TODO: MAke sure you fix tests after implementation

jariy17 (approved):

src/core/project/manager.tsx
```diff
@@ -123,6 +130,15 @@ export class FsProjectManager implements ProjectManager {
     return project;
   }
 
+  // eslint-disable-next-line require-yield
+  public async *addResource<TResource extends ProjectResource>(
```
notgitika, line 134: nice!

tejaskash (approved):

src/handlers/project/add/harness/index.ts
```diff
@@ -0,0 +1,449 @@
...
+  if ("agentCoreGateway" in c && c.agentCoreGateway) {
+    return {
+      type: tool.type,
+      name: tool.name,
+      config: {
+        agentCoreGateway: {
+          gatewayArn: requireField(c.agentCoreGateway.gatewayArn, "agentCoreGateway.gatewayArn"),
+        },
+      },
+    };
```
notgitika, line 238: what about when users set outbound auth or JWT here? are these being dropped or are they somewhere where I'm missing them. I am looking at `toAuthorizationConfig` below and I see those fields.

src/handlers/project/project.test.ts
```diff
@@ -98,21 +108,261 @@ describe("project create", () => {
...
+    ],
+  ])("%s", async (_label, flags) => {
+    await inProject();
+    await expect(run(["add", "harness", ...flags])).rejects.toBeInstanceOf(InputValidationError);
+  });
+});
+
+describe("project build", () => {
+  async function inBuildableProject(): Promise<string> {
+    const projectRoot = await inProject("MyAgent");
```
notgitika, line 358: nice!

notgitika (approved):

src/handlers/project/add/harness/index.ts
Hweinstock: good catch, looks like we want to wire up the field here https://github.com/aws/agentcore-cli/blob/a9d34be3ce4af79ed14b49a2f1b6a2d56d654af5/src/projectSchemas/harness.ts#L195 from https://docs.aws.amazon.com/bedrock-agentcore-control/latest/APIReference/API_HarnessAgentCoreGatewayConfig.html. I'll track this as a follow-up.

notgitika: thanks!

PR #2001, opened 2026-08-14
feat: implement project deploy command

notgitika:
This PR adds the project deploy command. it resolves the deployment target, synthesizes, bootstraps that target's environment, then deploys its stack. Synth reuses build's code path, so build's dependency check runs before anything touches AWS.

stack environments come from `agentcore/aws-targets.json`, which create() scaffolds empty. An empty list is an error naming the file to fill in, rather than an account resolved from the active credentials, deploy shouldn't guess where the user's infra belongs. It's checked before synth, so an empty list fails without running anything.

one deploy ships one target. `--target` selects it and defaults to default, so a project with a staging and a prod target can't reach prod by accident. The target is resolved before synth, so a misspelled --target costs no build and the error lists the configured names. which stack belongs to the target comes from the synthesized manifest, matched on the `agentcore:target-name` tag the generated CDK app already writes, so the CLI never has to reproduce that app's stack-naming convention.

bootstrap runs only where it's needed: a deploy reads `CDKToolkit`'s `BootstrapVersion` first and bootstraps only when the stack is missing, unreadable, or older than version 30, the same minimum the published CLI requires. It still bootstraps with `createCustomerMasterKey: true` to match that CLI, so a first deploy provisions a KMS key for the staging bucket. probing first is what stops a later deploy from rewriting a `CDKToolkit` somebody else bootstrapped without one.

Not in this PR:

This is the synth -> bootstrap -> deploy spine. Everything else the existing deploy does is deliberately left out for now:

- No credential preflight. Nothing checks that credentials exist or that their account matches the target's, so a wrong-account credential surfaces as a CDK error partway through rather than an upfront message naming both accounts. Follow-up.
- An empty aws-targets.json is an error, not auto-populated. Filling it in from the active credentials is a follow-up.
- Nothing is written to deployed-state.json. The existing CLI provisions identity, OAuth, and payment credential providers before synth and records their ARNs there, and the generated CDK app reads them back to build its stacks. Because this PR writes none of it, a project declaring payments fails during synth asking for credential providers that nothing creates; projects needing no pre-deploy identity are unaffected. Follow-up.
- No post-deploy steps. No knowledge-base ingestion, dataset upload, observability setup, or online-eval config, and no deployment state recorded for later commands to read. Stack outputs still reach the user: the deploy prints them under its result line. Follow-up.
- No --diff. Follow-up. No --dry-run either, by design: build already synthesizes without touching AWS, which is what a dry run is here.
- One target per deploy. No `--all` or multi-target deploy. Follow-up.
- Only deploy writes to the shared log and prints where it is. build and dev don't yet. Follow-up.


src/io/cdk.ts
```diff
@@ -0,0 +1,118 @@
...
+import {
+  BaseCredentials,
+  BootstrapEnvironments,
+  BootstrapStackParameters,
+  StackSelectionStrategy,
+  Toolkit,
+  type IIoHost,
+  type IoMessageLevel,
+} from "@aws-cdk/toolkit-lib";
+
```
notgitika: this static import is increasing the binary size and startup latency by a lot. looking into solutions

notgitika:
increasing test coverage

src/handlers/project/deploy/index.ts
```diff
@@ -1,11 +1,44 @@
...
+        "skip-bootstrap",
+        "skip bootstrapping the target environment before deploying",
+        z.boolean().default(false),
+      ),
+    ],
+    handle: async (ctx, flags) => {
+      // withProject has already resolved the enclosing project.
+      const project = ctx.require(ProjectKey);
+
+      // Progress and the CDK toolkit's own output both go to stderr, keeping stdout
```
Hweinstock: does it make sense to avoid cdk specific language since we expect a terraform extension soon?

src/handlers/project/deploy/index.ts
```diff
@@ -1,11 +1,44 @@
...
-    handle: async () => {
-      throw new NotImplementedError("agentcore project deploy is not implemented yet");
+    flags: [
+      flag(
+        "target",
+        "name of the aws-targets.json entry to deploy to",
+        z.string().default("default"),
+      ),
+      flag(
+        "skip-bootstrap",
```
Hweinstock: out of curiosity, when would a customer want to skip bootstrap? My understanding is that if an account/region is already bootstrapped it no-ops fairly quickly.

src/handlers/project/types.ts
```diff
@@ -20,9 +20,26 @@ export type CreateProjectInput = {
   skipGit?: boolean;
 };
 
-/** A progress step reported while a long-running project operation runs. */
-export type ProjectEvent = {
-  message: string;
+/**
+ * Something worth showing the user while a long-running project operation runs.
+ *
+ * A union rather than two optional fields, so an event is always exactly one of the
```
Hweinstock: does this comment explain anything that isn't clear from the code?

src/handlers/project/types.ts
```diff
@@ -20,9 +20,26 @@ export type CreateProjectInput = {
...
+ * two and a consumer that only writes text can read `message` without checking
+ * which it got.
+ */
+export type ProjectEvent =
+  /** A progress step, in this CLI's own wording. */
+  | { kind: "step"; message: string }
+  /** A line of output, forwarded as the tool that produced it phrased it. */
+  | { kind: "output"; message: string };
+
+export type DeployProjectOptions = {
```
Hweinstock: since all of these options are marked as required does `input` make more sense? I think "options" usually implies "optional".

src/core/project/manager.tsx
```diff
@@ -158,8 +171,137 @@ export class FsProjectManager implements ProjectManager {
     // The generated package.json defines `cdk` as "npm run build && cdk", so this
```
Hweinstock: is this comment overly verbose? It feels a little distracting to me.

src/core/project/manager.tsx
```diff
@@ -158,8 +171,137 @@ export class FsProjectManager implements ProjectManager {
...
+      }
+    }
+  }
+
+  // Synthesizes, bootstraps the target's environment, then deploys its stack.
+  private async *deployWithCdk(
+    project: Project,
+    options: DeployProjectOptions,
+  ): AsyncGenerator<ProjectEvent, void> {
+    // Every stack's environment comes from aws-targets.json, so an empty list means
```
Hweinstock: this comment feels pretty confusing to me. I think its saying aws-targets is the source of truth, but its very verbose.

src/handlers/project/types.ts
```diff
@@ -20,9 +20,26 @@ export type CreateProjectInput = {
...
+  /** The resolved AWS region, forwarded to the CDK subprocesses. */
+  region: string;
+  /** Skip bootstrapping the target environment before deploying. */
+  skipBootstrap: boolean;
+  /** Name of the aws-targets.json entry to deploy. */
+  target: string;
```
Hweinstock, line 44: is there a use case for deploying multiple targets?

src/core/project/manager.tsx
```diff
@@ -158,8 +171,137 @@ export class FsProjectManager implements ProjectManager {
...
+      yield* this.streamCdk({ kind: "bootstrap", environments: [environment] }, run);
+    }
+
+    yield { kind: "step", message: `Deploying ${stackName}` };
+    yield* this.streamCdk({ kind: "deploy", stackName }, run);
+  }
+
+  // Drives a CDK operation, surfacing the toolkit's messages as project events and
+  // logging them. Anything below info stays in the debug log rather than on screen.
+  private async *streamCdk(
```
Hweinstock: would this make sense on the cdk runner directly?

src/core/project/manager.tsx
```diff
@@ -158,8 +171,137 @@ export class FsProjectManager implements ProjectManager {
...
+  }
+
+  // Reads aws-targets.json, treating a missing file as empty and surfacing a parse
+  // failure as an actionable error that names the file rather than a Zod dump.
+  private async readDeploymentTargets(projectRoot: string): Promise<AwsTarget[]> {
+    const path = this.targetsPath(projectRoot);
+    if (!existsSync(path)) return [];
+    try {
+      return await this.json.read(path, AwsTargetsSchema);
+    } catch (cause) {
```
Hweinstock, line 298: Would it be simpler to improve the error message on `json.read` instead of rewrapping here?

src/core/project/assembly.ts
```diff
@@ -0,0 +1,62 @@
...
+});
+
+/**
+ * The name of the stack in the synthesized assembly that belongs to `target`.
+ *
+ * The generated CDK app synthesizes one stack per deployment target and tags each
+ * with the target's name, so deploy asks the assembly which stack to ship rather
+ * than deriving the name itself and hoping the two agree.
+ */
+export async function stackForTarget(
```
Hweinstock, line 30: shouldn't the synth output the stack name? Is there a reason we need to re-derive it here?

src/core/project/manager.tsx
```diff
@@ -158,8 +171,137 @@ export class FsProjectManager implements ProjectManager {
...
-    yield { message: "Synthesizing CloudFormation templates" };
-    await this.run(["npm", "run", "cdk", "--", "synth", "--quiet"], cdkDir);
+    //
+    // --output is passed explicitly rather than left to cdk.json, because deploy
+    // reads this directory back. A project that sets cdk.json's `output` would
+    // otherwise send synth somewhere deploy never looks, and deploy would ship
+    // whatever stale assembly it found there while reporting success.
+    yield { kind: "step", message: "Synthesizing CloudFormation templates" };
+    await this.run(
+      ["npm", "run", "cdk", "--", "synth", "--quiet", "--output", this.assemblyPath(project)],
```
Hweinstock: is there a way we can move more of the CDK specific logic into a separate class/module (maybe the CDK runner that already exists)?

I feel like the project manager is going to grow a lot as we add more project functionality, and I think it makes sense as a orchestration layer over the project functionality that shouldn't be concerned with implementations. For example, I feel like the scaffolding code that @tejaskash did is a nice example of separation of concerns we could try to follow.

Hweinstock (commented):
I think this is one of the hardest pieces of functionality to implement in the project, but I think this is an awesome start. Had a few questions about ways we might be able to simplify, and maybe make it easier to support terraform in the future. 

Also, maybe just nit on my end, but I find some of the comments overly verbose and distracting. However, if we see value in them expressing ideas not already expressed in the code, then I'm open to keeping them.

src/handlers/project/deploy/index.ts
notgitika: true, editing the comment

src/handlers/project/types.ts
notgitika: you are right, removed

src/handlers/project/types.ts
notgitika: yes, I renamed this to `DeployProjectInput`, matching `CreateProjectInput`/`ResolveProjectInput`.

src/core/project/manager.tsx
notgitika: haha yes I am editing all comments now

src/core/project/manager.tsx
notgitika: yes, made it concise

src/core/project/manager.tsx
notgitika: agree, I will move this to a `backends/cdk.ts` file

src/handlers/project/deploy/index.ts
notgitika: it is a no-op for now. it is a pretty quick operation so honestly we can just remove it and see if there is a need to add it back

src/handlers/project/types.ts
notgitika: i dont see any really. `--target` is required and resolved before anything is built. so far, one deploy ships one target.
we can add a `--all` flag or the ability to deploy multiple ones in one go.

src/core/project/manager.tsx
notgitika: moved the logic

src/core/project/assembly.ts
notgitika: nothing is re-derived. `stackForTarget` reads synth's manifest.json and selects the stack whose `agentcore:target-name` tag matches, precisely so the CLI never reproduces the app's naming convention

src/handlers/project/types.ts
notgitika: planning to add --target DEFAULT as default in a follow up PR when I do the credential preflight threading. it is not in the scope for this PR

src/projectSchemas/aws-targets.ts
```diff
@@ -0,0 +1,12 @@
+import z from "zod";
+
+/**
+ * Mirrors AwsDeploymentTarget in @aws/agentcore-cdk: each entry names the account and
+ * region the CDK app synthesizes a stack for. `agentcore project create` scaffolds an
+ * empty list, so a project has no targets until the user fills them in.
+ */
```
notgitika, line 7: follow up PR would add the aws credential pre-flight that we do currently in the CLI, if needed

src/handlers/project/deploy/index.ts
notgitika: when the region's already bootstrapped. because for us that isn't quite a no-op. we bootstrap with `withExisting({ createCustomerMasterKey: true })`, so re-running it updates the shared CDKToolkit stack and rewrites a staging bucket that had no CMK to use a customer-managed key. It also needs CloudFormation/IAM write access on every deploy, which a deploy-only role doesn't have. (I could recreate this that is why i'd added the flag).

but that's better handled by not bootstrapping than by a flag, so the flag is gone and we probe like main branch does: one DescribeStacks on CDKToolkit, read BootstrapVersion, require >= 30, bootstrap only if it's missing or older

src/core/project/manager.tsx
notgitika: `DeserializationError` already names the path, so the message isn't the issue. the rewrap adds the `TARGETS_EXAMPLE` (schema-specific, can't live in json.read) and the error source: DeserializationError is a plain Error, so `fromError` marks it INTERNAL and a user's typo counts as a CLI bug. ProjectStateError is USER.

that said, DeserializationError should probably be USER-sourced for every caller. what do you think? if so, I can raise a follow-up PR but it would affect everything so I don't wanna lump it in this already massive one :)

src/io/cdk.ts
notgitika: fixed it

src/io/cdk.ts
```diff
@@ -0,0 +1,282 @@
...
+ * Worth one read: bootstrapping updates a CloudFormation stack every deploy in the account
+ * shares, so an environment somebody has already prepared is left exactly as they left it.
+ */
+export const isBootstrapCurrent = async (
+  region: string,
+  read: (region: string) => Promise<BootstrapStack> = describeBootstrapStack,
+): Promise<boolean> => {
+  // A stack that cannot be read counts as absent: bootstrap then reports the real
+  // problem — no such stack, no permission — better than a probe could.
+  const { version, usable } = await read(region).catch(() => ({ version: 0, usable: false }));
```
tejaskash, line 280: [P1] Could we only convert the CloudFormation stack-not-found `ValidationError` into `false` and rethrow all other errors? Right now AccessDenied, throttling, and timeouts all look like "stack absent", so deploy proceeds to `cdk bootstrap`. That can fail deploy-only roles and, after a transient read failure, update the shared `CDKToolkit` stack with `createCustomerMasterKey: true`.

src/handlers/project/deploy/index.ts
```diff
@@ -1,11 +1,66 @@
...
+            continue;
+          }
+          logger.info(event.message);
+          config.io.stderr.write(`${event.message}\n`);
+        }
+
+        // The outputs are logged as they are reported rather than as the lines below print
+        // them, so a later run can be compared against this one.
+        logger.child({ outputs }).info("deployed project");
+        config.io.stderr.write(`Deployed project '${project.name}'\n`);
```
tejaskash, line 53: [P2] Could the final result use stdout, and the global JSON renderer when `--json` is set? Every success/output write here goes to stderr, so `agentcore project deploy --json` leaves stdout empty and callers cannot pipe the deployed endpoints or ARNs. Progress can remain on stderr.

src/logging/location.tsx
```diff
@@ -0,0 +1,52 @@
...
+ * The prefix the rotating file transport names its files after: it writes
+ * `<prefix>-<date>.log`, and a further `.<n>` when a day's log passes the size limit.
+ */
+export function logFilePrefix({ home = homedir() }: Location = {}): string {
+  return join(home, ...LOG_DIRECTORY, LOG_PREFIX);
+}
+
+/** The file the rotating transport is writing for this run's day. */
+export function logFilePath({ home, at = RUN_STARTED_AT }: Location = {}): string {
+  return `${logFilePrefix({ home })}-${day(at)}.log`;
```
tejaskash, line 41: [P2] `logFilePath` always returns the base dated path, but `DailyRotateFile` moves active writes to `.1.log`, `.2.log`, and so on after 5 MB, and the transport does not create a symlink. A failure after rotation therefore prints an older file that may not contain the failure details. Could we report the active rotated file, or print a stable symlink/directory instead?

src/projectSchemas/aws-targets.ts
```diff
@@ -0,0 +1,19 @@
...
+export const AwsTargetsSchema = z.array(
```
tejaskash, line 8: [P2] Could this schema reuse or fully mirror the pinned `AwsDeploymentTargetsSchema`? `@aws/agentcore-cdk@0.1.0-alpha.45` enforces target-name syntax and length, supported regions, and unique target names. The current schema accepts empty or invalid names, unsupported regions, and duplicates, so CLI preflight succeeds and synthesis fails later.

tejaskash (commented):
Four correctness issues found during deploy-path validation.

src/testing/logging.tsx
```diff
@@ -38,6 +38,40 @@ export function createSilentLogger(): Logger {
...
+  message: string;
+  /** The bindings in effect on the child it was logged through. */
+  bindings: LoggerBindings;
+}
+
+/**
+ * Create a logger that records what was logged instead of writing it anywhere, so a
+ * test can assert on the level and bindings a line carries rather than only its text.
+ */
+export function createRecordingLogger(): { logger: Logger; logs: RecordedLog[] } {
```
Hweinstock, line 53: can't we already do this with the log query functionality below? I believe we get the whole structured object.

src/testing/TestCoreClient.tsx
```diff
@@ -1178,6 +1179,20 @@ export class TestGatewayClient implements CoreGatewayClient {
...
+   *
+   * Emitted messages are collected and then yielded, rather than delivered while
+   * the operation is still running as the real runner does. That the real one
+   * delivers them as they arrive is covered directly, in `src/io/cdk.test.ts`.
+   */
+  onCdkOperation?: (operation: CdkOperation, emit: (event: CdkEvent) => void) => void;
+  /** The outputs a deploy produces, as the real toolkit reports a deployed stack's. */
+  cdkOutputs?: CdkOutputs;
+  /** Whether the target environment already has a bootstrap stack a deploy can use. */
+  alreadyBootstrapped?: boolean;
```
Hweinstock, line 1195: is there a way we can inject implementations that exhibit this behavior, rather than special test flags?

src/testing/TestCoreClient.tsx
```diff
@@ -1947,14 +1962,49 @@ export class TestCoreClient implements Core {
...
+      this.projectCommands.push({ command, cwd });
+    };
+    const checkTool = async () => {}; // CI hosts don't have uv installed
     this.projectManager = new FsProjectManager({
-      logger: options?.logger ?? createSilentLogger(),
+      logger,
       json: options?.json,
-      runner: async (command, { cwd }) => {
-        this.projectCommands.push({ command, cwd });
+      runner,
```
Hweinstock, line 1981: i'm not sure I follow what this code is for. Is the idea that we want to test a non-mocked project manager? If so, could we leverage the real one with test dependencies?

src/index.ts
```diff
@@ -30,8 +30,10 @@ process.exit(
     // generate a unique identifier corresponding to this process of this CLI. (ex. one command invoke, one TUI session)
     const cliSessionId = crypto.randomUUID();
 
+    // The rotating log every command writes to, shared with the commands that print where
```
Hweinstock, line 33: I feel like this comment exposes implementation details that are only relevant to the logger.

src/logging/location.tsx
```diff
@@ -0,0 +1,52 @@
+import { homedir } from "node:os";
```
Hweinstock, line 1: im confused why we need to make changes to logging. Could we simplify by pointing the customer to `~./agentcore/logs/`?

src/handlers/project/deploy/index.ts
```diff
@@ -1,11 +1,66 @@
...
+
+      // Progress goes to stderr, keeping stdout for machine output. The deployment
+      // tooling's own narration goes to the log, so a deploy shows only its steps.
+      try {
+        let outputs: Record<string, string> = {};
+        for await (const event of config.projectManager.deploy(project, {
+          region: ctx.require(RegionKey),
+          target: flags.target,
+        })) {
+          if (event.kind === "outputs") {
```
Hweinstock, line 40: would it be simpler to have outputs as part of the return value?

src/core/project/manager.tsx
Hweinstock: do we need `TARGETS_EXAMPLE`? Letting the deserialization bubble up with the correct error source feels like the simplest implementation, but unsure if I am missing a requirement?

src/core/project/manager.tsx
```diff
@@ -49,13 +58,22 @@ export class FsProjectManager implements ProjectManager {
...
   private readonly json: ReadWriteJson;
+  private readonly backends: Partial<Record<ManagedBy, ProjectBackend>>;
 
   constructor(config: ProjectManagerConfig) {
     this.logger = config.logger;
     this.source = config.source ?? defaultSource();
     this.runner = config.runner ?? runProcess;
     this.checkTool = config.checkTool ?? requireTool;
     this.json = config.json ?? new FsReadWriteJson({ logger: config.logger });
+    this.backends = config.backends ?? {
```
Hweinstock, line 69: Would IaC be a more descriptive name than backend?

src/core/project/assembly.ts
Hweinstock: I guess re-derive is the wrong word, but I'm wondering why the build doesn't output the stack name allowing us to avoid inspecting the assembly directly.

Hweinstock (commented):
I wasn't able to review all of the latest changes, but I think this PR is getting too big and complex. Do you see an opportunity to split this up into smaller simpler chunks?

src/core/project/manager.tsx
notgitika: the reason I chose backends is because there would be a `non-iac` implementation in the future. furthermore, the abstraction represents whatever builds and deploys the project’s artifacts. 

I could change it to `ProjectDeploymentBackend` or something along those sorts if we want a more specific name

notgitika:
decomposing into PRs
#2055 #2056 #2057 #2058

src/core/project/manager.tsx
notgitika: Addressed in #2058 (c76b6380): deploy now lets the shared user-sourced DeserializationError from json.read() propagate directly. TARGETS_EXAMPLE remains only for the distinct missing-file and valid-but-empty-list cases.

src/core/project/assembly.ts
notgitika: Kept the manifest lookup in #2058 because generic build has no deployment target and may synthesize multiple stacks; the cloud assembly is CDK synth’s structured output, so returning one stack from build would still require this lookup while leaking CDK-specific semantics into the build contract. Clarified the implementation in c76b6380 by renaming stackForTarget/stackName to stackArtifactIdForTarget/stackArtifactId.

PR #2004, opened 2026-08-14
feat(project): implement add functionality. 

Hweinstock:
### Problem 
Add functionality is not implemented. 

### Solution
- nest project spec attributes within `spec` so `project.managedBy` becomes `project.spec.managedBy` to make updating the spec more ergonomic. We keep name at the top level. 
- implement add for harness using the generator pattern (will be more relevant for runtime). 
- swap to a discriminated union vs a flag config for more ergonomic typing. This allows a single parameter, and allows TS to fully infer the shape. 

### Testing 
- added unit tests that verify all the flags work as expected. 
- tested e2e by creating a project and adding a harness.


src/core/types.tsx
```diff
@@ -37,7 +37,7 @@ export type CoreFetch = (
 // full ClientConfig so callers can request any client customization (region,
 // endpoint, ...).
 export interface AwsClients {
-  control(config: ClientConfig): BedrockAgentCoreControlClient
+  control(config: ClientConfig): BedrockAgentCoreControlClient;
```
Hweinstock, line 40: perhaps a rebase issue, but this is failing ci on main https://github.com/aws/agentcore-cli/actions/runs/31832012548/job/94869692163.

src/handlers/eval/ondemand/ondemand.test.tsx
```diff
@@ -24,7 +24,9 @@ const TRACE: SessionTrace = {
 const RESULT: EvaluateResult = {
   sessionsRequested: 1,
   sessionsEvaluated: 1,
-  results: [{ evaluatorId: "Builtin.Helpfulness", value: 0.9 } as EvaluateResult["results"][number]],
```
Hweinstock, line 27: same as above.

src/handlers/project/types.ts
```diff
@@ -46,12 +36,23 @@ export type Project = {
...
 };
 
+/** Discriminated union input for {@link ProjectManager.addResource}. */
+export type AddResourceInput =
+  | {
+      resourceType: "harness";
+      resourceConfig: z.input<typeof HarnessSpecSchema>;
+    }
+  | {
+      resourceType: "runtime";
```
Hweinstock, line 50: added runtime just to start the pattern.

src/core/project/manager.tsx
```diff
@@ -130,26 +126,69 @@ export class FsProjectManager implements ProjectManager {
...
+        `a ${resourceType} with name '${resourceConfig.name}' already exists`,
+      );
+
+    const newResources = [...existingResources];
+
+    switch (resourceType) {
+      case "harness": {
+        yield { message: `Scaffolding harness in project` };
+        const harnessPath = await this.scaffoldHarness(project.rootPath, input.resourceConfig);
+        newResources.push({ name: input.resourceConfig.name, path: harnessPath });
```
tejaskash: Could we keep this path relative to the project root? `scaffoldHarness` returns an absolute path, so this writes the developer's local path into `agentcore.json`. It works on that machine, but after the project is moved or cloned, CDK still tries to read the old path. Writing `app/${name}` here would match the existing project format and keep the project portable.

src/core/project/templates.ts
```diff
@@ -90,3 +92,17 @@ export async function createProjectTreeFromTemplate(
...
   ]);
 }
+
+const DEFAULT_HARNESS_SYSTEM_PROMPT = "You are a helpful assistant";
+
+export async function createHarnessTreeFromSpec(
+  spec: z.input<typeof HarnessSpecSchema>,
+): Promise<FsTreeNode> {
+  return FsTreeNode.createDirectory(".", [
+    FsTreeNode.createFile("harness.json", async () => json(HarnessSpecSchema.parse(spec))),
```
tejaskash: I think the Dockerfile needs to be handled as part of this scaffold too. Right now `--dockerfile` only puts the filename in `harness.json`; this tree creates `harness.json` and `system-prompt.md`, but never copies the Dockerfile into `app/<harness>`. The command succeeds, then CDK uses this directory as the build context and cannot find the file. Can we validate and copy the source Dockerfile here, and store its basename like the existing CLI does?

src/handlers/project/add/harness/index.ts
```diff
@@ -146,11 +146,10 @@ export const createAddHarnessHandler = (config: AddProjectResourceConfig) =>
       };
 
       const project = ctx.require(ProjectKey);
-      for await (const event of config.projectManager.addResource(
-        project,
-        "harness",
-        harnessConfig,
-      )) {
+      for await (const event of config.projectManager.addResource(project, {
+        resourceType: "harness",
```
tejaskash, line 167: One part of the requested tool configuration gets lost before this call. For an `agentcore_gateway` tool, `toTool` copies `gatewayArn` but drops `outboundAuth`. For example, asking for `{ none: {} }` produces a harness config with no outbound auth, which changes the behavior back to the default AWS IAM mode. Could we preserve the `awsIam`, `none`, and `oauth` variants?

src/handlers/project/add/harness/index.ts
```diff
@@ -146,11 +146,10 @@ export const createAddHarnessHandler = (config: AddProjectResourceConfig) =>
...
+        resourceConfig: harnessConfig,
```
tejaskash, line 168: Private Git skill auth has a field mismatch here. The SDK input gives us a `credentialArn`, but `toSkill` writes that value into `credentialName`. During synth, CDK treats `credentialName` as a project credential key, so it cannot resolve the ARN-shaped value and deployment fails. Could we either accept a project credential name for this project command, or carry the ARN separately instead of renaming it?

src/handlers/project/add/harness/index.ts
```diff
@@ -146,11 +146,10 @@ export const createAddHarnessHandler = (config: AddProjectResourceConfig) =>
       };
 
       const project = ctx.require(ProjectKey);
-      for await (const event of config.projectManager.addResource(
-        project,
-        "harness",
-        harnessConfig,
-      )) {
+      for await (const event of config.projectManager.addResource(project, {
```
tejaskash, line 166: There is currently no way to add a Dockerfile-backed harness in VPC mode. The environment conversion carries subnets and security groups, but the harness schema also requires `networkConfig.vpcId` for Dockerfile builds, and this handler has no `--vpc-id` input. The command always fails validation for that combination. Could we add an explicit VPC ID option and thread it into the resource config?

src/core/project/manager.tsx
```diff
@@ -130,26 +126,69 @@ export class FsProjectManager implements ProjectManager {
...
+        break;
+      }
+      // TODO: add limited special casing for runtime and default for other resources that proxy directly to spec changes.
```
tejaskash, line 174: Could we either implement the runtime branch here or leave runtime out of `AddResourceInput` for now? At the moment a runtime call falls through this switch, emits the `Updating project spec` message, writes the unchanged runtime list, and returns successfully. That silent success will be hard for callers to diagnose.

tejaskash (commented):
I found a few cases where the new add path reports success but writes a project that will not behave as requested. I left the details inline.

src/core/project/manager.tsx
Hweinstock: good catch, let me swap that to the relative path.

src/core/project/manager.tsx
notgitika: I think that's fine right? assuming we are implementing `add runtime` right after this.

src/core/project/templates.ts
```diff
@@ -90,3 +92,17 @@ export async function createProjectTreeFromTemplate(
...
+    FsTreeNode.createFile(
+      "system-prompt.md",
```
notgitika, line 109: are we making the changes on the cdk side as well? `resolveSystemPrompt()` on there uses the inline systemPrompt first and onlyl reads `system-prompt.md` when that field is absent. so there are these 2 input sources. did we want that behavior to still exist?

what we are losing here is that there is no way a user will be warned if they make changes in their local md file here if they don't update the value from harness.json.

src/core/project/manager.tsx
```diff
@@ -130,26 +126,69 @@ export class FsProjectManager implements ProjectManager {
...
+    }
+
+    yield { message: `Updating project spec file at '${agentCoreSpecPath}'` };
+    const newProjectSpec = await this.json.write(agentCoreSpecPath, {
```
notgitika: so the order of operations here is:
create harness.json -> create system-prompt.md -> update agentcore.json

if the write fails, the harness files remain and retrying would fail because it "already exists". can we add some cleanup to address this edge case?

notgitika (changes requested):

src/core/project/templates.ts
Hweinstock: wasn't aware thats how it worked, but that makes sense since it keeps the project portable. let me add that.

src/handlers/project/add/harness/index.ts
Hweinstock: Yeah I think @notgitika spotted the same in https://github.com/aws/agentcore-cli/pull/1998#discussion_r3785940770. Was going to address as follow-up, but let me bring in here.

src/core/project/manager.tsx
Hweinstock: Going to be the next PR,  but understand the current behavior is confusing. I'll have it throw early.

src/handlers/project/add/harness/index.ts
Hweinstock: wow great edge case find, was not aware vpc id was required with VPC + docker. Let me add the explicit flag for this and make it clear.

src/core/project/templates.ts
Hweinstock: oh, good callout, was not aware of that behavior. I feel like having two sources of system prompts might be unnecessary and confusing if we scaffold a `system-prompt.md` for them so I think stripping the system prompt field from the json to maintain backwards compatibility, and treat the file as source of truth might be the simplest path forward.

src/handlers/project/add/harness/index.ts
Hweinstock: yeah, good catch, I don't think what I have is quite right. It looks like using name here is going against the pattern for harness, since gateway, browser, and memory references all leverage ARN here. AFAICT credential is the only name reference in the input shape. I think we should accept an ARN here to be consistent so I'll add it to the schema, and follow up with the CDK change to support.

src/core/project/templates.ts
notgitika: I agree!

notgitika (approved):
LGTM

tejaskash (approved):
LGTM

PR #2007, opened 2026-08-14
refactor(eval): address on-demand review feedback

jariy17:
Follow-up to #1983: makes on-demand evaluation failures easier to classify and removes review-identified noise.

## Changes

- Missing runtime telemetry now throws `ResourceNotFoundError`.
- Failed CloudWatch queries now throw `CloudWatchQueryError` with query status and ID metadata.
- Malformed telemetry produces at most one warning per command.
- Repeated validation tests now use `test.each`.
- Redundant and drift-prone comments were removed.

## Decisions

1. **Fail fast:** one session can require multiple `Evaluate` calls. Continuing after one fails could return incomplete results while appearing successful.
2. **Keep results flat:** `context.spanContext.sessionId` supports per-session filtering without changing the SDK response shape.
3. **Defer a generic poller:** extract one when another production caller needs the same behavior.
4. **Keep window resolvers separate:** batch feeds the API's `DataSourceConfig`, including CloudWatch log groups and `onlineEvaluationConfigArn`; on-demand feeds client-side Logs Insights.

## Verification

- Build, typecheck, lint, changed-file formatting, and focused tests pass.
- Full suite: 1,373 tests pass.


src/handlers/eval/ondemand/ondemand.test.tsx
```diff
@@ -41,6 +71,85 @@ async function run(args: string[], configure?: (core: TestCoreClient) => void) {
   return { core, stdout: io.stdout(), stderr: io.stderr() };
 }
 
+async function runWithRealCore(options: LogsOptions, logger = createSilentLogger()) {
```
Hweinstock, line 81: should this use the fixture pattern? Might be a little harder to trigger these cases, but seems possible unless i'm missing something.

Hweinstock (commented):
thanks for coming back to this! one question on the testing pattern, otherwise lgtm.

Hweinstock:
is this from this PR? 
```

Run bun run format:check
$ prettier --check .
Checking formatting...
[warn] src/core/types.tsx
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
error: script "format:check" exited with code 1
```

src/handlers/eval/ondemand/ondemand.test.tsx
jariy17: Kept it on inline stubs on purpose. The fixture harness records against a live account; these cases can't be provoked there:

1. Query resolving `Failed`/`Cancelled`/`Timeout`
2. Runtime log group missing while `aws/spans` exists
3. Malformed telemetry rows — this test also asserts logger warnings, which the fixture `run()` helper (silent logger, stdout only) can't capture

Same `.send()` seam, same real `CoreClient`, so it's the deterministic counterpart for the negative paths, not a weaker test. Added a comment saying so in `ab60ff5`.

jariy17:
> is this from this PR?

Yes — dropped semicolon in `src/core/types.tsx:40`. Restored it in `ab60ff5`; `bun run format:check` is clean now.

Hweinstock (approved):

aidandaly24 (approved):

PR #2011, opened 2026-08-17
fix: fix the build command to build a project that has no deployment targets

notgitika:
Error:

`project build` failed on every freshly created project:

```
AgentCore CDK synthesis failed: No deployment targets configured.
Please define targets in agentcore/aws-targets.json
```

`project create` scaffolds `agentcore/aws-targets.json` as an empty list, but the generated CDK app threw on an empty list and synthesized one stack per target — so a new project had nothing to synthesize.

Fix:

Build does not need to know where a project deploys. An empty target list now synthesizes a single environment-agnostic stack, resolving account and region from CloudFormation pseudo-parameters instead of pinning them at synth time.

That stack is named `AgentCore-<project>` and carries no `agentcore:target-name` tag, so it is not a deploy candidate. A project that does configure targets is unaffected.

### DIFF
--- a/src/assets/cdk/bin/cdk.ts
+++ b/src/assets/cdk/bin/cdk.ts
@@ -110,9 +110,12 @@ async function main() {
 
   const spec = await configIO.readProjectSpec();
   const targets = await configIO.readAWSDeploymentTargets();
-  if (targets.length === 0) {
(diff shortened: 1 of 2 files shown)


tejaskash (approved):

Hweinstock (approved):
thanks for the fix!

PR #2013, opened 2026-08-17
feat(dev): open the agent inspector from project dev

tejaskash:
## What this does

`agentcore project dev` now opens the Agent Inspector — the local web UI from the current CLI — by default. The browser opens automatically on an interactive terminal (never in scripts or `--json`), the URL always prints, and `--no-ui` restores plain log streaming. From the UI you can see every runtime in the project, start any of them with a click, chat with a running agent, and browse its traces as they arrive. Edits to `agentcore.json` reload the agent list live. The web UI itself is the unmodified `@aws/agent-inspector` package — this PR provides everything behind it.

## Scope: local dev loop only, deliberately

This PR is at **full parity with the current CLI for everything reachable in the refactor today**: supervise and lazily start all runtimes, chat over HTTP/SSE, local MCP + A2A + AGUI proxying, live traces, resources view, config-watch reload, the same security guards (loopback Host check, origin allowlist, `X-Agentcore-Local` on POSTs).

What it intentionally does **not** include: the deployed-target features of the current CLI's dev UI (invoke deployed agents, deployed memory records, CloudWatch traces, harness invocation). `project deploy` is still a stub on `refactor`, so those features would be dead code behind unreachable routes — an earlier draft carried them as unwired interfaces and a review rightly called that out (~1,200 lines removed). The SPA already degrades gracefully on their 404s; they return together with real wiring when deploy lands. One known cosmetic difference: agent ports scan per-protocol defaults rather than mainline's `uiPort+1+index` scheme.

## How it fits together

- **`DevSupervisor`** (`core/dev/supervisor.ts`) — agents start lazily when the browser asks, each in an abort scope chained to Ctrl-C for its whole lifetime; all runner output merges into one agent-attributed stream (`[hello_world] ...` in the terminal, an `agent` field in NDJSON). Hot reload stays inside the child (`uvicorn --reload` / `tsx watch`); `setRuntimes` applies config edits without touching running agents.
- **Inspector server** (`core/dev/inspector/`) — the local route surface ported wire-compatibly from the current CLI's web-ui: status, start, protocol-aware `/invocations`, traces, `/api/mcp`, A2A agent card, resources, static SPA with SPA-routing fallback and per-process asset caching.
- **Storage/presentation split** (from #1980's review): `TraceStore` is raw per-trace persistence; frontend shaping (flattening, noise filtering) lives in `core/dev/inspector/`. `io/httpServer` streams `AsyncIterable` bodies for SSE, skips body buffering on GET/HEAD, and caps request sizes.
- **Assets** — the SPA (4 text files, 616KB) is staged from node_modules into the ordinary asset pipeline at build time (gitignored), so the npm bundle and all six compiled binaries ship it with zero new distribution code. Bun bundles bare `.html` entrypoints through its HTML-imports pipeline, so staged files carry a neutral `.asset` suffix stripped on read. `AGENT_INSPECTOR_PATH` overrides for SPA development (uncached).

## How it was tested

Beyond the unit/route suite (1,595 tests green): a full end-to-end session against a **real project and a real Bedrock call** — built the npm bundle, scaffolded a project with `project create` (real `uv sync`), ran `project dev`, and drove the Inspector with Playwright from the browser origin: SPA renders, agent starts via the UI (CORS + local-header guards exercised), an invocation streams back from Bedrock through the proxy, traces appear in the timeline with span counts, trace detail resolves by id, and foreign-origin POSTs are rejected 403. 10/10 checks, re-run after the trim. A separate headed session via the Playwright MCP server confirmed the same flow interactively (chat answered live, span waterfall rendered).

## Stack

Stacks on #1980 (`feat/dev-otel-collector`); retarget as the stack merges. Together they complete the local dev-loop plan: #1980 collects traces, this PR shows them. Follow-up (blocked on deploy landing in the refactor): wire the deployed-target Inspector capabilities.


tejaskash:
Closing to re-land as five smaller serial PRs (<1,000 lines each), per review feedback on size. The code here is fully verified (unit + browser E2E against a real project and Bedrock call) and will be extracted from this branch unchanged where possible:

1. **B1** — DevSupervisor + multi-agent terminal mode for `project dev`
2. **B2** — trace storage/presentation split + SSE-capable HTTP responses
3. **B3** — Inspector UI walking skeleton: security guards, status/start/traces routes, SPA asset pipeline, browser auto-open, `--no-ui`/`--ui-port`
4. **B4** — chat: `/invocations` SSE proxy + MCP/A2A/AGUI translations + agent card
5. **B5** — `/api/resources` + agentcore.json live-reload + perf fixes + E2E script

The `feat/agent-inspector` branch stays as the extraction source until B5 lands. Each PR ships working behavior (no interface-only skeletons), submitted serially into `refactor` after #1980 merges.

PR #2018, opened 2026-08-18
feat(project): scaffold add runtime handler

Hweinstock:
## Project 
runtimes cannot be added to a project post creation. 

## Solution 
We mostly follow the pattern establish by harness, but with a few exceptions:
- mirroring API shapes doesn't work as well here since API shapes don't support local BYO, but we do. Also, API supports s3 code artifacts, but we don't. Therefore we mirror where possible, but diverge when describing code configuration. 
- we extend the templates idea to runtime as well. This simplifies the command significantly. Rather than multiple flags for memory, language, protocol, etc. We can have a single one for template, that renders a preloaded template. This allows users to get started faster, without giving up on advanced configuration. This also leaves us flexible for folks to share runtime definitions across projects through custom templates (future feature). In the future we can add our templates here like `strands-py-bedrock` or `strands-ts-gemini-no-memory`. The help page would list all of these out so customers know their options. 

## Testing 
Setup a table based test for verifying flag validation and parsing. Will follow up on these tests once functionality is implemented in a follow-up. 

```
> agentcore project add runtime --name t --template hello-world-python
Error: runtime case not yet implemented in FsProjectManager.addResource

> agentcore project add runtime --help
Usage: agentcore project add runtime [options]
adds a runtime to the current project either from a template or from existing local code

Options:
  --name <name>                                                  the name of the runtime
  --description <description>                                    an optional description of the runtime
  --template <template>                                          template to scaffold from
  --role-arn <role-arn>                                          IAM role ARN that provides permissions for the runtime
  --code-location <code-location>                                path to existing agent source code (BYO path)
  --build <build>                                                build type: CodeZip or Container
  --entrypoint <entrypoint>                                      entrypoint file, e.g. main.py:handler (BYO only)
  --protocol <protocol>                                          server protocol ex. HTTP, MCP, A2A, AGUI
  --api-key <api-key>                                            API key source for non-bedrock model providers: '-' for stdin, 'file://path' for file
  --model-provider <model-provider>                              model provider (template only)
  --runtime-version <runtime-version>                            language runtime, e.g. PYTHON_3_13, NODE_22 (BYO CodeZip only)
  --dockerfile <dockerfile>                                      dockerfile path for the container build (BYO Container only)
  --build-context-path <build-context-path>                      docker build context directory relative to project root (BYO Container only)
  --custom-docker-build-args <custom-docker-build-args>          docker build args as JSON key/value object (BYO Container only)
  --additional-policies <additional-policies...>                 additional IAM policy ARNs or policy document paths for the execution role
  --network-configuration <network-configuration>                network configuration (JSON NetworkConfiguration)
  --vpc-id <vpc-id>                                              VPC ID for Container builds in VPC mode (CodeBuild cannot infer it from subnets)
  --authorizer-configuration <authorizer-configuration>          inbound authorizer configuration (JSON AuthorizerConfiguration)
  --protocol-configuration <protocol-configuration>              protocol configuration (JSON ProtocolConfiguration)
  --request-header-configuration <request-header-configuration>  request header passthrough configuration (JSON RequestHeaderConfiguration)
  --lifecycle-configuration <lifecycle-configuration>            lifecycle configuration (JSON LifecycleConfiguration)
  --environment-variables <environment-variables>                environment variables (JSON object of key/value strings)
  --filesystem-configurations <filesystem-configurations>        filesystem mount configurations (JSON FilesystemConfiguration[])
  --memory <memory>                                              memory configuration (JSON with mode: none | create | existing ) (template only)
  --tags <tags>                                                  tags to apply (JSON object of key/value strings)
  -h, --help       
```

## Follow-up items
- implement add runtime add the project manager level. 
- split out test file into resource specific cases. 
- make entrypoint optional. Its weird that we require it for container agents. The old cli defaulted to `main.py` which is unnecessary.

### DIFF
--- a/src/handlers/project/project.test.ts
+++ b/src/handlers/project/project.test.ts
@@ -10,7 +10,7 @@ import {
   TestGlobalConfigAccessor,
   testIO,
 } from "../../testing";
-import { InputValidationError } from "../../errors";
+import { InputValidationError, NotImplementedError } from "../../errors";
 import { FsReadWriteJson, type ReadWriteJson } from "../../io";
 
 async function run(args: string[], opts?: { core?: TestCoreClient }) {
@@ -112,7 +112,7 @@ describe("project create", () => {
 
 describe("project add harness", () => {
   const defaultModel = { provider: "bedrock", modelId: "global.anthropic.claude-sonnet-4-6" };
-  /** Verify error case for different flags **/
+  // Verifies each flag combination produces the expected harness config.
   test.each<[string, string[], Record<string, unknown>]>([
     ["minimal — name only", ["--name", "x"], { model: defaultModel }],
     [
@@ -610,6 +610,7 @@ describe("project add harness", () => {
     );
   });
 
+  // Rejects invalid flag combinations with InputValidationError.
   test.each([
     ["missing --name", ["--model", '{"bedrockModelConfig":{"modelId":"x"}}']],
     ["model without modelId", ["--name", "x", "--model", '{"bedrockModelConfig":{}}']],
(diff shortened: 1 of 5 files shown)


src/projectSchemas/runtime.ts
```diff
@@ -49,7 +49,7 @@ export const EntrypointSchema = z
   .regex(
     /^[a-zA-Z0-9_][a-zA-Z0-9_/.-]*\.(py|ts|js)(:[a-zA-Z_][a-zA-Z0-9_]*)?$/,
     'Must be a Python (.py) or TypeScript (.ts/.js) file path with optional handler (e.g., "main.py:handler" or "index.ts")',
-  ) as unknown as z.ZodType<FilePath>;
```
Hweinstock, line 52: this forced me to cast in the handler, so I removed it. I think its a relic from old CLI.

src/handlers/project/types.ts
```diff
@@ -40,6 +55,44 @@ export type Project = {
   spec: z.infer<typeof ProjectSpecSchema>;
 };
 
+/** Shared infrastructure config fields for a runtime (independent of source type). */
```
Hweinstock: because so many schema fields are determined at scaffold time, its useful to define this intermediate representation of the data.

src/handlers/project/project.test.ts
```diff
@@ -10,7 +10,7 @@ import {
   TestGlobalConfigAccessor,
```
Hweinstock, line 10: as a follow-up i think I'll break up this file.

src/handlers/project/types.ts
```diff
@@ -69,3 +125,25 @@ export interface ProjectManager {
...
+
+export const runtimeModelProviderSchema = z.enum(["bedrock", "anthropic", "openai", "gemini"]);
+export type RuntimeModelProvider = z.infer<typeof runtimeModelProviderSchema>;
+
+export type RuntimeModelProviderConfig = {
+  provider?: RuntimeModelProvider;
+  apiKey?: string;
+};
+
+export const runtimeMemoryConfigSchema = z.discriminatedUnion("mode", [
```
Hweinstock, line 137: this proposed structure would allow us to support existing memories (not supported today), and more advanced configuration. By default, I think we give them the same default memory as old CLI, and the customer can customize here.

Hweinstock:
moved to https://github.com/aws/agentcore-cli/pull/2035

PR #2019, opened 2026-08-18
fix(memory): use --id for memory selectors

notgitika:
This PR standardize the parent Memory selector on `--id` across event, record, actor, and session commands. We see this convention in the rest of the CLI.

I also add a regression assertion that imperative Memory selectors do not use `--memory`

bun build, typecheck, lint, format and test pass


aidandaly24 (approved):
No comments this LGTM, thanks!

nborges-aws (approved):
Nice work!

PR #2025, opened 2026-08-18
feat(project): add `project add memory`

notgitika:
This PR adds `agentcore project add memory`, which validates and appends a memory resource to `spec.memories` in `agentcore.json`. Memory resources do not scaffold any application files and are deployed through the generated CDK application.

Example:
```
  agentcore project add memory \
    --name UserFacts \
    --strategies SEMANTIC,EPISODIC
```

To maintain parity with the xisting CLI functionality, while also adding more configurability and customization, `--strategies` supports 2 input types:

  - Comma-separated managed strategy types using default namespaces
  - The memory's `strategies` array as JSON, exactly as it is stored in `agentcore.json`, for explicit names, descriptions, and namespaces (there is a parameter help in `--help` for ease of understanding for the user). It is parsed with the project schema's own `MemoryStrategySchema`, so the flag cannot drift from the schema.

The command also supports expiry duration, indexed keys, stream delivery, encryption and execution roles, descriptions, and tags (description is a new optional field. Here is the CDK PR for it https://github.com/aws/agentcore-l3-cdk-constructs/pull/325 )

### Callouts:

  - CUSTOM strategies are not supported because their extraction configuration cannot yet be represented by the project schema.

  ### Validation

I tested it myself e2e with different combinations, flags, strategy conversions and validation failures. It all works well. I also tested it with `agetncore project build` 

TODO: check compatibility with `harness` resource


src/core/project/manager.tsx
```diff
@@ -166,6 +167,21 @@ export class FsProjectManager implements ProjectManager {
         });
         break;
       }
+      case "memory": {
```
Hweinstock: could this be a default case that just pushes the resources? I think we're gonna end up writing the same code for every primitive.

src/core/project/manager.tsx
Hweinstock: Would adding validation against `ProjectSpecSchema` here be simpler than per resource validation? I probably should have set this up in my earlier PR.

src/handlers/project/add/memory/index.ts
```diff
@@ -0,0 +1,293 @@
...
+
+// The service default for raw event retention, applied when --event-expiry-duration
+// is omitted so the common case is a single --name.
+const DEFAULT_EVENT_EXPIRY_DURATION = 30;
+
+// TODO: drop the deploy-time caveat once the generated CDK app pins an
+// @aws/agentcore-cdk release containing the memory `description` field
+// (aws/agentcore-l3-cdk-constructs#325). Its MemorySchema is non-strict, so
+// until then the field is stripped at synth rather than rejected.
+const descriptionHelp = `(string)
```
Hweinstock: where does this come from? Are we responsible for updating it manually?

src/handlers/project/add/memory/index.ts
```diff
@@ -0,0 +1,293 @@
...
+  }
+  // CUSTOM is left out on purpose. Its point is the extraction configuration --
+  // prompt and model overrides, or a self-managed pipeline -- and neither the
+  // project spec nor the CDK's memory schema can carry one. Offering the type
+  // without it produced aws/agentcore-cli#241 ("select custom memory strategy,
+  // note there is no option to add prompts"), so it was removed in #266 and again
+  // in #713. Re-enable it alongside the configuration, not before: #676 tracks it.
+  if ("customMemoryStrategy" in strategy && strategy.customMemoryStrategy)
+    throw new InputValidationError(
+      `customMemoryStrategy is not supported: a custom strategy's extraction configuration cannot be expressed yet (see aws/agentcore-cli#676). Expected one of ${MemoryStrategyTypeSchema.options.join(", ")}`,
```
Hweinstock: I don't think we want PR references in user facing error messages.

src/handlers/project/add/memory/index.ts
```diff
@@ -0,0 +1,293 @@
...
+    );
+  throw new InputValidationError("Unrecognized memory strategy variant");
+}
+
+/** The fields every SDK memory strategy variant shares, in project-schema terms. */
+function commonStrategyFields(strategy: {
```
Hweinstock: what exactly is this function doing?

src/handlers/project/add/memory/index.ts
notgitika: this was added accidentally while I was still working on the PR. this one is removed now.

the other `strategiesHelp` takes inspiration from https://github.com/aws/agentcore-cli/blob/refactor/src/handlers/harness/parameterHelp.tsx

src/handlers/project/add/memory/index.ts
notgitika: I renamed it to `toProjectStrategyFields` for clarity now and improved the comment. 
basically, this function allowlist-picks the fields the project schema supports from SDK strategy objects to avoid spreading since the SDK types carry some extra fields (memoryRecordSchema, etc.) that could leak in silently.

src/handlers/project/add/memory/index.ts
notgitika: Yeah it was added by mistake previously, I just removed it and just made the error `customMemoryStrategy is not supported`

custom memory strategy is something we must think about once we reach parity to the existing functionality in the CLI.

src/core/project/manager.tsx
notgitika: good call out! about to add it.

src/core/project/manager.tsx
notgitika: hmm interesting I see what you mean. let me explore it, I think it might make things simpler.

src/core/project/manager.tsx
notgitika: I rebased onto #2031 and now reuse its `ProjectSpecSchema` validation for the complete updated spec

src/core/project/manager.tsx
notgitika: done, harness and runtime remain special cases, while the default case now adds spec-only resources directly

notgitika:
My agent ran focused testing for `project add memory` after the latest validation changes:

- 122 focused project tests pass, plus build/typecheck/lint/format
- Rebuilt CLI verified valid JSON persistence and atomic rejection for the new edge cases
- Covered strategy unions, unsupported fields, indexed keys, stream delivery, tags, malformed JSON, and boundary inputs
- Live AWS lifecycles passed for minimal memory, all four managed strategies, and Kinesis stream delivery (`CREATE_COMPLETE` followed by cleanup)
- Confirmed no test stacks, memories, or streams remain

The testing surfaced and we fixed empty shorthand entries, recursive unsupported-field stripping (including `__proto__`), missing JSON strategy names, and the JSON-object diagnostic. The remaining long default strategy-name boundary is in the existing project schema/CDK naming layer rather than this handler.

src/core/project/manager.tsx
```diff
@@ -139,7 +139,7 @@ export class FsProjectManager implements ProjectManager {
         `a ${resourceType} with name '${resourceConfig.name}' already exists`,
       );
 
-    const newResources = [...existingResources];
+    const newResources: unknown[] = [...existingResources];
```
nborges-aws, line 161: Was discussing this with @jariy17, the default makes sense to avoid having cases that all have no scaffolding and directly push. But at the same time, having the default loosens our validations since new resourceTypes will silently fall through to the default.

Meet in the middle: a resourceType enum, stack cases for new resourceTypes that do not need scaffolding

notgitika: Addressed in the rebase. `addResource` now stacks explicit `case "config-bundle"` and `case "memory"` arms for spec-only resources, while retaining the exhaustive `never` default. Adding a new resource type will therefore fail typechecking until its handling is made explicit.

src/handlers/project/add/memory/index.ts
```diff
@@ -0,0 +1,312 @@
...
+  createHandler({
+    name: "memory",
+    description: "adds a memory to the current project",
+    flags: [
+      flag("name", "the name of the memory", z.string().optional()),
+      flag("description", "a description of what the memory stores", z.string().optional()),
+      flag(
+        "event-expiry-duration",
+        "how long raw events are retained, in days (3-365)",
+        z.number().int().min(3).max(365).default(DEFAULT_EVENT_EXPIRY_DURATION),
```
nborges-aws, line 117: Is 3 day min here correct? I thought min was 7 in CDK

notgitika: it is 3 in the service, 7 in the L3 CDK, which is incorrect. I have this issue open to track the correction in the CDK 
https://github.com/aws/agentcore-l3-cdk-constructs/issues/334

nborges-aws (dismissed):
LGTM thanks for updates!

src/handlers/project/add/memory/index.ts
```diff
@@ -0,0 +1,312 @@
...
+const DEFAULT_EVENT_EXPIRY_DURATION = 30;
+
+const strategyFields = {
+  name: MemoryStrategyNameSchema,
+  description: z.string().optional(),
+  namespaces: z.array(z.string()).optional(),
+  namespaceTemplates: z.array(z.string()).optional(),
+};
+
+function projectMemoryObject<T extends z.ZodRawShape>(shape: T, label: string) {
```
jariy17, line 20: Why didn't we use [zod.strictObject](https://zod.dev/api?id=zstrictobject#zstrictobject)

nborges-aws (dismissed):

src/handlers/project/add/memory/index.ts
```diff
@@ -0,0 +1,312 @@
...
+  const supportedFields = new Set(Object.keys(shape));
```
jariy17, line 21: I feel like there is a lot going on here just for parse and converting strategy inputs. Maybe we should create an adapter like this 
```
xport interface MemoryStrategyAdapter<TInput = unknown> {
    readonly type: MemoryStrategyType;
    readonly memberKey: string;
    readonly inputSchema: z.ZodType<TInput>;

    fromInput(input: TInput): MemoryStrategy;
    toInput(strategy: MemoryStrategy): TInput;

    fromShorthand?(): MemoryStrategy;
    canUseShorthand(strategy: MemoryStrategy): boolean;
  }

  The codec indexes registered adapters:

  export class MemoryStrategyFlagCodec {
    private readonly byType: Map<string, MemoryStrategyAdapter>;
    private readonly byMember: Map<string, MemoryStrategyAdapter>;

    constructor(adapters: readonly MemoryStrategyAdapter[]) {
      this.byType = new Map(
        adapters.map((adapter) => [adapter.type, adapter]),
      );
      this.byMember = new Map(
        adapters.map((adapter) => [adapter.memberKey, adapter]),
      );
    }

    parse(raw: string): MemoryStrategy[] {
      return raw.trimStart().startsWith("[")
        ? this.parseJson(raw)
        : this.parseShorthand(raw);
    }

    toFlag(strategies: readonly MemoryStrategy[]): string {
      const adapters = strategies.map((strategy) => {
        const adapter = this.byType.get(strategy.type);

        if (!adapter) {
          throw new InputValidationError(
            `Unsupported memory strategy '${strategy.type}'`,
          );
        }

        return adapter;
      });

      if (
        strategies.every((strategy, index) =>
          adapters[index].canUseShorthand(strategy),
        )
      ) {
        return strategies.map((strategy) => strategy.type).join(",");
      }

      return JSON.stringify(
        strategies.map((strategy, index) => {
          const adapter = adapters[index];

          return {
            [adapter.memberKey]: adapter.toInput(strategy),
          };
        }),
      );
    }

    private parseShorthand(raw: string): MemoryStrategy[] {
      return raw.split(",").map((value) => {
        const type = value.trim();
        const adapter = this.byType.get(type);

        if (!adapter?.fromShorthand) {
          throw new InputValidationError(
            `Unsupported shorthand strategy '${type}'`,
          );
        }

        return adapter.fromShorthand();
      });
    }

    private parseJson(raw: string): MemoryStrategy[] {
      const inputs = JSON.parse(raw) as unknown[];
      return inputs.map((input) => this.parseJsonMember(input));
    }

    private parseJsonMember(input: unknown): MemoryStrategy {
      // Validate that input is an object containing exactly one member.
      const [memberKey] = Object.keys(input as object);
      const adapter = this.byMember.get(memberKey);

      if (!adapter) {
        throw new InputValidationError(
          `Unsupported memory strategy member '${memberKey}'`,
        );
      }

      const value = adapter.inputSchema.parse(
        (input as Record<string, unknown>)[memberKey],
      );

      return adapter.fromInput(value);
    }
  }

  Registration is centralized:

  const strategyCodec = new MemoryStrategyFlagCodec([
    new StandardStrategyAdapter("SEMANTIC", "semanticMemoryStrategy"),
    new StandardStrategyAdapter(
      "SUMMARIZATION",
      "summaryMemoryStrategy",
    ),
    new StandardStrategyAdapter(
      "USER_PREFERENCE",
      "userPreferenceMemoryStrategy",
    ),
    new EpisodicStrategyAdapter(),
  ]);

```

jariy17 (changes requested):

src/handlers/project/add/memory/index.ts
notgitika: good suggestion!

notgitika:
Per @jariy17's ask for a 1-by-1 comparison — every flag is one `agentcore.json` memory field, and `MemorySchema` has exactly these 9 fields, so there is no flag without a field and no field without a flag.

| flag | `memories[]` field |
| --- | --- |
| `--name` | `name` |
| `--description` | `description` |
| `--event-expiry-duration` | `eventExpiryDuration` |
| `--strategies` | `strategies[]` |
| `--indexed-keys` | `indexedKeys[]` |
| `--stream-delivery-resources` | `streamDeliveryResources` |
| `--encryption-key-arn` | `encryptionKeyArn` |
| `--execution-role-arn` | `executionRoleArn` |
| `--tags` | `tags` |

One command using all 9 flags:

```console
agentcore project add memory \
  --name UserFacts \
  --description "Durable facts and preferences for each end user." \
  --event-expiry-duration 45 \
  --strategies '[{"type":"SEMANTIC","name":"facts","description":"Durable user facts","namespaceTemplates":["/users/{actorId}/facts"]},{"type":"EPISODIC","name":"episodes","namespaceTemplates":["/episodes/{actorId}/{sessionId}"],"reflectionNamespaceTemplates":["/episodes/{actorId}"]}]' \
  --indexed-keys '[{"key":"tenant","type":"STRING"}]' \
  --stream-delivery-resources '{"resources":[{"kinesis":{"dataStreamArn":"arn:aws:kinesis:us-east-1:123456789012:stream/memory","contentConfigurations":[{"type":"MEMORY_RECORDS","level":"FULL_CONTENT"}]}}]}' \
  --encryption-key-arn arn:aws:kms:us-east-1:123456789012:key/abc \
  --execution-role-arn arn:aws:iam::123456789012:role/MyMemoryRole \
  --tags '{"team":"ml"}'
```

The resulting `memories[0]`, verbatim from the written `agentcore.json`:

```json
{
  "name": "UserFacts",
  "description": "Durable facts and preferences for each end user.",
  "eventExpiryDuration": 45,
  "strategies": [
    {
      "type": "SEMANTIC",
      "name": "facts",
      "description": "Durable user facts",
      "namespaceTemplates": ["/users/{actorId}/facts"]
    },
    {
      "type": "EPISODIC",
      "name": "episodes",
      "namespaceTemplates": ["/episodes/{actorId}/{sessionId}"],
      "reflectionNamespaceTemplates": ["/episodes/{actorId}"]
    }
  ],
  "indexedKeys": [{ "key": "tenant", "type": "STRING" }],
  "tags": { "team": "ml" },
  "encryptionKeyArn": "arn:aws:kms:us-east-1:123456789012:key/abc",
  "executionRoleArn": "arn:aws:iam::123456789012:role/MyMemoryRole",
  "streamDeliveryResources": {
    "resources": [
      {
        "kinesis": {
          "dataStreamArn": "arn:aws:kinesis:us-east-1:123456789012:stream/memory",
          "contentConfigurations": [{ "type": "MEMORY_RECORDS", "level": "FULL_CONTENT" }]
        }
      }
    ]
  }
}
```

The `--strategies` JSON above is the `strategies` array as stored, character for character — `da37fedf` drops the flag's own copy of that schema and parses with the project schema's `MemoryStrategySchema` itself (wrapped only to report unsupported fields), so `type`/`name`/`description`/`namespaceTemplates`/`reflectionNamespaceTemplates` and all of the schema's cross-field rules come from one place and can't drift. `--strategies SEMANTIC,EPISODIC` remains as the shorthand that fills in the default namespaces.

src/handlers/project/add/memory/index.ts
```diff
@@ -0,0 +1,220 @@
...
+      flag("name", "the name of the memory", z.string().optional()),
+      flag("description", "a description of what the memory stores", z.string().optional()),
+      flag(
+        "event-expiry-duration",
+        "how long raw events are retained, in days (3-365)",
+        z.number().int().min(3).max(365).default(DEFAULT_EVENT_EXPIRY_DURATION),
+      ),
+      flag(
+        "strategies",
+        "long-term memory strategies: comma-separated types, or the JSON strategies[] as stored in agentcore.json",
```
jariy17, line 121: This should use SourceResolver.

src/handlers/project/add/memory/index.ts
```diff
@@ -0,0 +1,220 @@
...
+function toDefaultStrategy(type: string): MemoryStrategy {
+  const parsed = MemoryStrategyTypeSchema.safeParse(type);
+  if (!parsed.success)
+    throw new InputValidationError(
+      `unrecognized memory strategy '${type}'; expected one of ${MemoryStrategyTypeSchema.options.join(", ")}`,
+    );
+
+  return {
+    type: parsed.data,
+    namespaceTemplates: DEFAULT_STRATEGY_NAMESPACE_TEMPLATES[parsed.data],
```
jariy17, line 213: shift maps into handler. could be a follow up.

nborges-aws (approved):
LGTM. Agree with TJs latest comments for a follow up

src/handlers/project/add/memory/index.ts
```diff
@@ -0,0 +1,220 @@
...
+    flags: [
+      flag("name", "the name of the memory", z.string().optional()),
+      flag("description", "a description of what the memory stores", z.string().optional()),
+      flag(
+        "event-expiry-duration",
+        "how long raw events are retained, in days (3-365)",
+        z.number().int().min(3).max(365).default(DEFAULT_EVENT_EXPIRY_DURATION),
+      ),
+      flag(
+        "strategies",
```
jariy17, line 120: Please use SourceResolver in follow up PR

src/handlers/project/add/memory/index.ts
```diff
@@ -0,0 +1,220 @@
...
+/** Expands a bare strategy type into a strategy carrying its default namespaces. */
+function toDefaultStrategy(type: string): MemoryStrategy {
+  const parsed = MemoryStrategyTypeSchema.safeParse(type);
+  if (!parsed.success)
+    throw new InputValidationError(
+      `unrecognized memory strategy '${type}'; expected one of ${MemoryStrategyTypeSchema.options.join(", ")}`,
+    );
+
+  return {
+    type: parsed.data,
```
jariy17, line 212: Shift templates to here because its not part of the schema type.

jariy17 (approved):
Follow Up PR

PR #2028, opened 2026-08-18
feat(project): add gateway resources

aidandaly24:
## Summary

Adds project-based Gateway resource management:

```text
agentcore project add gateway
agentcore project add gateway-target
agentcore project add gateway-connector

agentcore project remove gateway
agentcore project remove gateway-target --gateway <name>
agentcore project remove gateway-connector --gateway <name>
```

Project-command JSON maps directly to the existing version 1 `agentcore.json` schemas:

- `--authorizer-configuration` accepts `GatewayAuthorizerConfigSchema`;
- `--target-configuration` accepts one complete `AgentCoreGatewayTargetSchema` object;
- `--connector-configuration` accepts the same complete Target object and requires `targetType: "connector"`;
- JSON sources work inline, through `file://`, and through stdin;
- SDK request shapes are not accepted or translated.

Scalar endpoint, Runtime, and curated Connector shortcuts construct the same
project-schema objects accepted through JSON. Connector removal is an alias over
Gateway Target removal because Connectors are stored as ordinary Targets.

Gateway protocol behavior preserves the existing schema and L3 contract:

- omitting `--protocol-type` persists `protocolType: "None"` and creates an unrestricted Gateway;
- `--protocol-type MCP` persists `protocolType: "MCP"` and creates an MCP-only Gateway;
- `"None"` is not exposed as a flag value because omission already selects it;
- semantic search requires `--protocol-type MCP`.

The change also:

- makes project-owned compute Target branches reachable through direct Target JSON;
- validates OAuth/API-key credential-name references in complete Target JSON and OAuth shortcuts;
- validates Runtime and Runtime endpoint references;
- validates the complete candidate project before one `agentcore.json` write;
- uses the existing generic project-resource mutation flow, with only nested Target placement and removal handled specially;
- rejects duplicate Target names across all Gateways and `unassignedTargets[]`, matching the existing L3 Target-output contract.

The CLI does not create or copy files referenced by Target JSON. Paths such as
`schemaSource.inline.path` and `lambdaFunctionArn.toolSchemaFile` remain explicit
in `agentcore.json`.

## Testing

Verification on rebased CLI head `5fc696f2`:

- focused Gateway plus affected project deploy/Memory tests: 179 passed;
- `bun test src --coverage` outside the restricted sandbox: 1,948 passed;
- project line coverage: 97.90%;
- modified Gateway add handlers, remove handler, and Gateway schema: 100% line coverage;
- `bun run typecheck`;
- `bun run lint:check`;
- `bun run format:check`;
- `bun run build`;
- built CLI help verified for Gateway, Gateway Target, and removal commands.

### Live deployment

Earlier end-to-end verification of the persisted Gateway and Target shapes used
a generated project with `@aws/agentcore-cdk@0.1.0-alpha.45`:

- Runtime and direct MCP Targets reached `READY`;
- Runtime invocation through Gateway returned HTTP 200;
- MCP `initialize` and `tools/list` returned HTTP 200;
- the stack and all resources were destroyed and verified absent.

Credential-provider authoring is owned by the separate project credentials
commands. This PR validates Gateway Target references to those project
credentials but does not create them.


src/core/project/manager.tsx
```diff
@@ -195,9 +198,25 @@ export class FsProjectManager implements ProjectManager {
...
         newResources.push(resourceConfig);
         break;
-
+      case "gateway-target": {
+        const gatewayIndex = existingProjectSpec.agentCoreGateways.findIndex(
+          (gateway) => gateway.name === input.gatewayName,
+        );
+        if (gatewayIndex < 0) {
+          throw new InputValidationError(
+            `gateway '${input.gatewayName}' does not exist in agentCoreGateways[]`,
```
jariy17: nit: does not exist in project. Please check agentCoreGateway in agentcore.json

src/core/project/manager.tsx
```diff
@@ -283,6 +302,22 @@ export class FsProjectManager implements ProjectManager {
     };
   }
 
+  private assertUniqueGatewayTargetName(project: Project["spec"], name: string): void {
```
jariy17: Are all gateway target names required to be unique across every gateway, or must they only be unique inside a single gateway?

src/handlers/project/add/gateway-connector/index.ts
```diff
@@ -0,0 +1,127 @@
...
+        z.enum(["web-search", "bedrock-knowledge-bases"]).optional(),
+      ),
+      flag(
+        "connector-configuration",
+        "complete connector agentCoreGateways[].targets[] object (JSON; inline, file://<path>, or - for stdin)",
+        z.string().optional(),
+      ),
+      flag(
+        "knowledge-base",
+        "project Knowledge Base name or external ten-character Knowledge Base ID",
```
jariy17: add that this could only be used with `bedrock-knowledge-base`

src/handlers/project/add/gateway-target/index.ts
```diff
@@ -0,0 +1,204 @@
...
+  AgentCoreGatewayTargetSchema,
+  type AgentCoreGatewayTarget,
+  type OutboundAuth,
+} from "../../../../projectSchemas/gateway";
+import { createHandler, flag, ProjectKey } from "../../../../router";
+import { parseJsonFlagWithSchema } from "../../../utils";
+import type { Project } from "../../types";
+import type { AddProjectResourceConfig } from "../types";
+
+export const createAddGatewayTargetHandler = (config: AddProjectResourceConfig) =>
```
jariy17, line 15: If you check how the help looks for the old cli, you will notice that it divides based on target

```
Target types and their options:

  http-runtime — Route to an AgentCore runtime
    --runtime <name>               Runtime from your project
    --runtime-endpoint <name>      Endpoint / version alias (optional)

  mcp-server — Connect to an MCP-compatible server
    --endpoint <url>               Server endpoint URL
    --host <host>                  Lambda or AgentCoreRuntime
    --language <lang>              Python, TypeScript, or Other

  api-gateway — Connect to an Amazon API Gateway REST API
    --rest-api-id <id>             REST API ID
    --stage <stage>                Deployment stage

  open-api-schema / smithy-model — Auto-derive tools from a schema
    --schema <path>                Schema file path or S3 URI
    --schema-s3-account <id>       S3 bucket owner account ID

  lambda-function-arn — Connect to an AWS Lambda function
    --lambda-arn <arn>             Lambda function ARN
    --tool-schema-file <path>      Tool schema JSON file

  connector — Wire a managed AWS connector (bedrock-knowledge-bases, web-search)
    --connector <id>               bedrock-knowledge-bases or web-search
    --knowledge-base-id <id>       Project KB name or 10-char external KB id (for KB connectors)
    --exclude-domains <list>       Comma-separated domains to exclude (for web-search connector)

  passthrough — Route to an external HTTPS endpoint
    --passthrough-endpoint <url>   HTTPS endpoint URL
    --stickiness-identifier <expr> Session routing expression (optional)
    --stickiness-timeout <seconds> Sticky session timeout in seconds (optional)

  Auth (--outbound-auth <type>, --credential-name <name>) — valid types per target:
    mcp-server                     oauth or none
    lambda-function-arn            oauth or none
    http-runtime                   oauth or none
    open-api-schema                oauth or api-key (required)
    api-gateway                    api-key or none
    smithy-model                   none (uses gateway IAM role)
    passthrough                    gateway-iam-role, oauth, or jwt-passthrough
```

src/handlers/project/add/gateway-connector/index.ts
```diff
@@ -0,0 +1,127 @@
...
+  switch (connectorId) {
+    case "web-search":
+      return {
+        name,
+        targetType: "connector",
+        connectorId,
+        configurations: [{ name: "WebSearch", parameterValues: { maxResults: 10 } }],
+      };
+    case "bedrock-knowledge-bases":
+      if (!knowledgeBase) {
```
jariy17, line 115: Don't we do this validation at line 60?

src/handlers/project/add/gateway/index.ts
```diff
@@ -0,0 +1,154 @@
...
+    name: "gateway",
+    description: "adds a Gateway to the current project",
+    flags: [
+      flag("name", "the Gateway name", z.string().optional()),
+      flag(
+        "role-arn",
+        "IAM role the Gateway assumes; a default role is created when omitted",
+        z.string().optional(),
+      ),
+      flag(
```
jariy17: lets remove this flag and also omitt this protocol. The console doesn't allow customers to set this value anymore.

src/handlers/project/add/gateway/index.ts
```diff
@@ -0,0 +1,154 @@
...
+        "policy-engine-mode",
+        "Policy Engine mode: log-only or enforce",
+        z.enum(["log-only", "enforce"]).optional(),
+      ),
+      flag("exception-level", "exception detail level: debug", z.enum(["debug"]).optional()),
+      flag(
+        "tags",
```
jariy17: Check runtime and harness, we don't do SourceResolver for this.

src/handlers/project/add/gateway/index.ts
```diff
@@ -0,0 +1,154 @@
...
+        "role-arn",
+        "IAM role the Gateway assumes; a default role is created when omitted",
+        z.string().optional(),
+      ),
+      flag(
+        "protocol",
+        "restrict Target protocols to MCP; omitted allows every Target protocol",
+        z.enum(["mcp"]).optional(),
+      ),
+      flag(
```
jariy17, line 24: nit: The description should say enable semantic search to search tools on gateway

src/handlers/project/add/gateway/index.ts
```diff
@@ -0,0 +1,154 @@
...
+        z.enum(["AWS_IAM", "CUSTOM_JWT", "NONE"]).optional(),
+      ),
+      flag("description", "Gateway description", z.string().optional()),
+      flag(
+        "authorizer-configuration",
+        "project authorizerConfiguration (JSON; inline, file://<path>, or - for stdin)",
+        z.string().optional(),
+      ),
+      flag(
+        "policy-engine-name",
```
jariy17, line 41: Should we expand this for policy engines outside of this project? What is cons of doing that?

src/handlers/project/add/gateway-test-support.ts
```diff
@@ -0,0 +1,61 @@
...
+import { tmpdir } from "node:os";
+import { createRootHandler } from "../../index";
+import {
+  createSilentLogger,
+  TestCoreClient,
+  TestGlobalConfigAccessor,
+  testIO,
+} from "../../../testing";
+
+export function createGatewayProjectTestHarness(directoryPrefix: string) {
```
jariy17, line 12: combine this with index.test.ts. If no one other test file is using it, it should be combine with the test file

src/handlers/project/types.ts
```diff
@@ -84,12 +85,21 @@ export type AddResourceInput =
...
+      resourceType: "gateway-target";
+      gatewayName: string;
+      resourceConfig: AgentCoreGatewayTarget;
     };
 
 export type ProjectResource = AddResourceInput["resourceType"];
 
 export type RemoveResourceInput = {
-  resourceType: ProjectResource;
+  resourceType: Exclude<ProjectResource, "gateway" | "gateway-target">;
```
jariy17: Why are we excluding it from remove?

jariy17 (changes requested):

Hweinstock:
```

Run ./dist/bin/agentcore-linux-x64 --help
ResourceUnavailable: C:\codebuild\tmp\output\src2220336954\src\actions-runner\_work\_temp\6e195c49-dc39-49f6-be19-336eeffd6cd8.ps1:2
Line |
   2 |  ./dist/bin/agentcore-linux-x64 --help
     |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Program 'agentcore-linux-x64' failed to run: Class not registeredAt
     | C:\codebuild\tmp\output\src2220336954\src\actions-runner\_work\_temp\6e195c49-dc39-49f6-be19-336eeffd6cd8.ps1:2 char:1 + ./dist/bin/agentcore-linux-x64 --help + ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~.

```

Is this failure related to the changes here? I don't see this failure on main.

src/core/project/manager.tsx
aidandaly24: This is due to a limitation with the current CDK. The name technically only has to be unique to the gateway but the L3 CDKs require them to be globally unique. This is a change we could make on the L3 CDK but for now I was just trying to make the implementaion compatible.

aidandaly24: I'm going to inline this function

aidandaly24: Confirmed that the service only requires uniqueness within the parent Gateway. I’m keeping project-global uniqueness in this PR because the current L3 Target outputs and old CLI output parser are keyed only by Target name. I inlined the temporary validation and filed #2083 for the coordinated L3 and parser change.

src/handlers/project/add/gateway-connector/index.ts
aidandaly24: These checks cover opposite cases. The earlier check rejects `--knowledge-base` when the selected Connector is not `bedrock-knowledge-bases`. This check requires `--knowledge-base` when `bedrock-knowledge-bases` is selected. We need both directions.

src/handlers/project/add/gateway/index.ts
aidandaly24: I kept the flag but aligned it with the existing schema and old CLI as `--protocol-type MCP`. Omitting the flag persists `protocolType: "None"` for an unrestricted Gateway; passing it creates an MCP-only Gateway. We cannot omit the project field today because schema v1 and the current L3 interpret omission as MCP. I filed #2084 for the coordinated compatibility migration to change that later.

src/handlers/project/add/gateway/index.ts
aidandaly24: The current project schema models `policyEngineName` as a logical reference to a sibling `policyEngines[]` resource, which the L3 resolves to an ARN. Supporting external Policy Engines would require a new ARN representation in `agentcore.json` and corresponding L3 handling, rather than only another flag. I kept this PR aligned with the existing schema.

src/handlers/project/add/gateway-test-support.ts
aidandaly24: It is used by `gateway/index.test.ts`, `gateway-target/index.test.ts`, and `gateway-connector/index.test.ts`, so the “if no other test file is using it” condition does not apply. I left it in `gateway-test-support.ts` as shared setup.

src/handlers/project/types.ts
jariy17: I want to make sure that ur agent didn't find a issue with agentcore remove's function because this is the first project case where gateway-target is part of agentcoreGateways in agentcore.json. Please verify this

jariy17 (approved):

PR #2030, opened 2026-08-18
feat(capacity-provider): add capacity-provider resource (Journey 1)

xxuam:
## Description

  Adds the `capacity-provider` resource — a customer-managed pool of AWS-managed
  EC2 compute that agent runtimes can run on instead of the default managed fleet.

  Changes:
  - **`CapacityProviderPrimitive`** (create/remove) with CLI flags (`--name`,
    `--operator-role-arn`, `--subnets`, `--security-groups`, `--os`,
    `--instance-types`, `--volume`/`--volume-encrypted`/`--volume-kms-key`,
    `--instance-profile-arn`, `--idle-instance-timeout`, `--max-lifetime`,
    `--description`).
  - **Zod schema** for the capacity-provider spec (nested compute/VPC/volume/
    lifecycle config), added to the project spec.
  - **Interactive TUI** — add wizard and remove picker.
  - **Telemetry** — add/remove command-run instrumentation + `OperatingSystem` enum.
  - **Deploy/status plumbing** — CloudFormation output parsing, deployed-state
    schema, preflight empty-project guard (so a CP-only project isn't misread as a
    teardown), and `ResourceGraph` rendering.
  - **Docs** (AGENTS.md, README, docs/commands.md) and integration tests.

  This is the CLI half of Capacity Provider Journey 1 (create/remove a standalone
  capacity provider). The `AgentCoreCapacityProvider` L3 construct lands in a
  companion PR in aws/agentcore-l3-cdk-constructs.

  The `@aws/agentcore-cdk` pin stays at `0.1.0-alpha.45` in this PR until the L3
  construct publishes a new alpha; a follow-up commit bumps it.

  ## Related Issue

  Closes #

  ## Documentation PR

  

  ## Type of Change

  - [x] New feature

  ## Testing

  Ran the full unit suite (6128 passing) and integration suite (344 passing,
  including `integ-tests/add-remove-capacity-provider.test.ts` which exercises
  add → inspect JSON → remove against the real CLI binary). Also live-validated
  the full create → deploy → update → remove lifecycle against real AWS.

  - [x] I ran `npm run lint`

  ## Checklist

### DIFF
--- a/src/cli/commands/remove/command.tsx
+++ b/src/cli/commands/remove/command.tsx
@@ -27,6 +27,7 @@ async function handleRemoveAll(options: RemoveAllOptions): Promise<RemoveResult>
       for (const e of current.evaluators ?? []) items.push(`evaluator: ${e.name}`);
       for (const g of current.agentCoreGateways ?? []) items.push(`gateway: ${g.name}`);
       for (const pe of current.policyEngines ?? []) items.push(`policy-engine: ${pe.name}`);
+      for (const cp of current.capacityProviders ?? []) items.push(`capacity-provider: ${cp.name}`);
       return {
         success: true,
         message: items.length > 0 ? `Would remove: ${items.join(', ')}` : 'Nothing to remove',
@@ -84,6 +85,7 @@ async function handleRemoveAll(options: RemoveAllOptions): Promise<RemoveResult>
       harnesses: [],
       datasets: [],
       payments: [],
+      capacityProviders: [],
     });
 
     // Preserve aws-targets.json and deployed-state.json so that
--- a/src/cli/operations/deploy/preflight.ts
+++ b/src/cli/operations/deploy/preflight.ts
@@ -112,6 +112,7 @@ export async function validateProject(selectedTarget?: AwsDeploymentTarget): Pro
   // Check for gateways in agentcore.json
   const hasGateways = projectSpec.agentCoreGateways && projectSpec.agentCoreGateways.length > 0;
   const hasPayments = projectSpec.payments && projectSpec.payments.length > 0;
+  const hasCapacityProviders = projectSpec.capacityProviders && projectSpec.capacityProviders.length > 0;
 
   if (
     !hasAgents &&
@@ -122,7 +123,8 @@ export async function validateProject(selectedTarget?: AwsDeploymentTarget): Pro
     !hasPolicyEngines &&
     !hasHarnesses &&
     !hasDatasets &&
-    !hasPayments
+    !hasPayments &&
+    !hasCapacityProviders
   ) {
     if (!hasExistingStack) {
       throw new ValidationError(
--- a/src/cli/operations/deploy/preflight.ts
+++ b/src/cli/operations/deploy/preflight.ts
@@ -68,6 +68,43 @@ export function formatError(err: unknown): string {
 }
 
 /**
+ * Spec arrays whose presence means the project has something to deploy to CloudFormation.
+ * Keep in sync with the resource types that emit CFN outputs (see cloudformation/outputs.ts
+ * parse*Outputs). A project with none of these is empty: deploy either errors ("No resources
+ * defined") or, when a stack already exists, tears it down.
+ *
+ * This is the single source of truth for "is the project deployable" — a new deployable primitive
+ * adds its key here and every consumer stays correct. Previously this was a hand-maintained boolean
+ * chain that silently drifted (configBundles and onlineEvalConfigs were both missing from it, so a
+ * project containing only those was misclassified as empty). The `satisfies` clause makes a typo'd
+ * or renamed key a compile error.
+ */
+export const DEPLOYABLE_RESOURCE_KEYS = [
+  'runtimes',
+  'agentCoreGateways',
+  'memories',
+  'knowledgeBases',
+  'evaluators',
+  'onlineEvalConfigs',
+  'policyEngines',
+  'configBundles',
+  'datasets',
+  'capacityProviders',
+  'harnesses',
+  'payments',
+] as const satisfies readonly (keyof AgentCoreProjectSpec)[];
+
+/**
+ * True when the project defines at least one resource that deploys to CloudFormation.
+ */
+export function hasDeployableResources(spec: AgentCoreProjectSpec): boolean {
+  return DEPLOYABLE_RESOURCE_KEYS.some(key => {
+    const value = spec[key];
+    return Array.isArray(value) && value.length > 0;
+  });
+}
+
+/**
  * Validates the CDK project and loads configuration.
  * Also validates AWS credentials are configured before proceeding.
  * Returns the project context needed for subsequent steps.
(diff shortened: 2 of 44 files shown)


src/schema/schemas/primitives/capacity-provider.ts
```diff
@@ -0,0 +1,213 @@
...
+// Operator Role ARN Validation
+// ============================================================================
+
+/**
+ * Pattern for the capacity provider operator role ARN, matching the CFN
+ * resource contract exactly. The account segment is OPTIONAL — the service
+ * accepts role ARNs without an account id, so we must not force 12 digits.
+ */
+// eslint-disable-next-line security/detect-unsafe-regex -- anchored ARN pattern, no backtracking risk
+export const CAPACITY_PROVIDER_OPERATOR_ROLE_ARN_PATTERN = /^arn:aws(-[^:]+)?:iam::([0-9]{12})?:role\/.+$/;
```
tejaskash: Could we use the repository partition-agnostic ARN matcher (`arn:[^:]+:`) here and for the instance-profile and KMS patterns below? The current `aws(-[^:]+)?` form only accepts partition names beginning with `aws`, so otherwise valid configurations in a supported partition can fail schema validation. Please keep the mirrored CDK schema in sync as well.

tejaskash (commented):
Pulled this down together with the L3 PR (#328, wired in as a local tarball) and ran the full lifecycle against a real account in us-west-2: create → add capacity-provider → deploy → status → remove → teardown. Everything worked end to end — the CFN resource came up, the Id/Arn outputs landed in deployed-state, status showed it as deployed, and removing the last capacity provider correctly tore the stack down instead of tripping the empty-project check.

A few findings below, roughly in priority order. One note from the live test that isn't tied to a line: when the capacity provider fails asynchronously (e.g. the operator role is missing `events:PutRule`), CloudFormation only says `NotStabilized` — the real reason is only visible via `GetCapacityProvider.statusReason`. Not this PR's bug, but worth raising with the service team.

src/cli/operations/deploy/preflight.ts
```diff
@@ -112,6 +112,7 @@ export async function validateProject(selectedTarget?: AwsDeploymentTarget): Pro
   // Check for gateways in agentcore.json
   const hasGateways = projectSpec.agentCoreGateways && projectSpec.agentCoreGateways.length > 0;
   const hasPayments = projectSpec.payments && projectSpec.payments.length > 0;
+  const hasCapacityProviders = projectSpec.capacityProviders && projectSpec.capacityProviders.length > 0;
```
tejaskash: **P1** — This hand-maintained `hasX` chain has already drifted: `configBundles` and `onlineEvalConfigs` are missing from it today. A project containing only config bundles (which are CFN-deployed — deploy parses `parseConfigBundleOutputs`) falls into the empty branch and either throws "No resources defined" or, worse, gets silently treated as a teardown deploy when a stack already exists.

Adding an 11th flag here keeps the trap armed for the next resource type. Could we replace the chain with a typed `hasDeployableResources(spec)` helper that lists the deployable spec arrays in one place? Then the next primitive can't forget this file.

src/cli/primitives/CapacityProviderPrimitive.ts
```diff
@@ -0,0 +1,271 @@
...
+  }
+
+  /**
+   * Build a validated CapacityProvider config from CLI options.
+   * Zod validation (via CapacityProviderSchema.parse) rejects bad input here,
+   * at `add` time, rather than late at deploy/CFN time.
+   */
+  private buildCapacityProvider(options: AddCapacityProviderOptions): CapacityProvider {
+    const volumes = (options.volume ?? []).map(entry => {
+      const [volName, sizeRaw] = entry.split(':');
```
tejaskash: **P2** — This parse quietly accepts malformed input instead of rejecting it:

- `--volume data:20:gp3` — the destructuring drops everything after the second colon, so a user guessing a volume-type suffix silently gets a 20 GiB volume with no error.
- `--volume data:0x14` and `--volume data:2e1` — `Number()` happily parses hex and exponent notation, so both are silently reinterpreted as `sizeGiB: 20`.

Suggest splitting once, requiring exactly two segments, and validating the size with `/^[0-9]+$/` so anything that isn't literally `name:sizeGiB` hits the existing "Expected format name:sizeGiB" error.

src/schema/schemas/primitives/capacity-provider.ts
tejaskash: **P2** — This pins the `aws` partition prefix, which AGENTS.md explicitly rules out: "ARN regex patterns must use `arn:[^:]+:` (not `arn:aws:`) to match any partition." Same pattern shape at lines 108 and ~141 (instance profile, KMS key).

It happens to match all three current partitions, so nothing breaks today — but it's exactly the drift the rule exists to prevent, and since this file is duplicated verbatim into the L3 repo, the fix needs to land in both copies.

src/cli/commands/remove/command.tsx
```diff
@@ -27,6 +27,7 @@ async function handleRemoveAll(options: RemoveAllOptions): Promise<RemoveResult>
       for (const e of current.evaluators ?? []) items.push(`evaluator: ${e.name}`);
       for (const g of current.agentCoreGateways ?? []) items.push(`gateway: ${g.name}`);
       for (const pe of current.policyEngines ?? []) items.push(`policy-engine: ${pe.name}`);
+      for (const cp of current.capacityProviders ?? []) items.push(`capacity-provider: ${cp.name}`);
```
tejaskash, line 30: **P3** — The CLI dry-run enumeration now lists capacity providers, but the TUI remove-all confirmation (`useRemoveFlow.ts`) still doesn't — so a TUI user resetting a project full of capacity providers confirms a destructive action whose summary never mentions them. The "All schemas" wording softens it, but since this PR updated one of the two paths, it'd be good to update both so they enumerate consistently.

src/cli/tui/screens/add/AddScreen.tsx
```diff
@@ -39,6 +40,11 @@ const BASE_ADD_RESOURCES: { id: AddResourceType; title: string; description: str
     title: 'Payment Connector [preview]',
     description: 'Link payment provider credentials to a manager',
   },
+  {
+    id: 'capacity-provider',
+    title: 'Capacity Provider',
```
tejaskash: **P3** — The add screen says "Capacity Provider" but RemoveScreen says "Capacity Provider [preview]". Same resource, same release, two different maturity signals — worth picking one.

src/schema/schemas/primitives/capacity-provider.ts
xxuam: switched all patterns to the partition-agnostic

src/cli/operations/deploy/preflight.ts
xxuam: Fixed. Replaced the hand-maintained boolean chain with a typed hasDeployableResources(spec) backed by a single DEPLOYABLE_RESOURCE_KEYS list (as const satisfies readonly (keyof AgentCoreProjectSpec)[], so a  typo'd/renamed key is a compile error). This also fixes the latent bug flagged — configBundles and onlineEvalConfigs were both missing, so a project containing only those was misclassified.

src/cli/primitives/CapacityProviderPrimitive.ts
xxuam: Fixed

src/schema/schemas/primitives/capacity-provider.ts
xxuam: Fixed

src/cli/tui/screens/add/AddScreen.tsx
xxuam: Fixed

src/cli/commands/remove/command.tsx
xxuam: Fixed — useRemoveFlow.ts now lists capacity providers in the remove-all confirmation, matching the CLI dry-run enumeration.

tejaskash (approved):
LGTM

PR #2031, opened 2026-08-18
fix(project): simplify agentcore spec serialization errors

Hweinstock:
## Problem 
This PR solves two related problems with project spec serialization. 
- the agentcore spec is not validated before writing out. The harness piece is already validated as a side-effect of scaffolding, but we should validate the whole config. 
- `DeserializationError` does not extend the base error. (think it preceded it)

## Solution
- validate entire agentcore spec before writing out. 
- make deserialization error a user error to avoid rewrapping. 

## Testing 
added a unit test to verify the case by corrupting the agentcore.json and validating we fail early.

### DIFF
--- a/src/core/project/manager.tsx
+++ b/src/core/project/manager.tsx
@@ -24,7 +24,6 @@ import { ProjectSpecSchema } from "../../projectSchemas/project";
 import { enclosingProjectRoot } from "./fsUtils";
 import {
   AgentCoreCLIError,
-  DeserializationError,
   InputValidationError,
   NotImplementedError,
   ProjectStateError,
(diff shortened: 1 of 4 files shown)


src/core/project/manager.tsx
```diff
@@ -63,22 +62,12 @@ export class FsProjectManager implements ProjectManager {
...
-        });
-      }
-      throw error;
-    }
+    const spec = await this.json.read(configPath, ProjectSpecSchema);
+    return {
+      name: spec.name,
+      rootPath,
+      spec,
+    };
```
aidandaly24, line 70: Verrry nice👌

aidandaly24 (approved):
This looks good to me

notgitika (approved):
Thanks LGTM!

PR #2032, opened 2026-08-18
feat(eval): batch-evaluation simulate — each example owns its invoker

jariy17:
## What

Adds `batch-evaluation simulate` to replay a dataset against a runtime and grade the resulting sessions.

> **Each dataset example is a self-running object that owns its own invoker (`RunContext`). `invokeDataset` builds one invoker per session and passes it into the example's `run()`. The example performs the invocation; the machine supplies the invoker and collects the result.**

## The invoker seam

A dataset example knows what to send and how to turn the responses into ground truth, but not how to reach the runtime. Its `RunContext` owns that runtime-specific invocation:

```ts
// src/core/eval/invokeDataset/example/types.ts
export type TurnResult = { text: string };
export type RunContext = { invokeOnce(input: string): Promise<TurnResult> };

export interface Example {
  readonly exampleId: string;
  readonly schemaType: DatasetSchemaType;
  // Replay this example against the runtime `ctx` reaches, return neutral ground truth.
  run(ctx: RunContext): Promise<InlineGroundTruth | undefined>;
}
```

`invokeDataset` resolves the runtime once, then builds a distinct invoker for each client-generated session and passes it to that example:

```ts
const results = await runExamples(examples, async (example) => {
  const ctx: RunContext = {
    invokeOnce: async (input) => {
      const res = await invokeRuntime(
        deps,
        { /* resolved runtime, session, and rendered payload */ },
        options,
        signal,
      );
      return { text: /* drained response body */ };
    },
  };

  return example.run(ctx);
});
```

The example decides how many times to call `invokeOnce` and in what order. Its invoker decides how each call reaches the runtime for that session.

## Layout

```text
src/core/eval/invokeDataset/
├── load.ts                 DatasetLoader — pure JSONL parse → shape-classify → new
├── run.ts                  runExamples — bounded-concurrency pool, failure isolation
├── invokeDataset.test.ts   end-to-end golden coverage of the whole path
└── example/
    ├── types.ts            Example interface, RunContext (the invoker), TurnResult
    ├── predefined.ts       PredefinedExample — replays scripted turns, builds ground truth
    └── simulated.ts        SimulatedExample — not shipped; throws at construction
```

Plus `src/core/invokeRuntime.ts`, which extracts runtime invocation from `runtime.tsx` for reuse by both `RuntimeClient` and `invokeDataset`, and `src/handlers/eval/batch-evaluation/simulate/index.tsx`, which composes `invokeDataset` with `startBatchEvaluation`.

## Testing

- `bun run typecheck` passes.
- `bun test`: **1544 pass, 0 fail**.
- `invokeDataset.test.ts` drives the real `EvalClient.invokeDataset` over a fake AWS layer with golden fixtures. Its snapshot covers created sessions and inline ground truth across every supported variation while exercising the loader, example classes, concurrency pool, template rendering, and runtime invocation.
- `simulate.test.tsx` snapshots the handler's wrapped `sessionMetadata`.


src/io/template.ts
```diff
@@ -0,0 +1,33 @@
+import { InputValidationError } from "../errors";
```
Hweinstock: should we keep this inside evals until there is an opportunity to re-use it? I feel like the `io` directory should be reserved for shared abstractions.

src/io/template.ts
```diff
@@ -0,0 +1,33 @@
...
+export function renderJsonTemplate(
+  template: string,
+  values: Record<string, string>,
+  flagName = "payload-template",
+): Uint8Array {
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(template);
+  } catch {
+    throw new InputValidationError(`--${flagName} must be valid JSON`);
```
Hweinstock: should we wire the cause here?

src/handlers/eval/types.tsx
```diff
@@ -205,6 +206,39 @@ export type StartBatchEvaluationInput = {
...
+// InvokeDatasetInput is the runtime-level shape for replaying a dataset: invoke each
+// example against the runtime, one client-generated session per example. Runtime fields
+// only — no evaluator/name/kms (those belong to the grader the handler composes on top,
+// e.g. startBatchEvaluation). Invoke fields mirror `runtime invoke`.
+export type InvokeDatasetInput = {
+  runtimeId: string;
+  qualifier?: string;
+  payloadTemplate: string; // e.g. {"prompt":"{input}"} — {input} is the example's turn input
+  headers?: [string, string][];
+  bearerToken?: string;
```
Hweinstock, line 229: q: is there a reason bearerToken is treated different from other headers?

src/handlers/eval/batch-evaluation/simulate/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+        "bearer-token",
+        "CUSTOM_JWT bearer token (for JWT-auth runtimes)",
+        z.string().optional(),
+      ),
+      flag("user-id", "runtime user id", z.string().optional()),
+      flag("dataset", "dataset source: local JSONL path or a dataset id", z.string().optional()),
+      flag("dataset-version", "dataset version (with a dataset id)", z.string().optional()),
+      flag("evaluator", "evaluator id(s) to apply", z.array(z.string()).optional()),
+      flag("name", "batch evaluation name (unique in the account)", z.string().optional()),
+      flag("description", "optional description", z.string().optional()),
```
Hweinstock: what exactly is this describing? the simulation itself?

src/handlers/eval/batch-evaluation/simulate/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+      if (!flags["evaluator"]?.length) {
+        throw new InputValidationError(
+          "required option '--evaluator <evaluator...>' not specified",
+        );
+      }
+      if (!flags["name"])
+        throw new InputValidationError("required option '--name <name>' not specified");
+
+      // Ctrl-C aborts the run (invokes, the ingestion wait, the dataset download).
+      const controller = new AbortController();
```
Hweinstock, line 56: ahhh, theres a shared abstraction for this, but looks like it hasn't been merged yet :( 

https://github.com/aws/agentcore-cli/pull/1986

maybe we can swap it over as a follow-up once its merged.

src/handlers/eval/batch-evaluation/simulate/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+        );
+
+        ctx.require(JsonRendererKey).renderJson({
+          batchEvaluationId: job.batchEvaluationId,
+          status: job.status,
+          examplesInvoked: r.invoked,
+          examplesFailed: r.failed,
+        });
+      } catch (error) {
+        // A Ctrl-C exits quietly; the half-created sessions grade nothing.
```
Hweinstock: don't we want this to bubble up if the user cancelled? I think this could be useful data in telemetry.

src/handlers/eval/batch-evaluation/simulate/simulate.test.tsx
```diff
@@ -0,0 +1,182 @@
...
+        "--evaluator",
+        "E",
+      ],
+      /--name/,
+    ],
+  ])("rejects missing required flag", async (args, expected) => {
+    await expect(run(["eval", "batch-evaluation", "simulate", ...args])).rejects.toThrow(expected);
+  });
+
+  test("passes runtime-level flags to invokeDataset (no evaluator/name leak)", async () => {
```
Hweinstock: could this be enforced at the type level with a stricter input type?

src/core/runtime.tsx
```diff
@@ -11,178 +10,37 @@ import {
...
-        qualifier: request.qualifier,
-        region: options.region,
-      });
-      return this.invokeRuntimeWithCustomJwt(request, bearerToken, options, logger, signal);
-    }
-    return this.invokeRuntimeWithIam(request, options, signal);
-  }
-
-  private async invokeRuntimeWithCustomJwt(
+  // invokeRuntime delegates to the free function so EvalClient.invokeDataset can reuse
```
Hweinstock: this comment feels out of place (and may likely become stale if consumption patterns change).

src/core/invokeRuntime.ts
```diff
@@ -0,0 +1,221 @@
+import { randomUUID } from "node:crypto";
+import { InvokeAgentRuntimeCommand } from "@aws-sdk/client-bedrock-agentcore";
+import { InputValidationError } from "../errors";
+import type { Logger } from "../logging";
+import type { AwsClients, CoreFetch, CoreOptions } from "./types";
+import { abortable } from "./abortable";
+import { toClientConfig } from "./utils";
+
+// Core's own invoke DTOs — duplicated from, not imported out of, the runtime handler so
```
Hweinstock: Is this comment stale? I thought we pulled the common functionality out instead of duplicating.

src/core/invokeRuntime.ts
```diff
@@ -0,0 +1,221 @@
...
+  signal?: AbortSignal,
+): Promise<RuntimeInvokeResponse> {
+  const client = deps.clients.data(toClientConfig(options));
+  const endpoint = client.config.endpointProvider({
+    Region: options.region,
+    Endpoint: options.endpointUrl,
+  });
+  const url = new URL(endpoint.url);
+  if (url.protocol !== "https:") {
+    throw new TypeError("CUSTOM_JWT requires an HTTPS endpoint");
```
Hweinstock: OOS here, but we should update the errors in the file to use errors that extend our base class.

src/core/eval.tsx
```diff
@@ -507,6 +515,113 @@ export class EvalClient implements CoreEvalClient {
...
+    try {
+      return await readLocalDatasetFile(path, signal);
+    } finally {
+      await unlink(path).catch(() => {});
+    }
+  }
+
+  // downloadDatasetToTemp streams a dataset version's JSONL to a temp file so
+  // readDatasetText can read it — reuses downloadDataset rather than re-fetching.
+  private async downloadDatasetToTemp(
```
Hweinstock, line 641: would it be simpler to inline this above? The extra step adds some unnecessary complexity imo.

src/core/eval.tsx
```diff
@@ -507,6 +515,113 @@ export class EvalClient implements CoreEvalClient {
     };
   }
 
+  async invokeDataset(
+    input: InvokeDatasetInput,
+    options: CoreOptions,
+    signal?: AbortSignal,
```
Hweinstock, line 540: maybe OOS here, but should we move the signal inside options? I would think all core clients would care about cancellations.

src/core/eval.tsx
```diff
@@ -507,6 +515,113 @@ export class EvalClient implements CoreEvalClient {
...
+          for await (const chunk of response.body) text += decoder.decode(chunk, { stream: true });
+          text += decoder.decode();
+          return { text };
+        },
+      };
+      try {
+        const groundTruth = await example.run(ctx);
+        return { exampleId: example.exampleId, sessionId, groundTruth };
+      } catch (error) {
+        this.logger.debug(
```
Hweinstock: instead of logging and rethrowing, is there a way to enrich the error thrown to avoid noise?

src/core/eval.tsx
```diff
@@ -507,6 +515,113 @@ export class EvalClient implements CoreEvalClient {
...
+      }
+    });
+
+    if (failed > 0) {
+      this.logger.warn(`invokeDataset: ${failed} example(s) failed to invoke and were dropped`);
+    }
+
+    // AgentCore emits spans ~30s-3min after invoke; grade too early and it reads an empty
+    // log group and fails every session. Disabled via SIMULATE_INGESTION_WAIT_MS=0 (tests).
+    const waitMs = Number(process.env.SIMULATE_INGESTION_WAIT_MS ?? 180_000);
```
Hweinstock, line 611: is there anything we can poll on instead of a static wait time?

src/core/eval/invokeDataset/invokeDataset.test.ts
```diff
@@ -0,0 +1,235 @@
...
+const RUNTIME_ARN = "arn:aws:bedrock-agentcore:us-west-2:123456789012:runtime/rt-1";
+const row = (o: object) => JSON.stringify(o);
+
+async function* replyBytes(text: string): AsyncGenerator<Uint8Array> {
+  yield new TextEncoder().encode(text);
+}
+
+// A fake AWS layer: control resolves the runtime, data answers each invoke. Records every
+// payload it was asked to send, and per `opts` can fail or delay specific invokes.
+function fakeClients(opts: { fail?: (payload: string) => boolean; delayMs?: number } = {}): {
```
Hweinstock, line 29: this feels like a really complex testing setup. Is there a simpler way? I'm wondering if there's a structural change we could make to simplify here.

src/core/eval/invokeDataset/load.ts
```diff
@@ -0,0 +1,69 @@
...
+  static load(text: string): Example[] {
+    const examples: Example[] = [];
+    const seen = new Set<string>();
+    for (const line of text.split("\n")) {
+      const trimmed = line.trim();
+      if (!trimmed) continue;
+      let parsed: unknown;
+      try {
+        parsed = JSON.parse(trimmed);
+      } catch {
```
Hweinstock: should we wire the cause here?

src/core/eval/invokeDataset/example/types.ts
```diff
@@ -0,0 +1,17 @@
+import type { DatasetSchemaType } from "@aws-sdk/client-bedrock-agentcore-control";
+import type { InlineGroundTruth } from "@aws-sdk/client-bedrock-agentcore";
+
+// A record, not a bare string, so a future tool-branching type can widen it by a field.
```
Hweinstock: i feel like the code explains this comment.

Hweinstock (commented):
didn't get to all the examples work yet, but had a few small comments and a question on how we can simplify testing, because it feels pretty gnarly rn.

src/io/template.ts
jariy17: Yeah, I was thinking the invoke handler could use this but lets leave it in the simulate handler for now.

src/handlers/eval/batch-evaluation/simulate/index.tsx
jariy17: This should be batch-evaluation description. I'll rename this `batch-eval-desc`, same with `name` too.

src/core/eval.tsx
jariy17: Yeah i would think so too, I'll add it as a follow up PR.

src/core/eval.tsx
jariy17: I didn't poll right away because some traces might still be arriving, and I didn't want us to end up with incomplete session data.

src/core/eval.tsx
jariy17: sure we can

src/core/invokeRuntime.ts
jariy17: I'll update it now.

src/core/runtime.tsx
jariy17: I'll remove it

src/core/invokeRuntime.ts
jariy17: We did pull out the invoke Function but I did not pull it the original interface. This comment is explaining why I duplicated the runtime client's interface.

src/handlers/eval/batch-evaluation/simulate/index.tsx
jariy17: Sure that make sense.

src/handlers/eval/batch-evaluation/simulate/index.tsx
jariy17: Sure thing

src/handlers/eval/batch-evaluation/simulate/simulate.test.tsx
jariy17: Yeah i don't know why we did this test. It should be caught with proper golden tests.

src/handlers/eval/types.tsx
jariy17: It's the discriminator that selects the auth path, not an application header. Its presence routes the invoke to CUSTOM_JWT path (src/core/invokeRuntime.ts:73)

src/core/eval/invokeDataset/load.ts
jariy17: yes

src/core/eval/invokeDataset/example/types.ts
jariy17: removing

src/core/eval/invokeDataset/invokeDataset.test.ts
jariy17: Can I fixed these tests as a follow up?

src/handlers/eval/batch-evaluation/simulate/index.tsx
Hweinstock: looks like it was merged, so we should be able to leverage that shared abstraction now. 
https://github.com/aws/agentcore-cli/pull/1986

src/handlers/eval/batch-evaluation/simulate/index.tsx
```diff
@@ -0,0 +1,115 @@
+import z from "zod";
+import { createHandler, flag } from "../../../../router";
+import { InputValidationError } from "../../../../errors";
+import { JsonRendererKey } from "../../../../tui";
+import type { AppIO } from "../../../../io";
+import type { Core } from "../../../types";
+import { coreOptsFromCtx } from "../../../utils";
+import { parseRuntimeInvokeHeaders } from "../../../runtime/invoke/request";
+
+// Composes invokeDataset (replay) → startBatchEvaluation (grade). Invoke flags mirror
```
Hweinstock, line 10: is the comment describing the implementation of the handler? What value does it provide over the code?

src/core/eval.tsx
```diff
@@ -507,6 +516,121 @@ export class EvalClient implements CoreEvalClient {
...
+        },
+      };
+      try {
+        const groundTruth = await example.run(ctx);
+        return { exampleId: example.exampleId, sessionId, groundTruth };
+      } catch (error) {
+        // Enrich with the example identity so the dropped-invoke reason is self-describing
+        // in firstError, instead of a bare transport message logged separately.
+        const cause = error instanceof Error ? error : new Error(String(error));
+        throw new Error(
```
Hweinstock, line 598: to confirm my understanding, this error never hits telemetry because an individual example failing does not bubble up so this error is purely for the user message it creates?

src/core/eval.tsx
Hweinstock: how do we know that there aren't traces still coming in after 3 min mark?

src/core/invokeRuntime.ts
```diff
@@ -0,0 +1,221 @@
...
+  url.pathname = `${url.pathname.replace(/\/?$/, "/")}runtimes/${encodeURIComponent(request.runtimeId)}/invocations`;
+  url.search = new URLSearchParams({
+    accountId: request.accountId,
+    qualifier: request.qualifier,
+  }).toString();
+  const headers = new Headers(request.applicationHeaders);
+  try {
+    headers.set("Authorization", `Bearer ${bearerToken}`);
+  } catch {
+    throw new InputValidationError("Invalid bearer token");
```
Hweinstock, line 112: should we wire the cause up to these errors to get the stack trace? (same with 132, 155 below)

src/core/invokeRuntime.ts
```diff
@@ -0,0 +1,221 @@
...
+      headers,
+      body: request.payload as RequestInit["body"],
+      signal,
+    });
+  } catch (error) {
+    if (signal?.aborted) throw signal.reason ?? error;
+    logger
+      .child({
+        errorName:
+          error instanceof TypeError
```
Hweinstock, line 148: is there a reason `error instanceof error ? error.name : 'error'` doesn't work here?

src/handlers/eval/types.tsx
```diff
@@ -205,6 +206,39 @@ export type StartBatchEvaluationInput = {
...
+// ground truth. Grader-agnostic — the batch handler wraps `groundTruth` as
+// SessionMetadataShape; a future ondemand handler adapts it to EvaluationReferenceInput.
+export type InvokedSession = {
+  exampleId: string;
+  sessionId: string;
+  groundTruth?: InlineGroundTruth;
+};
+
+// InvokeDatasetResult reports the created sessions plus how many examples were invoked
+// vs dropped (a failed invoke is skipped, not fatal). firstError explains a total failure.
```
Hweinstock, line 245: does firstError explain a total failure? I thought it stored the first failed example's error?

src/core/eval/invokeDataset/invokeDataset.test.ts
```diff
@@ -0,0 +1,263 @@
+// Disables the post-invoke span-ingestion wait so the replay returns immediately.
+process.env.SIMULATE_INGESTION_WAIT_MS = "0";
```
Hweinstock, line 2: would this be a good application of global config? this might help this behavior be more discoverable by other tests that might need the same.

src/core/eval/invokeDataset/example/predefined.ts
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  // Sequential and awaited: the turns share one session, so racing them would interleave
+  // the conversation and misalign per-turn traces with the ground truth.
+  async run(ctx: RunContext): Promise<InlineGroundTruth | undefined> {
+    for (const turn of this.turns) await ctx.invokeOnce(turn.input);
+    return this.groundTruth();
+  }
+
+  // Emit every turn (carrying its prompt), not just those with an expectation: filtering
```
Hweinstock, line 58: i'm not quite sure what this comment is saying?

Hweinstock (commented):
only blocking question is why we do an explicit wait, rest could be polished later/in a follow-up. Is there a way we can be sure the spans we need exist before evaluating or is there a reason we don't have to be?

src/core/eval/invokeDataset/example/predefined.ts
jariy17: Ill remove it.

src/core/eval/invokeDataset/invokeDataset.test.ts
jariy17: So add to the bunfig.toml to preload a script that run this: 
./global-mocks.ts
```typescript
// Global mocks
import { mock } from "bun:test";

// Mock environment variables
process.env.SIMULATE_INGESTION_WAIT_MS = "0";
```

bunfig.toml
```typescript
[test]
preload = ["./global-mocks.ts"]
```
[ref](https://bun.com/docs/test/configuration)

Yes I could do that in the follow up Pr.

src/core/eval.tsx
jariy17: Correct. An individual example failing is caught and dropped by runExamples rather than rethrown, so it never reaches telemetry — it only feeds the user-facing message on a total failure. That was intentional: a run can be hundreds of examples, and I didn't want per-example failures flooding our telemetry.

src/core/invokeRuntime.ts
jariy17: Sure thing

src/handlers/eval/types.tsx
jariy17: WE will remove it later. Yes lets do a follow up on this.

src/core/invokeRuntime.ts
jariy17: I'll look into this in the next PR.

src/core/eval/invokeDataset/invokeDataset.test.ts
Hweinstock: we should be able to inject the global config down through core with the data we want.

Hweinstock:
> Is there a way we can be sure the spans we need exist before evaluating or is there a reason we don't have to be?

from chatting offline, sounds like the wait approach is shared by the eval time. Still unsure how they can be confident the evaluator results are not incomplete, but approving since its an existing solution.

Hweinstock (dismissed):

src/core/eval/invokeDataset/invokeDataset.test.ts
jariy17: ah that make sense

Hweinstock (approved):

src/handlers/eval/batch-evaluation/simulate/index.tsx
nborges-aws: +1

src/core/eval.tsx
```diff
@@ -525,6 +534,121 @@ export class EvalClient implements CoreEvalClient {
...
+    const deps = { clients: this.clients, fetch: this.fetch, logger: this.logger };
+    const accountId = accountIdFromRuntimeArn(runtime.agentRuntimeArn);
+
+    const { ok, failed, firstError } = await runExamples(examples, async (example) => {
+      // One session per example; the id is a client-owned input per the AgentCore docs,
+      // reused across turns so the conversation and its per-turn traces stay in order.
+      const sessionId = randomUUID();
+      const ctx: RunContext = {
+        invokeOnce: async (payload) => {
+          const response = await invokeRuntime(
```
nborges-aws, line 566: Comparing this to `src/handlers/runtime/invoke/request.ts`, and I notice that the code calls `normalizeRuntimeInvokeRequest` to do some validation before invoking. Do you think we need something similar here?

nborges-aws (approved):
LGTM. One comment with a question for you

PR #2034, opened 2026-08-19
refactor(add): swap harness to accept project schema directly instead of api shape. 

Hweinstock:
## Problem
The existing convention of mirroring API shapes has some issues. 
- evaluators experience diverges significantly from the APIs. (ex. [this](https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_StartBatchEvaluation.html) single API is responsible for multiple resources in the project). 
- for customers looking at their agentcore.json, its not obvious what flags to pass. 
- complexity in rejecting unsupported API fields. 
- does not extend well to supporting `update`/`edit` in the future. 

## Solution
swap to mirroring the schemas directly. This approach is simpler (removes ~300 lines of mapping code from harness) and will extend better to agentcore project resources that have diverged from the API. 

mirroring the schemas with json blobs also reduces the amount of flags significantly. 

The PR also ports the harness tests to their own file. 

## Testing 
Existing Unit Tests ported over. 

Helping screen becomes:
```
Usage: agentcore project add harness [options]

adds a harness to the current project

Options:
  --name <name>                                          the name of the harness
  --execution-role-arn <execution-role-arn>              IAM role the harness assumes; a default role is created when omitted
  --system-prompt <system-prompt>                        the agent's system prompt
  --model <model>                                        model configuration (JSON)
  --tools <tools>                                        tools available to the agent (JSON)
  --skills <skills>                                      skills available to the agent (JSON)
  --allowed-tools <allowed-tools...>                     tool allowlist patterns (e.g. * or @serverName/toolName)
  --memory <memory>                                      memory configuration (JSON)
  --truncation <truncation>                              context truncation configuration (JSON)
  --network-mode <network-mode>                          network mode for the harness environment (PUBLIC or VPC)
  --network-config <network-config>                      VPC network configuration (JSON)
  --lifecycle-config <lifecycle-config>                  lifecycle configuration (JSON)
  --session-storage-path <session-storage-path>          mount path for session storage
  --efs-access-points <efs-access-points>                EFS access point configurations (JSON)
  --s3-access-points <s3-access-points>                  S3 access point configurations (JSON)
  --environment-variables <environment-variables>        environment variables (JSON object of key/value strings)
  --container-uri <container-uri>                        ECR container image URI
  --authorizer-type <authorizer-type>                    inbound authorizer type (AWS_IAM or CUSTOM_JWT)
  --authorizer-configuration <authorizer-configuration>  inbound authorizer configuration (JSON)
  --max-iterations <max-iterations>                      max agent loop iterations per invocation
  --max-tokens <max-tokens>                              max total output tokens per invocation
  --timeout-seconds <timeout-seconds>                    max duration in seconds per invocation
  --tags <tags>                                          tags to apply (JSON object of key/value strings)
  --dockerfile <dockerfile>                              path to local dockerfile to use as the container image for the harness
  -h, --help                                             display help for command
```

## Notes
This does introduce a discoverability gap where customers need to know our project schemas, which means we'll need solid documentation here.


src/handlers/project/project.test.ts
```diff
@@ -115,580 +112,6 @@ describe("project create", () => {
   });
 });
 
-describe("project add harness", () => {
```
jariy17: Why did you get rid of this test?

src/handlers/project/add/harness/index.ts
```diff
@@ -84,406 +62,56 @@ export const createAddHarnessHandler = (config: AddProjectResourceConfig) =>
...
-      }
-
-      const harnessConfig = {
+      const harnessInput = {
         name: flags.name,
-        model: inputModelConfig
-          ? toModelConfig(inputModelConfig)
-          : { provider: "bedrock" as const, modelId: "global.anthropic.claude-sonnet-4-6" },
+        model: parseJsonFlag("model", flags["model"]) ?? {
+          provider: "bedrock" as const,
```
jariy17: nit: Make it into a CONST

src/handlers/project/add/harness/index.ts
```diff
@@ -84,406 +62,56 @@ export const createAddHarnessHandler = (config: AddProjectResourceConfig) =>
...
-        networkMode: env?.networkMode,
-        networkConfig: env?.networkConfig
-          ? { ...env.networkConfig, ...(inputVpcId ? { vpcId: inputVpcId } : {}) }
-          : undefined,
-        lifecycleConfig: env?.lifecycleConfig,
-        sessionStoragePath: env?.sessionStoragePath,
-        efsAccessPoints: env?.efsAccessPoints,
-        s3AccessPoints: env?.s3AccessPoints,
-        containerUri: artifact?.containerUri,
+        tags: parseJsonFlag("tags", flags["tags"]),
```
jariy17: 3. Tags differ from the shared convention

  This command currently accepts only:
```
  --tags '{"team":"ml","env":"prod"}'
```

  Other refactor-branch commands use parseTags, allowing both:
```
  --tags team=ml env=prod
  --tags '{"team":"ml","env":"prod"}'
```

  They declare the flag as z.array(z.string()).optional() and convert the values into Record<string, string> with parseTags.

Should we support both? Even though the flags don't match in the imperative one, we should keep the tag conventions the same between both.

jariy17 (dismissed):

src/handlers/project/add/harness/index.ts
Hweinstock: I think to be consistent, we should swap it.

src/handlers/project/add/harness/index.ts
```diff
@@ -37,453 +22,98 @@ export const createAddHarnessHandler = (config: AddProjectResourceConfig) =>
...
         "authorizer-configuration",
-        "inbound authorizer configuration (JSON AuthorizerConfiguration)",
+        "inbound authorizer configuration (JSON)",
         z.string().optional(),
       ),
       flag("max-iterations", "max agent loop iterations per invocation", z.number().optional()),
       flag("max-tokens", "max total output tokens per invocation", z.number().optional()),
       flag("timeout-seconds", "max duration in seconds per invocation", z.number().optional()),
-      flag("tags", "tags to apply (JSON object of key/value strings)", z.string().optional()),
+      flag("tags", "tags as key=value (repeatable) or JSON object", z.array(z.string()).optional()),
```
notgitika, line 64: Could we add `--connections` here? HarnessSpecSchema supports it but this handler currently doesn't

notgitika (dismissed):
OOS for here but when I was playing around with it I noticed it didn't have a  parameterized help like we see in `harness create --help`. The JSON-valued flags are otherwise difficult to construct. 

the help should document the harness spec shapes specifically, since they differ from the API shapes used by harness create.

can we discuss having that as a future improvement?

src/handlers/project/add/harness/index.ts
Hweinstock: AFAICT this isn't used in the old CLI so I left it out.

jariy17 (approved):

notgitika (approved):

PR #2035, opened 2026-08-19
feat(project): scaffold add runtime handler

Hweinstock:
### Problem 
`add runtime` is missing. 

### Solution 
- add support for `project add runtime` mostly mirroring the schema shapes. 
- support two modes `template` and `byo` (import OOS here). We then use this input schema to scaffold the project, and resolve the config that goes into agentcore.json (to be implemented). 
- new `template` flag describes templates like `strands-python`, which can be adjusted with flags (ex. `--build Container`, or `--protocol mcp`, with defaults for each template. This will keep us flexible to support more templates in the future. 

### Testing 
- added unit tests for handler parsing, with implementation details to be filled in.


src/handlers/project/add/runtime/index.ts
```diff
@@ -0,0 +1,237 @@
...
+        "role-arn",
+        "IAM role ARN that provides permissions for the runtime",
+        z.string().optional(),
+      ),
+      flag("code-location", "path to existing agent source code (BYO path)", z.string().optional()),
+      flag("build", "build type: CodeZip or Container", BuildTypeSchema.optional()),
+      flag("entrypoint", "entrypoint file, e.g. main.py:handler (BYO only)", z.string().optional()),
+      flag("protocol", "server protocol: HTTP, MCP, A2A, AGUI", ProtocolModeSchema.optional()),
+      flag(
+        "api-key",
```
nborges-aws, line 40: We should mark this as sensitive

src/handlers/project/add/runtime/index.ts
```diff
@@ -0,0 +1,237 @@
...
+      if (isTemplate && byoOnlyFlags.length > 0)
+        throw new InputValidationError(
+          `--${byoOnlyFlags[0]} is only available on the BYO path (--code-location)`,
+        );
+      if (!isTemplate && templateOnlyFlags.length > 0)
+        throw new InputValidationError(
+          `--${templateOnlyFlags[0]} is only available on the template path (--template)`,
+        );
+
+      const inputEnvironmentVariables = parseJsonFlag<Record<string, string>>(
```
nborges-aws, line 145: Might be worth changing this to `parseJsonFlagWithSchema`, and passing in `z.record(EnvVarNameSchema, z.string())`. That way we aren't accepting null, 0, etc. into env vars

src/handlers/project/add/runtime/types.ts
```diff
@@ -0,0 +1,67 @@
...
+    strategies: z.array(z.enum(Object.values(MemoryStrategyType))),
+  }),
+  z.object({
+    mode: z.literal("existing"),
+    arn: z.string().min(1),
+  }),
+]);
+export type RuntimeMemoryConfig = z.input<typeof runtimeMemoryConfigSchema>;
+
+const RuntimeInfraConfigSchema = z.object({
```
nborges-aws, line 28: qq: I see `ProjectRuntimeSchema` has a bunch of superRefines validating/invalidating flags that need to be used in combination (etc. ensures VPC always has a network config). Is that something we'll account for in the implementation PR or do we need those here?

src/handlers/project/add/runtime/index.ts
Hweinstock: good call!

src/handlers/project/add/runtime/index.ts
Hweinstock: I believe this should get validated on line 209 below, since we borrow the env var shape from the existing schema.

src/handlers/project/add/runtime/types.ts
Hweinstock: we should still surface these errors to the customer since we validate the schema before write, but this made me realize a small bug where this wouldn't rollback the scaffolded files. Fixing that here.

nborges-aws: Nice catch on the rollback. With that fix I have no concerns about the validation happening later in the chain

nborges-aws (approved):

src/handlers/project/add/runtime/index.ts
```diff
@@ -0,0 +1,238 @@
...
+} from "../../../../projectSchemas/constants";
+import { SourceResolver } from "../../../../io";
+import { RUNTIME_TEMPLATES } from "../../types";
+import {
+  runtimeMemoryConfigSchema,
+  runtimeModelProviderSchema,
+  RuntimeResourceConfigSchema,
+} from "./types";
+
+export const createAddRuntimeHandler = (config: AddProjectResourceConfig) =>
```
notgitika, line 21: OOS but a suggestion that we should try to adopt across the board. Can we divide flags into "basic" and "advanced" so that it is decluttered for the user to see when they type in `agentcore project add runtime --help`

here is how vercel does it: 

Image

src/handlers/project/add/runtime/index.ts
```diff
@@ -0,0 +1,238 @@
...
+          flags["lifecycle-configuration"],
+        ),
+        filesystemConfigurations: parseJsonFlag(
+          "filesystem-configurations",
+          flags["filesystem-configurations"],
+        ),
+        tags: parseTags(flags["tags"]),
+      };
+
+      const runtimeInput = isTemplate
```
notgitika, line 187: it looks like `--build` is accepted for templates but never added to runtimeInput, and the template schema strips it as well. 

This means `--build Container` is silently ignored. Could we include build in the template config?

src/handlers/project/add/runtime/index.ts
jariy17: I like this idea

src/handlers/project/add/runtime/index.ts
```diff
@@ -0,0 +1,238 @@
...
+      const inputEnvironmentVariables = parseJsonFlag<Record<string, string>>(
+        "environment-variables",
+        flags["environment-variables"],
+      );
+      const memoryConfiguration = parseMemoryConfig(flags["memory"]);
+
+      const entrypoint = flags.entrypoint ?? "main.py";
+
+      const source = new SourceResolver({ stdin: config.io.stdin });
+      const apiKey = await source.resolveText("api-key", flags["api-key"]);
```
jariy17, line 154: After looking at Tejas's [PR](https://github.com/aws/agentcore-cli/pull/2046), we shouldn't allow inline secrets. I asked Tejas to shift this helper functino into SourceResolver.

jariy17 (approved):
Its good, the nits could be a follow up pr.

src/handlers/project/add/runtime/index.ts
Hweinstock: this is interesting!

src/handlers/project/add/runtime/index.ts
Hweinstock: It looks like there is some inconsistency here since runtime does allow this: https://github.com/aws/agentcore-cli/blob/d7f3fd4cdb51487fd080e053b42745d582f0ac35/src/handlers/runtime/invoke/request.ts#L46. 

I'm going to merge to avoid conflicts, and make a follow-up task to align on consistent behavior here.

src/handlers/project/add/runtime/index.ts
Hweinstock: yeah good catch. I was initially thinking build would go in here, but the existing asset templates don't actually leverage it and instead have separate templates for build.

I'm working on the template doc now, so this behavior will likely change. I'm making a note to come back to this when we implement the handler to make sure its consistent.

notgitika: that sounds good to me! thanks

PR #2037, opened 2026-08-19
feat(project): implement remove command

Hweinstock:
### Problem 
There is no remove command!

### Solution 
Implement remove as a single command, taking the resource as an argument. Looking at the old CLI https://github.com/aws/agentcore-cli/blob/main/src/cli/primitives/, we basically duplicated this pattern many times, but here we instead do it generically. 

Note: we don't yet support remove all, but one could imagine adding it as a resource type with special behavior. 
### Testing 
unit tests with harness and runtime (since those are the only ones working e2e).

### DIFF
--- a/src/core/project/manager.tsx
+++ b/src/core/project/manager.tsx
@@ -9,6 +9,7 @@ import type {
   ProjectManager,
   ProjectEvent,
   ProjectResource,
+  RemoveResourceInput,
 } from "../../handlers/project/types";
 import type { Logger } from "../../logging";
 import {
@@ -197,6 +198,34 @@ export class FsProjectManager implements ProjectManager {
       );
       throw err;
     }
+  }
+
+  public async removeResource(project: Project, input: RemoveResourceInput): Promise<Project> {
+    const agentCoreSpecPath = join(project.rootPath, "agentcore", "agentcore.json");
+    const projectSpecKey = toProjectSpecKey(input.resourceType);
+
+    const existingProjectSpec = await this.json.read(agentCoreSpecPath, ProjectSpecSchema);
+
+    const existingResources = existingProjectSpec[projectSpecKey];
+    const newResources = existingResources.filter((r) => r.name !== input.name);
+
+    if (newResources.length === existingResources.length)
+      this.logger.child({ input }).warn(`unable to find resource to delete, skipping.`);
+
+    const newSpec = { ...existingProjectSpec, [projectSpecKey]: newResources };
+    const newSpecParseResult = ProjectSpecSchema.safeParse(newSpec);
+
+    if (!newSpecParseResult.success)
+      throw new InputValidationError(z.prettifyError(newSpecParseResult.error), {
+        cause: newSpecParseResult.error,
+      });
+
+    const newProjectSpec = await this.json.write(agentCoreSpecPath, newSpecParseResult.data);
+
+    return {
+      ...project,
+      spec: newProjectSpec,
+    };
   }
 
   private async scaffoldHarness(
--- a/src/core/project/manager.tsx
+++ b/src/core/project/manager.tsx
@@ -230,6 +230,24 @@ export class FsProjectManager implements ProjectManager {
 
     const newProjectSpec = await this.json.write(agentCoreSpecPath, newSpecParseResult.data);
 
+    try {
+      switch (input.resourceType) {
+        case "runtime":
+        case "harness": {
+          const outputPath = join(project.rootPath, "app", input.name);
+          await rm(outputPath, { recursive: true, force: true });
+          break;
+        }
+        default:
+          break;
+      }
+    } catch (e) {
+      throw new ProjectStateError(
+        `unable to clean up scaffolded files for ${input.resourceType} with name ${input.name}.`,
+        { cause: e },
+      );
+    }
+
     return {
       ...project,
       spec: newProjectSpec,
--- a/src/core/project/manager.tsx
+++ b/src/core/project/manager.tsx
@@ -230,24 +230,6 @@ export class FsProjectManager implements ProjectManager {
 
     const newProjectSpec = await this.json.write(agentCoreSpecPath, newSpecParseResult.data);
 
-    try {
-      switch (input.resourceType) {
-        case "runtime":
-        case "harness": {
-          const outputPath = join(project.rootPath, "app", input.name);
-          await rm(outputPath, { recursive: true, force: true });
-          break;
-        }
-        default:
-          break;
-      }
-    } catch (e) {
-      throw new ProjectStateError(
-        `unable to clean up scaffolded files for ${input.resourceType} with name ${input.name}.`,
-        { cause: e },
-      );
-    }
-
     return {
       ...project,
       spec: newProjectSpec,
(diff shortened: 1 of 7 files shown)


src/handlers/project/remove/index.test.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+
+afterEach(async () => {
+  process.chdir(originalCwd);
+  await Promise.all(
+    tempDirectories.splice(0).map((directory) => rm(directory, { recursive: true, force: true })),
+  );
+});
+
+async function run(args: string[]) {
```
Hweinstock, line 31: as a follow-up, I think we can pull this and inProject below out into a common location `(src/testing)`. I have a few open PRs that develop similar utils for this.

src/core/project/manager.tsx
```diff
@@ -199,6 +200,42 @@ export class FsProjectManager implements ProjectManager {
...
 
+  private getProjectSpecPath(project: Project): string {
+    return join(project.rootPath, "agentcore", "agentcore.json");
+  }
+
+  public async removeResource(project: Project, input: RemoveResourceInput): Promise<Project> {
+    const agentCoreSpecPath = this.getProjectSpecPath(project);
+    const projectSpecKey = toProjectSpecKey(input.resourceType);
+
+    const existingProjectSpec = await this.json.read(agentCoreSpecPath, ProjectSpecSchema);
```
Hweinstock, line 211: we could probably re-use some logic from add, but want to wait before trying to generalize this.

src/core/project/manager.tsx
```diff
@@ -199,6 +200,42 @@ export class FsProjectManager implements ProjectManager {
...
+
+    const existingResources = existingProjectSpec[projectSpecKey];
+    const newResources = existingResources.filter((r) => r.name !== input.name);
+
+    if (newResources.length === existingResources.length)
+      this.logger
+        .child({ input })
+        .warn(`unable to remove resource from project that does not exist.`);
```
notgitika, line 219: can we throw right after this? at this point, the handler would still return a user message like `remove successful`. I'm okay to keep it this way like a no-op but it might be a bit misleading.

notgitika (dismissed):
This PR looks good thanks for setting it up. I have a suggestion I would like addressed that I left a comment below on.

separately, I have a question for the future implementations we do: will we also be deleting any code we scaffold (for runtime and harness for instance) when the user runs this command? 

I spent some time reviewing the code on `main` and it seems mixed.

For remove gateway target, it removes the code: https://github.com/aws/agentcore-cli/blob/main/src/cli/operations/remove/remove-gateway-target.ts#L201
Same for code based evals: https://github.com/aws/agentcore-cli/blob/main/src/cli/primitives/EvaluatorPrimitive.ts#L261

however, for agent/runtime, we don't do the same: https://github.com/aws/agentcore-cli/blob/main/src/cli/primitives/AgentPrimitive.tsx#L222


Can we decide and align on one experience before we start with the rest of the implementation?

Hweinstock:
good callout, I think we should remove the scaffolded code to be consistent.

src/core/project/manager.tsx
Hweinstock: I think an idempotent delete is easier, especially if customers are using this in some kind of automation.

src/core/project/manager.tsx
```diff
@@ -230,6 +230,24 @@ export class FsProjectManager implements ProjectManager {
...
+        case "harness": {
+          const outputPath = join(project.rootPath, "app", input.name);
+          await rm(outputPath, { recursive: true, force: true });
+          break;
+        }
+        default:
+          break;
+      }
+    } catch (e) {
+      throw new ProjectStateError(
```
Hweinstock: decided to throw here since it could lead to bad state. ex. i remove a runtime named 'X", then try to recreate (it would fail since the files still exist).

notgitika:
> good callout, I think we should remove the scaffolded code to be consistent.

just had an offline discussion on why we decided not to delete the code early on.

I would lean on not deleting the code, since the user could wanna reuse the code if they maybe make some changes. That is the pattern we established with with agent/runtime initially but the following code with gateway targets/code based evals were regressions, since it wasn't documented properly.

what do you think?

Hweinstock:
responded internally, swapping back to original for now, we can add clean up later if we see a need.

notgitika (approved):
Nice!

src/core/project/manager.tsx
nborges-aws: I agree with keeping the successful exit code; idempotency here is useful. Though the `removed successfully` error messaging is very misleading. 

As a nit I'd say we should maintain the success exit code, but add a mechanism for differentiating the result.

nborges-aws (approved):
Left one comment on the non-existent resource handling. I do think the success error message is misleading and we should consider differentiating.

Definitely non-blocking for this PR and something we can think through. LGTM otherwise.

src/core/project/manager.tsx
```diff
@@ -199,6 +200,42 @@ export class FsProjectManager implements ProjectManager {
...
+  }
+
+  public async removeResource(project: Project, input: RemoveResourceInput): Promise<Project> {
+    const agentCoreSpecPath = this.getProjectSpecPath(project);
+    const projectSpecKey = toProjectSpecKey(input.resourceType);
+
+    const existingProjectSpec = await this.json.read(agentCoreSpecPath, ProjectSpecSchema);
+
+    const existingResources = existingProjectSpec[projectSpecKey];
+    const newResources = existingResources.filter((r) => r.name !== input.name);
```
jariy17, line 214: Later, we can make this a helper function that could be used to prevent duplicates for `add` and update existing for `update`

src/handlers/project/remove/index.test.ts
```diff
@@ -0,0 +1,113 @@
...
+  label: string;
+  commands: string[][];
+  specKey: string;
+  expectedRemaining: string[];
+};
+
+describe("project remove", () => {
+  // Verifies that resources are removed from agentcore.json and the correct
+  // remaining resources are left.
+  test.each<RemoveCase>([
```
jariy17, line 61: nit: There should be a test that will make sure we don't delete customer's code

jariy17 (approved):

jariy17:
> This PR looks good thanks for setting it up. I have a suggestion I would like addressed that I left a comment below on.
> 
> separately, I have a question for the future implementations we do: will we also be deleting any code we scaffold (for runtime and harness for instance) when the user runs this command?
> 
> I spent some time reviewing the code on `main` and it seems mixed.
> 
> For remove gateway target, it removes the code: https://github.com/aws/agentcore-cli/blob/main/src/cli/operations/remove/remove-gateway-target.ts#L201 Same for code based evals: https://github.com/aws/agentcore-cli/blob/main/src/cli/primitives/EvaluatorPrimitive.ts#L261
> 
> however, for agent/runtime, we don't do the same: https://github.com/aws/agentcore-cli/blob/main/src/cli/primitives/AgentPrimitive.tsx#L222
> 
> Can we decide and align on one experience before we start with the rest of the implementation?

I discussed with Harrison, I think we shouldn't delete their application code because it's their code. It's an easy two door decision if customers want this feature.

PR #2039, opened 2026-08-19
feat(dev): OTLP trace storage for local dev

tejaskash:
First slice of the `project dev` tracing work (#1980 restacks onto this; see #2013's closing comment for the full split plan).

## What this does

The storage half of local trace collection, with no I/O beyond the filesystem:
- **`otel/transforms.ts`** — pure OTLP wire handling: split one export batch into per-trace payloads (SDKs batch by time, not trace — attributing a whole batch to its first trace id corrupts trace identity), normalize protobuf/JSON ids to hex, and shape spans/logs for display (attribute flattening, transport-noise filtering).
- **`otel/store.ts`** — append-only storage: one JSON Lines file per trace under a directory, raw reads back. Malformed lines are skipped, never fatal (a half-written line must not break reads).

Ported from the current CLI's `operations/dev/otel/transforms.ts` with one correctness fix over the original: the per-trace batch partitioning described above.

## What consumes it

#1980 (the OTLP/HTTP collector + `project dev` wiring) stacks directly on this branch — the consumer is one PR up, visible now.

## Verification

23 tests (happy + unhappy paths per layer, audited for redundancy), tsc/lint/format clean. End-to-end proof of the full pipeline lives in #1980's stack (real project, real Bedrock invocation, traces on disk).


src/core/dev/otel/types.ts
```diff
@@ -0,0 +1,65 @@
...
+  arrayValue?: { values?: OtlpAttributeValue[] };
+  kvlistValue?: { values?: OtlpAttribute[] };
+}
+
+export interface OtlpAttribute {
+  key: string;
+  value?: OtlpAttributeValue;
+}
+
+export type OtlpAttributes = OtlpAttribute[] | Record<string, unknown>;
```
Hweinstock: should this be `Record<string, unknown>[]`? Or is there a reason we want a list or a record here?

PR #2041, opened 2026-08-19
feat(dev): supervise every runtime in project dev

tejaskash:
## What this does

`agentcore project dev` without `--agent` runs all of the project's runtimes at once. Each agent's output is attributed (`[orders] ...` in the terminal, an `agent` field in NDJSON); one agent crashing reports its failure and leaves the others running; Ctrl-C tears the whole set down. Selecting one runtime (`--agent`, or a single-runtime project) keeps the direct path where a crash fails the command — scripts and CI rely on that exit code.

## How

`DevSupervisor` (`core/dev/supervisor.ts`) owns per-agent lifecycle: an abort scope per agent chained to the command signal, a per-agent state machine (idle/starting/running/failed) with start dedup, TCP readiness probing, and one merged attributed event stream. It's also what the Inspector UI slice (next PR) drives via its start-on-demand API. Ports resolve sequentially — concurrent resolution raced two HTTP agents onto the same port.

## Verification

- Supervisor: state machine, dedup, retry-after-failure, merge, crash isolation, abort chaining.
- Handler with fake runners + fake collector, including runners that bind real TCP listeners so the readiness probe runs for real.
- End to end: built the bundle, scaffolded a project, added a second runtime to `agentcore.json`, ran `project dev` — both agents started on distinct ports with attributed logs, both answered real Bedrock invocations, and the trace directory holds spans for both service names (per-runtime `OTEL_SERVICE_NAME` verified end to end). SIGTERM exits 130.

tsc/lint/format clean.

### DIFF
--- a/src/core/dev/supervisor.ts
+++ b/src/core/dev/supervisor.ts
@@ -1,5 +1,5 @@
-import { connect } from "node:net";
 import { ResourceNotFoundError } from "../../errors";
+import { waitForPort } from "../../io";
 import type { DevEvent, DevRunner } from "../../handlers/project/dev/types";
 import type { ProjectRuntime } from "../../projectSchemas/runtime";
 
@@ -14,9 +14,9 @@ export interface AgentStatus {
   error?: string;
 }
 
-/** A dev event attributed to the agent that produced it. */
+/** A dev event tagged with the name of the runtime that produced it. */
 export interface SupervisedEvent {
-  agent: string;
+  agentName: string;
   event: DevEvent;
 }
 
@@ -25,11 +25,15 @@ export type SupervisorConfig = {
   projectRoot: string;
   runners: { CodeZip: DevRunner; Container: DevRunner };
   /** Resolves the full child environment for a runtime (dev env + OTEL vars). */
-  environment: (runtime: ProjectRuntime) => Promise<Record<string, string>>;
+  getDevEnvVarsForRuntime: (runtime: ProjectRuntime) => Promise<Record<string, string>>;
   /** Resolves the port a runtime should serve on. */
   resolvePort: (runtime: ProjectRuntime) => Promise<number>;
-  /** Resolves once a started agent accepts connections on its port. */
-  waitReady?: (port: number, signal: AbortSignal) => Promise<void>;
+  /**
+   * Resolves once a started agent accepts connections on its port. `lastActivityAt`
+   * returns the time of the agent's most recent output, so a build that keeps
+   * logging is not timed out mid-flight.
+   */
+  waitReady?: (port: number, signal: AbortSignal, lastActivityAt: () => number) => Promise<void>;
   signal: AbortSignal;
 };
 
@@ -52,7 +56,11 @@ export class DevSupervisor {
   private readonly agents = new Map<string, AgentEntry>();
   private readonly queue: SupervisedEvent[] = [];
   private wake: (() => void) | undefined;
-  private readonly waitReady: (port: number, signal: AbortSignal) => Promise<void>;
+  private readonly waitReady: (
+    port: number,
+    signal: AbortSignal,
+    lastActivityAt: () => number,
+  ) => Promise<void>;
 
   constructor(private readonly config: SupervisorConfig) {
     for (const runtime of config.runtimes) {
@@ -107,7 +115,7 @@ export class DevSupervisor {
    * and repeated starts of the same agent share one attempt; a previously
    * failed agent may be started again.
    */
-  public start(name: string): Promise<{ name: string; port: number }> {
+  public async start(name: string): Promise<{ name: string; port: number }> {
     const entry = this.agents.get(name);
     if (!entry) {
       const available = [...this.agents.keys()].join(", ");
@@ -116,7 +124,7 @@ export class DevSupervisor {
       );
     }
     if (entry.phase === "running" && entry.port !== undefined) {
-      return Promise.resolve({ name, port: entry.port });
+      return { name, port: entry.port };
     }
     if (entry.starting) return entry.starting;
 
@@ -145,8 +153,8 @@ export class DevSupervisor {
     }
   }
 
-  private push(agent: string, event: DevEvent): void {
-    this.queue.push({ agent, event });
+  private push(agentName: string, event: DevEvent): void {
+    this.queue.push({ agentName, event });
     this.wake?.();
   }
 
@@ -156,7 +164,7 @@ export class DevSupervisor {
     entry.error = undefined;
 
     const controller = new AbortController();
-    const onParentAbort = () => controller.abort();
+    const onParentAbort = () => controller.abort(this.config.signal.reason);
     // Chained for the agent's whole lifetime (not just startup): the command's
     // Ctrl-C must tear down every running child. The pump removes it on exit.
     this.config.signal.addEventListener("abort", onParentAbort, { once: true });
@@ -164,14 +172,17 @@ export class DevSupervisor {
 
     try {
       const port = await this.config.resolvePort(entry.runtime);
-      const env = await this.config.environment(entry.runtime);
+      const env = await this.config.getDevEnvVarsForRuntime(entry.runtime);
       const runner = this.config.runners[entry.runtime.build];
 
       let ready = false;
-      const readiness = this.waitReady(port, controller.signal).then(() => {
+      const activity = { at: Date.now() };
+      const readiness = this.waitReady(port, controller.signal, () => activity.at).then(() => {
         ready = true;
       });
-      const earlyExit = this.pump(entry, runner, { port, env, signal: controller.signal })
+      const earlyExit = this.pump(entry, runner, { port, env, signal: controller.signal }, () => {
+        activity.at = Date.now();
+      })
         .finally(unchain)
         .then(() => {
           if (!ready)
@@ -205,6 +216,7 @@ export class DevSupervisor {
     entry: AgentEntry,
     runner: DevRunner,
     input: { port: number; env: Record<string, string>; signal: AbortSignal },
+    onActivity: () => void,
   ): Promise<void> {
     const name = entry.runtime.name;
     try {
@@ -215,6 +227,7 @@ export class DevSupervisor {
         env: input.env,
         signal: input.signal,
       })) {
+        onActivity();
         this.push(name, event);
       }
       if (entry.phase === "running") {
@@ -232,46 +245,4 @@ export class DevSupervisor {
       }
     }
   }
-}
-
-/** Generous enough for a cold dependency install before the server first binds. */
-const READY_TIMEOUT_MS = 120_000;
-
-/**
- * Poll until a loopback TCP connection to `port` succeeds, the signal aborts,
- * or the deadline passes — a child that stays alive without ever binding must
- * fail its start instead of blocking every later runtime.
- */
-export function waitForPort(
-  port: number,
-  signal: AbortSignal,
-  intervalMs = 250,
-  timeoutMs = READY_TIMEOUT_MS,
-): Promise<void> {
-  const deadline = Date.now() + timeoutMs;
-  return new Promise((resolve, reject) => {
-    const attempt = () => {
-      if (signal.aborted) {
-        reject(new Error("Aborted while waiting for the agent to become ready."));
-        return;
-      }
-      if (Date.now() > deadline) {
-        reject(
-          new Error(
-            `Agent did not accept connections on port ${port} within ${timeoutMs / 1000}s.`,
-          ),
-        );
-        return;
-      }
-      const socket = connect({ port, host: "127.0.0.1" }, () => {
-        socket.destroy();
-        resolve();
-      });
-      socket.on("error", () => {
-        socket.destroy();
-        setTimeout(attempt, intervalMs);
-      });
-    };
-    attempt();
-  });
 }
--- a/src/core/dev/supervisor.ts
+++ b/src/core/dev/supervisor.ts
@@ -185,6 +185,7 @@ export class DevSupervisor {
       return { name, port };
     } catch (error) {
       controller.abort();
+      unchain(); // idempotent alongside the pump's cleanup; covers setup failures before the pump exists
       entry.phase = "failed";
       entry.error = error instanceof Error ? error.message : String(error);
       this.push(name, {
@@ -229,12 +230,33 @@ export class DevSupervisor {
   }
 }
 
-/** Poll until a loopback TCP connection to `port` succeeds, or the signal aborts. */
-function waitForPort(port: number, signal: AbortSignal, intervalMs = 250): Promise<void> {
+/** Generous enough for a cold dependency install before the server first binds. */
+const READY_TIMEOUT_MS = 120_000;
+
+/**
+ * Poll until a loopback TCP connection to `port` succeeds, the signal aborts,
+ * or the deadline passes — a child that stays alive without ever binding must
+ * fail its start instead of blocking every later runtime.
+ */
+export function waitForPort(
+  port: number,
+  signal: AbortSignal,
+  intervalMs = 250,
+  timeoutMs = READY_TIMEOUT_MS,
+): Promise<void> {
+  const deadline = Date.now() + timeoutMs;
   return new Promise((resolve, reject) => {
     const attempt = () => {
       if (signal.aborted) {
         reject(new Error("Aborted while waiting for the agent to become ready."));
+        return;
+      }
+      if (Date.now() > deadline) {
+        reject(
+          new Error(
+            `Agent did not accept connections on port ${port} within ${timeoutMs / 1000}s.`,
+          ),
+        );
         return;
       }
       const socket = connect({ port, host: "127.0.0.1" }, () => {
--- a/src/core/dev/supervisor.ts
+++ b/src/core/dev/supervisor.ts
@@ -1,0 +1,251 @@
+import { connect } from "node:net";
+import { ResourceNotFoundError } from "../../errors";
+import type { DevEvent, DevRunner } from "../../handlers/project/dev/types";
+import type { ProjectRuntime } from "../../projectSchemas/runtime";
+
+export type AgentPhase = "idle" | "starting" | "running" | "failed";
+
+export interface AgentStatus {
+  name: string;
+  buildType: ProjectRuntime["build"];
+  protocol: NonNullable<ProjectRuntime["protocol"]>;
+  phase: AgentPhase;
+  port?: number;
+  error?: string;
+}
+
+/** A dev event attributed to the agent that produced it. */
+export interface SupervisedEvent {
+  agent: string;
+  event: DevEvent;
+}
+
+export type SupervisorConfig = {
+  runtimes: ProjectRuntime[];
+  projectRoot: string;
+  runners: { CodeZip: DevRunner; Container: DevRunner };
+  /** Resolves the full child environment for a runtime (dev env + OTEL vars). */
+  environment: (runtime: ProjectRuntime) => Promise<Record<string, string>>;
+  /** Resolves the port a runtime should serve on. */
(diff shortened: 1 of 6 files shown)


tejaskash:
Review response, both findings addressed in `c5398368`:

- **Unbounded readiness poll** — `waitForPort` now has a 120s deadline (generous for a cold dependency install before first bind); a live-but-never-binding child fails its start with a clear message instead of blocking every later runtime. Tested against a dead port (rejects at deadline) and a live listener (resolves).
- **Leaked abort listeners on setup failure** — failures before the pump exists (port resolution, environment) now unchain the parent-abort listener like every other exit path; tested with a counting signal across three failed retries (adds − removes stays at the constructor's single wake listener).

Hweinstock:
> agentcore project dev without --agent runs all of the project's runtimes at once.

Is this the behavior on the publicly released CLI? I thought we always had to select one?

src/handlers/project/dev/index.ts
```diff
@@ -154,16 +137,54 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
...
+              .port,
           signal: controller.signal,
-        })) {
-          renderEvent(config.io, event, json);
+        });
+        // Sequential starts: concurrent port resolution would race two agents
+        // onto the same port. Failed starts surface as attributed status events.
+        for (const runtime of runtimes) {
+          await supervisor.start(runtime.name).catch(() => {});
+        }
```
notgitika: do we need to await for each runtime to be ready to start the output? 

had my agent try this out:

```
I checked this with a runner that emits a line immediately and never binds — stdout was still "" 1.5s later. And if the user Ctrl-Cs during startup, the handler never reaches the drain loop at all, so the whole queue is discarded and they get zero diagnostics.

Containers are the painful case, since container.ts:148-149 streams the entire docker build through this same event channel — a two-container project is a blank terminal for the duration of both cold builds. 
```

I think if we do it concurrently it should work.

```
const streaming = (async () => {
   for await (const { agent, event } of supervisor.events()) renderEvent(config.io, event, json, agent);
 })();

 for (const runtime of runtimes) await supervisor.start(runtime.name).catch(() => {});
   await streaming;
```

src/core/dev/supervisor.ts
```diff
@@ -0,0 +1,277 @@
...
+  error?: string;
+  starting?: Promise<{ name: string; port: number }>;
+};
+
+/**
+ * Owns the lifecycle of every dev-able runtime for the Inspector: agents start
+ * lazily (triggered from the browser), each in its own abort scope chained off
+ * the command's signal, and every runner's events merge into one attributed
+ * stream the dev handler renders. Restart-on-edit stays inside the child
+ * (uvicorn --reload / tsx watch) — the supervisor never restarts processes.
```
notgitika, line 53: FYI this PR starts all of them and doesnt do it lazily is that being introduced in the next PR? most probably the case I haven't checked but if not, lmk if we are planning to fix it

src/core/dev/supervisor.ts
```diff
@@ -0,0 +1,277 @@
...
+    this.waitReady = config.waitReady ?? waitForPort;
+    config.signal.addEventListener("abort", () => this.wake?.(), { once: true });
+  }
+
+  /**
+   * Replace the managed runtime set after a config change: new runtimes join
+   * idle, edited definitions apply on the next start, and removed runtimes
+   * drop unless they are currently starting or running.
+   */
+  public setRuntimes(runtimes: ProjectRuntime[]): void {
```
notgitika, line 78: confirmed setruntimes, running and snapshot are in the next PR

src/core/dev/supervisor.ts
```diff
@@ -0,0 +1,277 @@
...
+  /**
+   * Start an agent by name, resolving once it accepts connections. Concurrent
+   * and repeated starts of the same agent share one attempt; a previously
+   * failed agent may be started again.
+   */
+  public start(name: string): Promise<{ name: string; port: number }> {
+    const entry = this.agents.get(name);
+    if (!entry) {
+      const available = [...this.agents.keys()].join(", ");
+      throw new ResourceNotFoundError(
```
notgitika, line 122: start() returns a promise on every other path but throws synchronously here, so supervisor.start(name).catch(...) won't catch an unknown name, the throw escapes the .catch() in `index.ts:180`. 

safe today because the handler only passes names it just read from the spec, but an Inspector route taking a name from the browser would 500 instead of 404. Making the method async fixes it without changing any call site.

src/handlers/project/dev/index.ts
```diff
@@ -154,16 +137,54 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
...
+            region,
+          });
+          const otel =
+            collector && (runtime.instrumentation?.enableOtel ?? true)
+              ? otelEnvForRuntime(collector, runtime)
+              : {};
+          return { ...env, ...otel };
+        };
+
+        if (runtimes.length === 1) {
```
jariy17: Let's use DevSupervisor here. It will be less code to maintain.

src/core/dev/supervisor.ts
```diff
@@ -0,0 +1,277 @@
...
+  }
+
+  /**
+   * The merged event stream of every agent this supervisor has started. Ends
+   * when the supervisor's signal aborts and all pending events are drained.
+   */
+  public async *events(): AsyncGenerator<SupervisedEvent, void> {
+    while (true) {
+      for (const event of this.queue.splice(0)) yield event;
+      if (this.config.signal.aborted) return;
```
notgitika, line 144: this only returns on abort, so once every pump has finished there's nothing left to wake the consumer,the handler's for await waits forever. 

my agent tested this:
```
Verified with two runtimes that both exit immediately: their failures print, then the command was still running at 2s with no children and no work pending. Only Ctrl-C ends it, which exits 130, so CI sees a cancellation rather than a failure. The single-runtime path exits when its runner ends, so this is a divergence too. 
```

I think ending the stream when no agent is starting/running would cover it. or failing outright when zero agents started

src/handlers/project/dev/index.ts
```diff
@@ -92,40 +88,28 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
...
-        });
-        controller.signal.throwIfAborted();
-
-        let otelEnv: Record<string, string> = {};
-        if (flags.traces && (runtime.instrumentation?.enableOtel ?? true)) {
+        if (
+          flags.traces &&
+          runtimes.some((runtime) => runtime.instrumentation?.enableOtel ?? true)
+        ) {
           const tracesDirectory = join(project.rootPath, "agentcore", ".cli", "traces", "otlp");
```
Hweinstock, line 116: didn't we talk about removing this directory?

src/core/dev/supervisor.ts
```diff
@@ -0,0 +1,277 @@
...
+    this.config.signal.addEventListener("abort", onParentAbort, { once: true });
+    const unchain = () => this.config.signal.removeEventListener("abort", onParentAbort);
+
+    try {
+      const port = await this.config.resolvePort(entry.runtime);
+      const env = await this.config.environment(entry.runtime);
+      const runner = this.config.runners[entry.runtime.build];
+
+      let ready = false;
+      const readiness = this.waitReady(port, controller.signal).then(() => {
```
notgitika: so this starts the same instant as pump which means in 120s the READY_TIMEOUT_MS has to cover build + install + bind. if it fails to do so in the time, the catch would call controller.abort() which would kill a docker build and report it as "agent failed" is my understanding.

can we have each agent spend its own 120s while the rest wait? or maybe if the probe wait until the runner signals it's actually launching the server (rather than building)?

src/handlers/project/dev/index.ts
```diff
@@ -154,16 +137,54 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
...
+        // Several runtimes: supervise them all, streaming agent-attributed output.
+        const supervisor = new DevSupervisor({
+          runtimes,
           projectRoot: project.rootPath,
-          port: devPort.port,
-          env: { ...env, ...otelEnv },
+          runners: config.runners,
+          environment,
+          resolvePort: async (runtime) =>
+            (await resolveDevPort(runtime.protocol, undefined, config.checkPort, controller.signal))
```
jariy17: If we allow 1 runtime in a Dev Supervisor we don't need this helper function, and inline here.

notgitika (commented):
some comments

src/handlers/project/dev/index.ts
```diff
@@ -33,36 +34,31 @@ function otelEnvForRuntime(
...
+  const runtime = project.spec.runtimes.find((candidate) => candidate.name === name);
+  if (runtime) return [runtime];
+  const available = project.spec.runtimes.map((candidate) => candidate.name).join(", ");
+  throw new ResourceNotFoundError(
+    `Runtime '${name}' was not found. Available runtimes: ${available}.`,
   );
 }
 
-function renderEvent(io: AppIO, event: DevEvent, json?: JsonRenderer): void {
+function renderEvent(io: AppIO, event: DevEvent, json?: JsonRenderer, agent?: string): void {
```
jariy17: Let's always show the agentName. It simplifies the code here and more importantly customers will be know which agent they are talking.

src/handlers/project/dev/index.ts
Hweinstock: +1, i feel like DevSupervisor should be general enough to handle 1 or more agents.

src/handlers/project/dev/index.ts
```diff
@@ -154,16 +137,54 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
         }
         controller.signal.throwIfAborted();
 
-        const runner = config.runners[runtime.build];
-        for await (const event of runner.run({
-          runtime,
+        const environment = async (runtime: ProjectRuntime): Promise<Record<string, string>> => {
```
Hweinstock: `environment` feels pretty vague. Based on the implementation this is merging two sets of environment variables to be used in the dev environment for a given runtime. Is there a more descriptive name for this? maybe `getDevEnvironmentVariables` or `getDevEnvVarsForRuntime`?

src/handlers/project/dev/index.ts
```diff
@@ -154,16 +137,54 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
...
+            (await resolveDevPort(runtime.protocol, undefined, config.checkPort, controller.signal))
+              .port,
           signal: controller.signal,
-        })) {
-          renderEvent(config.io, event, json);
+        });
+        // Sequential starts: concurrent port resolution would race two agents
+        // onto the same port. Failed starts surface as attributed status events.
+        for (const runtime of runtimes) {
+          await supervisor.start(runtime.name).catch(() => {});
```
Hweinstock: does the `catch` here mean we ignore individual runtime failures? Is that the behavior we want, and if so should we log the warning here?

src/core/dev/supervisor.ts
```diff
@@ -0,0 +1,277 @@
...
+
+export type AgentPhase = "idle" | "starting" | "running" | "failed";
+
+export interface AgentStatus {
+  name: string;
+  buildType: ProjectRuntime["build"];
+  protocol: NonNullable<ProjectRuntime["protocol"]>;
+  phase: AgentPhase;
+  port?: number;
+  error?: string;
```
Hweinstock, line 14: Are we losing the error type by serializing it?

src/core/dev/supervisor.ts
```diff
@@ -0,0 +1,277 @@
...
+}
+
+/** A dev event attributed to the agent that produced it. */
+export interface SupervisedEvent {
+  agent: string;
```
Hweinstock: what does a string agent represent? is the agent name, agent id, agent type, or some serialization of the agent itself?

src/core/dev/supervisor.ts
```diff
@@ -0,0 +1,277 @@
...
+
+/** Generous enough for a cold dependency install before the server first binds. */
+const READY_TIMEOUT_MS = 120_000;
+
+/**
+ * Poll until a loopback TCP connection to `port` succeeds, the signal aborts,
+ * or the deadline passes — a child that stays alive without ever binding must
+ * fail its start instead of blocking every later runtime.
+ */
+export function waitForPort(
```
Hweinstock: Is there a way to make this general enough to move this to https://github.com/aws/agentcore-cli/blob/refactor/src/io/port.ts?

src/handlers/project/dev/index.ts
jariy17: These per-agent start failures never reach telemetry — they're emitted as status events and the .catch(() => {}) swallows the rejection, so nothing ever throws. Should we be capturing them? And if so, how do we report to telemetry without throwing, given that telemetry today seems to fire only when an error propagates? Mediation: Lets throw a error if we couldn't start any agents.

jariy17 (dismissed):
Small nits can be done in follow up. Good design I like it

src/core/dev/supervisor.ts
```diff
@@ -0,0 +1,277 @@
...
+          if (!ready)
+            throw new Error(entry.error ?? `Agent '${name}' exited before it became ready.`);
+        });
+      // Both branches outlive the race (the pump runs for the agent's lifetime);
+      // swallow their late rejections so losing branches never become unhandled.
+      readiness.catch(() => {});
+      earlyExit.catch(() => {});
+      await Promise.race([readiness, earlyExit]);
+
+      entry.phase = "running";
```
Hweinstock, line 197: I'm getting a little lost with all the tracked state. Is there a way to more explicitly map out the possible states and transitions? For example, an idle entry doesn't have a port, and a failed entry may have error information (but not a running entry).

src/core/dev/supervisor.ts
```diff
@@ -0,0 +1,277 @@
...
+    }
+    for (const [name, entry] of this.agents) {
+      if (!names.has(name) && entry.phase !== "running" && entry.phase !== "starting") {
+        this.agents.delete(name);
+      }
+    }
+  }
+
+  /** Current phase, port, and last error of every managed agent. */
+  public snapshot(): AgentStatus[] {
```
Hweinstock, line 93: I see in the handler we call events and start, but there is also `running`, `snapshot`, and `setRuntimes`. Are these for future support? Is there a way to make the contract we're trying to fulfill more explicit?

Hweinstock (commented):
mostly small naming things, but I'm a little lost by the state maintained by this supervisor.

src/handlers/project/dev/index.ts
tejaskash: Lets discuss

src/handlers/project/dev/index.ts
tejaskash: Good catch, fixed. All the agents start together now and their output streams as it comes in, so you see the build logs right away instead of waiting for the first one to bind. I also break out of the loop the instant Ctrl-C fires, so anything printed during startup still lands in the terminal. Went with the concurrent shape you suggested.

src/core/dev/supervisor.ts
tejaskash: Right, this PR starts them up front. The lazy, start-from-the-browser behavior lands in the Inspector PR.

src/core/dev/supervisor.ts
tejaskash: Yes. setRuntimes, running and snapshot go live in the Inspector PR. They are here now so that PR can stay small.

src/core/dev/supervisor.ts
tejaskash: Fixed. start() is async now, so an unknown name comes back as a rejected promise and the .catch picks it up. An Inspector route that reads a name from the browser will get a clean 404 instead of a 500.

src/handlers/project/dev/index.ts
tejaskash: Agreed. runSingleRuntime is gone and a single runtime goes through DevSupervisor like everything else. One path to maintain now.

src/core/dev/supervisor.ts
tejaskash: You are right, that was a real hang. The handler watches the snapshot now and ends the loop once nothing is starting or running. If every agent failed it exits non-zero, so CI sees a failure rather than a cancellation.

src/core/dev/supervisor.ts
tejaskash: Reworked this. It is an idle timeout now rather than a fixed deadline. The clock resets every time the agent prints a line, so a docker build that keeps logging never gets cut off. It only gives up after 120s of complete silence with no port, or when you Ctrl-C. Each agent tracks its own last-output time, so a slow one does not eat into another agent's budget.

src/handlers/project/dev/index.ts
tejaskash: Removed. The single case runs through the supervisor now, so the helper was no longer earning its place.

src/handlers/project/dev/index.ts
tejaskash: Agent output is always tagged with [name] now, since everything routes through the supervisor. The only lines left without a tag are the two command-level messages from the trace collector, which do not belong to any one agent.

src/handlers/project/dev/index.ts
tejaskash: Renamed to getDevEnvVarsForRuntime.

src/handlers/project/dev/index.ts
tejaskash: One agent failing no longer takes down its healthy siblings, and the failure prints to the terminal with the agent name in front of it. If none of them start, the command throws and exits non-zero, which is the path that reaches telemetry. More on that in the telemetry thread below.

src/core/dev/supervisor.ts
tejaskash: The thrown error keeps its class, so telemetry and callers still see the real type. The string on the snapshot is only there to show status in the agent list, it is not what we throw.

src/core/dev/supervisor.ts
tejaskash: It is the runtime name. I renamed the field to agentName and updated the comment to say so.

src/core/dev/supervisor.ts
tejaskash: Moved it to src/io/port.ts, right next to checkPort. The raw socket work lives in the IO layer now and the supervisor just calls into it.

src/handlers/project/dev/index.ts
tejaskash: Done. If we cannot start any agent the handler throws, so telemetry records it through the normal error path. Individual failures still print per agent without throwing, so one bad agent does not take the rest down with it.

src/core/dev/supervisor.ts
tejaskash: Agreed the tracked state is carrying a lot. I would rather settle its shape in the Inspector PR, where running and snapshot start reading it for real, and model it as one type per phase so an idle entry cannot hold a port and a running one cannot hold an error. Leaving it alone here to keep this PR focused.

src/core/dev/supervisor.ts
tejaskash: Those three are for the Inspector PR, which is why only start and events get called here. I will tighten the public surface when that PR wires them up for real.

src/handlers/project/dev/index.ts
tejaskash: Done in ec9377c9. `agent` is now required: `renderEvent` split into `renderAgentEvent(io, event, agent, json)` (always tags `[name]`) and `renderStatus(io, message, json)` for the two command-level collector lines (startup + persist-warning) that belong to no agent. The `agent === undefined` branches are gone. Every runtime, including the single-runtime case, now routes through the supervisor, so all agent output is attributed.

jariy17 (approved):

notgitika (approved):

PR #2043, opened 2026-08-19
feat(dev): OTLP trace storage for local dev

tejaskash:
First slice of the `project dev` tracing work; #1980 stacks on this.

## What this does

The storage half of local trace collection, filesystem I/O only:
- **`otel/transforms.ts`** — OTLP wire handling: split one export batch into per-trace payloads, normalize protobuf/JSON ids to hex, shape spans/logs for display (attribute flattening, transport-noise filtering). SDKs batch by time, not by trace, so attributing a whole batch to its first trace id would corrupt trace identity — hence the per-trace partitioning.
- **`otel/store.ts`** — append-only storage: one JSON Lines file per trace, read back raw. Malformed lines are skipped so a half-written line can't break a read.

Ported from the current CLI's `operations/dev/otel/transforms.ts`, with the per-trace batch partitioning as the one correctness fix over the original.

## Verification

Per-layer tests (happy + unhappy paths), tsc/lint/format clean. End-to-end proof of the full pipeline is in #1980 (real project, real Bedrock invocation, traces on disk).

---
Replaces #2039, closed by a stacking-tool mishap (base flipped to `main`); identical branch and content.

### DIFF
--- a/src/core/dev/otel/store.test.ts
+++ b/src/core/dev/otel/store.test.ts
@@ -1,0 +1,135 @@
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { mkdtemp, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { TraceStore } from "./store";
+import type { OtlpPayload } from "./types";
+
+const TRACE_A = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
+const TRACE_B = "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb";
+
+function payload(
+  traceId: string,
+  options: { serviceName?: string; startNano?: string; name?: string } = {},
+): OtlpPayload {
+  return {
+    resourceSpans: [
+      {
+        resource: {
+          attributes: [
+            { key: "service.name", value: { stringValue: options.serviceName ?? "agent-1" } },
+          ],
+        },
+        scopeSpans: [
+          {
+            scope: { name: "test" },
+            spans: [
+              {
+                traceId,
+                spanId: "0123456789abcdef",
+                name: options.name ?? "invoke_agent strands",
+                kind: 1,
+                startTimeUnixNano: options.startNano ?? `${BigInt(Date.now()) * 1_000_000n}`,
+                endTimeUnixNano: options.startNano ?? `${BigInt(Date.now()) * 1_000_000n}`,
+              },
+            ],
+          },
+        ],
+      },
+    ],
+  };
+}
+
+let directory: string;
+let store: TraceStore;
+
+beforeEach(async () => {
+  directory = await mkdtemp(join(tmpdir(), "trace-store-"));
+  store = new TraceStore(directory);
+});
+
+afterEach(async () => {
+  await rm(directory, { recursive: true, force: true });
+});
+
+describe("TraceStore", () => {
+  test("append then list returns the trace with metadata", async () => {
+    await store.append(payload(TRACE_A));
+    const traces = await store.list();
+    expect(traces).toHaveLength(1);
+    expect(traces[0]!.traceId).toBe(TRACE_A);
+    expect(traces[0]!.spanCount).toBe("1");
+    expect(traces[0]!.resourceSpans).toBeDefined();
+  });
+
+  test("appends to the same trace accumulate spans", async () => {
+    await store.append(payload(TRACE_A));
+    await store.append(payload(TRACE_A, { name: "tool_use" }));
+    const traces = await store.list();
+    expect(traces).toHaveLength(1);
+    expect(traces[0]!.spanCount).toBe("2");
+  });
+
+  test("payloads without a trace id are dropped", async () => {
+    await store.append({ resourceSpans: [] });
+    expect(await store.list()).toEqual([]);
+  });
+
+  test("a batch carrying several traces lands in each trace's own file", async () => {
+    const batch = payload(TRACE_A);
+    batch.resourceSpans![0]!.scopeSpans![0]!.spans!.push({
+      ...batch.resourceSpans![0]!.scopeSpans![0]!.spans![0]!,
+      traceId: TRACE_B,
+      name: "tool_use",
+    });
+    await store.append(batch);
+
+    const traces = await store.list();
+    expect(traces.map((trace) => trace.traceId).sort()).toEqual([TRACE_A, TRACE_B]);
+    expect(traces.every((trace) => trace.spanCount === "1")).toBe(true);
+    expect(await store.get(TRACE_B)).toBeDefined();
+  });
+
+  test("list filters by service name", async () => {
+    await store.append(payload(TRACE_A, { serviceName: "agent-1" }));
+    await store.append(payload(TRACE_B, { serviceName: "agent-2" }));
+    const traces = await store.list({ serviceName: "agent-2" });
+    expect(traces.map((trace) => trace.traceId)).toEqual([TRACE_B]);
+  });
+
+  test("list filters by time window and sorts newest first", async () => {
+    const oldNano = `${BigInt(Date.now() - 24 * 60 * 60 * 1000) * 1_000_000n}`;
+    await store.append(payload(TRACE_A, { startNano: oldNano }));
+    await store.append(payload(TRACE_B));
+
+    expect((await store.list()).map((trace) => trace.traceId)).toEqual([TRACE_B]);
+
+    const all = await store.list({ startTime: 0 });
+    expect(all.map((trace) => trace.traceId)).toEqual([TRACE_B, TRACE_A]);
+  });
+
+  test("get returns the trace detail or undefined for unknown ids", async () => {
+    await store.append(payload(TRACE_A));
+    const detail = await store.get(TRACE_A);
+    expect(detail?.resourceSpans).toBeDefined();
+    expect(await store.get(TRACE_B)).toBeUndefined();
+  });
+
+  test("skips malformed lines and files without failing", async () => {
+    await store.append(payload(TRACE_A));
+    await writeFile(join(directory, `${TRACE_A}.otlp.jsonl`), "{not json}\n", {
+      flag: "a",
+    });
+    await writeFile(join(directory, "garbage.otlp.jsonl"), "also not json\n");
+
+    const traces = await store.list();
+    expect(traces).toHaveLength(1);
+    expect(traces[0]!.spanCount).toBe("1");
+  });
+
+  test("list on a directory that does not exist returns empty", async () => {
+    const empty = new TraceStore(join(directory, "missing"));
+    expect(await empty.list()).toEqual([]);
+    expect(await empty.get(TRACE_A)).toBeUndefined();
+  });
+});
--- a/src/core/dev/otel/store.test.ts
+++ b/src/core/dev/otel/store.test.ts
@@ -1,5 +1,5 @@
 import { afterEach, beforeEach, describe, expect, test } from "bun:test";
-import { mkdtemp, rm, writeFile } from "node:fs/promises";
+import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { TraceStore } from "./store";
@@ -104,7 +104,7 @@ describe("TraceStore", () => {
     expect(await store.list({ serviceName: "agent-3" })).toEqual([]);
   });
 
-  test("list filters by time window and sorts newest first", async () => {
+  test("list filters by time window, sorts newest first, and caps to limit", async () => {
     const oldNano = `${BigInt(Date.now() - 24 * 60 * 60 * 1000) * 1_000_000n}`;
     await store.append(payload(TRACE_A, { startNano: oldNano }));
     await store.append(payload(TRACE_B));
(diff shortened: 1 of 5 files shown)


tejaskash:
P2, trace filtering: `extractTraceMeta` now collects every participating `service.name`, so the filter matches any of them. Test added.

src/core/dev/otel/store.ts
```diff
@@ -0,0 +1,129 @@
...
+      if (!meta.traceId) continue;
+      if (meta.lastSeen < start || meta.firstSeen > end) continue;
+      if (options.serviceName && !meta.serviceNames.includes(options.serviceName)) continue;
+
+      summaries.push({
+        traceId: meta.traceId,
+        timestamp: new Date(meta.lastSeen).toISOString(),
+        sessionId: meta.sessionId,
+        spanCount: String(meta.spanCount),
+        ...buildTraceDetail(trace.resourceSpans, trace.resourceLogs),
```
notgitika: every summary carries the full detail, and there's no cap on trace count. and the inspector re-polls this after every invocation. Nothing prunes the dir either. Since `ListTracesOptions` is new here, a limit for newest-N seems like the cheap fix we should try. 

Side note: `get()` ends up with no frontend caller because of this.

src/core/dev/otel/store.ts
```diff
@@ -0,0 +1,129 @@
...
+ */
+export class TraceStore {
+  constructor(private readonly directory: string) {}
+
+  /**
+   * Persist one OTLP export payload, partitioned by trace id so a batch that
+   * carries several traces lands in each trace's own file. Spans and log
+   * records without a trace id are dropped.
+   */
+  public async append(payload: OtlpPayload): Promise<void> {
```
notgitika, line 45: this now propagates fs errors where the old persistOtlp logged `Failed to persist OTLP: …` and moved on. In #1980, ingest doesn't catch, so respond turns it into a 500, project dev command survives, but the user sees nothing and the exporter just retries forever. can we catch it in one place or the other?

src/core/dev/otel/store.test.ts
notgitika, line 1: can we add a test for `get()` that asserts whether it merges spans across apends?

src/core/dev/otel/transforms.ts
```diff
@@ -0,0 +1,316 @@
...
+export function nanoToMs(nano: string | undefined): number {
+  if (!nano) return 0;
+  return Math.floor(Number(nano) / 1_000_000);
+}
+
+/**
+ * Normalize a trace/span id that may be base64 (protobuf JSON conversion) or
+ * already hex (JSON ingest) into lowercase hex.
+ */
+export function hexFromB64OrString(value: string | undefined): string {
```
notgitika, line 227: unrelated, i thought this said `B640r` and was wondering what type of input it was :(

src/core/dev/otel/transforms.ts
```diff
@@ -0,0 +1,316 @@
...
+      }
+    }
+  }
+
+  for (const resourceLog of resourceLogs) {
+    const service = getResourceAttribute(resourceLog.resource, "service.name");
+    if (service) services.add(service);
+    for (const scopeLog of resourceLog.scopeLogs ?? []) {
+      for (const record of scopeLog.logRecords ?? []) {
+        meta.spanCount++;
```
notgitika: this counts log records into spanCount, and above in line 33 counts spans before buildTraceDetail filters transport noise. was confused so had my agent try it out.

so a trace with 1 agent span + 4 http send spans reports 5. The inspector prints this straight out as {trace.spanCount} spans (agent-inspector/.../TraceRow.tsx:23), so the list says "5 spans" over a 1-span waterfall. 

can we count after filtering, or rename to something like recordCount?

src/core/dev/otel/transforms.ts
```diff
@@ -0,0 +1,316 @@
...
+  const flat: Record<string, unknown> = {};
+  for (const attribute of attributes) {
+    if (!attribute.value) continue;
+    const value = attribute.value;
+    if (value.stringValue !== undefined) flat[attribute.key] = value.stringValue;
+    else if (value.intValue !== undefined) flat[attribute.key] = Number(value.intValue);
+    else if (value.doubleValue !== undefined) flat[attribute.key] = value.doubleValue;
+    else if (value.boolValue !== undefined) flat[attribute.key] = value.boolValue;
+    else if (value.arrayValue?.values) {
+      flat[attribute.key] = value.arrayValue.values.map(
```
notgitika: `extractAnyValue` below already handles all of this properly, could we just call it here?

notgitika: as written this branch turns array ints into strings (["42"] instead of [42]), while line 255 does Number() for scalar ints, so the two come out as different types. There's also no kvlist branch anywhere in the chain, so kvlist attributes vanish entirely. no key in the output at all.


the test would need updating too: `transforms.test.ts:220` asserts a: ["y", "7"], should be a: ["y", 7].

src/core/dev/otel/transforms.test.ts
```diff
@@ -0,0 +1,243 @@
...
+    expect(
+      flattenAttributes([
+        { key: "s", value: { stringValue: "x" } },
+        { key: "i", value: { intValue: "42" } },
+        { key: "d", value: { doubleValue: 1.5 } },
+        { key: "b", value: { boolValue: true } },
+        { key: "a", value: { arrayValue: { values: [{ stringValue: "y" }, { intValue: "7" }] } } },
+        { key: "skipped" },
+      ]),
+    ).toEqual({ s: "x", i: 42, d: 1.5, b: true, a: ["y", "7"] });
```
notgitika: this needs fixing based on comment in the `transforms.ts` file

src/core/dev/otel/types.ts
```diff
@@ -0,0 +1,65 @@
...
+  arrayValue?: { values?: OtlpAttributeValue[] };
+  kvlistValue?: { values?: OtlpAttribute[] };
+}
+
+export interface OtlpAttribute {
+  key: string;
+  value?: OtlpAttributeValue;
+}
+
+export type OtlpAttributes = OtlpAttribute[] | Record<string, unknown>;
```
Hweinstock: why do we need to accept both formats?

src/core/dev/otel/types.ts
```diff
@@ -0,0 +1,65 @@
...
+export interface OtlpResource {
+  attributes?: OtlpAttributes;
+}
+
+export interface OtlpSpan {
+  traceId?: string;
+  spanId?: string;
+  parentSpanId?: string;
+  name?: string;
+  kind?: number | string;
```
Hweinstock, line 32: same question here

src/core/dev/otel/store.ts
```diff
@@ -0,0 +1,129 @@
...
+  endTime?: number;
+}
+
+/**
+ * Append-only local trace storage: one JSON Lines file per trace (named by its
+ * trace id), each line a per-trace slice of an OTLP export payload. No in-memory
+ * state — reads go to disk on demand, which is fine because the inspector only
+ * fetches traces on user actions. Malformed files and lines are skipped, never fatal.
+ */
+export class TraceStore {
```
Hweinstock, line 37: would it make sense for this to implement an interface and have consumers depend on that? My thinking is that would make it easier for us to change the implementation here without breaking consumers.

src/core/dev/otel/store.ts
```diff
@@ -0,0 +1,129 @@
...
+    const start = options.startTime ?? now - DEFAULT_LIST_WINDOW_MS;
+    const end = options.endTime ?? now;
+
+    const summaries: TraceSummary[] = [];
+    for (const file of await this.traceFiles()) {
+      const trace = await this.readTraceFile(file);
+      if (!trace) continue;
+
+      const meta = extractTraceMeta(trace.resourceSpans, trace.resourceLogs);
+      if (!meta.traceId) continue;
```
Hweinstock, line 73: should we log a warning or something here? Also, is there a reasonable explanation for a trace missing a traceId?

src/core/dev/otel/store.ts
```diff
@@ -0,0 +1,129 @@
...
+  public async get(traceId: string): Promise<TraceDetail | undefined> {
+    const trace = await this.readTraceFile(`${sanitize(traceId)}${OTLP_EXT}`);
+    if (!trace) return undefined;
+    return buildTraceDetail(trace.resourceSpans, trace.resourceLogs);
+  }
+
+  private async traceFiles(): Promise<string[]> {
+    try {
+      return (await readdir(this.directory)).filter((file) => file.endsWith(OTLP_EXT));
+    } catch {
```
Hweinstock: i assume this is to handle the case where the directory doesn't exist yet, but could it mask other failures (i.e. permissions, wrong path, etc.)?

src/core/dev/otel/store.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+
+  private async readTraceFile(
+    fileName: string,
+  ): Promise<{ resourceSpans: OtlpResourceSpan[]; resourceLogs: OtlpResourceLog[] } | undefined> {
+    let content: string;
+    try {
+      content = await readFile(join(this.directory, fileName), "utf8");
+    } catch {
```
Hweinstock: same comment here. I think we should handle the failures explicitly and let the other bubble up.

src/core/dev/otel/store.ts
```diff
@@ -0,0 +1,129 @@
...
+    const resourceSpans: OtlpResourceSpan[] = [];
+    const resourceLogs: OtlpResourceLog[] = [];
+    for (const line of content.split("\n")) {
+      if (!line.trim()) continue;
+      try {
+        const payload = JSON.parse(line) as OtlpPayload;
+        if (payload.resourceSpans) resourceSpans.push(...payload.resourceSpans);
+        if (payload.resourceLogs) resourceLogs.push(...payload.resourceLogs);
+      } catch {
+        // Skip malformed lines — a partially written line must not break reads.
```
Hweinstock, line 130: would this be a useful place to log?

src/core/dev/otel/store.ts
tejaskash: Added `limit`. Kept `get()` — the inspector reads detail through it. Can slim `list()` to summaries if youd

src/core/dev/otel/store.ts
tejaskash: Handled in #1980 — ingest catches, acks 200 so the exporter stops retrying, warns once. Store only swallows ENOENT.

src/core/dev/otel/store.test.ts
tejaskash: Added a get test — two appends, both spans back, plus the unknown-id case.

src/core/dev/otel/transforms.ts
tejaskash: Fixed — count comes from the filtered detail now, so it matches the waterfall.

src/core/dev/otel/transforms.ts
tejaskash: Done — `flattenAttributes` runs every value through `extractAnyValue`, so top-level kvlist no longer drops. Test updated.

src/core/dev/otel/transforms.test.ts
tejaskash: Fixed.

src/core/dev/otel/types.ts
tejaskash: Fixed to just `OtlpAttribute[]`.

src/core/dev/otel/types.ts
tejaskash: Fixed, `normalizeSpanKind` handles both.

src/core/dev/otel/store.ts
tejaskash: Interfaces live consumer-side here, not in Core. One impl doesn't warrant a producer-side interface imo, but happy to discuss

src/core/dev/otel/store.ts
tejaskash: No traceId means an empty/corrupt file, so we skip it. Core has no logger by design; the real failure (can't persist) warns in #1980.

src/core/dev/otel/store.ts
tejaskash: Good catch, fixed.

src/core/dev/otel/store.ts
tejaskash: Fixed

src/core/dev/otel/store.ts
tejaskash: Addressed in #1980

src/core/dev/otel/store.ts
Hweinstock: > Core has no logger by design

what exactly do you mean here? I thought we did wire it in at the top level https://github.com/aws/agentcore-cli/blob/e99b413582f387dcce420f73625156812b36fe8c/src/index.ts#L73?

src/core/dev/otel/transforms.ts
Hweinstock: i thought the same too. Is it worth explicitly writing out Base64 to make this clearer?

Hweinstock (approved):

src/core/dev/otel/store.ts
notgitika: ah yes, you're right on `get()`, `useTraces` does call the detail endpoint, I was looking at `useInvocationTraces` mb

notgitika: your reply was cut off here btw.  what's still open there is that `limit` slices after every file is read and built +  nothing prunes the dir. can we track that as a fast follow?

src/core/dev/otel/transforms.ts
notgitika: thanks for fixing. can we add a spanCount test with actual noise spans (1 agent + 4 `http send` → "1") would keep that one from coming back

notgitika (approved):
I replied to 2 of my earlier comments. they are non-blocking and can be follow ups. 
@Hweinstock and I had a nit on a function name for readability if that could also be addressed in the follow up that would be great!

src/core/dev/otel/store.ts
tejaskash: Good catch. Tracked as a fast follow: `list()` builds every file before slicing, and nothing prunes old files. Lands with the Inspector PR.

src/core/dev/otel/transforms.ts
tejaskash: Added in #1980. A trace with 1 agent span and 4 http-send spans now asserts a count of 1.

src/core/dev/otel/store.ts
tejaskash: Fair, I was wrong that Core has no logger. But `TraceStore` still isn't given one. Write failures already surface through `onError`, and nothing reads traces until the Inspector lands, which is where per-file logging belongs.

src/core/dev/otel/transforms.ts
tejaskash: Renamed `hexFromB64OrString` to `hexFromBase64OrHex` in #1980.

PR #2046, opened 2026-08-19
feat(project): add 'project add credentials' (AgentCore Identity api-key + oauth)

tejaskash:
## What

Adds the project-spec authoring path for AgentCore Identity credential providers:

```
agentcore project add credentials api-key --name <name> [--api-key -|file://p] [--api-key-secret-reference <json>]
agentcore project add credentials oauth   --name <name> [--discovery-url u --client-id c --scopes ...]
                                          [--vendor V --provider-configuration <json>]
                                          [--client-secret -|file://p] [--client-secret-reference <json>]
```

- Custom OAuth uses guided flags. Vendored OAuth uses `--vendor` (a free string, so new vendors need no CLI release) plus a complete `--provider-configuration` JSON.
- Secret flags accept only stdin (`-`) or `file://` sources. Inline values are rejected so keys never reach shell history. A single trailing newline is stripped. Multi-line values are rejected.
- An omitted secret writes a commented placeholder to `agentcore/.env.local` (`AGENTCORE_CREDENTIAL_<NAME>[_CLIENT_SECRET]`) and prints a fill-before-deploy notice. Existing keys are never overwritten.
- Secrets Manager references (`{"secretId","jsonKey"}`) are recorded in the spec (`secretRef` / `clientSecretRef`) and skip `.env.local`.
- Schema reshape in `projectSchemas/credential.ts`: `providerConfig` (rejects nested `clientSecret` / `apiKey`), `clientId`, `secretRef` / `clientSecretRef`. The `usage` field is dropped. Deep validation lives in the schema so every writer path enforces it.

## Structure

- Each leaf command is its own file (`add/credentials/api-key`, `add/credentials/oauth`), matching the command tree. Shared helpers live in `add/credentials/shared.ts`.
- `.env.local` writes go through an `EnvLocalFile` class with `upsert` and `rollback`, so a failed spec write reverses the secrets file in the same cleanup path as scaffolded files.

## Reuse

`SourceResolver` (stdin and file sources), `parseSecretReference`, and the identity oauth2 `config.ts` helpers (`parseProviderConfigFlags`, `validateProviderConfigMode`). `buildProviderConfigInput` is not used here, because SDK input construction happens at deploy time.

## Not in this PR

- Deploy-side work. `project deploy` already throws `NotImplementedError`. The lifecycle design (CFN-owned providers via the new `CfnApiKeyCredentialProvider` / `CfnOAuth2CredentialProvider` L1s, convergent pre and post-deploy secret sync from `.env.local`, hash tracking in deployed state) lands with deploy.
- The `@aws/agentcore-cdk` credential construct and the schema mirror in that repo (L3 PR 333).
- README. The `project` command tree is undocumented there today, so no partial documentation was added.
- Migration for providers created imperatively by the current CLI (a name collision on first CFN create) is a deploy-task edge case.

## Testing

- Command-seam tests drive the real router, handler, `FsProjectManager`, and filesystem across the three credential shapes and the failure paths. Schema tests cover the reshape. `EnvLocalFile` has unit tests for rollback (delete when created, restore prior content, no-op when nothing written). Full suite: 1605 pass.
- E2E on the built artifact (`bun run build`, `node dist/index.js`): a real project, all three credential shapes, and the inline-secret and duplicate-name failures exit non-zero with no partial writes.


tejaskash:
Additional E2E verification: built the artifact and ran 'project build' (real npm install + CDK synth via published @aws/agentcore-cdk 0.1.0-alpha.45) with all three credential shapes in agentcore.json. Synth passes; the published ConfigIO strips the new fields harmlessly since nothing consumes them before deploy work lands.

One live contract drift found and contained in 1cd5d140: ConfigIO requires credential names of 3-255 chars while the CLI allowed 1-128, so a 2-char name passed 'add' and then broke 'build'. The CLI now enforces the intersection (3-128). The L3 refactor-branch follow-up should align its name rule with the service's 1-128 and add the new fields (providerConfig, clientId, secretRef, clientSecretRef).

src/core/project/manager.tsx
```diff
@@ -197,6 +204,23 @@ export class FsProjectManager implements ProjectManager {
       );
       throw err;
     }
+
+    if (input.resourceType === "credential" && input.envEntries?.length) {
```
Hweinstock: Is there a way we can handle this in the same pattern as scaffolding such that it:
- rolls back entire process on failure. 
- avoids a second conditional branch in this function. 

One potential way I see to do this would be to implement some interface:
```
interface LocalEnvAccessor {
   upsert(entries: LocalEnvEntries[]): Promise<LocalEnvAccessor>;
   rollback(silent?: boolean): Promise<void>
}
```

Then here, we instantiate it with the project, call upsert in the credential scaffold case, then call rollback where we rollback the scaffolding. 

This would also move the envLocal.ts to a class implementation, which feels consistent with how we've done things like `SourceResolver`.

src/handlers/project/add/credentials/index.ts
```diff
@@ -0,0 +1,233 @@
...
+import { SourceResolver } from "../../../../io";
+import { parseSecretReference } from "../../../identity/parser";
+import {
+  parseProviderConfigFlags,
+  validateProviderConfigMode,
+} from "../../../identity/oauth2-credential-provider/config";
+import type { AddProjectResourceConfig } from "../types";
+import type { AddResourceInput, EnvLocalEntry } from "../../types";
+
+export function createAddCredentialsHandler(config: AddProjectResourceConfig): Router {
```
Hweinstock, line 6: should we continue to match the command tree with the file tree + 1:1 of file to handler pattern?

src/projectSchemas/credential.ts
```diff
@@ -1,8 +1,11 @@
 import z from "zod";
 import { PaymentProviderSchema } from "./payment";
+// Min 3 keeps names inside what @aws/agentcore-cdk's ConfigIO accepts (3-255)
```
Hweinstock: i feel like this comment is at risk of becoming stale if the referenced implementation changes.

src/core/project/manager.tsx
```diff
@@ -197,6 +204,23 @@ export class FsProjectManager implements ProjectManager {
...
+      const envPath = join(project.rootPath, ENV_LOCAL_RELATIVE_PATH);
+      yield { message: `Updating secrets file at '${envPath}'` };
+      const { skipped } = await upsertEnvLocalEntries(envPath, input.envEntries);
+      for (const key of skipped) {
+        yield { message: `'${key}' already exists in ${ENV_LOCAL_RELATIVE_PATH}; left unchanged` };
+      }
+      if (await ensureGitignoreCoversEnvLocal(project.rootPath)) {
```
Hweinstock: don't we already scaffold the project with it in the .gitignore here https://github.com/aws/agentcore-cli/blob/14e04408d23d3ec3b9fc822a75355c540d3344ee/src/assets/templates/shared/gitignore.template#L2? If the user removes it later, that feels like their own (wrong) choice.

src/handlers/project/project.test.ts
```diff
@@ -13,8 +14,10 @@ import {
 import { DeserializationError, InputValidationError } from "../../errors";
 import { FsReadWriteJson, type ReadWriteJson } from "../../io";
 
-async function run(args: string[], opts?: { core?: TestCoreClient }) {
+async function run(args: string[], opts?: { core?: TestCoreClient; stdin?: string }) {
   const io = testIO();
+  // testIO's stdin is a PassThrough; pre-filling and ending it simulates piped input.
+  if (opts?.stdin !== undefined) (io.io.stdin as unknown as PassThrough).end(opts.stdin);
```
Hweinstock: whats the difference between this pattern and https://github.com/aws/agentcore-cli/blob/0cd32ba1f9d8a52f4309c1418baafa1b2b4ee0d9/src/cli/commands/invoke/__tests__/resolve-prompt.test.ts#L30? 

Maybe OOS here, but is there a way we can standardize on something in https://github.com/aws/agentcore-cli/blob/33e2d2f8c5e698fb19d270c033fa013161477a78/src/testing/testIO.tsx#L36 so every consumer doesn't need to reinvent?

src/core/project/manager.tsx
tejaskash: Fixed

src/handlers/project/add/credentials/index.ts
tejaskash: Fixed

src/projectSchemas/credential.ts
tejaskash: Fixed

src/core/project/manager.tsx
tejaskash: Fixed

src/handlers/project/project.test.ts
tejaskash: Fixed

src/core/project/envLocal.ts
```diff
@@ -0,0 +1,80 @@
...
+    if (this.snapshot === undefined) return;
+    if (this.snapshot === null) await rm(this.path, { force: true });
+    else await atomicWrite(this.path, this.snapshot);
+  }
+}
+
+/** Reads a file, returning null when it does not exist so callers can tell empty from absent. */
+async function readOrNull(path: string): Promise<string | null> {
+  try {
+    return await readTextFile(path);
```
aidandaly24: I tried this through the built CLI with `node dist/index.js`, and it fails here with `Bun is not defined` because `readTextFile()` uses `Bun.file()`. I think this needs a Node-compatible read path, otherwise the Bun tests pass but the published artifact cannot add a credential that writes to `.env.local`.

src/core/project/envLocal.ts
```diff
@@ -0,0 +1,80 @@
...
+    const written: string[] = [];
+    const skipped: string[] = [];
+    let content = existing ?? "";
+    for (const entry of entries) {
+      if (existingKeys.has(entry.key)) {
+        skipped.push(entry.key);
+        continue;
+      }
+      const separator = content === "" || content.endsWith("\n") ? "" : "\n";
+      content += `${separator}# ${entry.comment}\n${entry.key}=${entry.value ?? ""}\n`;
```
aidandaly24: I think we need to quote or escape these values before writing them as dotenv. For example, `left#right` is written as-is, but `parseEnv()` reads it back as only `left`. Surrounding spaces are changed too. Silently changing credential material will probably make the provider fail later and be difficult to diagnose.

src/handlers/project/add/credentials/shared.ts
```diff
@@ -0,0 +1,70 @@
+import { ProjectKey, type Context } from "../../../../router";
+import { InputValidationError } from "../../../../errors";
+import { SourceResolver } from "../../../../io";
+import { parseSecretReference } from "../../../identity/parser";
+import type { AddProjectResourceConfig } from "../types";
+import type { AddResourceInput } from "../../types";
+
+/** Derives the .env.local variable name a credential's secret is stored under. */
+export function credentialEnvVarName(credentialName: string, suffix = ""): string {
+  return `AGENTCORE_CREDENTIAL_${credentialName.replace(/-/g, "_").toUpperCase()}${suffix}`;
```
aidandaly24, line 9: One edge case here: `foo-bar` and `foo_bar` are both valid credential names, but they both become `AGENTCORE_CREDENTIAL_FOO_BAR`. I tried adding both and the second command succeeds, but it reuses the first secret because the existing key is skipped. Maybe normalized-name collisions should be rejected, or the mapping should remain one-to-one? What do you think?

src/core/project/manager.tsx
aidandaly24: I think there is still one rollback case here. `.env.local` is updated before the complete spec is validated, but rollback only runs when `json.write` fails. I tried the invalid name `ab`: the command failed validation and left `agentcore.json` unchanged, but the environment entry remained. Could validation happen before the env write, or be included in the same rollback boundary?

src/core/project/envLocal.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    const written: string[] = [];
+    const skipped: string[] = [];
+    let content = existing ?? "";
+    for (const entry of entries) {
+      if (existingKeys.has(entry.key)) {
+        skipped.push(entry.key);
+        continue;
+      }
+      const separator = content === "" || content.endsWith("\n") ? "" : "\n";
```
jariy17, line 53: nit: Leave a code comment show the format if you can.
// # <entry.comment>
<entry.key>=<entry.value>

src/core/project/envLocal.ts
```diff
@@ -0,0 +1,80 @@
...
+  /** Restores the file to its pre-`upsert` state; a no-op when `upsert` wrote nothing. */
+  async rollback(): Promise<void> {
+    if (this.snapshot === undefined) return;
+    if (this.snapshot === null) await rm(this.path, { force: true });
+    else await atomicWrite(this.path, this.snapshot);
+  }
+}
+
+/** Reads a file, returning null when it does not exist so callers can tell empty from absent. */
+async function readOrNull(path: string): Promise<string | null> {
```
jariy17: private this function if no class uses.

src/core/project/envLocal.ts
```diff
@@ -0,0 +1,80 @@
...
+  get path(): string {
+    return join(this.rootPath, ENV_LOCAL_RELATIVE_PATH);
+  }
+
+  /**
+   * Appends entries, creating the file when missing. Keys that already exist
+   * are left unchanged so user-managed values survive re-runs. Returns the keys
+   * written and those skipped.
+   */
+  async upsert(entries: EnvLocalEntry[]): Promise<{ written: string[]; skipped: string[] }> {
```
jariy17: This doesn't update existing keys, it just skips. So the word upsert doesn't work here. Maybe insertIfNew...

src/handlers/project/add/credentials/api-key/index.ts
```diff
@@ -0,0 +1,61 @@
...
+        "api-key-secret-reference",
+        'external secret reference JSON: {"secretId":"<arn>","jsonKey":"<key>"}',
+        z.string().optional(),
+      ),
+    ],
+    handle: async (ctx, flags) => {
+      if (!flags.name)
+        throw new InputValidationError("required option '--name <name>' not specified");
+
+      const secretRef = parseExclusiveSecretRef(
```
jariy17, line 31: Should we combine these flags because they do the same thing and only one of them can be used?

src/handlers/project/add/credentials/oauth/index.ts
```diff
@@ -0,0 +1,115 @@
+import z from "zod";
```
jariy17, line 1: This would be a good example of using templates. We can provide default templates for Google Oauth, Okta and so on.

src/core/project/manager.tsx
```diff
@@ -269,6 +296,8 @@ function toProjectSpecKey(resourceType: ProjectResource) {
       return "harnesses";
     case "runtime":
       return "runtimes";
+    case "credential":
```
jariy17, line 337: Side Note: I really want to abstract this.

src/handlers/project/add/credentials/shared.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+  return parseSecretReference(refFlag, refValue);
+}
+
+/**
+ * Resolves a secret flag while refusing inline values: an inline secret leaks
+ * into shell history and process listings, so only stdin and files are allowed.
+ * A single trailing newline is stripped (echo and editors add one).
+ */
+export async function resolveSecretFlag(
```
jariy17: I feel like this should be part of the SourceResolver named resolveSecretValueFlags because it could be use in other locations

jariy17 (changes requested):
Pretty good beside the upsert Name function

src/core/project/envLocal.ts
tejaskash: Good catch. The root cause was in `readTextFile` itself (`src/io/fileRead.ts`), which called `Bun.file`, a global that exists only under the Bun runtime. The published bundle targets node, so this broke every caller of that function, not only this path (`eval.tsx` reads files the same way). Fixed it there by reading through `node:fs/promises`, which keeps the `AbortSignal` support. Verified the built artifact with `node dist/index.js`.

src/core/project/envLocal.ts
tejaskash: Fixed. Values are single-quoted before they are written. `node:util` `parseEnv` treats everything inside single quotes as literal, so `left#right`, leading and trailing spaces, `$`, and backslashes all read back byte for byte. The one value single quotes cannot represent is a value that itself contains a single quote, so that case is rejected with a message that points the caller at a Secrets Manager reference instead. A `parseEnv` round-trip test covers it.

src/handlers/project/add/credentials/shared.ts
tejaskash: Fixed. `addCredentialToProject` now rejects a new name that derives the same environment variable as an existing credential, and the error names both. Test added in `project.test.ts`.

src/core/project/manager.tsx
tejaskash: Fixed. The whole-spec `safeParse` now runs inside the same `try` as the write, so a rejected spec runs the `.env.local` rollback together with the scaffolded-file cleanup rather than leaving the staged secret behind.

src/core/project/envLocal.ts
tejaskash: Fixed.

src/core/project/envLocal.ts
tejaskash: Fixed. `readOrNull` is now a private method on `EnvLocalFile`.

src/core/project/envLocal.ts
tejaskash: Fixed. Renamed to `insertIfNew`.

src/handlers/project/add/credentials/api-key/index.ts
tejaskash: The two flags carry different meanings, so I kept them apart. `--api-key` takes the secret value itself (from stdin or a file) and the CLI stores it in `.env.local`. `--api-key-secret-reference` names a secret the caller already keeps in Secrets Manager and stores no value. `parseExclusiveSecretRef` makes sure only one is given. Folding them into one flag would force the CLI to guess whether the argument is a reference or a raw secret, and that ambiguity is how a real secret ends up read as a reference or the reverse.

src/handlers/project/add/credentials/oauth/index.ts
tejaskash: Agreed that vendor templates for Google, Okta, and the rest would be a good addition. It is out of scope for this change, which lands the guided-custom and vendored paths. Noted as a follow-up.

src/core/project/manager.tsx
tejaskash: Leaving it for now. The switch is exhaustive over `ProjectResource`, and the inferred return type is what keeps unrelated spec keys (name, managedBy) from leaking in, which the doc comment above it calls out. An abstraction here would need to preserve that per-case return typing to earn its place, so I would rather keep the direct mapping until a second use appears.

src/handlers/project/add/credentials/shared.ts
tejaskash: Leaving `resolveSecretFlag` local for now. Its only callers today are the two credential handlers, which share it through `shared.ts`. Moving it onto `SourceResolver` would widen that class for a single use. When a second, unrelated caller needs the same stdin-or-file-only rule, that is the point to lift it.

jariy17: It's should be used in runtime too now: https://github.com/aws/agentcore-cli/pull/2035/changes#r3833097791. Shift it into the SourceResolver when you rebase :)

jariy17 (dismissed):

src/handlers/project/add/credentials/shared.ts
tejaskash: Done. Moved it onto `SourceResolver` as `resolveSecret` (commit a37b46d5). It throws `SourceResolutionError`, which is already a subtype of `InputValidationError`, so the io package stays free of handler-level error types. Both credential handlers now call `resolver.resolveSecret(...)`, and the runtime handler in #2035 can do the same.

jariy17 (approved):

aidandaly24 (approved):

PR #2048, opened 2026-08-19
feat(project): add `online-eval` to `project add`

jariy17:
## Command structure

`agentcore project add online-eval --help`

```
Usage: agentcore project add online-eval [options]

adds an online evaluation config to the current project

Options:
  --name <name>                          the name of the online evaluation config
  --agent <agent>                        harness/runtime name whose traffic to sample (mutually exclusive with --log-group-name)
  --endpoint <endpoint>                  the agent endpoint qualifier to scope monitoring to (requires --agent)
  --log-group-name <log-group-name...>   CloudWatch log group name(s) for custom data sources (1-5; mutually exclusive with --agent)
  --service-name <service-name...>       service name(s) to filter traces for custom data sources (requires --log-group-name)
  --evaluator <evaluator...>             evaluator name(s), Builtin.* IDs, or ARNs to apply
  --sampling-rate <sampling-rate>        percentage of sessions to sample (0.01-100)
  --description <description>            a description of the config's monitoring purpose
  --enable-on-create <enable-on-create>  enable evaluation immediately after deploy (default true; pass false to add it paused)
  --tags <tags>                          tags to apply (JSON object of key/value strings)
  -h, --help                             display help for command
```

---
**Related:** aws/agentcore-l3-cdk-constructs#335 — `OnlineEvaluationConfig` `agent` resolves runtimes only (harness targets unsupported). The `--agent` help text here says "runtime" to match.

### DIFF
--- a/src/core/project/manager.tsx
+++ b/src/core/project/manager.tsx
@@ -161,6 +161,7 @@ export class FsProjectManager implements ProjectManager {
         );
       }
       case "config-bundle":
+      case "online-eval":
         newResources.push(resourceConfig);
         break;
 
@@ -289,5 +290,7 @@ function toProjectSpecKey(resourceType: ProjectResource) {
       return "runtimes";
     case "config-bundle":
       return "configBundles";
+    case "online-eval":
+      return "onlineEvalConfigs";
   }
 }
--- a/src/handlers/project/add/online-eval/index.ts
+++ b/src/handlers/project/add/online-eval/index.ts
@@ -1,0 +1,93 @@
+import z from "zod";
+import { createHandler, flag, ProjectKey } from "../../../../router";
+import { InputValidationError } from "../../../../errors";
+import { OnlineEvalConfigSchema } from "../../../../projectSchemas/online-eval-config";
+import { parseJsonFlag } from "../../../utils";
+import type { AddProjectResourceConfig } from "../types";
+
+export const createAddOnlineEvalHandler = (config: AddProjectResourceConfig) =>
+  createHandler({
+    name: "online-eval",
+    description: "adds an online evaluation config to the current project",
+    flags: [
+      flag("name", "the name of the online evaluation config", z.string().optional()),
+      flag(
+        "agent",
+        "harness/runtime name whose traffic to sample (mutually exclusive with --log-group-name)",
+        z.string().optional(),
+      ),
+      flag(
+        "endpoint",
+        "the agent endpoint qualifier to scope monitoring to (requires --agent)",
+        z.string().optional(),
+      ),
+      flag(
+        "log-group-name",
+        "CloudWatch log group name(s) for custom data sources (1-5; mutually exclusive with --agent)",
+        z.array(z.string()).optional(),
+      ),
+      flag(
+        "service-name",
+        "service name(s) to filter traces for custom data sources (requires --log-group-name)",
+        z.array(z.string()).optional(),
+      ),
+      flag(
+        "evaluator",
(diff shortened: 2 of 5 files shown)


src/core/project/manager.tsx
```diff
@@ -155,6 +166,12 @@ export class FsProjectManager implements ProjectManager {
         });
         break;
       }
+      case "online-eval": {
```
nborges-aws: Having a default for this makes sense to avoid having cases that all have no scaffolding and directly push. But at the same time, having the default loosens our validations since new resourceTypes will fall through to the default.

Meet in the middle: a resourceType enum, stack cases for new resourceTypes that do not need scaffolding

Hweinstock:
```

Run bun run format:check
$ prettier --check .
Checking formatting...
[warn] src/handlers/project/add/online-eval/index.test.ts
[warn] src/handlers/project/add/online-eval/index.ts
[warn] Code style issues found in 2 files. Run Prettier with --write to fix.
error: script "format:check" exited with code 1
```

I think format is failing.

src/core/project/manager.tsx
```diff
@@ -155,6 +166,12 @@ export class FsProjectManager implements ProjectManager {
...
+        // Pure spec-level resource: no files to scaffold, just record the config.
```
Hweinstock: nit: i feel like this comment is expressed by the code.

src/handlers/project/add/online-eval/index.ts
```diff
@@ -0,0 +1,93 @@
...
+          flags["enable-on-create"] === undefined
+            ? undefined
+            : flags["enable-on-create"] === "true",
+        tags: parseJsonFlag<Record<string, string>>("tags", flags["tags"]),
+      };
+
+      // OnlineEvalConfigSchema owns every cross-field rule (agent XOR log groups,
+      // evaluators XOR insights, endpoint/serviceNames prerequisites). Validate once
+      // here so the user gets one actionable message rather than a raw stack later at
+      // deploy time — json.write does not re-validate on the way to disk.
```
Hweinstock: I don't think this comment is fully accurate since we do actually validate before writing: https://github.com/aws/agentcore-cli/blob/14e04408d23d3ec3b9fc822a75355c540d3344ee/src/core/project/manager.tsx#L169. 

Its also referencing other parts of the codebase that may change, so its high risk to become stale.

Hweinstock (commented):
two nits on comments, otherwise lgtm (once format is fixed).

src/core/project/manager.tsx
jariy17: I'll remove it

src/core/project/manager.tsx
jariy17: I'll look at your config bundle PR and follow that pattern.

src/handlers/project/add/online-eval/index.ts
jariy17: I'm just gonna tell claude to stop putting code comments unless there is a good reason too.

src/core/project/manager.tsx
```diff
@@ -155,6 +155,9 @@ export class FsProjectManager implements ProjectManager {
         });
         break;
       }
+      case "online-eval":
+        newResources.push(resourceConfig);
+        break;
```
notgitika: lets just stack this on config bundle

notgitika (approved):
LGTM

nborges-aws (approved):
LGTM!

PR #2055, opened 2026-08-20
refactor(project): move build behind a project backend

notgitika:
## Summary

- introduce the `ProjectBackend` boundary selected by `managedBy`
- move CDK build and synthesis behavior behind `CdkBackend`
- preserve existing project build behavior and error handling

## Stack

This is PR 1 of 4 decomposing #2001. Review and merge bottom-up.

1. **#2055 - backend build boundary (this PR)**
2. #2056 - deploy command contract
3. #2057 - CDK Toolkit adapter
4. #2058 - safe deployment implementation

typecheck, lint, format, test all pass

PR #2056, opened 2026-08-20
feat(project): add deploy command contract

notgitika:
## Summary

- add the `project deploy` handler and `--target` contract
- define deployment target validation and deployment result types
- synchronize the closed AgentCore region enum with the current supported-region list
- wire manager/backend deploy interfaces with an intentionally unimplemented CDK operation

## Stack

This is PR 2 of 4 decomposing #2001. The command remains nonfunctional until the final PR. Review and merge bottom-up.

2. **#2056 - deploy command contract (this PR)**
3. #2057 - CDK Toolkit adapter
4. #2058 - safe deployment implementation

### DIFF
--- a/src/core/project/manager.tsx
+++ b/src/core/project/manager.tsx
@@ -36,6 +36,9 @@ import type { HarnessSpecSchema } from "../../projectSchemas/harness";
 import z from "zod";
 import { CdkBackend } from "./backends/cdk";
 import type { ProjectBackend } from "./backends/types";
+import { AwsDeploymentTargetsSchema } from "../../projectSchemas/aws-targets";
+
+const TARGETS_EXAMPLE = '[{ "name": "default", "account": "111122223333", "region": "us-east-1" }]';
 
 type ProjectManagerConfig = {
   logger: Logger;
@@ -311,12 +314,39 @@ export class FsProjectManager implements ProjectManager {
     yield* this.backendFor(project).build(project);
   }
 
+  // Resolves the named target from aws-targets.json before handing off, so the
+  // backend receives a fully resolved account and region and never has to know
+  // how targets are stored. The backend owns everything after that point.
   public async *deploy(
-    _project: Project,
-    _input: DeployProjectInput,
+    project: Project,
+    input: DeployProjectInput,
   ): AsyncGenerator<ProjectEvent, DeployResult> {
-    yield* [];
-    throw new NotImplementedError("agentcore project deploy is not implemented yet");
+    const targetsPath = join(project.rootPath, "agentcore", "aws-targets.json");
+    if (!existsSync(targetsPath)) {
+      throw new ProjectStateError(
+        `No deployment targets are configured for project '${project.name}'. ` +
+          `Add ${targetsPath}, for example:\n\n${TARGETS_EXAMPLE}`,
+      );
+    }
+
+    const targets = await this.json.read(targetsPath, AwsDeploymentTargetsSchema);
+
+    if (targets.length === 0) {
+      throw new ProjectStateError(
+        `No deployment targets are configured for project '${project.name}'. ` +
+          `Add at least one to ${targetsPath}, for example:\n\n${TARGETS_EXAMPLE}`,
+      );
+    }
+
+    const target = targets.find((candidate) => candidate.name === input.target);
+    if (!target) {
+      throw new ProjectStateError(
+        `Project '${project.name}' has no deployment target named '${input.target}'. ` +
+          `${targetsPath} defines: ${targets.map(({ name }) => name).join(", ")}.`,
+      );
+    }
+
+    return yield* this.backendFor(project).deploy(project, { target });
   }
 
   private backendFor(project: Project): ProjectBackend {
--- a/src/core/project/manager.tsx
+++ b/src/core/project/manager.tsx
@@ -4,6 +4,8 @@ import { join, relative } from "node:path";
 import type {
   AddResourceInput,
   CreateProjectInput,
+  DeployProjectInput,
+  DeployResult,
   ResolveProjectInput,
   Project,
   ProjectManager,
@@ -307,6 +309,14 @@ export class FsProjectManager implements ProjectManager {
 
   public async *build(project: Project): AsyncGenerator<ProjectEvent, void> {
     yield* this.backendFor(project).build(project);
+  }
+
+  public async *deploy(
+    _project: Project,
+    _input: DeployProjectInput,
+  ): AsyncGenerator<ProjectEvent, DeployResult> {
+    yield* [];
+    throw new NotImplementedError("agentcore project deploy is not implemented yet");
   }
 
   private backendFor(project: Project): ProjectBackend {
--- a/src/handlers/project/deploy/index.test.ts
+++ b/src/handlers/project/deploy/index.test.ts
@@ -1,64 +1,53 @@
-import { describe, expect, test } from "bun:test";
+import { afterEach, describe, expect, test } from "bun:test";
+import { mkdtemp, rm, writeFile } from "node:fs/promises";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
 import { createRootHandler } from "../../index";
-import { ProjectSpecSchema } from "../../../projectSchemas/project";
 import {
   createSilentLogger,
   TestCoreClient,
   TestGlobalConfigAccessor,
   testIO,
 } from "../../../testing";
-import type {
-  AddResourceInput,
-  CreateProjectInput,
-  DeployProjectInput,
-  DeployResult,
-  Project,
-  ProjectEvent,
-  ProjectManager,
-} from "../types";
+import type { DeployBackendInput, ProjectBackend } from "../../../core/project";
+import type { AwsDeploymentTarget } from "../../../projectSchemas/aws-targets";
+import type { DeployResult, Project, ProjectEvent } from "../types";
 
-const project: Project = {
-  name: "orders",
-  rootPath: "/workspace/orders",
-  spec: ProjectSpecSchema.parse({ name: "orders", version: 1 }),
+const DEFAULT_TARGET: AwsDeploymentTarget = {
+  name: "default",
+  account: "111122223333",
+  region: "us-east-1",
 };
+const STAGING_TARGET: AwsDeploymentTarget = {
+  name: "staging",
+  account: "444455556666",
+  region: "eu-west-1",
+};
+const TARGETS = [DEFAULT_TARGET, STAGING_TARGET];
 
 /**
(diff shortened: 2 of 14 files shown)


src/handlers/project/project.test.ts
```diff
@@ -921,3 +921,15 @@ describe("project build", () => {
     await expect(run(["build"])).rejects.toThrow(/npm install/);
   });
 });
+
+describe("project deploy", () => {
+  test("requires an AgentCore project", async () => {
```
Hweinstock, line 618: likely OOS here, but we should probably have a test that does this for all project commands not just deploy.

src/handlers/project/types.ts
```diff
@@ -28,6 +28,16 @@ export type ProjectEvent = {
   message: string;
 };
 
+export type DeployProjectInput = {
+  /** Name of the aws-targets.json entry to deploy. */
+  target: string;
+};
+
+export type DeployResult = {
+  /** Outputs returned by the deployed stack, keyed by CloudFormation output name. */
```
Hweinstock: for non-IAC / terraform backends, what would this represent? I don't think we need to know exactly, but want to make sure this api is not coupling us to CDK.

src/handlers/project/deploy/index.test.ts
```diff
@@ -0,0 +1,93 @@
...
+} from "../types";
+import { createDeployProjectHandler } from ".";
+
+const project: Project = {
+  name: "orders",
+  rootPath: "/workspace/orders",
+  spec: ProjectSpecSchema.parse({ name: "orders", version: 1 }),
+};
+
+function fakeProjectManager(result: DeployResult, events: ProjectEvent[] = []) {
```
Hweinstock: is it possible to re-use https://github.com/aws/agentcore-cli/blob/33e2d2f8c5e698fb19d270c033fa013161477a78/src/testing/TestCoreClient.tsx#L1951-L1960 like how https://github.com/aws/agentcore-cli/blob/33e2d2f8c5e698fb19d270c033fa013161477a78/src/handlers/project/project.test.ts#L88 does it?

Hweinstock (commented):
small questions, but approach lgtm. I can take another look once the dependent changes are merged.

src/handlers/project/project.test.ts
notgitika: yes! I will add it to all as a follow up :)

src/handlers/project/types.ts
notgitika: yeah let me change the comment. 

The type is neutral for those backends so the comment is wrong at this point. It's just a string map so any backend can fill it, CDK from CFN outputs, Terraform from terraform output etc. nothing reads specific keys out of it, we just print them or dump them with --json. I'll reword the doc so it stops saying cfn

src/handlers/project/deploy/index.test.ts
notgitika: ah yes I should have done this. let me switch it

src/projectSchemas/aws-targets.ts
```diff
@@ -0,0 +1,59 @@
...
+  "us-west-2",
+  "us-gov-west-1",
+]);
+
+export const DeploymentTargetNameSchema = z
+  .string()
+  .min(1)
+  .max(64)
+  .regex(
+    /^[a-zA-Z][a-zA-Z0-9_-]*$/,
```
nborges-aws: Can we remove underscores from the regex? On main we do not accept underscores: [aws-targets.ts](https://github.com/aws/agentcore-cli/blob/main/src/schema/schemas/aws-targets.ts#L42)

Our CDK normalizes all underscores to hyphens, so this is more of a nit. But for robustness we should disallow upfront.

Hweinstock:
I think there are some merge conflicts: `src/handlers/project/project.test.ts`.

src/projectSchemas/aws-targets.ts
notgitika: nice catch! just getting it done now

src/core/project/manager.tsx
```diff
@@ -281,6 +286,41 @@ export class FsProjectManager implements ProjectManager {
...
+  ): AsyncGenerator<ProjectEvent, DeployResult> {
+    const targetsPath = join(project.rootPath, "agentcore", "aws-targets.json");
+    if (!existsSync(targetsPath)) {
+      throw new ProjectStateError(
+        `No deployment targets are configured for project '${project.name}'. ` +
+          `Add ${targetsPath}, for example:\n\n${TARGETS_EXAMPLE}`,
+      );
+    }
+
+    const targets = await this.json.read(targetsPath, AwsDeploymentTargetsSchema);
```
tejaskash, line 332: These validation messages never actually reach the user here. `json.read` wraps any schema failure in `DeserializationError`, whose message is just `Failed to deserialize file at "<path>"` (io/json.ts:55, errors.tsx:98) — the zod detail only goes to the debug log, and the top-level reporter prints `error.message`. So someone who typos a region or duplicates a name in this hand-edited file (which our own error text above tells them to write) just gets an opaque "failed to deserialize" with no idea which field is wrong.

We already do the right thing one method up: `addResource` rethrows `InputValidationError(z.prettifyError(...))` (manager.tsx:190). Worth mirroring that — catch `DeserializationError` here and rethrow with the prettified cause, or fix it once in `FsReadWriteJson.read` so every hand-edited file benefits. Either way we should add a handler-level test for what the user actually sees; the current schema tests pass in isolation and dont catch this.

Hweinstock: I think we should fix in the source, such that the error message includes the schema failure information (perhaps extending `DeserializationError` to take an optional detail field that corresponds to the prettified zod error in this case) , but I think that could be a follow-up.

notgitika: thanks for the catch, fixed! I also added the tests

nborges-aws (dismissed):
LGTM thanks for fixes!

src/io/json.ts
```diff
@@ -52,7 +52,10 @@ export class FsReadWriteJson implements ReadWriteJson {
           errorMessage: parseResult.error.message,
         })
         .error(`failed to validate parsed json file`);
-      throw new DeserializationError(filePath, { cause: parseResult.error });
+      throw new DeserializationError(filePath, {
+        cause: parseResult.error,
+        details: z.prettifyError(parseResult.error),
```
Hweinstock, line 57: haha I refreshed and you implemented exactly what I was thinking, crazy!

src/testing/TestCoreClient.tsx
```diff
@@ -1179,6 +1180,9 @@ export class TestGatewayClient implements CoreGatewayClient {
 type TestCoreClientOptions = {
   logger?: Logger;
   json?: ReadWriteJson;
+  // Stubs one backend rather than the whole manager, so tests keep the real
```
Hweinstock: nit: I think this is communicated by the partial and doesn't warrant a comment.

src/handlers/project/deploy/index.test.ts
```diff
@@ -0,0 +1,202 @@
...
+    await inProjectWithTargets(subject, []);
+
+    await expect(subject.run()).rejects.toThrow(/No deployment targets are configured/);
+    expect(subject.calls).toEqual([]);
+  });
+});
+
+// aws-targets.json is hand-edited, so these assert on the message the user
+// actually sees rather than on the schema in isolation: the reporter prints only
+// error.message, so a validation detail left in `cause` may as well not exist.
```
Hweinstock: wait didn't we fix this by appending the detail to the message?

src/handlers/project/deploy/index.test.ts
```diff
@@ -0,0 +1,202 @@
...
+    async *deploy(project, input) {
+      calls.push({ project, input });
+      yield* events;
+      return result;
+    },
+  };
+  return { calls, backend };
+}
+
+function harness(result: DeployResult, events: ProjectEvent[] = []) {
```
Hweinstock: what does this have to with `harness` or is this like a testing harness (unrelated to AgentCore harness)? If so, maybe we can rename to avoid confusion.

src/handlers/project/deploy/index.test.ts
```diff
@@ -0,0 +1,202 @@
...
+  const directory = await inTempDirectory();
+  await subject.create(["create", "--name", "orders", "--skip-install", "--skip-git"]);
+  const projectRoot = join(directory, "orders");
+  await writeFile(join(projectRoot, "agentcore", "aws-targets.json"), contents);
+  process.chdir(projectRoot);
+  return projectRoot;
+}
+
+/** Scaffolds a project with deployment targets and cds into it. */
+function inProjectWithTargets(
```
Hweinstock: nit: this feels like unnecessary indirection.

src/handlers/project/deploy/index.ts
```diff
@@ -1,11 +1,53 @@
...
+import { JsonKey } from "../../keys";
+import type { DeployResult, ProjectManager } from "../types";
 
-export const createDeployProjectHandler = () =>
+type DeployProjectHandlerConfig = {
+  projectManager: ProjectManager;
+  io: AppIO;
+};
+
+async function runDeploy(
```
Hweinstock: nit: is there a reason these should live in a separate functions, rather than in-lined? Do we expect them to be re-used?

Hweinstock (commented):
few comments on testing and some nits, but I like the direction!

src/io/json.ts
notgitika: 🧙🏼

src/testing/TestCoreClient.tsx
notgitika: removed

src/handlers/project/deploy/index.test.ts
notgitika: stale comment block i didnt catch this, removed

src/handlers/project/deploy/index.test.ts
notgitika: it was a testing harness I renamed it to testDeployCommand

src/handlers/project/deploy/index.test.ts
notgitika: collapsed to one helper taking file contents. now, callers pass JSON.stringify(...), so the malformed-JSON case is just another call.

src/handlers/project/deploy/index.ts
notgitika: no resuse planned so I just made them to be inline functions

tejaskash (approved):
All actionable review comments are addressed. I verified the final head against current refactor and reran the full local test, typecheck, lint, and formatting checks.

nborges-aws (approved):

PR #2057, opened 2026-08-20
feat(project): add CDK Toolkit adapter

notgitika:
## Summary

- add a backend-local adapter for `@aws-cdk/toolkit-lib`
- load the Toolkit lazily and route notifications through the existing logger at debug level
- keep the Toolkit external in the npm bundle
- add direct adapter tests without activating deployment

## Stack

This is PR 3 of 4 decomposing #2001. The adapter is backend-local rather than part of shared `io`. Review and merge bottom-up.

2. #2056 - deploy command contract
3. **#2057 - CDK Toolkit adapter (this PR)**
4. #2058 - safe deployment implementation

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -11,8 +11,16 @@ const DIST = join(REPO_ROOT, "dist");
 
 const ASSET_NAMING = "agentcore-assets/[dir]/[name].[ext]";
 
-// The Toolkit reads files relative to its package directory at runtime, so keep
-// that package intact when producing the npm bundle.
+// The Toolkit reads files relative to its own package directory at runtime
+// (bundledPackageRootDir(__dirname)), so the npm bundle keeps that package
+// intact rather than inlining it.
+//
+// This covers the npm bundle only. compile() below cannot use the same trick:
+// a standalone binary ships no node_modules, so marking the Toolkit external
+// would leave an unresolvable require. Bun instead inlines it and rewrites
+// __dirname to the *build machine's* path, which resolves on the builder and
+// fails everywhere else. Binaries therefore have to hand the Toolkit an
+// explicit copy of anything it would otherwise read from its package.
 const EXTERNAL = ["@aws-cdk/toolkit-lib"];
 
 // Shrink whitespace/syntax but keep identifiers: minified names make stack
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -11,6 +11,10 @@ const DIST = join(REPO_ROOT, "dist");
 
 const ASSET_NAMING = "agentcore-assets/[dir]/[name].[ext]";
 
+// The Toolkit reads files relative to its package directory at runtime, so keep
+// that package intact when producing the npm bundle.
+const EXTERNAL = ["@aws-cdk/toolkit-lib"];
+
 // Shrink whitespace/syntax but keep identifiers: minified names make stack
 // traces unreadable and erase error names telemetry keys on.
 const MINIFY = { whitespace: true, syntax: true, identifiers: false } as const;
@@ -54,6 +58,7 @@ async function bundle(): Promise<void> {
     outdir: DIST,
     target: "node",
     minify: MINIFY,
+    external: EXTERNAL,
   });
 
   // Mirror assets beside the emitted module for resolveAssetsRoot().
--- a/src/core/project/backends/cdk/toolkit.ts
+++ b/src/core/project/backends/cdk/toolkit.ts
@@ -1,0 +1,110 @@
+import type { IIoHost, IoMessage, Toolkit } from "@aws-cdk/toolkit-lib";
+import type { Logger } from "../../../../logging";
+
+export type CdkOperation =
+  | { kind: "bootstrap"; environments: string[]; templateFile?: string }
+  | { kind: "deploy"; stackName: string };
+
+export type CdkRunOptions = {
+  /** Synthesized cloud assembly used by deploy operations. */
+  assemblyDirectory: string;
+  /** Region used for the Toolkit's own AWS SDK calls. */
+  region: string;
+};
+
+export type CdkOutputs = Record<string, string>;
+
+export type CdkRunner = (operation: CdkOperation, options: CdkRunOptions) => Promise<CdkOutputs>;
+
+export type CdkToolkit = Pick<Toolkit, "bootstrap" | "deploy" | "fromAssemblyDirectory">;
+
+export type CdkToolkitLib = Pick<
+  typeof import("@aws-cdk/toolkit-lib"),
+  | "BaseCredentials"
+  | "BootstrapEnvironments"
+  | "BootstrapSource"
+  | "BootstrapStackParameters"
+  | "StackSelectionStrategy"
+>;
+
+export type LoadedCdkToolkit = {
+  lib: CdkToolkitLib;
+  toolkit: CdkToolkit;
+};
+
+export type CdkToolkitLoader = (ioHost: IIoHost, region: string) => Promise<LoadedCdkToolkit>;
+
+export function createCdkIoHost(logger: Logger): IIoHost {
+  const toolkitLogger = logger.child({ component: "cdk-toolkit" });
+  const notify = async (message: IoMessage<unknown>): Promise<void> => {
+    toolkitLogger
+      .child({
+        action: message.action,
+        level: message.level,
+        ...(message.code && { code: message.code }),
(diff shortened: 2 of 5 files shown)


src/core/project/backends/cdk/toolkit.ts
```diff
@@ -0,0 +1,110 @@
...
+export type CdkRunOptions = {
+  /** Synthesized cloud assembly used by deploy operations. */
+  assemblyDirectory: string;
+  /** Region used for the Toolkit's own AWS SDK calls. */
+  region: string;
+};
+
+export type CdkOutputs = Record<string, string>;
+
+export type CdkRunner = (operation: CdkOperation, options: CdkRunOptions) => Promise<CdkOutputs>;
```
tejaskash, line 18: Just curious, did you consider making this a class-based adapter rather than exposing `CdkRunner` as a function? The function works fine here, but I'd like to understand what drove this choice.

scripts/build.ts
```diff
@@ -54,6 +58,7 @@ async function bundle(): Promise<void> {
     outdir: DIST,
     target: "node",
     minify: MINIFY,
+    external: EXTERNAL,
```
tejaskash, line 69: This keeps `@aws-cdk/toolkit-lib` external for the npm bundle, but `compile()` below (the six platform binaries) doesn't pass `external`. So the moment #2058 makes this reachable from `src/index.ts`, Bun will statically bundle the toolkit into every standalone binary — and that breaks the exact thing the comment above warns about: the toolkit reads data files relative to its own package dir at runtime (e.g. `lib/api/bootstrap/bootstrap-template.yaml`), and that path doesn't exist inside a compiled binary. Marking it external in `compile()` won't rescue it either, since the binaries ship no `node_modules` to resolve against.

Since this PR is where both the dependency and the asymmetry come in, I'd rather we settle the compiled-binary deploy story here than discover it in #2058. A few options: embed the toolkit's data files through the existing asset pipeline, make deploy fail with a clear message on compiled binaries, or at the very least narrow the comment to say it only covers the npm bundle.

src/core/project/backends/cdk/toolkit.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  const source = await toolkit.fromAssemblyDirectory(options.assemblyDirectory);
+  const result = await toolkit.deploy(source, {
+    stacks: {
+      strategy: lib.StackSelectionStrategy.PATTERN_MUST_MATCH_SINGLE,
+      patterns: [operation.stackName],
+    },
+  });
+  return result.stacks[0]?.outputs ?? {};
```
tejaskash, line 114: `PATTERN_MUST_MATCH_SINGLE` throws on a non-match, so the only way `stacks` comes back empty is the toolkit's empty-template path — and in 1.38.2 that path will delete an existing stack and still return normally. When that happens, `?? {}` quietly turns it into a success with no outputs, and a caller expecting `RuntimeArn` just gets `undefined` with nothing pointing at where it went. This branch also isn't covered by a test.

Suggest asserting `result.stacks.length === 1` and throwing otherwise, then keeping `?? {}` only for the genuine case where a stack is present but has no outputs.

src/core/project/backends/cdk/toolkit.ts
notgitika: Considered it — the deciding factor was the shape of this seam rather than a preference for functions.

`CdkBackendConfig` has eight injectable seams and seven are already function types: `ProcessRunner`, `typeof requireTool`, `CdkRunner`, `CdkCredentialResolver`, `BootstrapProbe`, `AccountResolver`, `BootstrapTemplateLoader`. The one that isn't is `ReadWriteJson`, which has several methods. `CdkRunner` has one operation, so a class would be a single-method interface plus a constructor, and every test would build an object to say what a lambda already says.

The operation is a discriminated union (`bootstrap` | `deploy`), so a class wouldn't split into separate methods either — it would still be one `run` dispatching on `kind`, just behind another indirection.

Where state does exist it's already factored out: `loadCdkToolkit` owns construction, and `performCdkOperation` is a free function over an already-loaded toolkit, so the branching logic is testable without constructing anything or stubbing a class. On #2058 the loader gains a third parameter for shared credentials, and that stayed a signature change instead of constructor plumbing.

Worth revisiting if this grows a second operation that has to share state across calls — that's where a class starts earning its keep.

scripts/build.ts
notgitika: Good catch on the asymmetry — `compile()` gets no `external`, and it can't. I checked this against a real compiled binary rather than reasoning from the source, and the failure mode is slightly worse than "the path doesn't exist inside the binary": Bun rewrites `__dirname` for inlined deps to the **build machine's** absolute path. So `bundledPackageRootDir(__dirname)` resolves fine on the builder and throws `PackageManifestNotFound` everywhere else, which is exactly the shape of bug that passes CI and fails for users.

I traced every package-relative read in toolkit-lib 1.38.2 to size the actual surface:

- `lib/api/bootstrap/bootstrap-environment.js:340` and `lib/actions/bootstrap/index.js:218` — the default bootstrap template. On our path, but bypassed entirely when `templateFile` is passed.
- `lib/api/aws-auth/user-agent.js:47` — calls `bundledPackageRootDir(__dirname, false)` and degrades to `<unknown>`. Harmless.
- `lib/util/package-info.js:49,53` — reached only via `lib/api/cloud-assembly/private/prepare-source.js:185` `defaultEnvVars()`, which is the `fromCdkApp` synth path. We use `fromAssemblyDirectory`, so it's off our path.

So the bootstrap template is the only real one, and option (a) is what #2058 does: `scripts/build.ts` resolves the template out of toolkit-lib at build time and embeds it (`:55`), throws if toolkit-lib stops shipping it (`:59`), and throws again if the finished binary doesn't contain it (`:73`). At runtime `loadBootstrapTemplate()` materializes it to a temp file and passes `templateFile`, so the toolkit never reads its own package directory.

Two changes came out of this:

1. **Here** (5cc76e54): narrowed the comment to say it covers the npm bundle only, and why `compile()` can't use the same trick — marking the toolkit external in a binary that ships no `node_modules` just moves the failure to an unresolvable require.
2. **On #2058** (8484de46): `loadBootstrapTemplate()` returned `undefined` both when nothing is embedded and when the embed is missing from a binary. The first is correct — scripts and the npm bundle have `node_modules`, so the toolkit reading its own package is the right fallback. The second silently fell through to that same fallback and produced the confusing `PackageManifestNotFound`. It now treats other embedded files as proof it's a binary and fails naming the missing file.

The build-time guards mean a binary can't ship without the template, so (2) only covers the narrower case where the embed exists but stops matching the lookup — but that's the one that would otherwise surface as an unrelated error message.

src/core/project/backends/cdk/toolkit.ts
notgitika: Fixed as suggested in 5cc76e54 — asserts `result.stacks.length === 1` and throws otherwise, keeping `?? {}` only for the genuine case of a stack that deployed but declares no outputs.

Confirmed the mechanism in 1.38.2 before writing it: `toolkit.js` hits `resourceCount === 0` and either skips with `W5021` when the stack is absent, or emits `W5022` and calls `this._destroy(...)` when it exists — both return before `ret.stacks.push`. So an empty `stacks` really can mean a deletion, and the error says so, since that's the part a user can't infer from "deploy succeeded, no outputs".

Two tests cover the branch: the empty-`stacks` path fails instead of returning `{}`, and a deployed stack with no outputs still succeeds. Checked they have teeth by removing the assertion — only the new test fails.

tejaskash (approved):

src/core/project/backends/cdk/toolkit.ts
```diff
@@ -0,0 +1,126 @@
...
+    notify,
+    requestResponse: async (request) => {
+      await notify(request);
+      return request.defaultResponse;
+    },
+  };
+}
+
+/** Loads the Toolkit only when a deploy operation needs it. */
+export const loadCdkToolkit: CdkToolkitLoader = async (ioHost, region) => {
```
Hweinstock, line 60: I like how this is loaded lazily.

Hweinstock (approved):

src/core/project/backends/cdk/toolkit.ts
notgitika: yeah I had my agent test the latency with the static import and it was quite a bit higher. thanks!

Issue #2022, opened 2026-08-18
Credential provider is never deleted, despite remove credential saying the deploy will apply it


Agotfrid:
### Description

`agentcore remove credential --name <n> --yes` returns a note saying the removal will be applied to AWS on the next deploy:

```json
{"success":true,"resourceType":"credential","resourceName":"retiretest-fake-token",
 "message":"Removed credential 'retiretest-fake-token'",
 "note":"Your agent app source code has not been modified. Deploy with `agentcore deploy` to apply your removal changes to AWS."}
```

That does not happen. The API-key credential provider is never deleted from AWS. It survives the removal, the following deploy, and a full stack teardown. There appears to be no CLI path that deletes one at all, so the only way to remove it is a manual `aws bedrock-agentcore-control delete-api-key-credential-provider` call.

The misleading note is the reason I am filing this as a bug rather than a feature request. An operator who follows the CLI's own guidance will reasonably believe the credential is gone. It is not, and neither is its backing secret, which is a security-relevant leak whenever the credential holds a real token.

Two details suggest the missing deletion is an oversight rather than a deliberate design decision:

- The CLI **does** implement provider deletion for **payment** credential providers, via a direct `POST /identities/DeletePaymentCredentialProvider` call. There is no equivalent call for API-key providers anywhere in the CLI.
- The credential provider is not a CloudFormation resource, so `cdk destroy` cannot remove it even in principle. On a two-runtime Container project the stack contained 24 resources (Runtime, ECR, KMS, Lambda, Logs, IAM, CodeBuild, CustomResource, CDK Metadata) and no credential-provider resource of any type. Only an explicit API call can delete it.

### Steps to Reproduce

**1. Scaffold a project with one runtime and one API-key credential.**

```bash
agentcore create --no-agent --project-name RetireTest

agentcore add agent --type byo --name alpha_amp --language TypeScript \
  --framework Strands --model-provider Bedrock --code-location app/alpha_amp \
  --entrypoint dist/main.js --build Container --protocol HTTP

agentcore add credential --name retiretest-fake-token --type api-key --api-key not-a-real-token
```

**2. Deploy.** The output includes a `Creating credentials...` step.

```bash
agentcore deploy --target <target>
```

**3. Confirm the provider now exists in AWS.**

```bash
aws bedrock-agentcore-control list-api-key-credential-providers --region us-east-1
```

`retiretest-fake-token` is present. Correct so far.

[...]

Issue #1799, opened 2026-07-21
Hard-coded `CDKToolkit` stack name causes false bootstrap prompt


pluralistix:
### Description

`agentcore deploy` checks for a CloudFormation stack named exactly `CDKToolkit`:

```ts
export const CDK_TOOLKIT_STACK_NAME = 'CDKToolkit';

new DescribeStacksCommand({
  StackName: CDK_TOOLKIT_STACK_NAME,
});
```

This breaks in environments where CDK bootstrap infrastructure is deployed centrally, for example through CloudFormation StackSets, which create stack names such as:

```text
StackSet-CDKToolkit-9afdf51a-b1cf-4dfc-9b0c-63ccafe70cbc
```

The environment is already bootstrapped, but AgentCore still prompts:

```text
Bootstrap AWS environment
```

In interactive mode, this prompt can be cancelled and deployment continues successfully. In CI/CD, there is no clean equivalent.

## Expected behavior

AgentCore should support externally managed bootstrap infrastructure without assuming ownership of it.

A minimal fix would be:

```bash
agentcore deploy --skip-bootstrap-check
```

or a configuration option such as:

```json
{
  "bootstrap": {
    "management": "external"
  }
}
```

In this mode, AgentCore must not:

* require a stack named `CDKToolkit`
* create bootstrap infrastructure
* update existing bootstrap infrastructure
* assume ownership of a detected bootstrap stack

Making the stack name configurable alone is not sufficient, because detection must not imply permission to manage that stack.

### Steps to Reproduce

1. Bootstrap an AWS account and region through CloudFormation StackSets 

2. Ensure the resulting bootstrap stack has a generated name, for example:

   ```text
   StackSet-CDKToolkit-9afdf51a-b1cf-4dfc-9b0c-63ccafe70cbc
   ```

3. Configure AgentCore to deploy to the same account and region.

4. Run:

   ```bash
   agentcore deploy
   ```

5. Observe that AgentCore prompts to bootstrap the AWS environment.

6. Cancel the bootstrap prompt.

7. Observe that the deployment continues successfully using the existing bootstrap infrastructure.

### Expected Behavior

agentcore deploy should not report that bootstrapping is required when a usable CDK bootstrap environment already exists.

The bootstrap check should not depend exclusively on a CloudFormation stack named exactly CDKToolkit.

AgentCore should also provide a way to mark bootstrap infrastructure as externally managed, ensuring that the CLI does not attempt to create or update it.

### Actual Behavior

[...]

PR #2058, opened 2026-08-20
feat(project): implement safe CDK deployment

notgitika:
## Summary

- resolve and validate one deployment target before backend dispatch
- resolve one CDK CLI-compatible credential provider and share it across STS, CloudFormation, and Toolkit operations
- reject credentials for the wrong AWS account before synthesis or mutation
- synthesize to a fixed assembly directory and select exactly one tagged stack
- bootstrap only absent or stably outdated environments, propagating unsafe and ambiguous states
- deploy through the backend-local Toolkit adapter and return stack outputs
- embed and clean up the Toolkit bootstrap template for standalone executables

## Known gap: projects that declare credentials

This PR deploys the stack; it does not run the pre-deploy identity setup, so it does
not write `agentcore/.cli/deployed-state.json`. Nothing on `refactor` writes that file
today, so this is a gap being carried forward rather than a regression, but it is worth
stating rather than leaving implied:

- Projects with no `credentials` and no `payments` deploy correctly. `credentials` is
  read as `undefined` by the vended app (`src/assets/cdk/bin/cdk.ts:150`), which is the
  same value it sees today.
- A project that declares an **api-key or OAuth credential** deploys and reports success,
  but the declaration is silently dropped: no credential resource appears in the template
  and no provider is created. Measured on a real deploy from this branch — the deploy
  exits 0, the template contains only the runtime, role, policy, metadata and SSM
  parameter, and `identity api-key-credential-provider get` answers
  `ApiKeyCredentialProvider not found`. Since `credentials` is `undefined` on `refactor`
  too, this is inherited behavior, not something this PR introduces.
- A project that declares a **payment connector** fails during synth, because the vended
  app requires a `credentialProviderArn` per connector (`cdk.ts:177-186`). Its error
  message also names `agentcore deploy`, which does not exist on this branch.

Provisioning credential providers, syncing their client secrets, and writing that state
file is a change of comparable size to this PR and lands next to two open provider
lifecycle bugs (#1673, #2022), so it is the immediate follow-up rather than part of the
deploy spine. That follow-up is now open as #2089, which turns the silent drop into a
created provider and the payment case into an accurate error, verified end-to-end against
a real account.

## Stack

This is PR 4 of 4 decomposing #2001 and is the first PR in the stack that activates deployment. Review and merge bottom-up.

2. #2056 - deploy command contract
3. #2057 - CDK Toolkit adapter
4. **#2058 - safe deployment implementation (this PR)**

## Follow-ups

- ~~Pre-deploy identity setup: create or update api-key and OAuth2 credential providers, sync client secrets, and write `agentcore/.cli/deployed-state.json` before synth. Also fixes the stale `agentcore deploy` guidance in `src/assets/cdk/bin/cdk.ts:184`.~~ Open as #2089, stacked on this PR.
- Revisit and document the inherited `createCustomerMasterKey: true` bootstrap policy, including recurring KMS cost and security rationale. This PR preserves the shipped CLI behavior.
- Define confirmation and ownership semantics before changing shared `CDKToolkit` stacks, including externally managed bootstrap infrastructure tracked by #1799.

did npm bundle and standalone executable smoke tests


src/core/project/backends/cdk.ts
```diff
@@ -36,17 +76,77 @@ export class CdkBackend implements ProjectBackend {
...
+    const credentials = await this.resolveCredentials(target.region);
+    const account = await this.resolveAccount(target.region, credentials);
+    if (account !== target.account) {
+      throw new ProjectStateError(
+        `Deployment target '${target.name}' expects AWS account ${target.account}, ` +
+          `but the active credentials belong to ${account}.`,
+      );
+    }
+
+    yield* this.build(project);
```
tejaskash, line 103: Could we do the identity setup before synthesizing here? The generated CDK app reads `agentcore/.cli/deployed-state.json` during synth to resolve configured credential names to provider ARNs, but this path never creates or updates those providers, syncs their secrets, or writes the target state. On a first deploy, payment projects fail synthesis because the ARN is missing, and other resources can carry a name for a provider that does not exist. We need to provision the configured credentials and persist their target state before calling `build()`.

scripts/build.ts
```diff
@@ -83,15 +111,19 @@ async function compile(target: string): Promise<void> {
...
-    entrypoints: [ENTRYPOINT, ...assets],
+    entrypoints: [ENTRYPOINT, ...assets, template],
     compile: { target: target as Bun.Build.CompileTarget, outfile },
     minify: MINIFY,
     root: REPO_ROOT,
     naming: { asset: ASSET_NAMING },
     plugins: [assetLoaderPlugin()],
   });
-  console.log(`Compiled ${target} → ${outfile} (${assets.length} assets embedded)`);
+  await assertTemplateIsEmbedded(outfile, template);
```
tejaskash, line 127: This is the reason the current Windows build is failing. Bun emits `agentcore-windows-x64.exe`, while `outfile` still points to the extensionless path. The build succeeds, then this assertion tries to read `agentcore-windows-x64` and fails with `ENOENT`. Could we pass the actual emitted path here for Windows targets?

notgitika: windows fix was gonna be a follow up but lemme fix it here itself so that it is not left in a failing state on main.

notgitika: fixed :)

src/core/project/backends/cdk.ts
notgitika: you are correct but the intention was to keep it out of this PR as to not inflate it for ease of review.

it will be done in a follow up right after this is merged. would you rather me do an interim fix that fails fast on specs containing creds?

tejaskash: Nopem, good for now. I would structure it as a post-deploy hook. Credentials are now created by CDK so the post-deploy hook will simply update the credential.

tejaskash (approved):
Approving with the credentials followup

src/core/project/backends/cdk/toolkit.ts
```diff
@@ -104,7 +195,7 @@ export async function performCdkOperation(
   // deletion a successful deploy.
   if (result.stacks.length !== 1) {
```
nborges-aws, line 196: We're running this check after `toolkit.deploy()` deletes an existing stack with no resources. Should we validate the template has resources before invoking toolkit deploy?

src/core/project/backends/cdk/assembly.ts
```diff
@@ -0,0 +1,58 @@
...
+  if (!existsSync(manifestPath)) {
+    throw new ProjectStateError(`No synthesized cloud assembly was found at ${manifestPath}.`);
+  }
+
+  const manifest = await json.read(manifestPath, AssemblyManifestSchema);
+  const stacks = Object.entries(manifest.artifacts).filter(
+    ([, artifact]) => artifact.type === STACK_ARTIFACT,
+  );
+  const matches = stacks.filter(
+    ([, artifact]) => artifact.properties?.tags?.[TARGET_TAG] === target,
```
nborges-aws, line 42: Do we need to validate environment here as well (account/region)? This selects on tag alone

src/core/project/backends/cdk/toolkit.ts
notgitika: I think it does, good catch. let me check to see if i can recreate this. if i can, lemme push in a fix here itself

notgitika:
good comments, I am raising a follow up addressing the 3 comments

nborges-aws (approved):

PR #2067, opened 2026-08-21
feat(eval): imperative online-insight commands + BYO execution role

jariy17:
> **Draft — Part 1.** Deploy bug-bash on the exploratory account: **PASSED** (results below). Scope TODOs remain.

## Command structure

```
agentcore eval online-insight create --help
  --name <name>                     (required)
  --execution-role-arn <arn>        (required) IAM role the config assumes — NO auto-provisioning
  --agent <agent>                   harness ID or runtime ID whose traffic to sample
  --endpoint <endpoint>             endpoint qualifier (default DEFAULT; requires --agent)
  --data-source-config <json>       traces to evaluate (alt. to --agent)
  --insight <insight...>            (required) Builtin.Insight.* IDs or ARNs
  --clustering-frequency <freq...>  DAILY | WEEKLY | MONTHLY
  --sampling-rate <rate>            (required) 0.01-100
  --session-timeout-minutes <n>     1-1440 (default 15)
  --filters <json>                  trace filters
  --enable-on-create <true|false>   default true
  --description <text>
```
Also: `eval online-insight get | list | pause | resume | delete` (by `--id`).

## What this does
- New `eval online-insight` command family (service-direct), sibling of `eval online-eval`.
- Dedicated client methods on `CoreEvalClient`: `createOnlineInsight`, `getOnlineInsight`, `listOnlineInsights`, `setOnlineInsightExecutionStatus`, `deleteOnlineInsight`.
- `create` sets the SDK `insights` (+ optional `clusteringConfig`), never `evaluators`, and **requires** `--execution-role-arn` — insight configs are **not** auto-provisioned a role.
- get/list/pause/resume/delete delegate to the online-eval control-plane ops (same underlying `OnlineEvaluationConfig` resource).

## Service client (`CoreEvalClient`)
`CreateOnlineInsightInput`: `name`, `samplingRate`, `insightIds` (required), `evaluationExecutionRoleArn` (required), `clusteringConfig?`, `sessionTimeoutMinutes?`, `filters?`, `enableOnCreate?`, `description?`, and a source union (`agent`+`endpoint?` | `dataSourceConfig`). The impl maps `insightIds → insights[{insightId}]` on `CreateOnlineEvaluationConfigCommand` and passes the caller role straight through (no `grantOnlineEvalScope`).

## Decisions made (flag for review)
1. New flag named `--execution-role-arn` (net-new, non-breaking). Online-eval's `--role-arn` rename is deferred.
2. Insight methods live on `CoreEvalClient` (not a separate class file).

## TODO (remaining scope)
- [ ] `eval online-insight update`
- [ ] Make `eval online-eval` `--role-arn` **required** + remove shared auto-provisioning (`onlineEvalExecutionRole.tsx`, the `?? autoProvision` fallback, `--update-role`, role-scope warning)
- [x] Deploy bug-bash on exploratory account — PASSED

## Verification
`tsc --noEmit` clean for these changes. Unit test added (`online-insight.test.tsx`). Note: `bun test`/unbundled `bun run` for the `eval` chain can't execute in the current sandbox (pre-existing missing `@opentelemetry/otlp-transformer` deep-import that only the bundler resolves) — bug-bash uses the built bundle.


src/handlers/eval/online-insight/create/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+const ARN_PREFIX = "arn:";
+
+export const createCreateOnlineInsightHandler = (core: Core, io: AppIO) =>
+  createHandler({
+    name: "create",
+    description: "create an online insight config",
+    flags: [
+      flag("name", "the name of the online insight config", z.string().optional()),
+      flag(
+        "execution-role-arn",
```
jariy17: it should be --role-arn

src/handlers/eval/online-insight/create/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+        "IAM role the online insight assumes (required; not auto-provisioned)",
```
jariy17: remove this: (required; not auto-provisioned)

src/handlers/eval/online-insight/create/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+        z.string().optional(),
+      ),
+      flag("agent", "harness ID or runtime ID whose traffic to sample", z.string().optional()),
```
jariy17: just runtime id

src/handlers/eval/online-insight/online-insight.test.tsx
```diff
@@ -0,0 +1,179 @@
+import { describe, expect, test } from "bun:test";
```
jariy17, line 1: It should use the golden test pattern that harness and runtime have.

src/core/eval.tsx
```diff
@@ -556,6 +557,57 @@ export class EvalClient implements CoreEvalClient {
...
+  getOnlineInsight(id: string, options: CoreOptions): Promise<GetOnlineEvaluationConfigResponse> {
+    return this.getOnlineEvaluationConfig(id, options);
+  }
+
+  listOnlineInsights(
+    nextToken: string | undefined,
+    maxResults: number | undefined,
+    options: CoreOptions,
+  ): Promise<ListOnlineEvaluationConfigsResponse> {
+    return this.listOnlineEvaluationConfigs(nextToken, maxResults, options);
```
nborges-aws: This needs a filter to check if any insights are present in the content returned. 

Since its the same API this will return online-evals and online-insights together. Should filter to just online-insights

src/core/eval.tsx
```diff
@@ -556,6 +557,57 @@ export class EvalClient implements CoreEvalClient {
...
+    options: CoreOptions,
+  ): Promise<UpdateOnlineEvaluationConfigResponse> {
+    return this.setOnlineEvaluationExecutionStatus(id, executionStatus, options);
+  }
+
+  deleteOnlineInsight(
+    id: string,
+    options: CoreOptions,
+  ): Promise<DeleteOnlineEvaluationConfigResponse> {
+    return this.deleteOnlineEvaluationConfig(id, options);
```
nborges-aws, line 670: More of a nit but same issue as above. If user passes an online-eval ID to the delete online-insight call it would work

nborges-aws (changes requested):
Need to filter results based on online-insight vs online-eval, since it is the same API and returns everything. Other than that PR LGTM.

I had to solve the same problem in my PR for batch insights. I added a handler that maybe can be repurposed to be shared here also. [Here's an example](https://github.com/aws/agentcore-cli/pull/2078/changes#diff-1ecc0593d91babf242a67e88b3aa0eef67c90db37645018648760a9189657d59R26) of how I used it

src/core/eval.tsx
notgitika: +1 good callout!

src/handlers/eval/online-insight/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export function createOnlineInsightHandler(core: Core, io: AppIO): Router {
+  return new Router("online-insight", "manage AgentCore online insight configs")
+    .handler(createCreateOnlineInsightHandler(core, io))
+    .handler(createGetOnlineInsightHandler(core))
+    .handler(createListOnlineInsightHandler(core))
+    .handler(createPauseOnlineInsightHandler(core))
+    .handler(createResumeOnlineInsightHandler(core))
+    .handler(createDeleteOnlineInsightHandler(core));
+}
```
notgitika, line 20: heads up, I think this opens a broken TUI for bare invocations. 

the eval router applies `withTuiOnEmptyFlagsAndArgs` and it gets inherited by all the leaves here, but since we don't call `.supportedTuiCommands(...)` every command stays TUI-eligible. and there are no online-insight screens/routes yet. 

so online-insight get with no flags pops the catch-all HelpScreen instead of printing "required --id", and in a non-TTY (scripts/CI) it actually throws "interactive mode requires a TTY". 

The test confirms it, it passes `--json` on list specifically to avoid opening the TUI. 

quick fix that matches online-eval: add an empty `.supportedTuiCommands()` so everything falls through to the real CLI handlers for now. When the TUI lands later, bump it to .supportedTuiCommands("get", "list") + add the routes, same as online-eval does today

src/handlers/eval/online-insight/create/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+const ARN_PREFIX = "arn:";
+
+export const createCreateOnlineInsightHandler = (core: Core, io: AppIO) =>
+  createHandler({
+    name: "create",
+    description: "create an online insight config",
+    flags: [
+      flag("name", "the name of the online insight config", z.string().optional()),
+      flag("role-arn", "IAM role the online insight assumes", z.string().optional()),
+      flag("agent", "runtime ID whose traffic to sample", z.string().optional()),
```
notgitika: might be wrong but doesnt this also resolve a harness arn? if so, can we change the description? 

I could see that  we still call the same `agentDataSource` online-eval uses, which resolves a harness ID too

jariy17: Yes Ill fix it.

src/handlers/eval/online-insight/index.tsx
jariy17: Okay ill fix it

src/core/eval.tsx
jariy17: Sure thing ill fix it

src/core/eval.tsx
jariy17: Sure thing I'll fix it

nborges-aws (approved):
LGTM thanks for fixes!

src/core/eval.tsx
```diff
@@ -556,6 +567,109 @@ export class EvalClient implements CoreEvalClient {
...
+    // Fill the requested page across underlying pages, since the shared List API
+    // returns eval configs too (mirrors listGatewayConnectors over Targets).
+    const pageSize = maxResults ?? DEFAULT_ONLINE_INSIGHT_PAGE_SIZE;
+    const items: InsightSummary[] = [];
+    let token = nextToken;
+    let filling = true;
+
+    for (let page = 0; page < MAX_ONLINE_INSIGHT_PAGES; page++) {
+      const requestToken = token;
+      const requestSize = filling ? pageSize - items.length : DEFAULT_ONLINE_INSIGHT_PAGE_SIZE;
```
notgitika, line 620: this ties the scan page to the result page, so --max-results 1 only scans one config per request. 

#2092  already landed listBatchInsights in this file doing it right (full scan page + boundary re-read), can we rebase onto refactor and share that helper?

PR #2045, opened 2026-08-19
revert: remove capacity-provider Journey 1

aidandaly24:
## Description

Reverts #2030 (`feat(capacity-provider): add capacity-provider resource (Journey 1)`).

Capacity-provider support needs to be coordinated with its L3 dependency and release sequencing before landing on CLI
`main`. This revert restores the repository tree exactly to commit `88a02ed7`, the parent of the reverted merge.

## Related Issue

Reverts #2030.

## Documentation PR

N/A - this removes the documentation introduced by the reverted change.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Other: revert

## Testing

- [x] Verified the resulting tree is byte-identical to `88a02ed7`
- [x] `npm run typecheck`
- [x] `npm run format:check`
- [x] `npm run build`
- [x] Commit hooks: ESLint, Prettier, secretlint, and typecheck

## Checklist

- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [x] I have updated the documentation accordingly
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the
terms of your choice.

PR #2068, opened 2026-08-22
feat(capacity-provider): capacity provider support

xxuam:
## Description

  Completes the AgentCore **Capacity Provider** feature end to end and brings its DevEx to parity with the rest of the CLI.

  This PR carries two commits — the revert-of-revert that **restores Journey 1**, plus a new commit adding Journeys 2–4 and the DevEx polish — so it delivers the full feature:

  - **J1 — create/remove a capacity provider** as a first-class project resource (`add capacity-provider` / `remove capacity-provider`), persisted to `agentcore.json` and synthesized to an
  `AWS::BedrockAgentCore::CapacityProvider` CloudFormation resource (VPC, launch parameters, named EBS volumes, lifecycle, auto-created operator role), with `deploy`/`status` integration. (Restored here via
  revert-of-revert; originally #2030, reverted in #2045 for release timing.)
  - **J2 — attach a runtime to a capacity provider** by in-project name or external ARN (`add agent --capacity-provider <name-or-arn>`), rendered as `capacityProviderConfiguration` on the runtime (network config
  is mutually exclusive and omitted).
  - **J3 — mount capacity-provider volumes** into a runtime (`--cp-volume-name` / `--cp-volume-mount-path`), routed through the existing filesystem-mount framework.
  - **J4 — `capacity-provider delete-session`** data-plane command to deprovision a live CP session (destructive confirm, `--yes`, name/id/ARN resolution).
  - **TUI parity:** full interactive `add capacity-provider` wizard (named volumes, EBS encryption + KMS key, instance profile, lifecycle timeouts); capacity-provider attach is now available in the agent
  **create/template** flow (and `agentcore create`) as well as the **BYO** flow.
  - **DevEx polish:** paired `--volume-name` / `--volume-size` flags (replacing `--volume name:sizeGiB`, for consistency with the other paired flags); a referential-integrity guard that blocks removing a capacity
  provider still referenced by a runtime; and removal of the `[preview]` label from the CP add/remove menus.
  - Bumps `@aws-sdk/client-bedrock-agentcore` to a version that ships `DeleteCapacityProviderSessionCommand`.

  ## Related Issue

  Closes #

  ## Documentation PR

  

  ## Type of Change

  - [ ] Bug fix
  - [x] New feature
  - [ ] Breaking change
  - [x] Documentation update

  ## Testing
  
  - [x] I ran `npm run lint`

  Details:
  - **Unit:** full suite green (6219 tests). Added coverage for the paired volume flags, the advanced parity flags (instance-profile / encryption+KMS / lifecycle), the `buildCreateAgentConfig` mapping, the
  `useGenerateWizard` CP flow, the wizard `buildSteps` sequencing, and a duplicate-menu guard.
  - **Integ:** capacity-provider add/remove and add-agent-attach integ tests pass against the real CLI binary.
  - **Typecheck + lint:** clean (no new warnings).
  - **Interactive TUI (harness):** drove the `add capacity-provider` wizard (forward + back-navigation, including encrypt→KMS→back→No), the agent create/template CP attach, `agentcore create` with a CP-by-ARN, and
  the BYO CP flow.
  - **Live end-to-end (us-west-2):** deployed a template agent on an arm64 capacity provider with a mounted volume → CP + runtime `READY`, invoke returned a response, `delete-session` deprovisioned the session,
  and teardown removed the stack cleanly.

  ## Checklist

  > ⚠️  **Dependent change / merge order:** this PR consumes capacity-provider support from `@aws/agentcore-cdk` (the L3 constructs), which is a separate PR (`feat/capacity-provider-attach` in
  `agentcore-l3-cdk-constructs`). That L3 PR must merge and publish a new alpha, and the `@aws/agentcore-cdk` pin here must be bumped to it, before capacity-provider deploys work off the pinned version. (Verified
  locally against a bundled build of the L3 branch.)
  
  ---
  
  terms of your choice.


src/cli/tui/screens/agent/useAddAgent.ts
```diff
@@ -118,7 +118,15 @@ export function mapByoConfigToAgent(config: AddAgentConfig): AgentEnvSpec {
           },
         }
       : {}),
-    ...buildFilesystemConfigurations(config.sessionStorageMountPath, config.efsAccessPoints, config.s3AccessPoints),
+    ...(config.capacityProviderConfiguration && {
```
tejaskash, line 124: Could we leave `networkMode` unset when this configuration is present? `AddAgentScreen` passes `undefined`, but `mapByoConfigToAgent()` turns that into `PUBLIC` at line 84. The resulting object has both `networkMode` and `capacityProviderConfiguration`, which `AgentEnvSpecSchema` rejects. That makes every BYO capacity-provider submission fail when we write `agentcore.json`.

src/cli/commands/capacity-provider/action.ts
```diff
@@ -0,0 +1,110 @@
...
+    throw new Error(
+      `No AgentCore project found. Run inside a project to reference "${name}" by name, or pass the capacity provider id or ARN.`
+    );
+  }
+  const configIO = new ConfigIO({ baseDir: configRoot });
+  const deployed = await configIO.readDeployedState();
+
+  // Deployed state is keyed per target; find the target that has this capacity provider.
+  let record: { capacityProviderId: string; capacityProviderArn: string } | undefined;
+  for (const target of Object.values(deployed.targets)) {
```
tejaskash: Can we resolve this name within the requested region or reject it when more than one target matches? The same provider name can exist in multiple deployment targets, and this loop always takes the first one. With `--region`, we can pair an ID from one target with a different region and send the delete request to the wrong place.

src/cli/primitives/AgentPrimitive.tsx
```diff
@@ -445,6 +500,9 @@ export class AgentPrimitive extends BasePrimitive<AddAgentOptions, RemovableReso
               efsMountPaths: efsPaths,
               s3AccessPointArns: s3Arns,
               s3MountPaths: s3Paths,
+              capacityProvider: cliOptions.capacityProvider,
```
tejaskash, line 503: These values reach `add()`, but the import branch returns through `handleImportPath()`, and its `executeImportAgent()` call never forwards the capacity-provider configuration or volumes. The create import path drops them too. As a result, `--type import --capacity-provider ...` reports success while persisting a runtime with no attachment. Can we add these fields to the shared import operation and both callers?

src/cli/commands/create/command.tsx
```diff
@@ -373,6 +382,33 @@ async function handleCreateCLI(options: CreateOptions): Promise<void> {
...
+      const cpVolumeCheck = validateCapacityProviderVolumeMounts(cpVolumePairs.mounts);
+      if (!cpVolumeCheck.success) {
+        throw new ValidationError(cpVolumeCheck.error);
+      }
+      if (cpVolumePairs.mounts.length > 0 && !options.capacityProvider) {
+        throw new ValidationError(
+          'Capacity provider volume mounts (--cp-volume-name / --cp-volume-mount-path) require attaching the runtime to a capacity provider (--capacity-provider <name-or-arn>).'
+        );
+      }
+      const capacityProviderConfiguration = options.capacityProvider
```
tejaskash, line 414: Can we reject the name form for new-project creation, or create the sibling provider first? A new project has no `capacityProviders` entry for `capacityProviderName` to resolve against, so `agentcore create --capacity-provider myPool` scaffolds the directory and then fails project validation. An external ARN is the only form this flow can persist today.

src/cli/tui/screens/generate/useGenerateWizard.ts
```diff
@@ -67,8 +70,10 @@ export function useGenerateWizard(options?: UseGenerateWizardOptions) {
       if (advancedSettings.has('dockerfile') && config.buildType === 'Container') {
         subSteps.push('dockerfile');
       }
-      // Network — always networkMode, plus subnets/securityGroups for VPC
-      if (advancedSettings.has('network')) {
+      // Network — always networkMode, plus subnets/securityGroups for VPC. A capacity provider
+      // supplies its own network topology, so the network steps are mutually exclusive with it:
+      // when a CP is also selected, CP wins and the network steps are skipped.
+      if (advancedSettings.has('network') && !advancedSettings.has('capacityProvider')) {
```
tejaskash, line 76: When Network and Capacity Provider are both selected, this removes `networkMode` from `steps`, but `setAdvanced()` below still routes to `networkMode` first. Finishing that absent step falls through to confirmation, so the capacity-provider screen is skipped. Can we use the same precedence in both places? The BYO wizard has the same mismatch.

src/cli/tui/screens/capacity-provider/AddCapacityProviderScreen.tsx
```diff
@@ -0,0 +1,597 @@
...
+      onComplete({
+        name,
+        operatorRoleArn: operatorRoleArn || undefined,
+        subnets,
+        securityGroups,
+        os,
+        instanceTypes,
+        instanceProfileArn: instanceProfileArn || undefined,
+        volumes,
+        volumeEncrypted: volumes.length > 0 && volumeEncrypted ? true : undefined,
```
tejaskash: Could we preserve the explicit `false` here? Choosing No encryption sets `volumeEncrypted` to false, but this expression converts it to `undefined`. The primitive then omits `Encrypted`, whose service default is true, so the created volume contradicts the selection and confirmation screen. `volumes.length > 0 ? volumeEncrypted : undefined` would retain the choice.

src/schema/schemas/primitives/capacity-provider.ts
```diff
@@ -0,0 +1,275 @@
...
+// The reference is EITHER an in-project sibling (by name) OR an external CP (by
+// ARN) — exactly one. At synth the CDK app resolves a sibling name to the created
+// CP's ARN, or passes the ARN through, then injects
+// CapacityProviderConfiguration.CapacityProviderArn onto the CfnRuntime.
+//
+// NOTE: This block is duplicated in @aws/agentcore-cdk. Keep the two in sync.
+// ============================================================================
+
+/** Partition-agnostic capacity-provider ARN (arn:[^:]+: per multi-partition rules). */
+export const CAPACITY_PROVIDER_ARN_PATTERN = /^arn:[^:]+:bedrock-agentcore:[a-z0-9-]+:\d{12}:capacity-provider\/.+$/;
```
tejaskash: Could we validate the full provider ID here instead of accepting any nonempty resource segment? For example, `capacity-provider/x` passes this schema, but the service requires `{name}-{10 alnum}`. That means bad external ARNs fail only during deployment. Please tighten the pattern and keep the L3 copy in sync.

src/cli/tui/screens/agent/useAddAgent.ts
xxuam: Fixed — mapByoConfigToAgent now leaves networkMode unset when a CP is attached (omitted, not defaulted to PUBLIC). Added a schema-mapper test.

src/cli/commands/capacity-provider/action.ts
xxuam: Fixed — by-name resolution now collects all matches and disambiguates by --region (rejects cross-region ambiguity). Added region tests.

src/cli/primitives/AgentPrimitive.tsx
xxuam: Fixed — added capacityProviderConfiguration + volumes to ExecuteImportAgentParams and forwarded from all three callers (AgentPrimitive, useAddAgent, create).

src/cli/commands/create/command.tsx
xxuam: Fixed — create now rejects the name form up front (non-ARN → ValidationError); only an external ARN attaches at create. Added a test.

src/cli/tui/screens/generate/useGenerateWizard.ts
xxuam: Fixed in both wizards — network branch now guarded with && !selected.has('capacityProvider') to match the steps memo. Added a test.

src/cli/tui/screens/capacity-provider/AddCapacityProviderScreen.tsx
xxuam: Fixed to volumes.length > 0 ? volumeEncrypted : undefined, preserving an explicit No.

src/schema/schemas/primitives/capacity-provider.ts
xxuam: Tightened to {name}-{10 alnum} so bad ARNs fail at validate time. L3 copy kept in sync (#336). Added accept/reject tests.

tejaskash (approved):
All review comments are addressed on this head, including the mirrored ARN validation in aws/agentcore-l3-cdk-constructs#336. Focused validation passes. The red web-search E2E check is unrelated to this change and failed on an internal service response.

PR #2069, opened 2026-08-22
feat(eval): online-insight update (P2)

jariy17:
> **Draft · P2 of the online-insight stack.** Stacked on #2067 (P1). Real-deploy bug bash pending before ready.

## Command
```
agentcore eval online-insight update --id <id> [--sampling-rate] [--session-timeout-minutes] [--filters]
  [--insight <i...>] [--clustering-frequency <f...>] [--agent | --data-source-config] [--endpoint | --clear-endpoint]
  [--role-arn]
```

## What
- New `eval online-insight update` handler + `updateOnlineInsight` client method.
- Merges over the current config (like online-eval update), sets `insights`/`clusteringConfig` (never evaluators), repoints the data source (agent | data-source | endpoint/clear-endpoint), and passes the caller `--role-arn` straight through.
- **No role re-scope machinery** (no `--update-role`, no role-scope warning) — insight configs are BYO-role only.

## Stack
P1 #2067 → **P2 (this)** → P3 online-eval BYO role → P4 TUI get/list (+ #2029). See the plan artifact.

## Verification
tsc clean; golden test hierarchy updated to include `update` (19/19 replay). Update golden fixture + real-deploy bug bash: in progress.

PR #2070, opened 2026-08-22
feat(eval): online-eval BYO execution role, remove auto-provisioning (P3)

jariy17:
> **Draft · P3 of the online-insight stack.** Stacked on #2069 (P2). Real-deploy bug bash done via RECORD; see below.

## What
`eval online-eval` now requires the caller to bring their own execution role, and the CLI-side auto-provisioning is deleted entirely.

- `create`: `--role-arn` is now **required** (guard rejects a missing value with `required option '--role-arn <role-arn>' not specified`). It is passed straight through to `CreateOnlineEvaluationConfig`.
- `update`: `--role-arn` stays optional (replace); when omitted the service keeps the existing role.

## Removed
- `src/core/onlineEvalExecutionRole.tsx` (+ its unit test) — the whole default-role provisioning module: `grantOnlineEvalScope` / `revokeOnlineEvalScope` / `onlineEvalExecutionRoleName` / `scopePolicyName` / `executionPolicy` / `accountIdFromRoleArn`.
- `create`'s auto-provision fallback and the `retryWhileRolePropagates` IAM-propagation retry (a caller role that cannot be assumed now fails fast).
- `update`'s `--update-role` flag, the managed-role grant/revoke/KMS re-scope block, and the entire role-scope-warning path (`RoleScopeWarning` type, stderr advisory). `updateOnlineEvaluationConfig` now returns the bare `UpdateOnlineEvaluationConfigResponse`.
- Now-dead helpers in `core/eval.tsx`: `evaluatorKmsKeys`, `logGroupNamesOf` (create no longer calls `GetEvaluator` or scopes a role).

## Stack
P1 #2067 → P2 #2069 → **P3 (this)** → P4 TUI get/list. See the plan artifact.

## Verification
- `tsc --noEmit` clean.
- online-eval goldens **re-recorded** against the AWS exploratory account (725476964917) — the real-deploy bug bash. Recording surfaced (and fixed) a real pause→delete `UPDATING` race by adding a `settle()` before delete. Goldens now show the caller-supplied role; no `AgentCoreOnlineEval-*` provisioned role, no IAM (`GetRole`/`PutRolePolicy`) fixtures. Verified no `agentcore_cli_online_eval*` residue in the account afterward.
- Offline replay green: online-eval 25/25, online-insight 19/19 (untouched). Full suite 1751/1751.


Hweinstock (approved):
is this vision here that we'll come back to support automatic role creation once we have a general solution for imperative?

jariy17:
| is this vision here that we'll come back to support automatic role creation once we have a general solution for imperative?

Yes, but after today's standup. I'll leave this feature for now and implement automatic role creation later.

PR #2071, opened 2026-08-22
feat(eval): ondemand simulate — replay a dataset, evaluate synchronously

jariy17:
## What

`eval ondemand simulate` — replay a dataset against a runtime, then evaluate the sessions **synchronously, client-side** (scores print inline). The on-demand twin of `batch-evaluation simulate`.

Pipeline: `invokeDataset` (replay) → `getTracesForAgent` (CloudWatch) → `evaluate` (Evaluate API).

## Follows the batch-evaluation simulate pattern

- Same invoke flags + `--ingestion-wait-ms` (default 180000; 0 to skip)
- Same SIGINT abort + `invokeDataset` composition
- Refuses when 0 invoked, naming the first failure (`failures[0]`)
- Output adds `sessions[]` (exampleId ↔ sessionId) + `failures[]`

**Differs by design:** no `--name` / `--kms-key-arn` (no async job); output is inline scores, not a job id.

## Output

```json
{
  "sessionsEvaluated": 1,
  "results": [ { "evaluatorId": "Builtin.Helpfulness", "value": 0.83, "label": "Very Helpful", "explanation": "…", "tokenUsage": {…} } ],
  "examplesInvoked": 1,
  "examplesFailed": 0,
  "sessions": [ { "exampleId": "greet", "sessionId": "…" } ],
  "failures": []
}
```

## Tests (batch pattern)

- **Edge tests** — `ondemand.test.tsx` (TestCoreClient): required-flag validation, refuse-when-nothing-invoked, `--ingestion-wait-ms` passthrough.
- **Fixture golden** — `ondemand.fixture.test.tsx`: real invoke → CloudWatch traces → Evaluate, recorded against a live agent, replayed offline via `matchGolden`. Deterministic via the injected `newSessionId` seam **plus a `now()` clock seam** — on-demand's trace-query window is otherwise `Date.now()`-based, so its `StartQuery` fixture key would drift between record and replay (batch has no client query; `evaluate` pins an explicit `--start-time/--end-time`).
- Live-validated across 10 dataset scenarios against a real agent in the EXPLORE account (all real Builtin.Helpfulness scores; edge/negative paths handled).

## Notes

- Builtin.Helpfulness ignores ground-truth refs (`ignoredReferenceInputFields`); the handler still forwards them (adapter exercised).
- Rebased onto current `refactor` (was stacked on the pre-merge batch-simulate work).


src/core/eval.tsx
```diff
@@ -582,7 +583,7 @@ export class EvalClient implements CoreEvalClient {
     const logGroupName = runtimeLogGroup(runtimeId, qualifier);
     const serviceName = runtimeServiceName(runtimeName, qualifier);
 
-    const endMs = input.window ? +input.window.endTime : Date.now();
+    const endMs = input.window ? +input.window.endTime : this.now();
```
notgitika, line 595: why did we make this change just curious

src/handlers/eval/ondemand/simulate/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+    description: "replay a dataset against a runtime, then evaluate the sessions client-side",
+    flags: [
+      flag("runtime-id", "runtime id to invoke per scenario", z.string().optional()),
+      flag("qualifier", "runtime endpoint qualifier (default DEFAULT)", z.string().optional()),
+      flag(
+        "payload-template",
+        'JSON payload template; {input} is the scenario input, e.g. {"prompt":"{input}"}',
+        z.string().optional(),
+      ),
+      flag("header", "an ordered application header (repeatable)", z.array(z.string()).optional()),
```
notgitika, line 25: wdym repeatable? 
like `--header <> --header <>` ?

src/handlers/eval/ondemand/simulate/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+      }
+      if (!flags["dataset"])
+        throw new InputValidationError("required option '--dataset' not specified");
+      if (!flags["evaluator"]?.length) {
+        throw new InputValidationError(
+          "required option '--evaluator <evaluator...>' not specified",
+        );
+      }
+
+      const controller = new AbortController();
```
nborges-aws: We should use the shared `withUserCancellation` helper here

src/handlers/eval/ondemand/simulate/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+          })),
+          failures: replay.failures,
+        });
+      } finally {
+        process.off("SIGINT", interrupt);
+      }
+    },
+  });
+
+function toReferenceInputs(s: InvokedSession): EvaluationReferenceInput[] {
```
nborges-aws: are we intentionally omitting expectedResponses here? Without it a dataset that only returns expected responses won't have reference inputs

src/handlers/eval/ondemand/simulate/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+          throw new InputValidationError(
+            `no examples could be invoked (${replay.failed} failed) — nothing to evaluate${detail}`,
+          );
+        }
+
+        const traces = await core.eval.getTracesForAgent(
+          {
+            agent: flags["runtime-id"],
+            endpoint: flags["qualifier"],
+            sessionIds: replay.sessions.map((s) => s.sessionId),
```
nborges-aws, line 90: If the dataset is very large and has lots of session Ids, would we need batching here?

nborges-aws (changes requested):

src/handlers/eval/ondemand/simulate/index.tsx
jariy17: Yes but I expect customers to use batch-evaluations simulate for heavier loads. We can add batching here if customers want it in the future.

src/handlers/eval/ondemand/simulate/index.tsx
jariy17: Ah nice catch. I didn't realized it was missing expectedResponses. I'll update my pr for this.

src/handlers/eval/ondemand/simulate/index.tsx
jariy17: Good Idea, Ill update it in my pr.

nborges-aws (dismissed):

src/handlers/eval/ondemand/simulate/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+  if (!gt) return [];
+  const refs: EvaluationReferenceInput[] = [];
+  if (gt.assertions?.length || gt.expectedTrajectory) {
+    refs.push({
+      context: { spanContext: { sessionId: s.sessionId } },
+      ...(gt.assertions?.length && { assertions: gt.assertions }),
+      ...(gt.expectedTrajectory && { expectedTrajectory: gt.expectedTrajectory }),
+    });
+  }
+  (gt.turns ?? []).forEach((turn, i) => {
```
notgitika, line 129: how confident are we that traceIds always has exactly one entry per term in the expected order? 

could partial or extra telemetry shift this mapping and attach an expected response to the wrong trace?

notgitika: actually I just checked this: 

https://github.com/aws/agentcore-cli/blob/ac5d4e11c71c315f57d3450856fc841a58d0ea38/src/handlers/eval/ondemand/__fixtures__/simulate-ds.jsonl
and
https://github.com/aws/agentcore-cli/blob/ac5d4e11c71c315f57d3450856fc841a58d0ea38/src/handlers/eval/ondemand/__fixtures__/simulate.golden.json#L15

see L15 where it says the user prompts "hi" when in fact we prompted "what is 2+2"

src/core/eval.tsx
jariy17: For the Golden tests, it requires the inputs to be deterministic. However, Date.now() is not deterministic. Therefore, I needed to inject a now function that could mocked by our unit tests. Look at our unit tests for this.

src/handlers/eval/ondemand/simulate/index.tsx
jariy17: Yes, its repeatable like that.

src/handlers/eval/ondemand/simulate/index.tsx
jariy17: Currently, we assume that each traceId is exactly one entry per turn in an expected order. HOWEVER, you did found issue with our golden test. 

Our golden tests uses the same session ids. Therefore, we could be invoking an existing session with the same conversation. To fix this, I'm creating a function that returns a unique identifier when running unit tests and generates a new one when we are recording.

notgitika (approved):

src/handlers/eval/ondemand/simulate/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+      flag("runtime-id", "runtime id to invoke per scenario", z.string().optional()),
+      flag("qualifier", "runtime endpoint qualifier (default DEFAULT)", z.string().optional()),
+      flag(
+        "payload-template",
+        'JSON payload template; {input} is the scenario input, e.g. {"prompt":"{input}"}',
+        z.string().optional(),
+      ),
+      flag("header", "an ordered application header (repeatable)", z.array(z.string()).optional()),
+      flag(
+        "bearer-token",
```
nborges-aws, line 27: this and `--header` should be marked sensitive

PR #2078, opened 2026-08-24
feat: add readonly batch insights TUI

nborges-aws:
## Description 
Adds an interactive, read-only TUI for the `eval batch-insights` command tree. `get` and `list` commands now expose TUI flows; run remains CLI only.

## Screens

| Path | Screen |
|---|---|
| `eval batch-insights` | `RouterScreen` menu |
| `eval batch-insights list` / `eval batch-insights get` | Insights-filtered `BatchEvaluationPicker` using `PaginatedTablePicker` |
| `eval batch-insights get/:batchEvaluationId` | Service-side insight report in `JsonDetail` |

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing
Added `batch-insights.screen.test.tsx` covering screen behavior

- [x] 27 focused tests pass
- [x] `bun run test` (1757 pass, 0 fail)
- [x] bun run typecheck
- [x] bun run lint:check
- [x] bun run format:check
- [x] If I modified src/assets/, I ran the snapshot update command and committed the results

## Checklist

- [x] I have added tests that prove the feature works
- [x] I have updated the documentation accordingly, or no documentation changes are needed
- [x] I have added an appropriate example, or no new example is needed

---

### DIFF
--- a/src/components/BatchEvaluationPicker.tsx
+++ b/src/components/BatchEvaluationPicker.tsx
@@ -41,6 +41,11 @@ function toRow(summary: BatchEvaluationSummary): BatchEvaluationRow {
 export interface BatchEvaluationPickerProps extends ScreenProps {
   breadcrumb: string[];
   description?: string;
+  queryKeyPrefix?: string;
+  include?: (summary: BatchEvaluationSummary) => boolean;
+  loadingMessage?: string;
+  emptyMessage?: string;
+  emptyPageMessage?: string;
   onSelect: (batchEvaluationId: string) => void;
   onEscape?: () => void;
 }
@@ -55,6 +60,11 @@ export function BatchEvaluationPicker({
   core,
   breadcrumb,
   description,
+  queryKeyPrefix = "batch-evaluations",
+  include,
+  loadingMessage = "Loading batch evaluations…",
+  emptyMessage = "No batch evaluations found in this Region.",
+  emptyPageMessage = "No batch evaluations on this page.",
   onSelect,
   onEscape,
 }: BatchEvaluationPickerProps) {
@@ -66,11 +76,11 @@ export function BatchEvaluationPicker({
     <PaginatedTablePicker
       breadcrumb={breadcrumb}
       description={description}
-      queryKey={["batch-evaluations", opts.region]}
+      queryKey={[queryKeyPrefix, opts.region]}
       loadPage={async (token, pageSize) => {
         const response = await core.eval.listBatchEvaluations(token, pageSize, opts);
         return {
-          items: response.batchEvaluations ?? [],
+          items: (response.batchEvaluations ?? []).filter((summary) => include?.(summary) ?? true),
           nextToken: response.nextToken,
         };
       }}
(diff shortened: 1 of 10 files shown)


src/components/BatchEvaluationPicker.tsx
```diff
@@ -41,6 +41,11 @@ function toRow(summary: BatchEvaluationSummary): BatchEvaluationRow {
 export interface BatchEvaluationPickerProps extends ScreenProps {
```
jariy17, line 41: Can you explain why we are reusing BatchEvaluationPicker constructs for BatchInsights. They are two different resource that will diverge in the future.

jariy17 (changes requested):

src/components/BatchEvaluationPicker.tsx
nborges-aws: The CLI experience and semantics are separate, but they aren't two different resources. Batch evals and insights have the same `listBatchEvaluations` API, response model, and display fields. Insights is just a filtered projection of BatchEvaluations, and has its own screen, routes, validations, etc.

If the resources diverge to separate APIs in the future, we'll need a significant refactor to the insights logic regardless of them sharing resources or not. Separating this constructs now creates more code to maintain upfront, and does not save us any effort in the future.

jariy17: I remember Swarmim said they were two different resources and I think we should seperate them but I'm fine just rename all shared compoenent like BatchEvalautionPicker to BatchEvaluation/InsightPicker so we know that its used for both.

nborges-aws: Our CLI experience does match the idea that they are two different resources to the customer/product side (screens, commands, routes, etc.). When I say they're the same resource I mean on the service/API abstraction side. 

The name `BatchEvaluationPicker` is accurate to the service resource it loads. It requests and renders all batch evaluations, which includes insights. I feel like renaming it to include insights implies that the picker operates on two service resources, when it doesn't. The picker does exactly what it is named for: gives all batch insights.

`src/handlers/eval/batch-insights/list/screen.tsx` has to filter out insights from the full list of batch evaluations. So that logic is distinct to the insights flow. I think the idea that insights is getting batch evals, and filtering down is accurate with whats really happening. Naming it otherwise would be misleading to whats actually happening

src/components/BatchEvaluationPicker.tsx
```diff
@@ -66,11 +76,11 @@ export function BatchEvaluationPicker({
     <PaginatedTablePicker
       breadcrumb={breadcrumb}
       description={description}
-      queryKey={["batch-evaluations", opts.region]}
+      queryKey={[queryKeyPrefix, opts.region]}
```
notgitika, line 79: nice!

notgitika (dismissed):

notgitika:
This is great! Approved.

jariy17 (approved):
Nico and I have discussed. We have decided to have dedicated BatchInsights function in evalClient. And rename BatchEvaluationPicker to BatchEvaluationAndInsightsPicker.

PR #2082, opened 2026-08-24
feat(dev): add Agent Inspector HTTP layer (server, security, assets)

tejaskash:
## What

First of three stacked PRs re-authoring the Agent Inspector from `feat/agent-inspector` against current `refactor` APIs. This one lands the **HTTP contract and SPA delivery only** — a pure request to response handler the dev command will compose with `io/startHttpServer`. It is not yet reachable from the CLI.

Stacked on #2041 (`feat/dev-supervisor`); retarget to `refactor` once that merges.

## Scope

- `src/core/dev/inspector/{types,respond,server,testkit}.ts` — DI interfaces (`InspectorSupervisor`, `InspectorTraces`, `InspectorAssets`, `InspectorDeps`), response helpers, and `createInspectorHandler`.
- `src/core/dev/inspectorAssets.ts` — reads the staged SPA through `AssetSource` with an `AGENT_INSPECTOR_PATH` override and a node_modules fallback.
- `src/io/packagedAssets.ts` — raw file reads and package-dir resolution, so `node:fs`/`node:module` stay out of `core/dev`.
- `scripts/build.ts` — `stageInspectorAssets()` copies `@aws/agent-inspector/dist-assets` into the asset tree before bundle and compile.

Routes registered this PR: `GET /api/status`, `POST /api/start`, `GET /api/traces`, `GET /api/traces/:id`, static SPA (with `index.html` fallback), and a graceful `{ success:false, error }` 404 for everything else. Agent-proxy routes (invocations, MCP, A2A, resources) and the CLI wiring land in the following PRs, so no stub routes appear here.

## Security model

Loopback-only Host check (accepts `localhost`, `127.0.0.1`, `[::1]`), server-side Origin allowlist (plus the Vite `:5173` dev origins), `X-Agentcore-Local` required on POSTs, CORS preflight, and a CSP on served HTML.

## Design notes vs the reference branch

- **Route matching hardened**: exact `/api/traces` for the list route and `slice` + `decodeURIComponent` for `:id`, so `/api/tracesXYZ` no longer matches the list route and encoded ids decode. Covered by tests.
- **Boundary of concern**: filesystem and package resolution moved into `src/io/packagedAssets.ts`; `inspectorAssets.ts` no longer imports `node:fs`/`node:module`.
- **Trace list capped** to the newest 200 per poll, matching `TraceStore.list`'s existing `limit` contract, since each summary carries full spans/logs.
- Security-guard rejection cases parameterized with `test.each`.

## Verification

- `bun test src/core/dev/inspector` + `inspectorAssets.test.ts` — 26 pass.
- Full `bun test` (1812 pass), `bun run typecheck`, `bun run lint:check`, `bun run format:check` all green.
- `bun run build` stages the four SPA files into `src/assets/agent-inspector/` (gitignored) and mirrors them into `dist/assets/`.

### DIFF
--- a/src/io/packagedAssets.ts
+++ b/src/io/packagedAssets.ts
@@ -1,0 +1,25 @@
+import { readFile } from "node:fs/promises";
+import { createRequire } from "node:module";
+import { dirname } from "node:path";
+
+/** Read a file's raw bytes, or undefined when it is absent or unreadable. */
+export async function readOptionalBytes(path: string): Promise<Uint8Array | undefined> {
+  try {
+    return await readFile(path);
+  } catch {
+    return undefined;
+  }
+}
+
+/**
+ * The directory a resolvable package file sits in, or undefined when the
+ * package is not installed. Keeps node:module out of the domain layer that
+ * resolves bundled dependencies at runtime.
+ */
+export function resolvePackageFileDir(specifier: string): string | undefined {
+  try {
+    return dirname(createRequire(import.meta.url).resolve(specifier));
+  } catch {
+    return undefined;
+  }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@ dist
 *.tgz
 *.bun-build
 
+# staged Agent Inspector SPA (copied from @aws/agent-inspector at build time)
+src/assets/agent-inspector/
+
 # code coverage
 coverage
 *.lcov
--- a/bun.lock
+++ b/bun.lock
@@ -12,6 +12,7 @@
         "@aws-sdk/client-cloudwatch-logs": "^3.1092.0",
         "@aws-sdk/client-iam": "^3.1080.0",
         "@aws-sdk/client-sts": "^3.1092.0",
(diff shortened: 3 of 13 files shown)


src/io/packagedAssets.ts
```diff
@@ -0,0 +1,25 @@
+import { readFile } from "node:fs/promises";
+import { createRequire } from "node:module";
+import { dirname } from "node:path";
+
+/** Read a file's raw bytes, or undefined when it is absent or unreadable. */
+export async function readOptionalBytes(path: string): Promise<Uint8Array | undefined> {
```
AlexanderRichey, line 6: Is this helper necessary? Looking at where it's used, it doesn't seem to be reducing complexity or cognitive load? In other words, the code above would be just as easy to follow, or even easier, without this little helper.

AlexanderRichey (approved):
Primarily copy/pasting stuff as is, right?

tejaskash:
> Primarily copy/pasting stuff as is, right?

Yep, split up into smaller chunks to make it easier to read

PR #2085, opened 2026-08-24
feat(dev): agent-proxy routes for the Agent Inspector

tejaskash:
## What

Adds the Agent Inspector routes that talk to a running agent or read the project spec. This layer is still a pure request-to-response handler and is not reachable from the CLI yet. The `project dev` wiring lands in #2086.

## Routes

- `POST /invocations` — protocol-aware proxy. HTTP, A2A, and AGUI agents are each normalized into the SPA's `data: <json>` SSE contract. MCP returns a clear error pointing at `/api/mcp` rather than being proxied as HTTP.
- `POST /api/mcp` — forwards a JSON-RPC body to the agent's `/mcp` endpoint, buffering the reply under a 10MB cap.
- `GET /api/a2a/agent-card` — fetches the running agent's A2A card.
- `GET /api/resources` — flattens `project.spec` into the resource graph the SPA renders.

## Tests

Every route is unit tested behind fake deps: invocation routing and SSE normalization (including A2A dedup and the non-streaming fallback), the MCP forward and its size cap, the agent-card paths, the resource graph, and `httpServer`'s streaming and abort handling.

`bun test`, `typecheck`, `lint:check`, `format:check` all green.

### DIFF
--- a/src/core/dev/inspector/invocations.test.ts
+++ b/src/core/dev/inspector/invocations.test.ts
@@ -1,5 +1,6 @@
 import { afterEach, describe, expect, test } from "bun:test";
-import type { HttpRequestHandler } from "../../../io/httpServer";
+import { type HttpRequestHandler, startHttpServer } from "../../../io/httpServer";
+import { parseAgentEvent } from "./invocations";
 import { ServerFarm, fakeSupervisor, post, runningAgent } from "./testkit";
 import type { InspectorDeps } from "./types";
 
@@ -36,6 +37,40 @@ describe("POST /invocations routing", () => {
       success: false,
       error: "MCP agents are invoked through POST /api/mcp, not /invocations.",
     });
+  });
+});
+
+describe("upstream connection failures", () => {
+  async function deadPort(): Promise<number> {
+    const handle = await startHttpServer(() => ({ status: 200 }));
+    await handle.close();
+    return handle.port;
+  }
+
+  test.each([
+    { protocol: "HTTP", errorPrefix: "Agent server error" },
+    { protocol: "A2A", errorPrefix: "A2A agent error" },
+  ])("returns 502 when a $protocol agent is unreachable", async ({ protocol, errorPrefix }) => {
+    const supervisor = fakeSupervisor({
+      agents: [runningAgent("orders", await deadPort(), protocol)],
+    });
+    const { url } = await farm.inspector({ supervisor });
+    const response = await post(url, "/invocations", { agentName: "orders", prompt: "hi" });
+    expect(response.status).toBe(502);
+    const body = (await response.json()) as { success: boolean; error: string };
+    expect(body.success).toBe(false);
+    expect(body.error).toContain(errorPrefix);
+  });
+});
+
+describe("parseAgentEvent drops non-renderable frames", () => {
+  test.each([
+    { name: "a JSON primitive that is not text", data: JSON.stringify(42) },
+    { name: "a blank error field", data: JSON.stringify({ error: "" }) },
+    { name: "a blank text field", data: JSON.stringify({ text: "" }) },
+    { name: "an empty non-JSON token", data: "" },
+  ])("returns null for $name", ({ data }) => {
+    expect(parseAgentEvent(data)).toBeNull();
   });
 });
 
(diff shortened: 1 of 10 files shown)


src/core/dev/inspector/server.ts
```diff
@@ -96,7 +94,6 @@ function handleStatus(deps: InspectorDeps): HttpResponse {
   return json(200, status);
 }
 
-/** POST /api/start — start an agent on demand; concurrent starts share one attempt. */
```
Hweinstock, line 99: q: were these removed on purpose?

src/core/dev/inspector/invocations.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  let agentResponse: Response;
+  try {
+    agentResponse = await fetch(`http://127.0.0.1:${port}/invocations`, {
+      method: "POST",
+      headers,
+      body,
+      signal,
+    });
+  } catch (error) {
```
Hweinstock, line 72: would there be a benefit to wire the logger here?

src/core/dev/inspector/invocations.ts
```diff
@@ -0,0 +1,267 @@
...
+  for await (const data of sseData(stream)) {
+    const payload = parseAgentEvent(data);
+    if (payload !== null) yield sseEvent(payload);
+  }
+}
+
+// Handles bedrock {text}, {error}, ConverseStream contentBlockDelta, bare JSON string, and non-JSON tokens.
+export function parseAgentEvent(data: string): string | { error: string } | null {
+  try {
+    const parsed: unknown = JSON.parse(data);
```
Hweinstock, line 100: would it simplify this code to use a zod schema that we parse with?

src/core/dev/inspector/invocations.ts
```diff
@@ -0,0 +1,267 @@
...
+        yield sseEvent(text);
+      }
+    } catch {
+      yield sseEvent(data);
+    }
+  }
+}
+
+// When streamedFromStatus is set, artifact-update text is skipped because status-update already streamed it.
+function extractSseEventText(
```
Hweinstock, line 196: should this function signature / name mention that its A2A specific?

src/core/dev/inspector/invocations.test.ts
```diff
@@ -0,0 +1,175 @@
...
+    expect(await response.text()).toBe(`data: ${JSON.stringify(expected)}\n\n`);
+  });
+
+  test("re-frames an agent error event as an error payload", async () => {
+    const { url } = await inspectorFor(sseAgent([JSON.stringify({ error: "boom" })]));
+    const response = await post(url, "/invocations", { agentName: "orders", prompt: "hi" });
+    expect(await response.text()).toBe(`data: ${JSON.stringify({ error: "boom" })}\n\n`);
+  });
+
+  test("passes a non-SSE response body through untouched", async () => {
```
Hweinstock, line 100: i like how readable these tests are.

src/io/httpServer.ts
```diff
@@ -100,6 +99,35 @@ async function respond(
...
+}
+
+export async function stream(
+  body: AsyncIterable<Uint8Array>,
+  response: ServerResponse,
+  signal: AbortSignal,
+): Promise<void> {
+  for await (const chunk of body) {
+    if (signal.aborted) break;
+    if (!response.write(chunk)) await drain(response, signal);
```
Hweinstock, line 113: nice and simple!

Hweinstock (approved):
LGTM! some small suggestions/questions but no blockers.

src/core/dev/inspector/invocations.ts
tejaskash: Good call — renamed to `extractA2aEventText` in #2086 (07915919). It only handles A2A artifact/status/task kinds, so the generic SSE name was misleading.

src/core/dev/inspector/server.ts
tejaskash: Yes, on purpose. Those route JSDocs just restated the method/path already visible in the handler, so the comment-cleanup pass dropped them per the no-restating-the-code guideline. The security and wire-contract comments stayed.

src/core/dev/inspector/invocations.ts
tejaskash: There's no logger in this layer today (the dev handler surfaces status via renderStatus and lets errors propagate). The upstream failure already reaches the user as the 502 body in the Inspector, so I left it. Happy to add structured logging if we introduce a logger dep for the dev command more broadly.

src/core/dev/inspector/invocations.ts
tejaskash: I looked at it — zod does not buy much here. parseAgentEvent takes loosely-typed passthrough tokens from arbitrary agent runtimes, handles several shapes ({text}, {error}, ConverseStream delta, bare string) AND a non-JSON fallback that returns the raw token. A zod union would still need the try/catch + raw fallback + empty-to-null glue, so the imperative form stays clearer. Left as is.

PR #2086, opened 2026-08-24
feat(dev): wire the Agent Inspector into project dev

tejaskash:
## What

Third and final PR of the Agent Inspector stack (C1 #2082 → C2 #2085 → **C3**). C1 built the Inspector HTTP layer and C2 added the agent-proxy routes, both tested in isolation behind fake dependencies. This PR wires that layer into `agentcore project dev` and adds the two IO leaves it needs, so the Inspector is reachable from the CLI for the first time.

## Changes

- **`project dev` is UI-by-default.** New `--ui` flag (default true, so `--no-ui` is auto-generated) and `--ui-port`. In UI mode the handler resolves a UI port (explicit port must be free, else scans from 8081), starts the Inspector server via `startHttpServer(createInspectorHandler({...}))`, hands the collector's `TraceStore` to it as `InspectorTraces`, watches `agentcore.json` to reload the supervised runtime set live, and opens the browser when interactive and not `--json`. `--no-ui` keeps the existing plain single-runtime log stream.
- **`src/io/openBrowser.ts`** — best-effort detached browser launch (`open` / `cmd start` / `xdg-open`), never blocks or fails the command.
- **`src/io/watchFile.ts`** — single-file watcher, debounced, closes on abort, fails quietly on a missing file or unsupported platform.
- **`OtelCollector.store` renamed to `traces`** so the collector hands its store to the Inspector without the Inspector knowing the collector exists. `DevTraceCollector` gains the matching `traces` accessor.

## Why the two IO primitives belong in `src/io`

The handler is a real second caller of both, and both are raw platform mechanics (child-process spawn, `fs.watch`) that the boundary rule keeps out of `src/core` and the handlers. They live beside `checkPort`, `startHttpServer`, and the other IO leaves.

## Lifecycle

The Inspector server rides the one `AbortController` shared with the collector, supervisor, and watcher, so Ctrl-C tears everything down through one cancellation domain. The collector closes only after runners return, so an agent's final spans still reach it. The config watcher's `onChange` swallows parse errors deliberately: a half-saved config parses on the next change event.

## Tests

- `project dev` UI-mode suite: starts the Inspector and prints the URL, opens the browser only on a TTY and never in `--json`, `/api/status` lists every runtime with none started (lazy), `agentcore.json` edits reload the set, `--agent` narrows it, a taken `--ui-port` fails fast, `--port` with several runtimes is rejected.
- `watchFile` suite: debounces a burst of edits to one callback and stops on abort, a missing file fails quietly.
- All existing trace-collection and interruption tests kept, updated for the `traces` rename.
- `bun test` full suite green, typecheck / lint:check / format:check clean.

## Verification altitudes

- **Per layer:** the suites above.
- **End to end:** pending on this branch after review — build the CLI, real project with two runtimes, confirm the browser opens the Inspector, invocations stream, MCP/A2A/resources answer, config reload is live, `--no-ui` gives the plain stream, `--json` never opens a browser.

Stacked on #2085. Merge C1 → C2 → C3 in order.

### DIFF
--- a/src/core/dev/inspector/invocations.ts
+++ b/src/core/dev/inspector/invocations.ts
@@ -181,7 +181,7 @@ async function* transformA2aSse(
   for await (const data of sseData(stream)) {
     try {
       const event = JSON.parse(data) as Record<string, unknown>;
-      const { text, kind } = extractSseEventText(event, streamedFromStatus);
+      const { text, kind } = extractA2aEventText(event, streamedFromStatus);
       if (text) {
         if (kind === "status-update") streamedFromStatus = true;
         yield sseEvent(text);
@@ -193,7 +193,7 @@ async function* transformA2aSse(
 }
 
 // When streamedFromStatus is set, artifact-update text is skipped because status-update already streamed it.
-function extractSseEventText(
+function extractA2aEventText(
   event: Record<string, unknown>,
   streamedFromStatus: boolean,
 ): { text: string | null; kind: string | undefined } {
--- a/src/core/dev/port.ts
+++ b/src/core/dev/port.ts
@@ -26,7 +26,19 @@ export async function resolveDevPort(
   checkPort: PortChecker,
   signal: AbortSignal,
 ): Promise<DevPort> {
-  const defaultPort = DEV_PORTS[protocol ?? "HTTP"];
+  return findFreePort(DEV_PORTS[protocol ?? "HTTP"], explicitPort, checkPort, signal);
+}
+
+/**
+ * Resolve a free port from `defaultPort`. An explicit port must be free or the
+ * call fails; otherwise the next free port from the default up is taken.
+ */
+export async function findFreePort(
+  defaultPort: number,
+  explicitPort: number | undefined,
+  checkPort: PortChecker,
+  signal: AbortSignal,
+): Promise<DevPort> {
   const requestedPort = explicitPort ?? defaultPort;
 
   if (await checkPort(requestedPort, signal)) {
--- a/src/core/dev/supervisor.ts
+++ b/src/core/dev/supervisor.ts
@@ -43,6 +43,10 @@ type AgentEntry = {
   port?: number;
   error?: string;
   starting?: Promise<{ name: string; port: number }>;
+  /** The running child's pump, so shutdown can await its final spans. */
+  running?: Promise<void>;
+  /** A reloaded definition held for a live agent, applied on its next start. */
+  pendingRuntime?: ProjectRuntime;
 };
 
 /**
@@ -79,8 +83,15 @@ export class DevSupervisor {
     const names = new Set(runtimes.map((runtime) => runtime.name));
     for (const runtime of runtimes) {
       const existing = this.agents.get(runtime.name);
-      if (existing) existing.runtime = runtime;
-      else this.agents.set(runtime.name, { runtime, phase: "idle" });
+      if (!existing) {
+        this.agents.set(runtime.name, { runtime, phase: "idle" });
+      } else if (existing.phase === "running" || existing.phase === "starting") {
+        // A live process keeps its current definition; the edit applies on next
+        // start, so the Inspector never proxies a running agent with stale metadata.
+        existing.pendingRuntime = runtime;
+      } else {
+        existing.runtime = runtime;
+      }
     }
     for (const [name, entry] of this.agents) {
       if (!names.has(name) && entry.phase !== "running" && entry.phase !== "starting") {
@@ -128,6 +139,10 @@ export class DevSupervisor {
     }
     if (entry.starting) return entry.starting;
 
+    if (entry.pendingRuntime) {
+      entry.runtime = entry.pendingRuntime;
+      entry.pendingRuntime = undefined;
+    }
     entry.starting = this.launch(entry).finally(() => {
       entry.starting = undefined;
     });
@@ -141,7 +156,14 @@ export class DevSupervisor {
   public async *events(): AsyncGenerator<SupervisedEvent, void> {
     while (true) {
       for (const event of this.queue.splice(0)) yield event;
-      if (this.config.signal.aborted) return;
+      if (this.config.signal.aborted) {
+        // Let every live child finish shutting down so its final spans reach the
+        // collector before the caller closes it, then drain what they emitted.
+        const running = [...this.agents.values()].map((entry) => entry.running).filter(Boolean);
+        await Promise.allSettled(running);
+        for (const event of this.queue.splice(0)) yield event;
+        return;
+      }
       await new Promise<void>((resolve) => {
         this.wake = resolve;
         // A push during the yields above ran while wake was undefined, so its
@@ -180,14 +202,14 @@ export class DevSupervisor {
       const readiness = this.waitReady(port, controller.signal, () => activity.at).then(() => {
         ready = true;
       });
-      const earlyExit = this.pump(entry, runner, { port, env, signal: controller.signal }, () => {
+      const pump = this.pump(entry, runner, { port, env, signal: controller.signal }, () => {
         activity.at = Date.now();
-      })
-        .finally(unchain)
-        .then(() => {
-          if (!ready)
-            throw new Error(entry.error ?? `Agent '${name}' exited before it became ready.`);
-        });
+      });
+      entry.running = pump;
+      const earlyExit = pump.finally(unchain).then(() => {
+        if (!ready)
+          throw new Error(entry.error ?? `Agent '${name}' exited before it became ready.`);
+      });
       // Both branches outlive the race (the pump runs for the agent's lifetime);
       // swallow their late rejections so losing branches never become unhandled.
       readiness.catch(() => {});
--- a/src/handlers/project/dev/index.ts
+++ b/src/handlers/project/dev/index.ts
@@ -16,7 +16,7 @@ import type { AppIO, BrowserOpener, FileWatcher, PortChecker, startHttpServer } 
 import { createHandler, flag, ProjectKey } from "../../../router";
 import { JsonRendererKey, type JsonRenderer } from "../../../tui";
 import { JsonKey, RegionKey } from "../../keys";
-import type { Project } from "../types";
+import type { Project, ProjectManager } from "../types";
 import type { DevEnvironmentLoader } from "./environment";
 import type { DevEvent, DevRunner, DevTraceCollector, DevTraceCollectorStarter } from "./types";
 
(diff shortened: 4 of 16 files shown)


src/handlers/project/dev/index.ts
```diff
@@ -86,6 +100,12 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
         z.coerce.number().int().min(1).max(65535).optional(),
       ),
       flag("traces", "disable local OTEL trace collection", z.boolean().default(true)),
+      flag("ui", "run without the Agent Inspector web UI", z.boolean().default(true)),
```
Hweinstock: are we planning to ever re-introduce the tui experience? I wonder if making this a `mode` enum would be clearer ex. `headless | tui | browser`.

src/handlers/project/dev/index.ts
```diff
@@ -103,6 +123,12 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
         const project = ctx.require(ProjectKey);
         const region = ctx.require(RegionKey);
         const runtimes = selectRuntimes(project, flags.agent);
+        if (!flags.ui && !flags.agent) {
+          const available = runtimes.map((runtime) => runtime.name).join(", ");
+          throw new InputValidationError(
+            `--no-ui runs a single agent in the terminal. Pass --agent <name> to choose which one. Available: ${available}.`,
```
Hweinstock: for my own understanding, is commander converting `--ui` to `--no-ui` because it defaults to true? Also, does the description then apply to the `no` version of the flag?

src/handlers/project/dev/index.ts
```diff
@@ -159,11 +185,26 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
           return { ...env, ...otel };
         };
 
+        if (!flags.ui) {
+          await runWithoutUi(
```
Hweinstock, line 198: is it intentional to limit no ui to a single runtime, if so why? Can't I do this on `refactor` today?

src/core/dev/inspector/invocations.ts
```diff
@@ -193,7 +193,7 @@ async function* transformA2aSse(
 }
 
 // When streamedFromStatus is set, artifact-update text is skipped because status-update already streamed it.
-function extractSseEventText(
+function extractA2aEventText(
```
notgitika, line 196: im confused why did we change this method name? isnt it still extracting from SSE?

notgitika: nvm I see this comment in the previous PR https://github.com/aws/agentcore-cli/pull/2085#discussion_r3866401089

src/core/dev/port.ts
```diff
@@ -26,7 +26,19 @@ export async function resolveDevPort(
...
 ): Promise<DevPort> {
-  const defaultPort = DEV_PORTS[protocol ?? "HTTP"];
+  return findFreePort(DEV_PORTS[protocol ?? "HTTP"], explicitPort, checkPort, signal);
+}
+
+/**
+ * Resolve a free port from `defaultPort`. An explicit port must be free or the
+ * call fails; otherwise the next free port from the default up is taken.
+ */
+export async function findFreePort(
```
notgitika, line 36: just a question, certain protocols reserve a port right? if that port is busy, are we still finding a new port for them?

src/handlers/project/index.ts
```diff
@@ -42,6 +43,16 @@ export function createProjectHandler(config: ProjectHandlerConfig): Router {
         loadDevEnvironment,
         checkPort,
         startTraceCollector: startOtelCollector,
+        startServer: startHttpServer,
+        openBrowser,
+        inspectorAssets: new InspectorAssets(),
+        isInteractive: () => process.stdout.isTTY === true,
+        watchFile,
+        reloadRuntimes: async (projectRoot) => {
```
Hweinstock: I feel like we're over injecting here. I understand that passing these in makes it easy to test, but I feel like there should be way to inject a project manager that gives us the ability to configure these behaviors in the test.

src/handlers/project/dev/index.ts
```diff
@@ -177,21 +218,45 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
...
+            const reloaded = await config.reloadRuntimes(project.rootPath);
+            supervisor.setRuntimes(
+              flags.agent ? reloaded.filter((runtime) => runtime.name === flags.agent) : reloaded,
+            );
+            renderStatus(config.io, "Reloaded agents from agentcore.json.", json);
+          } catch {
+            // A half-saved config parses on the next change event.
+          }
+        };
+        config.watchFile(
```
Hweinstock, line 257: I feel like file watching is unnecessary complexity. I've had experience with them behaving inconsistently on windows, and not working well with IDEs that use temp files behind the scenes. What do you think about manual reloading? (this could also be OOS here and taken as a follow-up if we want to remove watching for now)

src/handlers/project/dev/index.ts
```diff
@@ -177,21 +218,45 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
...
+            supervisor,
+            traces: collector?.traces,
+            assets: config.inspectorAssets,
+            project,
+            selectedAgent: flags.agent,
+          }),
+          { port: uiPort, signal: controller.signal },
+        );
+
+        const onConfigChange = async () => {
```
Hweinstock, line 244: I wonder if it makes sense for the manager to expose this kind of functionality ex. 
```
projectManager.onProjectSpecChange(project, listener)
```
I don't think we need this abstraction yet, but was just curious if you had thoughts.

src/handlers/project/dev/index.ts
```diff
@@ -177,21 +218,45 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
           signal: controller.signal,
         });
 
-        const starts = Promise.allSettled(
-          runtimes.map((runtime) => supervisor.start(runtime.name)),
+        const uiPort = (
+          await findFreePort(UI_DEFAULT_PORT, flags["ui-port"], config.checkPort, controller.signal)
```
notgitika, line 231: hmm i think 8081 will conflict with the default inspector port. the UI would try to bind to it first and agent would say its occupied. can ui port selection exclude the runtime part?

src/handlers/project/dev/index.ts
```diff
@@ -177,21 +218,45 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
...
+            assets: config.inspectorAssets,
+            project,
+            selectedAgent: flags.agent,
+          }),
+          { port: uiPort, signal: controller.signal },
+        );
+
+        const onConfigChange = async () => {
+          try {
+            const reloaded = await config.reloadRuntimes(project.rootPath);
```
notgitika: Collector setup only considers the initial runtimes. A reloaded Container or OTEL-enabled runtime may not have a reachable collector. Can this be updated during reload?

src/handlers/project/dev/index.ts
```diff
@@ -177,21 +218,45 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
...
+            supervisor.setRuntimes(
```
notgitika, line 249: reloading changes protocol/build metadata for running agents without restarting them. Inspector may then use the wrong protocol. can updates wait until restart?

src/handlers/project/dev/index.ts
```diff
@@ -177,21 +218,45 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
...
+          projectSpecPath(project.rootPath),
+          () => void onConfigChange(),
+          controller.signal,
         );
 
+        const url = `http://127.0.0.1:${server.port}`;
+        renderStatus(config.io, `Agent Inspector running at ${url}`, json);
+        if (config.isInteractive() && !json) await config.openBrowser(url);
+
         for await (const { agentName, event } of supervisor.events()) {
```
notgitika, line 267: final spans could be dropped probably since shutdown closes the trace collector before running agents finish stopping, let's await runner shutdown first?

src/handlers/project/dev/index.ts
tejaskash: Right, `ui` defaults true so Commander generates `--no-ui`, and it's going away with the `--mode` switch anyway.

src/handlers/project/dev/index.ts
tejaskash: Intentional, no UI means no lazy start, so you pick one agent with `--agent`.

src/handlers/project/dev/index.ts
tejaskash: Agreed, one caller today, not worth the abstraction yet.

notgitika (commented):
added some comments and just a question, why are we changing the flag name? why cant we just go ahead with what we have `-b` which means `--no-browser`?

src/handlers/project/dev/index.ts
tejaskash: Done, swapped to a `--mode <browser|headless|tui>` enum.

src/handlers/project/index.ts
tejaskash: Done, inject the project manager instead of the bespoke reload closure.

src/handlers/project/dev/index.ts
tejaskash: Keeping it in for now, I feel like it doesn't fail loudly. Worst case they have to do a manual restart later. If it does work, it adds some value.

src/core/dev/inspector/invocations.ts
tejaskash: Right, it only handles A2A event kinds, so the generic SSE name was misleading.

src/core/dev/port.ts
tejaskash: Yep, we scan for the next free port unless you pass an explicit `--port`.

src/handlers/project/dev/index.ts
tejaskash: No real conflict, the UI takes 8081 first and agents default to 8080 and skip occupied ports.

src/handlers/project/dev/index.ts
tejaskash: Fixed, we hold a running agent's edit until its next start instead of mutating it live.

src/handlers/project/dev/index.ts
tejaskash: Fixed, events() now waits for the child to finish before we close the collector.

src/handlers/project/dev/index.ts
tejaskash: Known limitation, a container added mid-session needs a restart, and the old CLI didn't handle it either.

src/core/dev/inspector/invocations.ts
Hweinstock: thanks for addressing this!

src/core/dev/supervisor.ts
Hweinstock: OOS here, but I wonder if there's an opportunity to simplify the state management here.

src/handlers/project/dev/index.ts
notgitika: are we going to file this as a follow up for post merge?

notgitika (approved):

src/handlers/project/dev/index.ts
tejaskash: Yep

PR #2089, opened 2026-08-24
feat(project): tear down the stack on a deploy of an empty project, behind --yes

notgitika:
## Summary

Deploying a project that declares no resources tore down its stack silently. This makes that outcome explicit, gated, and reported — and fixes the stack-selection gap found in the same review.

**A deploy of nothing was a silent delete.** The CDK Toolkit reads a template with no resources as an instruction to *delete* an existing stack of that name, and reports the run as an ordinary successful deploy. `#2058` caught it after the fact, by which point the stack was gone.

**The first attempt at a guard here was dead code.** CDK writes an `AWS::CDK::Metadata` resource into every stack unless version reporting is disabled, so a project whose spec declares nothing *still* synthesizes a template with one resource in it — `Object.keys(Resources).length === 0` is unreachable through the CLI. Verified against eight real synthesized assemblies. Counting only the resources the project asked for makes the check fire.

**Making it fire needs somewhere for that deploy to go.** `main` supports this: empty the project, deploy, and the stack is destroyed. The refactor dropped it, so refusing outright would trade a silent delete for a regression. Instead the deploy routes to an explicit `toolkit.destroy()`, gated on `--yes`:

```
$ agentcore project deploy
Project 'orders' declares no resources to deploy, so deploying to target 'default'
would delete stack 'AgentCore-orders-default' and every resource in it.
Re-run with --yes to confirm, or restore the resources the project should have.

$ agentcore project deploy --yes
Removing stack AgentCore-orders-default
Removed project 'orders' from target 'default'
```

Three states are distinguished, all before anything is destroyed:

| Project declares | Stack exists | `--yes` | Outcome |
| --- | --- | --- | --- |
| resources | — | — | deploy |
| nothing | yes | no | refuse, naming the stack that would be deleted |
| nothing | yes | yes | destroy, reported as *Removed* |
| nothing | no | — | refuse: add a resource, e.g. `project add runtime` |

Destroying explicitly rather than deploying the empty template is what makes the outcome reportable: `destroy` fails loudly when the stack cannot be removed, where a deploy of an empty template succeeds either way. The empty-outputs guard in `performCdkOperation` stays as the backstop for reaching that state some other way.

Detection reads the synthesized template rather than counting spec collections the way `main` does, so a resource type added to the spec later is covered without anyone remembering to extend a list.

**Stack selection also matched on the target-name tag alone**, never on the account and region the artifact was synthesized for. Those derive from the same target today and so cannot disagree, but nothing enforced it, and the Toolkit deploys where the artifact's environment points rather than where the tag says. Both fields were being stripped on read, since the manifest schema declared neither.

`stackArtifactIdForTarget` becomes `stackArtifactForTarget`, returning the template path and the deployed stack name alongside the id. The stack name is derived as `properties.stackName ?? artifactId`, which is [CDK's own derivation](https://github.com/aws/aws-cdk-cli/blob/main/packages/%40aws-cdk/cx-api/lib/artifacts/cloudformation-artifact.ts) — so the name handed to CloudFormation is the one the Toolkit would have used.

## Testing

`bun test src/` — 1985 pass, 0 fail against these files. Unit coverage for each state in the table, for the metadata-only template that made the original guard dead, and for the ordering that makes the guards meaningful: every check runs before the Toolkit is called, so a bad target or an unconfirmed teardown fails without touching AWS. `probeStack` is covered for a present stack, a stack mid-rollback, an absent stack, an empty response, and a permissions failure — the last matters because reporting "no stack" on an `AccessDeniedException` would turn a confirmed teardown into an unexplained "add a resource" error.

## Notes

`main` deletes orphaned harnesses and config bundles before destroying the stack, because it creates some of them imperatively. This branch creates everything through CloudFormation, so the stack deletion covers them; nothing is intentionally left behind, but flagging it in case a later imperative resource needs the same treatment.

### DIFF
--- a/src/core/project/backends/cdk.ts
+++ b/src/core/project/backends/cdk.ts
@@ -11,6 +11,8 @@ import {
 } from "../../../io";
 import type { Logger } from "../../../logging";
 import type { DeployBackendInput, ProjectBackend } from "./types";
+import { createCloudFormationClient } from "../../factories";
+import type { CreateCloudFormationClient } from "../../types";
 import {
   countDeployableResources,
   stackArtifactForTarget,
@@ -18,6 +20,8 @@ import {
 } from "./cdk/assembly";
 import { readDeployedState, removeTargetState, updateTargetState } from "./cdk/deployedState";
 import {
+  bootstrapStackReader,
+  createCloudFormationStackReader,
   probeBootstrap,
   probeStack,
   resolveAwsAccount,
@@ -40,6 +44,7 @@ export type CdkBackendConfig = {
   runner?: ProcessRunner;
   checkTool?: typeof requireTool;
   json?: ReadWriteJson;
+  createCloudFormationClient?: CreateCloudFormationClient;
   cdk?: CdkRunner;
   resolveCredentials?: CdkCredentialResolver;
   bootstrap?: BootstrapProbe;
@@ -69,8 +74,16 @@ export class CdkBackend implements ProjectBackend {
     this.cdk = config.cdk ?? createCdkRunner(config.logger);
     this.resolveCredentials =
       config.resolveCredentials ?? createCdkCredentialResolver(config.logger);
-    this.bootstrap = config.bootstrap ?? probeBootstrap;
-    this.stack = config.stack ?? probeStack;
+    const readStack = createCloudFormationStackReader(
+      config.createCloudFormationClient ?? createCloudFormationClient,
+    );
+    const readBootstrapStack = bootstrapStackReader(readStack);
+    this.bootstrap =
+      config.bootstrap ??
+      ((region, credentials) => probeBootstrap(region, credentials, readBootstrapStack));
+    this.stack =
+      config.stack ??
+      ((stackName, region, credentials) => probeStack(stackName, region, credentials, readStack));
     this.resolveAccount = config.resolveAccount ?? resolveAwsAccount;
     this.loadBootstrapTemplate = config.loadBootstrapTemplate ?? loadBootstrapTemplate;
   }
@@ -207,15 +220,13 @@ export class CdkBackend implements ProjectBackend {
       );
     }
 
-    const confirmed =
-      input.confirmTeardown ||
-      (await input.requestTeardownConfirmation?.({
-        projectName: project.name,
-        targetName: target.name,
-        stackName: artifact.stackName,
-        account: target.account,
-        region: target.region,
-      }));
+    const confirmed = await input.confirmTeardown({
+      projectName: project.name,
+      targetName: target.name,
+      resourceDescription: `stack '${artifact.stackName}' and every resource in it`,
+      account: target.account,
+      region: target.region,
+    });
     if (!confirmed) {
       throw new ProjectStateError(
         `Project '${project.name}' declares no resources to deploy, so deploying to target ` +
--- a/src/core/project/backends/cdk.ts
+++ b/src/core/project/backends/cdk.ts
@@ -207,7 +207,16 @@ export class CdkBackend implements ProjectBackend {
       );
     }
 
-    if (!input.confirmTeardown) {
+    const confirmed =
+      input.confirmTeardown ||
+      (await input.requestTeardownConfirmation?.({
+        projectName: project.name,
+        targetName: target.name,
+        stackName: artifact.stackName,
+        account: target.account,
+        region: target.region,
+      }));
+    if (!confirmed) {
       throw new ProjectStateError(
         `Project '${project.name}' declares no resources to deploy, so deploying to target ` +
           `'${target.name}' would delete stack '${artifact.stackName}' and every resource in ` +
--- a/src/core/project/backends/cdk.ts
+++ b/src/core/project/backends/cdk.ts
@@ -11,13 +11,19 @@ import {
 } from "../../../io";
 import type { Logger } from "../../../logging";
 import type { DeployBackendInput, ProjectBackend } from "./types";
-import { assertStackHasResources, stackArtifactForTarget } from "./cdk/assembly";
-import { readDeployedState, updateTargetState } from "./cdk/deployedState";
+import {
+  countDeployableResources,
+  stackArtifactForTarget,
+  type StackArtifact,
+} from "./cdk/assembly";
+import { readDeployedState, removeTargetState, updateTargetState } from "./cdk/deployedState";
 import {
   probeBootstrap,
+  probeStack,
   resolveAwsAccount,
   type AccountResolver,
   type BootstrapProbe,
+  type StackProbe,
 } from "./cdk/environment";
 import {
   createCdkCredentialResolver,
@@ -26,6 +32,7 @@ import {
   type BootstrapTemplateLoader,
   type CdkCredentialResolver,
   type CdkRunner,
+  type CdkRunOptions,
 } from "./cdk/toolkit";
 
 export type CdkBackendConfig = {
@@ -36,6 +43,7 @@ export type CdkBackendConfig = {
   cdk?: CdkRunner;
   resolveCredentials?: CdkCredentialResolver;
   bootstrap?: BootstrapProbe;
+  stack?: StackProbe;
   resolveAccount?: AccountResolver;
   loadBootstrapTemplate?: BootstrapTemplateLoader;
 };
@@ -49,6 +57,7 @@ export class CdkBackend implements ProjectBackend {
   private readonly cdk: CdkRunner;
   private readonly resolveCredentials: CdkCredentialResolver;
   private readonly bootstrap: BootstrapProbe;
+  private readonly stack: StackProbe;
   private readonly resolveAccount: AccountResolver;
   private readonly loadBootstrapTemplate: BootstrapTemplateLoader;
 
@@ -61,6 +70,7 @@ export class CdkBackend implements ProjectBackend {
     this.resolveCredentials =
       config.resolveCredentials ?? createCdkCredentialResolver(config.logger);
     this.bootstrap = config.bootstrap ?? probeBootstrap;
+    this.stack = config.stack ?? probeStack;
     this.resolveAccount = config.resolveAccount ?? resolveAwsAccount;
     this.loadBootstrapTemplate = config.loadBootstrapTemplate ?? loadBootstrapTemplate;
   }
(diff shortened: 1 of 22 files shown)


src/core/project/backends/cdk/assembly.ts
```diff
@@ -54,5 +83,88 @@ export async function stackArtifactIdForTarget(
...
+// correct tag from carrying a stack into the wrong account or region: the Toolkit
+// deploys where the artifact's environment points, not where the tag says.
+function assertEnvironmentMatches(
+  id: string,
+  environment: string | undefined,
+  expected: StackEnvironment,
+): void {
+  // An artifact with no environment is environment-agnostic: it deploys into
+  // whatever the credentials resolve to, which the account preflight checked.
+  if (environment === undefined) return;
```
nborges-aws: For my own understanding: wouldn't CDK write `unknown-account` and `unknown-region` if env-agnostic? The comment on L9 says thats the case, but this comment says no env = env-agnostic.

Hweinstock:
> Rejects a synthesized stack that declares no resources.

I kind of think we want the opposite. If I clear all of my resources, then deploy, I'd expect that to delete the stack. Without this functionality, how do I destroy the AWS resources I've created with the CLI?

I could see an alternative where we offer an explicit `destroy`, but that feels overly complex.

notgitika:
>Rejects a synthesized stack that declares no resources.

> I kind of think we want the opposite. If I clear all of my resources, then deploy, I'd expect that to delete the stack. Without this functionality, how do I destroy the AWS resources I've created with the CLI?

>I could see an alternative where we offer an explicit destroy, but that feels overly complex.

Discussed with folks offline we are going with a confirmation + flag once the condition is user runs deploy when spec is empty. this will just inform the user that we are also going to delete their stack. (`-y`)

src/handlers/project/deploy/index.ts
```diff
@@ -16,6 +16,11 @@ export const createDeployProjectHandler = (config: DeployProjectHandlerConfig) =
     description: "deploy the project to AWS",
     flags: [
       flag("target", "name of the aws-targets.json entry to deploy", z.string().default("default")),
+      flag(
+        "yes",
```
Hweinstock, line 26: to me `--yes` implies, answer yes to the prompt, but here there is no prompt and its instead signaling a bypassing of safety measures (kind of like `git push --force`). 

What do you think of naming this to `force`?

src/core/project/backends/cdk.ts
```diff
@@ -102,13 +112,19 @@ export class CdkBackend implements ProjectBackend {
...
+      account: target.account,
+      region: target.region,
+    });
     const options = { assemblyDirectory, credentials, region: target.region };
 
+    // Decided before the Toolkit is handed the assembly: it reads a template
+    // with nothing in it as an instruction to delete the stack, and reports that
+    // as an ordinary successful deploy.
+    if ((await countDeployableResources(this.json, assemblyDirectory, artifact)) === 0) {
+      return yield* this.teardown({ project, artifact, input, options });
```
Hweinstock, line 141: where does the need for an explicit teardown come in? I thought the behavior of toolkit lib was that it already deleted the stack if it was empty.

src/core/project/backends/cdk/assembly.ts
```diff
@@ -54,5 +96,101 @@ export async function stackArtifactIdForTarget(
...
+  return Object.values(template.Resources).filter(
+    (resource) => resource.Type !== METADATA_RESOURCE_TYPE,
+  ).length;
+}
+
+// Both the target tag and the stack's environment derive from the same target in
+// the synthesized app, so today they cannot disagree. Checking anyway keeps a
+// correct tag from carrying a stack into the wrong account or region: the Toolkit
+// deploys where the artifact's environment points, not where the tag says.
+function assertEnvironmentMatches(
```
Hweinstock: > the stack's environment derive from the same target in the synthesized app, so today they cannot disagree.

What could change that would make them disagree? I'm wondering if we can avoid this complexity, or if there is real concern we could deploy to the wrong environment.

Hweinstock (commented):
one question about the flag naming, and some internal questions about toolkit lib behavior for my own understanding.

nborges-aws (dismissed):
LGTM. Teardown looks sound; I'm indifferent on `--force` vs `--yes` naming convention Harrison raised.

src/core/project/backends/cdk.ts
notgitika: oh so toolkit lib deletes it automatically is the raw `Resources` is empty. for us, an empty project still synths `AWS::CDK::Metadata`. so we need this explicit destroy

src/core/project/backends/cdk/assembly.ts
notgitika: fixed

src/core/project/backends/cdk/assembly.ts
notgitika: it was a defensive check just in case but you are right lets remove this

src/handlers/project/deploy/index.ts
notgitika: discussed offline, keeping `--yes` + adding a prompt for user to explicitly put `y/N` if they dont put that flag up. some more work but I will rope it in this PR itself

src/core/project/backends/cdk/assembly.ts
```diff
@@ -16,19 +18,45 @@ const AssemblyManifestSchema = z.object({
...
+        // resource, so leaving Type optional counts it as deployable and keeps
+        // the teardown path from claiming such a stack is empty.
+        z.object({ Type: z.string().optional() }).passthrough(),
+      )
+      .default({}),
+  })
+  .passthrough();
+
+/** The synthesized stack a deploy selected, and where its template lives. */
+export interface StackArtifact {
```
Hweinstock: nit: should this be a type since its not expecting extensions or implementations?

src/handlers/project/deploy/index.ts
```diff
@@ -44,3 +72,37 @@ export const createDeployProjectHandler = (config: DeployProjectHandlerConfig) =
...
+function createTeardownConfirmationHandler(io: AppIO): TeardownConfirmationHandler {
+  return async (request) => {
+    if (!(await promptForTeardown(io, request))) {
+      throw new UserCancellationError();
+    }
+    return true;
+  };
+}
+
+async function promptForTeardown(
```
Hweinstock, line 89: very simple! I like the injection + how we re-use `UserCancellationError` too.

src/handlers/project/types.ts
```diff
@@ -104,9 +104,29 @@ export type ProjectEvent = {
...
 export type DeployProjectInput = {
   /** Name of the aws-targets.json entry to deploy. */
   target: string;
+  /**
+   * Permission to tear the target's stack down when the project no longer
+   * declares anything to deploy. Withheld by default.
+   */
+  confirmTeardown: boolean;
+  /** Requests approval after the backend discovers that this deploy is a teardown. */
+  requestTeardownConfirmation?: TeardownConfirmationHandler;
```
Hweinstock: Would this be simpler if we inject a teardown confirmer that returns true if --y was passed, and otherwise prompts?

src/handlers/project/types.ts
```diff
@@ -104,9 +104,29 @@ export type ProjectEvent = {
   message: string;
 };
 
+/** The destructive deployment discovered after a project has been synthesized. */
+export type TeardownConfirmationRequest = {
+  projectName: string;
+  targetName: string;
+  stackName: string;
```
Hweinstock: is this leaking CDK specifics into this interface? If we do support non-IaC or terraform, what would we fill here?

src/core/project/manager.test.ts
```diff
@@ -415,13 +422,29 @@ describe("FsProjectManager.deploy", () => {
...
+    ]);
     expect(deployed.events).toEqual([{ message: "Backend deployment started" }]);
     expect(deployed.result).toEqual({
       outputs: { RuntimeArn: "arn:runtime" },
     });
   });
 
+  // The backend decides whether a deploy of nothing is a teardown; the manager's
+  // job is only to carry the user's confirmation through to it.
+  test("passes the teardown confirmation through to the backend", async () => {
```
Hweinstock: is this property already covered by the handler tests where we inject the fake backend? If it didn't route, those tests wouldn't work.

src/core/project/backends/cdk/environment.ts
```diff
@@ -98,6 +108,41 @@ export async function probeBootstrap(
   }
 }
 
+const describeStack: StackReader = async (stackName, region, credentials) => {
+  const { CloudFormationClient, DescribeStacksCommand } =
+    await import("@aws-sdk/client-cloudformation");
+  const client = new CloudFormationClient({ credentials, region });
```
Hweinstock: I think we'll want to inject this client to follow the existing DI pattern.

Hweinstock (commented):
mostly nits or questions, but two things I'm hesitant about. 
- the cloudformation client deviates from the existing aws client pattern and isn't injected. This means we don't cache it and make use of persistent connections, and can't leverage the existing fixtures for testing. 
- we have a lot of testing at a lot of different levels, which is great, but I'm wondering if we can rely on handlers for most of this. I remember having a discussion with alex about how these tests can become friction when making changes, rather than giving us confidence we didn't break anything.

nborges-aws (dismissed):
LGTM

src/handlers/project/types.ts
notgitika: good idea!

src/core/project/backends/cdk/assembly.ts
notgitika: yep, changing to type

src/core/project/manager.test.ts
notgitika: yes it is let me remove this test

src/core/project/backends/cdk/environment.ts
notgitika: ahh good point, I’ll follow the existing injected and cached AWS client pattern here

src/handlers/project/types.ts
notgitika: you are right good catch let me keep it generic here

src/core/project/backends/cdk/environment.ts
```diff
@@ -27,6 +32,44 @@ export type AccountResolver = (
...
+  credentials: CdkCredentialProvider,
+) => Promise<Stack[] | undefined>;
+export type StackProbe = (
+  stackName: string,
+  region: string,
+  credentials: CdkCredentialProvider,
+) => Promise<boolean>;
+
+/** Shares CloudFormation connections for calls using the same credentials and region. */
+export function createCloudFormationStackReader(
```
Hweinstock, line 47: ah, for my own understanding, is this to handle deploying to multiple targets?

Hweinstock (approved):
lgtm, thanks for addressing earlier comments.

src/core/project/backends/cdk/environment.ts
notgitika: a single deploy is scoped to one target/region, so this isn't a multi-target thing.

 it's just memoizing one CloudFormationClient per (credentials, region) so the CloudFormation reads we make (the CDKToolkit bootstrap probe and the empty-project stack probe) reuse a client instead of building a new one each time. Keying by (credentials, region) means it'd also stay correct if a caller ever spanned regions/targets

PR #2094, opened 2026-08-24
refactor: align templates with old cli definition

Hweinstock:
### Problem 
The new CLI has a new definition of templates that is not fully flushed out and diverges from the old CLI. We want to get e2e functionality first, before rethinking this. 

### Solution 
- Templates now again refer to the asset rendering, and the flag for templates refers to flag presets. 
- remove byo support to simplify runtime handler for initial version. 
- bring parity to create flags to what was discussed in (basically what was in the old cli). 
- adjust runtime flags to match the create flags. 
- make minimal backend changes to support new interface. 

### Future Work
Want to refactor the templates to split up template parameters from template identifiers (i.e. what determines the base vs rendered in with handlebars. )

### Testing
- added unit tests that exercise the command e2e.


src/core/project/templates.ts
```diff
@@ -60,28 +59,38 @@ export const TEMPLATES: Record<ProjectTemplate, Template> = {
   },
 };
 
+function buildRuntimeTemplateKey(input: ScaffoldRuntimeInput): string {
```
Hweinstock, line 62: i plan to refactor this in a future PR. We're going to want some of these parameters to determine the asset templates to render from, and some of them to be rendered into it with handlebars. However, this felt like the smallest change I could make to keep it functional with the new interface. 

This PR is intended to focus on aligning the handlers with what we want, then we work backwards from there to implement the functionality we need to support it.

src/handlers/project/types.ts
```diff
@@ -6,33 +6,60 @@ import type { ProjectSpecSchema } from "../../projectSchemas/project";
...
-  template: ProjectTemplate;
   /** Skip installing dependencies (npm install, uv sync). */
   skipInstall?: boolean;
   /** Skip initializing a git repository. */
   skipGit?: boolean;
 };
 
+/** Set of flags needed to scaffold a new Runtime-based agent **/
+export const ScaffoldRuntimeInputSchema = z.object({
+  runtimeName: z.string().min(1),
```
notgitika: maybe not related to this PR but if we are only checking if it is a string, is "../MyAgent" a valid runtime name? it would scaffold cold into `<cwd>/..` outside the project then right?

src/handlers/project/project.test.ts
```diff
@@ -108,6 +114,52 @@ describe("project create", () => {
...
+        "--name",
+        "MyAgent",
+        "--template",
+        "hello-world-python",
+        "--build",
+        "Container",
+      ]),
+    ).rejects.toThrow(/--template and --build are mutually exclusive/);
+  });
+
```
notgitika, line 131: just wanted to confirm is our vision that we can probably add certain build time env vars into templates? so I can do like `hello-world-python[no-memory]` 
or do we just reject that idea altogether?

src/handlers/project/create/index.ts
```diff
@@ -17,9 +24,37 @@ export const createCreateProjectHandler = (config: CreateProjectHandlerConfig) =
...
+        "framework",
+        "agent framework for the scaffolded runtime code",
+        z.enum(["none"]).optional(),
+      ),
+      flag(
+        "model-provider",
+        "model provider for the scaffolded runtime code",
+        z.enum(["Bedrock"]).optional(),
+      ),
+      flag(
```
notgitika, line 50: api-key is still not supported fully yet right? like it gets resolved but nothing reads it so it is dropped (assuming its since we only have bedrock for now)

src/handlers/project/create/index.ts
```diff
@@ -28,16 +63,59 @@ export const createCreateProjectHandler = (config: CreateProjectHandlerConfig) =
...
-      // Progress and success go to stderr, keeping stdout for machine output.
-      for await (const event of config.projectManager.create({
+      const scaffoldingFlags = [
+        "build",
+        "language",
+        "framework",
+        "model-provider",
+        "api-key",
+        "memory",
+        "runtime-name",
```
notgitika, line 73: wait so is runtime-name also mutually exclusive? I cant have "mydemoagent" as my runtime name with the template? seems like we are forcing the user too much here folks might want custom runtime name but the basic hello world scaffolding in it. 

pulling it out of this list also gives you somewhere to hang the codeLocation fix.

src/handlers/project/create/index.ts
```diff
@@ -28,16 +63,59 @@ export const createCreateProjectHandler = (config: CreateProjectHandlerConfig) =
...
+
+      for await (const event of config.projectManager.create(createInput)) {
         config.io.stderr.write(`${event.message}\n`);
       }
 
       config.io.stderr.write(`Created project '${flags["name"]}' in ./${flags["name"]}\n`);
     },
   });
+
+function parseScaffoldRuntimeInput(input: Record<string, unknown>) {
```
notgitika, line 117: if i miss a flag, it would surface the raw Zod error.

for eg: `create --name MyAgent --runtime-name foo` prints 
``` Invalid input: expected "none" → at framework ```  

it points users to internal camelCase field paths rather than the `--kebab` flags they typed. can we check for the missing flags up front? or map it back to flag names

Comment (notgitika) [reply src/handlers/project/create/index.ts:117]: could be polished in follow up tbh

File: src/handlers/project/add/runtime/index.ts
Comment (notgitika) [inline src/handlers/project/add/runtime/index.ts:1]: nice!

**Review by notgitika — COMMENTED** [review]
still reviewing the manager and templates files in a couple mins

File: src/core/project/templates.ts
Comment (notgitika) [inline src/core/project/templates.ts]: this is hardcoded but line 103 now writes the assets to `app/${input.runtimeName}`. 

those are thes ame when `runtimeName` is `hello-world`, which is true for the presets but not for the custom flags path. I would say this is a blocker for this PR

**Review by notgitika — CHANGES_REQUESTED** [review]

File: src/handlers/project/types.ts (hunk 2 above)
Comment (Hweinstock) [reply src/handlers/project/types.ts]: good question! my understanding is this case is handled for the runtime case in the schema itself, so when the write fails, the scaffolding rollback, but we don't have the same on the create command, so its currently possible to do this already.

runtime schema: https://github.com/aws/agentcore-cli/blob/e51a676036252137ca976af04a79b42bf92ed073/src/projectSchemas/runtime.ts#L16-L23

To allow us to fail earlier and fix the create case, we can validate the same regex we validate in the schema here.

File: src/handlers/project/project.test.ts (hunk 3 above)
Comment (Hweinstock) [reply src/handlers/project/project.test.ts:131]: my implementation here is based on the old CLI. my understanding from our conversation yesterday was that we want to mirror existing behavior, before adding new functionality. Hopefully we get a chance to come back and make this more flexible.

File: src/handlers/project/create/index.ts (hunk 4 above)
Comment (Hweinstock) [reply src/handlers/project/create/index.ts:50]: yes exactly. I put it there so that its wired up in the handler once we're ready, but there is no way to provide it. 

Let me add some validation to reject this if we pass bedrock to make this more explicit.

File: src/handlers/project/create/index.ts (hunk 5 above)
Comment (Hweinstock) [reply src/handlers/project/create/index.ts:73]: yeah I think I agree we should allow this to be customized, but our templates are not flexible enough to support this yet. The follow-up https://github.com/aws/agentcore-cli/pull/2099 refactors the templates to add this functionality which should allow us to allow overrides like this.

File: src/handlers/project/create/index.ts (hunk 6 above)
Comment (Hweinstock) [reply src/handlers/project/create/index.ts:117]: good catch. I think it might make sense to circle back here, because this definitely isn't the only case of ugly error messages and I think we should establish some consistent standards.

File: src/core/project/templates.ts
Comment (Hweinstock) [reply src/core/project/templates.ts]: Yeah this is a hack to get the same behavior. let me be consistent with the hardcoding and properly remove it in the follow-up.

File: src/handlers/project/project.test.ts (hunk 3 above)
Comment (notgitika) [reply src/handlers/project/project.test.ts:131]: sounds good, we can revisit this later

File: src/handlers/project/create/index.ts (hunk 6 above)
Comment (notgitika) [reply src/handlers/project/create/index.ts:117]: 💯

File: src/handlers/project/types.ts [hunk 7]
```diff
@@ -43,15 +43,20 @@ type CreateProjectInputBase = {
...
+    runtimeName: AgentNameSchema,
+    build: BuildTypeSchema,
+    language: z.enum(["Python"]),
+    framework: z.enum(["none"]),
+    modelProvider: z.enum(["Bedrock"]),
+    apiKey: z.string().min(1).optional(),
+    memory: z.enum(["none"]),
+  })
+  .refine(({ modelProvider, apiKey }) => !(modelProvider === "Bedrock" && apiKey !== undefined), {
+    message: "API keys are not compatible with Bedrock model providers",
```
notgitika, line 57: noice

src/core/project/manager.test.ts
```diff
@@ -103,7 +103,7 @@ describe("FsProjectManager.create", () => {
         name: "hello_world",
         build: "CodeZip",
         entrypoint: "main.py",
-        codeLocation: "app/hello-world",
+        codeLocation: "app/hello_world",
```
notgitika, line 106: 😭 I forgot `-` is not supported in runtime

notgitika (approved):
looks great to me!

src/core/project/templates.ts
AlexanderRichey: What makes this complexity necessary at this stage? Do we expect more than a handful of templates?

src/handlers/project/create/index.ts
AlexanderRichey: Wouldn't these kinds of errors get thrown before the handler itself runs? The framework should handle the validations of the given flags assuming they're populated.

src/handlers/project/create/index.ts
```diff
@@ -28,16 +63,59 @@ export const createCreateProjectHandler = (config: CreateProjectHandlerConfig) =
...
+      const isTemplate = flags["template"] !== undefined;
+      if (presentScaffoldingFlags.length > 0 && isTemplate)
+        throw new InputValidationError(
+          `--template and --${presentScaffoldingFlags[0]} are mutually exclusive`,
+        );
+
+      const isCustom = presentScaffoldingFlags.length > 0;
+
+      const source = new SourceResolver({ stdin: config.io.stdin });
+      const apiKey = await source.resolveSecret("api-key", flags["api-key"]);
```
AlexanderRichey, line 86: Will this hang if the api-key param isn't passed? api-key is required only for non-Bedrock model providers, right?

AlexanderRichey (approved):
Looks good. Left a couple questions that can be addressed in a follow up if necessary.

src/handlers/project/create/index.ts
Hweinstock: Most of them yes, but I think the referenced message is coming from https://github.com/Hweinstock/agentcore-cli/blob/667fb0e60f750d4bed85ffb8c1fc790053375e69/src/handlers/project/create/index.ts#L117-L121 which happens after the flags are parsed and is part of a shared validation between this and the create handler for the scaffolding flags. 

We could get rid of this if we share the flags themselves rather than a separate schema, but originally wanted to keep the flags explicit.

src/handlers/project/create/index.ts
Hweinstock: I believe it handles the undefined case internally https://github.com/Hweinstock/agentcore-cli/blob/667fb0e60f750d4bed85ffb8c1fc790053375e69/src/io/source.ts#L62. 

And yeah exactly, we only support bedrock atm so this is effectively a placeholder.

PR #2098, opened 2026-08-25
feat(eval): batch simulate — --ingestion-wait-ms + per-example failures/sessions

jariy17:
## What

Batch `simulate` review follow-ups (base: `refactor`):

1. **`--ingestion-wait-ms` flag** (default `180000`, `0` skips) → `InvokeDatasetInput.waitIngestionMs`. **Drops the `SIMULATE_INGESTION_WAIT_MS` env var** — value flows through the input only.
2. **Per-example failures.** `runExamples` returns `failures: {item,error}[]`; `invokeDataset` surfaces `failures: [{exampleId, error}]` — names *which* examples dropped + why, not just a count.
3. **Output** now renders `sessions[]` + `failures[]` (below).
4. Uses Golden Test Pattern for simulate instead of dedicated unit tests for invokeDataset

## Output

```json
{
  "batchEvaluationId": "batch-eval-test", "status": "RUNNING",
  "examplesInvoked": 1, "examplesFailed": 1,
  "sessions": [ { "exampleId": "ok1", "sessionId": "s1" } ],
  "failures": [ { "exampleId": "bad", "error": "HTTP 500" } ]
}
```

- `sessions[]` — `exampleId ↔ sessionId`. Join key: a later `eval batch-evaluation get` returns `results[]` keyed by `sessionId`, mapping each score back to its dataset row.
- `failures[]` — always present (`[]` on a clean run); names which examples dropped + why.


src/handlers/eval/batch-evaluation/simulate/index.tsx
```diff
@@ -69,12 +74,14 @@ export const createSimulateBatchEvaluationHandler = (core: Core, _io: AppIO) =>
...
             dataset: flags["dataset"],
             datasetVersion: flags["dataset-version"],
+            waitIngestionMs: flags["ingestion-wait-ms"],
           },
           opts,
           controller.signal,
         );
         if (r.invoked === 0) {
-          const detail = r.firstError ? `; first error: ${r.firstError.message}` : "";
+          const first = r.failures[0];
```
Hweinstock, line 83: would it make sense to just join all the errors? Or are these already communicated to the customer somewhere else?

Just wanted to make sure they'd be aware of individual datasets failing even if the overall command doesn't fail.

src/handlers/eval/batch-evaluation/batch-evaluation.fixture.test.tsx
```diff
@@ -136,4 +139,49 @@ describe("eval batch-evaluation (fixture-backed)", () => {
     expect(job.batchEvaluationId).toBeTruthy();
     expect(job.status).toBeTruthy();
   });
+
+  test("simulate replays a dataset, then submits a batch job over the created sessions", async () => {
```
Hweinstock, line 143: wow is this the final test that replaced all those others? love how simple it is!

src/testing/fixtures.tsx
```diff
@@ -145,6 +145,31 @@ function reviveError(tagged: TaggedError): Error {
   return error;
 }
 
+const STREAM_TAG = "$stream";
+
+async function freezeStream(response: unknown): Promise<unknown> {
```
Hweinstock, line 150: took me a sec, but this makes sense and feels like a simple solution. 

correct me if I misunderstood, but for streaming apis the response is not json serializable, so we convert the response via the stream tag. In the process, we resolve the whole stream, so we need to convert the data back to a stream before passing it downstream to consumers who are expecting a stream. 

nice!

Hweinstock (approved):

src/testing/fixtures.tsx
```diff
@@ -145,6 +145,31 @@ function reviveError(tagged: TaggedError): Error {
...
+  const stream = (response as { response?: { transformToString?: () => Promise<string> } })
+    ?.response;
+  if (typeof stream?.transformToString !== "function") return response;
+  return {
+    ...(response as Record<string, unknown>),
+    response: { [STREAM_TAG]: await stream.transformToString() },
```
notgitika, line 156: Could we serialize transformToByteArray() as base64 instead? transformToString() decodes as UTF-8, so binary or invalid UTF-8 runtime responses are changed during replay.

nborges-aws (commented):
PR overall LGTM. One concern with a gap in testing. Tests are all either exercise happy path, or mock invokeDataset with a mocked failure object in the response. Nowhere is exercising actual invocation failures assemble that failure object correctly.

The deleted invokeDataset test is actually the only place that exercised this.

src/testing/fixtures.tsx
jariy17: Yep ur correct. The reviveStream name seems a little weird to me, maybe Ill use mimicStream.

notgitika (approved):
Left a comment, could be followup

src/testing/fixtures.tsx
jariy17: can do in a follow up.

src/handlers/eval/batch-evaluation/simulate/index.tsx
jariy17: Yes all the errors would show in the debug logs. I assumed that if all dataset examples have failed, it would be the same error. However, maybe we should convene these errors directly to the tui or cli. I'll look into this in a follow up

jariy17:
I'll address the missing bad paths in a follow up

PR #2099, opened 2026-08-25
feat(templates): support rendering strands http template. 

Hweinstock:
### Problem 
We are not able to render real templates yet. 

### Solution 
implement support for `add runtime` in core with the strands python http template as an example. 
- refactored addResources to modify an in-memory agentcore spec for stronger type safety and more flexibility in adding multiple resources. 
- refactored the `template.ts` file into its own directory to make it extensible to future templates. 
- define a `TemplateRenderer` interface that we can implement via handlebars. 

### Testing / Verification
create an agent from this template through create flow, and runtime add flow, then verify dev works for both:

```
bun run compile

./dist/bin/agentcore-linux-x64 project create --name testProject --template strands-python

cd testProject/

../dist/bin/agentcore-linux-x64 project add runtime --name agent2 --template strands-python

../dist/bin/agentcore-linux-x64 project dev

<seperate terminal>

 curl -X POST localhost:8080/invocations -H 'Content-type:
 application/json' -d '{ "prompt": "hello" }'

 curl -X POST localhost:8081/invocations -H 'Content-type:
 application/json' -d '{ "prompt": "hello" }'

```

we get responses on both

- unit tests for all of this. 

### Future Work
This PR does not yet wire up support for memory, credentials for non-bedrock providers, or other templates. These will all be follow-ups, but their extension should be hopefully obvious based on the shape setup by the PR. 

Also, we want to support overriding runtimeName on a template within the create flow (comment from previous PR that wasn't fully addressed here, but should be easy to support now with dynamic templates).


src/assets/templates/strands-http-python/pyproject.toml
```diff
@@ -0,0 +1,31 @@
...
+requires = ["hatchling ~= 1.27.0"]
+build-backend = "hatchling.build"
+
+[project]
+name = "{{ name }}"
+version = "0.1.0"
+description = "AgentCore Runtime Application using Strands SDK"
+readme = "README.md"
+requires-python = ">=3.10"
+dependencies = [
```
Hweinstock, line 11: tweaked some version numbers here, and swapped it to use compatible release so that templates shouldn't break. https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers-compatible-release

src/core/project/manager.tsx
```diff
@@ -217,27 +234,34 @@ export class FsProjectManager implements ProjectManager {
...
+      }
+      case "memory": {
+        projectSpec.memories.push(parseResource(MemorySchema, input.resourceConfig));
+        break;
+      }
       case "gateway":
-        newResources.push(resourceConfig);
+        projectSpec.agentCoreGateways.push(input.resourceConfig);
         break;
       case "gateway-target": {
```
Hweinstock, line 248: we have to explicitly case them to avoid casting here. This allows us to validate up-front, which is redundant, but does allow us to differentiate when our input causes a bad config, and when a config is already bad so I think its worth it since zod parsing is cheap.

Hweinstock:
Windows tests unrelated, tracked in https://github.com/aws/agentcore-cli/issues/2110.

src/assets/templates/strands-http-python/main.py
```diff
@@ -0,0 +1,714 @@
+from typing import Any
```
Hweinstock, line 1: i copied these source files as is, but I wonder if there's room to simplify. It'd be great if they were human readable.

src/core/project/templates/project.ts
```diff
@@ -0,0 +1,71 @@
...
+
+type CreateProjectConfig = {
+  assetSource: AssetSource;
+  templateRenderer: TemplateRenderer;
+};
+/** Scaffold a project from scratch, with optional support for rendering a runtime with the project. **/
+export async function createProjectTree(
+  config: CreateProjectConfig,
+  input: { projectName: string },
+  options?: { runtime?: ScaffoldRuntimeInput },
```
Hweinstock, line 17: if we decide to add harness here, it should be a pretty straightforward change since they both share the `TemplateResolver` return type.

src/core/project/manager.tsx
```diff
@@ -248,15 +270,11 @@ export class FsProjectManager implements ProjectManager {
...
-
-    // Validate and write inside the same boundary so a rejected spec rolls back
-    // staged side effects (.env.local, scaffolded files) rather than leaving them.
     let newProjectSpec: z.infer<typeof ProjectSpecSchema>;
     try {
-      const newSpecParseResult = ProjectSpecSchema.safeParse(newSpec);
+      const newSpecParseResult = ProjectSpecSchema.safeParse(projectSpec);
       if (!newSpecParseResult.success)
-        throw new InputValidationError(z.prettifyError(newSpecParseResult.error), {
+        throw new ProjectStateError(z.prettifyError(newSpecParseResult.error), {
```
Hweinstock, line 272: we now know based on above that our new entries parsed, but the existing config is invalid so we don't throw an input validation anymore.

src/assets/templates/strands-http-python/README.md
```diff
@@ -0,0 +1,46 @@
...
+If installation was successful, a virtual environment is already created with dependencies installed.
+
+Activate the environment with `source .venv/bin/activate` on macOS/Linux, `.venv\Scripts\activate.bat` in Windows
+Command Prompt, or `.\.venv\Scripts\activate.ps1` in Windows PowerShell.
+
+`agentcore project dev` will start a local server on 0.0.0.0:8080.
+
+In a new terminal, you can invoke that server with:
+
+`agentcore project invoke --dev "What can you do"`
```
tejaskash: Did we implement this?

src/assets/templates/strands-http-python/main.py
tejaskash: Let's discuss this today, but we've been back and forth on what the right mix is between human readable templates and flexibility to compose different features on the fly

src/assets/templates/strands-http-python/mcp_client/client.py
```diff
@@ -0,0 +1,116 @@
+import os
```
jariy17, line 1: I feel like we could use Strand's [tools from folder feature](https://strandsagents.com/docs/user-guide/concepts/tools/#loading-tools-from-files) to simplify this process here

src/core/project/templates/runtime.ts
```diff
@@ -0,0 +1,122 @@
...
+
+function buildResolverKey(
+  framework: ScaffoldRuntimeInput["framework"],
+  language: ScaffoldRuntimeInput["language"],
+): `${ScaffoldRuntimeInput["framework"]}/${ScaffoldRuntimeInput["language"]}` {
+  return `${framework}/${language}`;
+}
+
+const getTemplateResolvers = (assetSource: AssetSource, templateRenderer: TemplateRenderer) => ({
+  [buildResolverKey("none", "Python")]: async (input: RuntimeResourceConfig) => {
```
aidandaly24, line 60: I tried `agentcore project add runtime --name mcp_agent --template hello-world-python --protocol MCP`. It succeeds and writes `protocol: "MCP"`, but the generated `main.py` still runs `BedrockAgentCoreApp` HTTP. I think the `none/Python` resolver should reject non-HTTP protocols the same way `strands-python` does.

src/assets/templates/strands-http-python/README.md
aidandaly24: nit: `agentcore project invoke --dev` is not registered on the refactor branch. We do have local invocation through the Inspector /invocations endpoint, so should this example use dev or the Agent Inspector instead? Deployed invocation will be the top-level `agentcore project invoke` command once #2115 lands. I can update this `README.md` in that PR or those commands could be included here too.

src/core/project/manager.tsx
```diff
@@ -145,66 +160,66 @@ export class FsProjectManager implements ProjectManager {
...
       }
       case "runtime": {
-        throw new NotImplementedError(
-          "runtime case not yet implemented in FsProjectManager.addResource",
-        );
+        yield { message: "Scaffolding runtime in project" };
+        const outputPath = join(project.rootPath, "app", input.resourceConfig.name);
+        scaffoldedPaths.push(outputPath);
+
+        const spec = await this.scaffoldRuntimeResources(outputPath, input.resourceConfig);
```
tejaskash, line 207: Could we generate `uv.lock` during this runtime scaffold? `hello-world-python-container` emits a Dockerfile that copies `uv.lock` and runs `uv sync --frozen`, but this path never creates the lockfile. A new container runtime fails on the next `project dev` or deploy unless the user runs `uv sync` first.

src/assets/templates/strands-http-python/pyproject.toml
```diff
@@ -0,0 +1,31 @@
+[build-system]
+requires = ["hatchling ~= 1.27.0"]
+build-backend = "hatchling.build"
+
+[project]
+name = "{{ name }}"
```
tejaskash, line 6: Could we normalize the runtime name before using it as the package name here? The CLI accepts names such as `agent_`, but uv rejects that value in `[project].name`, so `uv lock` and `uv sync` fail after the add command succeeds.

src/core/project/templates/harness.ts
```diff
@@ -0,0 +1,49 @@
...
+        dockerfile: spec.dockerfile ? "Dockerfile" : undefined,
+      });
+
+      const tree = FsTreeNode.createDirectory(parsed.name, [
+        FsTreeNode.createFile("harness.json", async () => json(parsed)),
+        FsTreeNode.createFile(
+          "system-prompt.md",
+          async () => systemPrompt ?? DEFAULT_SYSTEM_PROMPT,
+        ),
+        ...(spec.dockerfile ? [FsTreeNode.fromTextFile("Dockerfile", spec.dockerfile)] : []),
```
tejaskash, line 33: Could we load or validate the Dockerfile before returning this tree? `fromTextFile` is not read until `tree.write`, after `harness.json` and `system-prompt.md` have already been written. A missing `--dockerfile` then leaves a partial harness directory behind even though the command fails.

src/core/project/templates/runtime.ts
Hweinstock: good catch, I wasn't really testing the hello world python since i wasn't sure if we were going to remove it, but we should fix this.

src/assets/templates/strands-http-python/README.md
Hweinstock: Let me remove for now and we can add back once we know what the command will look like.

Hweinstock: removed, we can re-add once its implemented.

src/handlers/project/add/runtime/index.ts
notgitika, line 1: the schema was widened to ["strands", "none"] and the resolver is registered, but both flag declarations still say `z.enum(["none"])`.

so `--framework strands` is rejected and `--help` only lists none. is `--template` the only scope for this PR?

src/core/project/templates/harness.ts
```diff
@@ -0,0 +1,49 @@
...
+      ]);
+
+      return {
+        tree,
+        spec: {
+          harnesses: [{ name: parsed.name, path: join("app", parsed.name) }],
```
notgitika: join("app", name) uses the platform separator, and the result is written into agentcore.json, which gets committed and shared.

on Windows that's `"path": "app\\h1"`. if that is pulled down on macOS or Linux CI, it resolves that to a single filename with a backslash in it, and cdk deploy breaks for everyone except the author. can we always emit `/` and convert to platform separators only when touching the filesystem?

src/core/project/manager.tsx
notgitika, line 1: there is an edge case that when scaffolding fails, cleanup runs `rm -rf app/<name>`. But that's the whole directory, not just the files this command created. If the user already had something there, it's gone.

is there a way to track the files actually written and delete only those. can be a followup

Hweinstock: This might take some work to get right, lets do it as a follow-up.

notgitika: sounds good to me!

notgitika (approved):
thanks for addressing the comments!

aidandaly24 (approved):

tejaskash (approved):
Verified the requested fixes on the current head. Runtime additions now generate `uv.lock`, Python names are normalized before rendering, and missing Dockerfiles fail before any harness files are written. Focused tests and the full current-base merge validation passed.

PR #2102, opened 2026-08-25
feat(eval): read-only ab-test commands + TUI

jariy17:
Adds the full read-only + write `agentcore eval ab-test` command family, plus the interactive TUI. Five commits.

## Command + TUI structure

```
agentcore eval ab-test                      TUI menu shows get + list only
├── get     --id                            read  · JSON to stdout
├── list    --next-token --max-results      read  · JSON page to stdout
├── pause   --id                            write · executionStatus → PAUSED
├── resume  --id                            write · executionStatus → RUNNING
├── stop    --id                            write · executionStatus → STOPPED
└── delete  --id                            write · DeleteABTestCommand
```

## How it works

A/B tests are **imperative jobs** (like `batch-evaluation`), not project resources. All commands are one-liners through the data-plane SDK client `@aws-sdk/client-bedrock-agentcore`:

- `get`/`list` → `GetABTestCommand` / `ListABTestsCommand`. `get` returns per-evaluator metrics **inline** (no CloudWatch).
- `pause`/`resume`/`stop` → `UpdateABTestCommand({ abTestId, executionStatus })` — the same shape `online-eval` pause/resume uses.
- `delete` → `DeleteABTestCommand`.
- TUI: menu → picker → detail hub + JSON, modeled on `runtime`. Write commands are CLI-only and kept out of the interactive menu via `.supportedTuiCommands("get","list")` (matches `online-eval`).

## Known / deferred

1. **`stop` gateway-rule suggestion** — after stopping, the doc calls for printing a suggested (never executed) `update-gateway-rule` to promote the winner. Left as a `// TODO` this PR; tracked as a follow-up.
2. **`--json` error path** — failures print plain-text `Error: …` to stderr, not a JSON error object. Shared-renderer issue, not ab-test-specific; follow-up.
3. **Transition validity** (pause-on-RUNNING, delete-on-RUNNING, etc.) is enforced server-side; the CLI forwards and surfaces the service error. No client-side state machine, matching `online-eval`.

## Files

| Area | Files |
| --- | --- |
| CLI handlers | `handlers/eval/ab-test/{get,list,pause,resume,stop,delete}/index.tsx`, `index.tsx` |
| TUI screens | `handlers/eval/ab-test/{screen,list/screen,get/screen}.tsx`, `components/AbTestPicker.tsx`, 5 routes in `Root.tsx` |
| Core | `core/eval.tsx` (`getABTest`/`listABTests`/`setABTestExecutionStatus`/`deleteABTest`), `handlers/eval/types.tsx`, `testing/TestCoreClient.tsx` |

## Tests

- **Golden fixture** (`ab-test.fixture.test.tsx`, like `harness`): real router → CoreClient → recorded SDK → `matchGolden`. get / list / not-found.
- **Screen** (`ab-test.screen.test.tsx`, like `runtime`): menu (read-only), picker, hub, json, retry.
- **Command-flow** (`ab-test.write.test.tsx`, like `batch-evaluation`): hierarchy = get/list/pause/resume/stop/delete; each transition's `executionStatus`; delete; required `--id`.

## Bug bash — all green

- `bun test src/handlers/eval src/components` → 338+ pass / 0 fail; `bun run typecheck` → 0 errors; `oxlint` → clean.
- 4-agent `/review-pr` + a live non-destructive bug-bash on the exploratory account (us-west-2): all 6 commands dispatch correctly; write verbs return a clean not-found on a bogus id (no mutation); `--id` guard fires pre-network. One real bug found & fixed in this PR: the TUI menu listed write commands (dead-end) — now `.supportedTuiCommands`.

## Follow-ups (separate PRs)

- `stop`'s gateway-rule promotion suggestion
- structured `--json` error output (shared renderer)
- remove stale `projectSchemas/ab-test.ts` (contradicts the job model)

### DIFF
--- a/src/components/AbTestPicker.tsx
+++ b/src/components/AbTestPicker.tsx
@@ -1,0 +1,74 @@
+import type { ABTestSummary } from "@aws-sdk/client-bedrock-agentcore";
+import { useNavigate } from "react-router";
+import type { ScreenProps } from "../handlers/types";
+import { coreOptsFromCtx } from "../handlers/utils";
+import { formatTimestamp } from "./formatTimestamp";
+import { PaginatedTablePicker } from "./PaginatedTablePicker";
+import type { DataTableColumn } from "./ui/data-table";
+
+interface AbTestRow extends Record<string, unknown> {
+  abTestId: string;
+  name: string;
+  status: string;
+  executionStatus: string;
+  updatedAt: string;
+}
+
+export const abTestColumns = [
+  { key: "name", header: "name", flex: true },
+  { key: "status", header: "status", width: 14 },
+  { key: "executionStatus", header: "execution", width: 12 },
+  { key: "updatedAt", header: "updated UTC", width: 16, render: formatTimestamp },
+] satisfies DataTableColumn<AbTestRow>[];
+
+function toRow(summary: ABTestSummary): AbTestRow {
+  const id = summary.abTestId ?? "";
+  return {
+    abTestId: id,
+    name: summary.name ?? id,
+    status: summary.status ?? "-",
+    executionStatus: summary.executionStatus ?? "-",
+    updatedAt: summary.updatedAt?.toISOString() ?? "-",
+  };
+}
+
+export interface AbTestPickerProps extends ScreenProps {
+  breadcrumb: string[];
+  description?: string;
+  onSelect: (abTestId: string) => void;
+  onEscape?: () => void;
+}
+
+export function AbTestPicker({
+  ctx,
+  core,
+  breadcrumb,
+  description,
+  onSelect,
+  onEscape,
+}: AbTestPickerProps) {
+  const opts = coreOptsFromCtx(ctx);
+  const navigate = useNavigate();
+  const goBack = onEscape ?? (() => navigate("/" + breadcrumb.slice(0, -1).join("/")));
+
+  return (
+    <PaginatedTablePicker
+      breadcrumb={breadcrumb}
+      description={description}
+      queryKey={["ab-tests", opts.region]}
+      loadPage={async (token, pageSize) => {
+        const response = await core.eval.listABTests(token, pageSize, opts);
+        return { items: response.abTests ?? [], nextToken: response.nextToken };
+      }}
+      toRow={toRow}
+      columns={abTestColumns}
+      getValue={(row) => row.abTestId}
+      onSelect={onSelect}
+      onBack={goBack}
+      loadingMessage="Loading A/B tests…"
+      errorMessage={(error) => `Error: ${error.message}`}
+      emptyMessage="No A/B tests found in this Region."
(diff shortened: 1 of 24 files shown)


src/components/AbTestPicker.tsx
```diff
@@ -0,0 +1,74 @@
...
+  { key: "updatedAt", header: "updated UTC", width: 16, render: formatTimestamp },
+] satisfies DataTableColumn<AbTestRow>[];
+
+function toRow(summary: ABTestSummary): AbTestRow {
+  const id = summary.abTestId ?? "";
+  return {
+    abTestId: id,
+    name: summary.name ?? id,
+    status: summary.status ?? "-",
+    executionStatus: summary.executionStatus ?? "-",
```
notgitika, line 30: discussed offline and we see that this is the output from the API. i would imagine as customer would be confused seeing these 2 fields in the data table in TUI. we can try to unify it in the future maybe

src/handlers/eval/ab-test/ab-test.screen.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+          evaluatorArn: "arn:aws:bedrock-agentcore:us-east-1:123456789012:evaluator/quality",
+          controlStats: { treatmentName: "C", sampleSize: 100, mean: 0.8 },
+          variantResults: [
+            { treatmentName: "T1", sampleSize: 100, mean: 0.86, isSignificant: true },
+          ],
+        },
+      ],
+    },
+    ...overrides,
+  } as GetABTestResponse;
```
notgitika: is the reason we are not using `satisfies` that we have `treatmentname` and not `variantname`? I see that in the fixture

src/core/eval.tsx
```diff
@@ -395,6 +389,20 @@ export class EvalClient implements CoreEvalClient {
       .send(new ListBatchEvaluationsCommand({ nextToken, maxResults }));
   }
 
+  async getABTest(id: string, options: CoreOptions): Promise<GetABTestResponse> {
+    return this.clients.data(toClientConfig(options)).send(new GetABTestCommand({ abTestId: id }));
+  }
+
+  async listABTests(
+    nextToken: string | undefined,
+    maxResults: number | undefined,
```
notgitika, line 403: nice that we have these as just passthroughs :)

notgitika (dismissed):

src/handlers/eval/ab-test/ab-test.screen.test.tsx
jariy17: Good catch — switched to `satisfies GetABTestResponse` in c2e0c18. It immediately surfaced that the stub had `treatmentName` (the real metric field is `variantName`, matching the recorded fixture) and was missing `Variant.variantConfiguration`. Both fixed; the cast was hiding them.

notgitika (dismissed):

AlexanderRichey (approved):

Issue #2104, opened 2026-08-25
Project deploy: same project name + target collides on one CloudFormation stack


notgitika:
## Summary
Two separate projects that share the same `name` (in `agentcore.json`) and target, deployed to the same account+region, resolve to the **same** CloudFormation stack. The second deploy updates/overwrites the first project's stack instead of creating its own.

## Repro
1. Create two projects in different directories, both `name: "demo"`.
2. `agentcore deploy --target default` from each, same account+region.
3. Both map to stack `AgentCore-demo-default`; the second deploy overwrites the first's stack.

## Root cause
Stack identity is derived solely from `(project name, target name)` — `AgentCore-<name>-<target>` (`src/assets/cdk/bin/cdk.ts` `toStackName`). The project spec has no stable unique id distinct from the human-chosen `name`, and the `agentcore:project-name` tag is just the name, so nothing disambiguates two same-named projects. The deploy path only verifies the account matches the target (`src/core/project/backends/cdk.ts`), not project identity.

## Impact
Silent overwrite of another project's deployed resources when names collide in a shared account+region.

## Suggested fix
Generate a stable project id at `project init`, store it committed in `agentcore.json`, stamp it as a stack tag (e.g. `agentcore:project-id`), and refuse to deploy when an existing stack of that name carries a different id.

PR #2105, opened 2026-08-25
feat: persist minimal deploy state to deployed-state.json

notgitika:
Moves the deploy-state file to a committed `agentcore/.cli/deployed-state.json`, and stops storing a full snapshot of every resource.

State is now keyed per target and holds only the deployed CloudFormation **stack ARN** (captured from the toolkit deploy result) plus the imperatively-created **credential ARNs** the synth step needs before the stack exists. Everything else is meant to be read live from CloudFormation, so the file can't go stale.

## Changes
- New `DeployedState` schema + `readDeployedState` / `updateTargetState` (merge-not-clobber: preserves sibling targets and keys this CLI doesn't own).
- CDK toolkit runner now surfaces `stackArn`; `CdkBackend.deploy` persists it per target after a successful deploy.
- Vended CDK app reads the new top-level path.

## Notes
- The live CloudFormation reader (for `status` and resource ARNs) and the credentials preflight that populates the credential ARNs are follow-ups.
- Related: a pre-existing same-name stack-collision bug is tracked in #2104 (orthogonal to this change).

### DIFF
--- a/src/assets/cdk/bin/cdk.ts
+++ b/src/assets/cdk/bin/cdk.ts
@@ -124,10 +124,11 @@ async function main() {
   const connectorParametersByFile = resolveConnectorParametersByFile(specAny, projectRoot);
   const harnessConfigs = resolveHarnessConfigs(specAny, projectRoot);
 
-  // Read deployed state for credential ARNs (populated by pre-deploy identity setup)
+  // Read deployed state for credential ARNs (populated by pre-deploy identity setup).
+  // Lives top-level under agentcore/ (committed), not the gitignored .cli/ dir.
   let deployedState: Record<string, unknown> | undefined;
   try {
-    deployedState = JSON.parse(fs.readFileSync(path.join(configRoot, '.cli', 'deployed-state.json'), 'utf8'));
+    deployedState = JSON.parse(fs.readFileSync(path.join(configRoot, 'deployed-state.json'), 'utf8'));
   } catch {
     // Deployed state may not exist on first deploy
   }
--- a/src/assets/cdk/bin/cdk.ts
+++ b/src/assets/cdk/bin/cdk.ts
@@ -129,8 +129,11 @@ async function main() {
   let deployedState: Record<string, unknown> | undefined;
   try {
     deployedState = JSON.parse(fs.readFileSync(path.join(configRoot, 'deployed-state.json'), 'utf8'));
-  } catch {
-    // Deployed state may not exist on first deploy
+  } catch (err) {
+    // A missing file is the normal first-deploy case. A malformed one is not:
+    // surface it rather than silently synthesizing without the credential ARNs
+    // it holds (which would drop them from the stack).
+    if ((err as NodeJS.ErrnoException).code !== 'ENOENT') throw err;
   }
 
   const app = new App();
--- a/src/assets/cdk/bin/cdk.ts
+++ b/src/assets/cdk/bin/cdk.ts
@@ -125,10 +125,10 @@ async function main() {
   const harnessConfigs = resolveHarnessConfigs(specAny, projectRoot);
 
   // Read deployed state for credential ARNs (populated by pre-deploy identity setup).
-  // Lives top-level under agentcore/ (committed), not the gitignored .cli/ dir.
+  // Under agentcore/.cli/ to match the released CLI's location.
   let deployedState: Record<string, unknown> | undefined;
   try {
-    deployedState = JSON.parse(fs.readFileSync(path.join(configRoot, 'deployed-state.json'), 'utf8'));
+    deployedState = JSON.parse(fs.readFileSync(path.join(configRoot, '.cli', 'deployed-state.json'), 'utf8'));
   } catch (err) {
     // A missing file is the normal first-deploy case. A malformed one is not:
     // surface it rather than silently synthesizing without the credential ARNs
--- a/src/core/project/backends/cdk/deployedState.test.ts
+++ b/src/core/project/backends/cdk/deployedState.test.ts
@@ -119,4 +119,34 @@ describe("updateTargetState", () => {
       },
     });
   });
+
+  test("preserves unknown fields inside a credential entry across an update", async () => {
+    const root = await projectRoot();
+    // A field a newer CLI (or the CDK app) records that this code doesn't model.
+    await Bun.write(
+      statePath(root),
+      JSON.stringify({
+        targets: {
+          default: {
+            resources: {
+              credentials: { k: { credentialProviderArn: "arn:a", futureField: "keep" } },
+            },
+          },
+        },
+      }),
+    );
+
+    await updateTargetState(json, root, "default", { stackArn: "arn:stack:default" });
+
+    expect(await readRaw(root)).toEqual({
+      targets: {
+        default: {
+          stackArn: "arn:stack:default",
+          resources: {
+            credentials: { k: { credentialProviderArn: "arn:a", futureField: "keep" } },
+          },
+        },
+      },
+    });
+  });
 });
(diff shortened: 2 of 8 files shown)


notgitika:
just ignore the harness reviewer :/

src/core/project/backends/cdk/deployedState.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+function statePath(root: string): string {
+  return join(root, DEPLOYED_STATE_RELATIVE_PATH);
+}
+
+async function readRaw(root: string): Promise<unknown> {
+  return JSON.parse(await Bun.file(statePath(root)).text());
+}
+
+describe("readDeployedState", () => {
```
Hweinstock, line 37: is there a way to test this behavior through the handlers? The advantage being that implementation details here could change, but we still observe the same behavior e2e. 

If that isn't really possible, than maybe one level higher in cdk?

Hweinstock (dismissed):

src/assets/cdk/bin/cdk.ts
```diff
@@ -124,10 +124,11 @@ async function main() {
   const connectorParametersByFile = resolveConnectorParametersByFile(specAny, projectRoot);
   const harnessConfigs = resolveHarnessConfigs(specAny, projectRoot);
 
-  // Read deployed state for credential ARNs (populated by pre-deploy identity setup)
+  // Read deployed state for credential ARNs (populated by pre-deploy identity setup).
+  // Lives top-level under agentcore/ (committed), not the gitignored .cli/ dir.
   let deployedState: Record<string, unknown> | undefined;
   try {
-    deployedState = JSON.parse(fs.readFileSync(path.join(configRoot, '.cli', 'deployed-state.json'), 'utf8'));
```
AlexanderRichey, line 130: This is code that's on the customer's side, right? We write this `cdk.ts` file to their disk, correct? If so, does that mean the existing `agentcore status` command depends on this file? Or that the existing `agentcore deploy` does?

src/core/project/backends/cdk/deployedState.ts
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Merges a patch into a single target's entry and writes the whole file back,
+ * preserving every other target and every resource kind this code does not own.
+ *
+ * The merge is shallow except for `resources`, which is merged one level deep so
+ * updating one resource kind (e.g. `credentials`) leaves the others in place.
+ * A resource map provided in the patch replaces the previous map for that kind
+ * wholesale, so a credential dropped from the spec stops being advertised.
+ */
+export async function updateTargetState(
```
AlexanderRichey, line 88: It's a bit hard to see if we need these utilities without seeing where they're used. It seems like the real purpose of this PR is to add the needed DeployedState types. Is that right?

src/assets/cdk/bin/cdk.ts
```diff
@@ -124,10 +124,11 @@ async function main() {
...
+    deployedState = JSON.parse(fs.readFileSync(path.join(configRoot, 'deployed-state.json'), 'utf8'));
```
aidandaly24: I think malformed deployed state can leave us in a bad partial state here. This catches the parse error and synthesis continues, so AWS deployment can complete before `updateTargetState()` rereads the same invalid file and throws. The command then reports failure after mutating AWS without recording the new stack ARN. We should only treat a missing file as absent and validate existing state before deployment.

src/core/project/backends/cdk/deployedState.ts
```diff
@@ -0,0 +1,104 @@
...
+  json: ReadWriteJson,
+  projectRoot: string,
+  targetName: string,
+  patch: Partial<TargetState>,
+): Promise<DeployedState> {
+  const statePath = statePathFor(projectRoot);
+  const state = await readDeployedState(json, projectRoot);
```
aidandaly24, line 95: The “never drops another target” guarantee only holds for sequential updates. I reproduced two concurrent updates both reading empty state and the final file containing only `prod`. If concurrent deployments are not supported, I think we should remove or qualify this guarantee rather than imply the merge is generally safe.

src/core/project/backends/cdk/toolkit.ts
```diff
@@ -28,7 +28,13 @@ export type CdkOutputs = Record<string, string>;
 export type CdkCredentialProvider = SdkBaseConfig["credentialProvider"];
 export type CdkCredentialResolver = (region: string) => Promise<CdkCredentialProvider>;
 
-export type CdkRunner = (operation: CdkOperation, options: CdkRunOptions) => Promise<CdkOutputs>;
+/**
+ * Result of a CDK operation. `stackArn` is the ARN of the deployed stack (only
+ * a deploy produces one); bootstrap leaves it undefined.
+ */
+export type CdkRunResult = { outputs: CdkOutputs; stackArn?: string };
```
aidandaly24, line 35: `stackArn` should probably be required for deploy results. CDK’s `DeployedStack` defines it as required, but `CdkRunResult` makes it optional and the backend silently skips persistence when it is absent. That leaves a successful deployment that later commands cannot resolve. I think we should fail instead of accepting a deploy result without the ARN.

src/core/project/backends/cdk/deployedState.ts
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Project-relative path of the state file the synthesized CDK app reads.
+ *
+ * Deliberately top-level under `agentcore/` (not the gitignored `.cli/`): the
+ * file is meant to be committed so a target's stack binding and imperatively
+ * created credential ARNs are shared across everyone who deploys that target.
+ */
+export const DEPLOYED_STATE_RELATIVE_PATH = join("agentcore", "deployed-state.json");
+
+const CredentialStateSchema = z.object({
```
aidandaly24: One edge case here: Zod strips fields not listed in `CredentialStateSchema`. A stack-ARN-only update reads and rewrites the whole file, so it can delete future or unowned fields inside an existing credential entry. I think this nested schema should use `.passthrough()` too.

src/core/project/backends/cdk/deployedState.ts
```diff
@@ -0,0 +1,104 @@
...
+    ...patch,
+    ...(resources && { resources }),
+  };
+
+  const next: DeployedState = {
+    ...state,
+    targets: { ...state.targets, [targetName]: merged },
+  };
+
+  await json.write(statePath, next);
```
aidandaly24: This state file should use the existing `atomicWrite()` helper. `FsReadWriteJson.write()` truncates the authoritative file directly, so an interruption or disk failure can leave malformed JSON that blocks later deploys.

aidandaly24 (changes requested):
I found several deployed-state integrity cases that I think need to be addressed before this merges.

src/assets/cdk/bin/cdk.ts
notgitika: fixed!

src/core/project/backends/cdk/deployedState.ts
notgitika: i qualified the comment

src/assets/cdk/bin/cdk.ts
notgitika: the existing `status` command does. we decided to keep it under `.cli/` for the time being

src/core/project/backends/cdk/toolkit.ts
notgitika: done

src/core/project/backends/cdk/deployedState.ts
notgitika: fixed

src/core/project/backends/cdk/deployedState.ts
notgitika: done! thanks for the thorough review!

aidandaly24 (approved):

PR #2108, opened 2026-08-25
refactor(core): shared paginateFiltered for client-side-filter listing

jariy17:
**One helper for every `list` that filters a broader API client-side.**

## Why
Three `list*` methods over-fetch a broad service API and filter client-side:

## What changed
1. New **`FilteredPaginator`** class (`src/core/filteredPaginator.ts`) — scans **full** service pages, fills the requested page across them, bounded by 101 scans (`ResultTruncationError` past that).
2. `listOnlineInsights` / `listBatchInsights` / `listGatewayConnectors` each delegate to the static `FilteredPaginator.paginate({ ... })`, passing their predicate + page sizes inline.
3. Deleted the 3 inline loops + their per-method scan-cap constants.

## Behavior changes
- **A page seam can duplicate one item.** At a boundary the paginator hands back the scan page's *own* token, so its matches reappear on the next page (not a re-fetched cursor). Deliberate: it never re-requests a token at a different `maxResults`, so list APIs that bind `maxResults` into the token can't reject the continuation.
- **A page may slightly exceed `--max-results`** when one service page already holds a full page of matches (guards against an infinite loop).
- **`listGatewayConnectors` now validates `maxResults`** (rejects `<1` / non-integer) — it didn't before.
- Net for `listBatchInsights` + `listGatewayConnectors`: exact-size pages → dup-on-boundary.

## Tests
- `filteredPaginator.test.ts` — 10 unit tests: validation, under-fill, overshoot+dup, guard over-return, truncation cap, token seeding, scan sizing.
- Per-method tests trimmed to **filter/wiring only** — pagination mechanics now live once in the helper.
- Full suite **2019 pass / 0 fail**, `tsc` clean.
- Live bug-bashed on the exploratory account: 5/5 scenarios (list matrix, id-op guard, connector list, batch-insights list, TUI), residue-free.


src/core/eval.tsx
```diff
@@ -440,49 +436,18 @@ export class EvalClient implements CoreEvalClient {
...
-      if (response.nextToken === undefined) {
-        return { ...response, batchEvaluations, nextToken: undefined };
-      }
-      batchEvaluationToken = response.nextToken;
-    }
-
-    throw new ResultTruncationError(
-      `Batch Insights discovery exceeded ${MAX_BATCH_INSIGHTS_SCAN_REQUESTS} Batch Evaluation scan requests; results are incomplete`,
-    );
+    const page = await FilteredPaginator.paginate({
```
tejaskash, line 439: Like the design of this

tejaskash (approved):

src/core/filteredPaginator.ts
```diff
@@ -0,0 +1,65 @@
...
+    for (let scan = 0; scan < MAX_SCAN_REQUESTS; scan++) {
+      const requestToken = token;
+      const page = await fetchPage(requestToken, scanPageSize);
+      const matches = page.items.filter(predicate);
+      results.push(...matches);
+
+      if (results.length >= pageSize) {
+        // Page holds >= pageSize matches by itself. Return every match found (the
+        // page may exceed maxResults) and advance past it: replaying its token would
+        // loop, and skipping the surplus would drop matches — so we over-return.
```
nborges-aws, line 47: Doesn't over returning here just drop resources then? We only render up to `pageSize` rows, but we continue from where the service token leaves off. Everything beyond maxResults that was included by the service becomes permanently unreachable.

src/core/filteredPaginator.ts
```diff
@@ -0,0 +1,65 @@
...
+        if (matches.length >= pageSize) {
+          return { items: results, nextToken: page.nextToken };
+        }
+        // Partial page: replaying its token is safe — the taken matches just repeat.
```
nborges-aws, line 51: I get how replaying the service page here prevents dropping matched items. But for the inverse case, wouldn't this duplicate every match already consumed from the first partial page? If page A matches 2 items and page B matches 8, we're still using the token that returns page B. So all 8 from the second page would be duplicated and rendered twice. 

I think this needs to change so that if A + B fills the max results, we return nextToken?

nborges-aws (changes requested):
Love that we're adding a shared utility for this. A few issues with our filtering algorithm that we should iron out first

src/core/filteredPaginator.ts
jariy17: This is best case scenario without introduce a composite token. I think its fine customers seen a duplicate. If this is an issue, I can introduce now.

src/core/filteredPaginator.ts
jariy17: This is best case scenario. It's not the best solution we can find. We can generate a composite token ({ serviceToken, skip }) so we can prevent this and the replay issue but its too complicated for a small inconvenience.

nborges-aws (approved):

PR #2109, opened 2026-08-26
fix(dev): cancel readiness retries on abort

Hweinstock:
## Problem

`waitForPort` left active sockets and scheduled retries alive after abort. In CI, the supervisor's 60ms abort test instead ran for about 60 seconds and timed out; the original #2041 Windows run eventually lost its runner.

## Solution

Abort now destroys the active socket, clears the pending retry, removes its listener, and rejects immediately. The TCP test also closes accepted sockets and awaits listener shutdown.

## Testing

- Windows unit CI passed in 3m24s.
- Supervisor suite passed 25 consecutive runs; the abort test completes in about 60ms.
- Supervisor, project dev, and port tests passed: 31/31.
- Prettier and `tsc --noEmit` passed.

### DIFF
--- a/src/core/dev/supervisor.test.ts
+++ b/src/core/dev/supervisor.test.ts
@@ -292,11 +292,13 @@ describe("DevSupervisor", () => {
       "produced no output and did not accept connections on port 1 within 0.1s",
     );
 
-    const server = createServer();
+    const server = createServer((socket) => socket.destroy());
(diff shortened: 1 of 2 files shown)


Hweinstock:
Related: https://github.com/aws/agentcore-cli/issues/2110

tejaskash (approved):
Thanks for the fix!

jariy17 (approved):

PR #2111, opened 2026-08-26
feat: support imperative eval recommendation command

nborges-aws:
## Description

Adds imperative command-line support for AgentCore evaluation recommendations:

* `eval recommendation start`
* `eval recommendation get`
* `eval recommendation list`
* `eval recommendation delete`

**Summary of Changes**

* Supports `SYSTEM_PROMPT_RECOMMENDATION` and `TOOL_DESCRIPTION_RECOMMENDATION` recommendation types
* `start` accepts recommendation configuration as inline JSON, a `file://` path, or stdin (`-`) through `SourceResolver` class
* Supports the service API fields for description, KMS key ARN, and tags
* Adds the recommendation operations to `EvalClient`, the shared handler types, and `TestCoreClient`
* Adds recorded golden fixtures covering the complete start, get, list, and delete flow

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Added unit tests and golden fixture tests exercising recommendation logic.

- [x] `bun run test` (2052 pass, 0 fail)
- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.


src/handlers/eval/recommendation/get/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+import type { Core } from "../../../types";
+import { coreOptsFromCtx } from "../../../utils";
+
+export const createGetRecommendationHandler = (core: Core) =>
+  createHandler({
+    name: "get",
+    description: "get a recommendation by id",
+    flags: [flag("id", "the ID of the recommendation", z.string().optional())],
+    handle: async (ctx, flags) => {
+      if (!flags["id"]) throw new InputValidationError("required option '--id <id>' not specified");
```
jariy17, line 14: we can definitely abstracted this so we have a common error message for this type of thing.  OBO

src/core/eval.tsx
```diff
@@ -341,6 +351,39 @@ export class EvalClient implements CoreEvalClient {
       .send(new DeleteEvaluatorCommand({ evaluatorId: id }));
   }
 
+  async startRecommendation(
+    request: StartRecommendationRequest,
```
jariy17: We create our own types here because handlers define the interface for the core client. Even if it's a one-to-one mapping, we should still define our own type.

src/core/recommendation.test.tsx
```diff
@@ -0,0 +1,119 @@
+import { describe, expect, test } from "bun:test";
```
jariy17: We don't need this. Use Golden tests please

src/handlers/eval/recommendation/start/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+      flag("description", "a description of the recommendation", z.string().optional()),
+      flag(
+        "kms-key-arn",
+        "customer managed KMS key ARN for recommendation data",
+        z.string().optional(),
+      ),
+      flag("tags", "tags as key=value (repeatable) or JSON object", z.array(z.string()).optional()),
+    ],
+    handle: async (ctx, flags) => {
+      if (!flags["name"]) {
```
jariy17:
```
const requiredFlags: Array<[string, string]> = [
  ["name", "--name <name>"],
  ["type", "--type <type>"],
  ["recommendation-config", "--recommendation-config <recommendation-config>"],
];

for (const [key, usage] of requiredFlags) {
  if (!flags[key]) {
    throw new InputValidationError(`required option '${usage}' not specified`);
  }
}
```

jariy17: nit: we should use something like this.

src/handlers/eval/recommendation/recommendation.test.tsx
```diff
@@ -0,0 +1,332 @@
...
+  };
+}
+
+function callArgs(core: TestCoreClient, method: string): unknown[] {
+  const call = core.eval.calls.find((candidate) => candidate.method === method);
+  if (!call) throw new Error(`${method} was not called`);
+  return call.args;
+}
+
+describe("eval recommendation command hierarchy", () => {
```
jariy17: do we need these tests?

src/handlers/eval/recommendation/get/index.tsx
Hweinstock: yeah I think marking flags as required and supporting it directly in the framework would be ideal, but likely OOS here.

src/handlers/eval/recommendation/recommendation.fixture.test.tsx
```diff
@@ -0,0 +1,273 @@
...
+    throw new Error("start fixture did not return a recommendation id");
+  }
+  return recommendationId;
+}
+
+function isNotFound(error: unknown): boolean {
+  return (error as Error).name === "ResourceNotFoundException";
+}
+
+async function waitForTerminal(
```
Hweinstock, line 123: is there a way to leverage https://github.com/aws/agentcore-cli/blob/01c9317785c01bad268cd8356517bcb45de746ec/src/testing/timing.tsx#L16 for these?

src/handlers/eval/recommendation/recommendation.test.tsx
Hweinstock: +1, I've seen these tests in other places, but they feel pretty low value. The tests on the handlers would fail if the hierarchy is incorrect. Also, what coverage do we get on the tests below that we don't get on the fixture based tests?

src/handlers/eval/recommendation/recommendation.test.tsx
```diff
@@ -0,0 +1,332 @@
...
+    const { core, stdout, route } = testRecommendationCommand();
+    const response = {
+      recommendationId: "recommendation-1",
+      recommendationArn: "arn:recommendation-1",
+      name: "support-prompt",
+      type: "SYSTEM_PROMPT_RECOMMENDATION",
+      status: "PENDING",
+      createdAt: new Date("2026-08-26T12:00:00.000Z"),
+      updatedAt: new Date("2026-08-26T12:00:00.000Z"),
+    } as StartRecommendationResponse;
```
Hweinstock: why do we need to cast here?

src/core/recommendation.test.tsx
Hweinstock: are the handler tests sufficient? I think these are testing the same things at a lower level.

src/core/eval.tsx
nborges-aws: sure will update

src/handlers/eval/recommendation/recommendation.fixture.test.tsx
nborges-aws: Good call out I'll switch to use this util

src/handlers/eval/recommendation/start/index.tsx
nborges-aws: Sure will add something like this

src/handlers/eval/recommendation/recommendation.test.tsx
nborges-aws: I added these tests to cover behavior the fixtures don't exercise. E.g. malformed json, exercising source resolver, etc. But consensus is against unit test so I've removed this + the other one entirely

src/handlers/eval/recommendation/recommendation.test.tsx
nborges-aws: N/A since test file was removed altogether

jariy17 (approved):

jariy17 (approved):

PR #2112, opened 2026-08-26
feat: read project stack state live from CloudFormation

notgitika:
Adds the read-side of the deploy-state refactor: a small helper to describe a project's CloudFormation stack, so resource details can be read live instead of from a local snapshot that goes stale.

## What's here
- `describeStack(region, credentials, stackName)` — a `DescribeStacks` call that returns the stack, or `undefined` when it doesn't exist. Accepts a stack name or ARN. The AWS call is injectable (and lazy-loaded like `environment.ts`), so it's unit-tested without a real client.
- Generalized the existing bootstrap `isBootstrapStackNotFound` → `isStackNotFound` and reused it.

## Not here (by design)
Interpreting the stack — status classification (deployed / in-progress / failed) and which outputs to surface — is a `project status` decision and is left to whoever builds it, rather than baked in ahead of the consumer.

### DIFF
--- a/src/core/project/backends/cdk/stackReader.test.ts
+++ b/src/core/project/backends/cdk/stackReader.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, test } from "bun:test";
 import type { Stack } from "@aws-sdk/client-cloudformation";
-import { classifyStack, readStackState, type StackReader } from "./stackReader";
+import { describeStack, type DescribeStacks } from "./stackReader";
 import type { CdkCredentialProvider } from "./toolkit";
 
 const credentials: CdkCredentialProvider = async () => ({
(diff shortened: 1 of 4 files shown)


src/core/project/backends/cdk/stackReader.ts
```diff
@@ -0,0 +1,97 @@
...
+  if (READY_STATUSES.has(status)) {
+    return { kind: "ready", status, outputs: outputsToRecord(stack.Outputs) };
+  }
+  return { kind: "failed", status };
+}
+
+const describeStack: StackReader = async (region, credentials, stackName) => {
+  const { CloudFormationClient, DescribeStacksCommand } =
+    await import("@aws-sdk/client-cloudformation");
+  const client = new CloudFormationClient({ credentials, region });
```
Hweinstock: Is there a way we can inject this client from core clients?

Maybe define cloudformation client like the others here https://github.com/aws/agentcore-cli/blob/01c9317785c01bad268cd8356517bcb45de746ec/src/core/index.tsx#L47-L57

and inject it into the project manager?

src/core/project/backends/cdk/stackReader.test.ts
```diff
@@ -0,0 +1,128 @@
...
+function stack(status: Stack["StackStatus"], outputs?: Stack["Outputs"]): Stack {
+  return {
+    StackName: "AgentCore-example-default",
+    CreationTime: new Date(0),
+    StackStatus: status,
+    ...(outputs && { Outputs: outputs }),
+  };
+}
+
+describe("classifyStack", () => {
```
Hweinstock: I feel like it would be stronger to test this through the handler (once it exists), but since it doesn't exist yet this seems reasonable

Hweinstock (commented):
I wonder if we're coming at this problem from the wrong side. I understand the end goal is to have a status handlers that reads off the resources and whether they are deployed or not, but it feels strange to build the functionality before the interface. 

I feel like we're inevitably going to be reworking the functionality to match the interface or derive the interface from the functionality (which I don't think we want).

notgitika:
> I wonder if we're coming at this problem from the wrong side. I understand the end goal is to have a status handlers that reads off the resources and whether they are deployed or not, but it feels strange to build the functionality before the interface.
> I feel like we're inevitably going to be reworking the functionality to match the interface or derive the interface from the functionality (which I don't think we want).

I see your point and I agree. I thiknk the classification is really a status-interface decision so it can be worked on together with it. I can pull it back in this PR and only have  the raw API (describeStackByName → DescribeStacks, returning the stack or undefined) implemented.

notgitika:
Trimmed it down to just the raw describeStack call, dropped the status classification since that's really a status decision, not this PR's.

src/core/project/backends/cdk/stackReader.ts
notgitika: I made the call injectable so it's testable without a real client. Left it lazy-importing the SDK for now rather than going through CoreClient. figured that's cleaner to wire up when status actually consumes it. Lmk if you'd rather do it now.

notgitika:
patch coverage being low is expected for now. the tests deliberately bypass it by injecting a fake describe

src/core/project/backends/cdk/stackReader.ts
```diff
@@ -0,0 +1,52 @@
...
+export async function describeStack(
+  region: string,
+  credentials: CdkCredentialProvider,
+  stackName: string,
+  describe: DescribeStacks = cloudFormationDescriber(region, credentials),
+): Promise<Stack | undefined> {
+  try {
+    // Not-found is a thrown ValidationError, not an empty result; every other
+    // error (auth, throttling, malformed request) is real and propagates.
+    return (await describe(stackName))?.[0];
```
aidandaly24: nit: An empty successful `DescribeStacks` response is different from a missing stack. CloudFormation reports not-found through the thrown `ValidationError`. An empty successful response is malformed. Returning `undefined` here makes callers report “not deployed” instead of the actual service-response problem. This should throw `MalformedServiceResponseError`, like the bootstrap reader does.
Could be a followup or other PR though.

aidandaly24 (dismissed):

src/core/project/backends/cdk/stackReader.ts
notgitika: good point, fixed it in this PR itself

aidandaly24 (approved):

PR #2113, opened 2026-08-26
feat(eval): ab-test config-based run (create)

jariy17:
## Command structure

```
agentcore eval ab-test config-based run
  -n, --name <name>              required   A/B test name
  -g, --gateway <id>             required   deployed gateway id
  --control <json>               required   {"config-bundle","bundle-version"}  (inline · file:// · -)
  --treatment <json>             required   {"config-bundle","bundle-version"}  (inline · file:// · -)
  --online-eval <id>             required   online-evaluation config id
  --treatment-weight <1-99>                 control weight = 100 - this (default 50)
  --gateway-filter <json>                   GatewayFilter {"targetPaths":[...]}  (inline · file:// · -)
  --role-arn <arn>                          execution-role override (else auto-provisioned)
  --enable-on-create <true|false>           start immediately (default true; false = create PAUSED)
```

Starts an A/B test that splits gateway traffic between two config-bundle versions on one runtime. `config-based` is one of the two `ab-test` run modes (the other is `target-based`, #2135). First **create** command in the `ab-test` family (get/list/pause/resume/stop/delete already merged in #2102).

## Actual input → output

**Input** (run against the exploratory account, us-west-2):

```bash
agentcore eval ab-test config-based run \
  --name orders-prompt-v2-rollout \
  --gateway orders-gateway-demo-000 \
  --control '{"config-bundle":"orders-agent-prompt-abc","bundle-version":"1111...1111"}' \
  --treatment '{"config-bundle":"orders-agent-prompt-abc","bundle-version":"2222...2222"}' \
  --online-eval online-eval-abc123 \
  --treatment-weight 20 \
  --gateway-filter '{"targetPaths":["/orders/checkout"]}'
```

**Actual output** — the CLI resolves + validates the gateway *before* any create, so an unreachable gateway fails fast and creates nothing:

```
Error: User: arn:aws:sts::…:assumed-role/Admin/… is not authorized to perform: bedrock-agentcore:GetGateway
```

**Output on success** (shape returned by `CreateABTest` — a real create was intentionally not run: it provisions IAM + routes live traffic on a shared account):

```json
{
  "abTestId": "orders-prompt-v2-rollout-abc123",
  "abTestArn": "arn:aws:bedrock-agentcore:us-west-2:…:ab-test/orders-prompt-v2-rollout-abc123",
  "name": "orders-prompt-v2-rollout",
  "status": "CREATING",
  "executionStatus": "RUNNING",
  "createdAt": "2026-08-26T15:00:00.000Z"
}
```

With `--enable-on-create false`, `executionStatus` is `NOT_STARTED` (created PAUSED, no traffic).

## How it works

- **One SDK call** — `CreateABTestCommand` on the data-plane client. Builds `C`/`T1` variants (weights `100 - w` / `w`), `evaluationConfig = { onlineEvaluationConfigArn }`, and passes `--gateway-filter` straight through as the SDK `GatewayFilter`.
- **`--control` / `--treatment` / `--gateway-filter`** each accept inline JSON, `file://<path>`, or `-` (stdin) via `SourceResolver` — same pattern as `online-eval create`'s `--filters`. The `--control`/`--treatment` JSON keys stay `config-bundle`/`bundle-version` because they name the actual configuration-bundle resource + version being pinned.
- **Execution role** — required by the API. When `--role-arn` is omitted, Core provisions one (`abTestExecutionRole.ts`: confused-deputy-scoped trust + gateway-rule/get inline policy), retries `CreateABTest` while IAM propagates (data-plane 403 too), and rolls the role back if create fails.
- **Validation is server-side** — `CreateABTest` validates gateway READY, bundle versions, and the online-eval (enabled iff starting) by assuming the role. The CLI resolves ids → ARNs and surfaces the service's 4xx; no duplicated client-side state machine.
- **`--runtime` dropped** — not a `CreateABTest` field (deviates from the doc on purpose).

## Tests

- `ab-test.test.tsx` — command hierarchy + flag→request mapping, `--gateway-filter` passthrough, `--enable-on-create`/`--role-arn`, identical-variant rejection, weight bounds, required-flag matrix, malformed/mis-shaped JSON.
- `ab-test.fixture.test.tsx` — golden create→delete flow (`config-based run`) replayed offline, plus the read-only get/list goldens.
- `abTestExecutionRole.test.ts` — 64-char name cap, trust + inline policy bodies, create-vs-reuse.

## Bug bash — 4 agents + live exploratory (us-west-2)

- **Standard / Munger / AI-slop (3/10) / live**: all green. Live bash ran 8 non-destructive cases — every path errored **before** any create; no state mutated.
- **Fixed from review (commit 2):** identical-variant guard now compares the `(bundle, version)` pair (was version-only); `CreateABTest` retries on data-plane `AccessDenied`; `accountId` extraction throws instead of a silent `"*"`; `--treatment-weight` must be an integer; added the role-module unit tests.
- `bun test` → all pass / 0 fail · `typecheck` → 0 errors · `oxlint` → clean.

## Follow-ups (separate PRs)

- Structured `--json` error output (errors print plain text to stderr — cross-cutting, all commands).
- TUI wizard for `config-based run`.
- `stop`'s gateway-rule promotion suggestion (still a TODO).


nborges-aws (approved):
LGTM no findings

PR #2115, opened 2026-08-26
feat(project): add project-aware invoke

aidandaly24:
## Description

Add project-aware invocation for Runtimes and Harnesses declared by the project enclosing the current directory.

- add `agentcore project invoke` as an interactive picker for deployed project Runtimes and Harnesses on the default target
- add explicit headless child commands:
  - `agentcore project invoke runtime --name <name> --payload <payload>`
  - `agentcore project invoke harness --name <name> --prompt <prompt>`
- allow `--name` to be omitted when the project declares exactly one resource of that type
- resolve logical project names to deployed physical IDs and regions through the selected deployment target and live CloudFormation stack
- preserve the existing Runtime payload, content-type, response, output-file, and JSON behavior
- preserve the existing Harness prompt, session, transcript, and chat behavior
- reuse the Runtime and Harness invoke operations and TUI consoles instead of adding a third transport or presentation path
- document project invoke in the CLI and generated Runtime templates

This PR intentionally does not define a project-specific Runtime payload or response contract. Runtime responses retain the existing raw invoke behavior, including raw SSE. Gateway project invoke remains outside this scope.

The project deployment prerequisites are already present on `refactor`:

- #2099 for the current Runtime templates
- #2105 for per-target deployed stack state
- #2112 for live CloudFormation stack reads

## Related Issue

N/A

## Documentation PR

N/A - documentation is included.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Other

## Testing

- `bun test src`: `2560 pass, 0 fail`
- focused project invoke/menu/CDK stack suites: `70 pass, 0 fail`
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- `bun run secrets:check`
- `bun run build`
- `git diff --check`

Live verification in account `603141041947`, region `us-west-2`, using the retained `InvokeMatrix827` project:

- Runtime headless invoke omitted `--name`, preserved the native raw SSE response, and returned `REBASED_RUNTIME_OK`
- Harness headless invoke omitted `--name`, preserved the transcript response, and returned `REBASED_HARNESS_OK`
- the 120x40 TUI harness listed only deployed project Runtime and Harness resources
- an undeployed project showed one deployment error with no selectable configured resources
- Runtime selection opened the existing endpoint picker and JSON console
- Harness selection opened the existing chat
- Escape returned from both embedded consoles to the project picker without leaving the TUI process
- Escape returned from the project invoke picker and its error state to the project command menu

## Checklist

- [x] I have added tests that prove the feature works
- [x] I have updated relevant documentation
- [x] I have verified direct Runtime and Harness behavior remains green
- [x] I have run live deployment and TUI verification
- [x] Dependent changes have been merged and published


src/handlers/project/invoke/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+          flags["bearer-token"],
+          io.stdin,
+        );
+        const launchContext: ProjectInvokeLaunchContext = {
+          target: flags.target,
+          sessionId: flags["session-id"],
+          qualifier: flags.qualifier,
+          bearerToken,
+        };
+        await renderInvokeTui(
```
Hweinstock: wouldn't this mean if I do `agentcore project invoke --session-id .. --qualifier .. --target .. --bearer-token` we would open the TUI? 

I thought the convention was that no flags => open TUI. 

I'm open to breaking this, but I do think we need to be consistent since it might confuse a customer (or their agent) if it varies by command what determines if a TUI opens. 

On the flip side, I do think this existing convention helps keep things simple. For example, we don't need the branching logic here to conditionally render the tui in a few places.

src/handlers/project/invoke/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+        );
+        return;
+      }
+
+      const selected = selectResource(project, flags.runtime, flags.harness);
+      if (selected.resourceType === "harness" && flags["bearer-token"] !== undefined) {
+        throw new InputValidationError("--bearer-token is only valid with --runtime");
+      }
+      if (
+        selected.resourceType === "harness" &&
```
Hweinstock: doesn't a similar constraint also apply to runtime (with a higher cap)? https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_InvokeAgentRuntime.html#API_InvokeAgentRuntime_RequestSyntax

unreleated to changes here but I wonder why they are different max length?

src/handlers/project/invoke/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+  throw new InputValidationError(
+    `Project has multiple invokable resources. Specify one:\n` +
+      `  --runtime: ${runtimes.join(", ") || "none"}\n` +
+      `  --harness: ${harnesses.join(", ") || "none"}`,
+  );
+}
+
+function targetContext(ctx: Context, region: string): Context {
+  return ctx.withValue(RegionKey, region);
+}
```
Hweinstock: what's the advantage of making this a function? I feel like it adds an unnecessary layer of indirection.

src/handlers/project/invoke/index.tsx
```diff
@@ -0,0 +1,236 @@
+import z from "zod";
+import { InputValidationError, ResourceNotFoundError } from "../../../errors";
+import type { AppIO } from "../../../io";
+import { withUserCancellation } from "../../../runnable";
+import { argument, createHandler, flag, ProjectKey, type Context } from "../../../router";
+import { JsonRendererKey, renderTuiAt } from "../../../tui";
+import { RuntimeInvokeLaunchContextKey } from "../../runtime/invoke/launchContext";
+import { invokeRuntimeTarget } from "../../runtime/invoke/operation";
```
Hweinstock: this feels strange to me. I'm not sure we want handlers from project depending on imperative handlers. I feel like `core` should be where this shared functionality lives. 

My understanding of the setup was that handlers serve as a "ui shell" over the functionality that handle routing and validation, with `core` implementing the *core* logic. However, I think @AlexanderRichey would know better.

src/handlers/project/invoke/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+        let path = `/agentcore/harness/invoke/${encodeURIComponent(deployed.id)}`;
+        if (flags["session-id"]) path += `/${encodeURIComponent(flags["session-id"])}`;
+        if (flags.qualifier) path += `?qualifier=${encodeURIComponent(flags.qualifier)}`;
+        await renderInvokeTui(path, invokeCtx, core, io);
+        return;
+      }
+
+      if (selected.resourceType === "harness") {
+        const result = await invokeHarnessTurn(
```
Hweinstock: where we do handle user cancellations for harness? I see the wrapper for runtime below.

src/handlers/project/invoke/index.test.tsx
```diff
@@ -0,0 +1,490 @@
...
+
+const HARNESS = { name: "support", path: "app/support" } as const;
+
+function body(...chunks: Uint8Array[]): AsyncIterable<Uint8Array> {
+  return (async function* () {
+    yield* chunks;
+  })();
+}
+
+async function inProject(resources: {
```
Hweinstock, line 55: OOS here, but I noticed we write very similiar functions in all the project tests. I'm gonna put a task on my board to move this in `src/testing/project`.

src/handlers/project/invoke/index.test.tsx
```diff
@@ -0,0 +1,490 @@
...
+  const root = await mkdtemp(join(tmpdir(), "agentcore-project-invoke-"));
+  temporaryDirectories.push(root);
+  await mkdir(join(root, "agentcore"), { recursive: true });
+  const spec = ProjectSpecSchema.parse({
+    name: "orders",
+    version: 1,
+    runtimes: resources.runtimes ?? [],
+    harnesses: resources.harnesses ?? [],
+  });
+  await writeFile(join(root, "agentcore", "agentcore.json"), JSON.stringify(spec));
```
Hweinstock, line 68: Instead of writing the files we need, what if wrote a real project by running `create`? We're already coupling this to create behavior here so may as well make it explicit and get the full coverage.

src/handlers/project/invoke/index.test.tsx
```diff
@@ -0,0 +1,490 @@
...
+  args: string[],
+  resources: { runtimes?: unknown[]; harnesses?: unknown[] },
+  configure?: (core: TestCoreClient) => void,
+  io: TestIO = testIO(),
+  command: "project" | "alias" = "project",
+) {
+  await inProject(resources);
+  const resolved = testBackend();
+  const core = new TestCoreClient({ backends: { CDK: resolved.backend } });
+  core.runtime
```
Hweinstock, line 101: for my own understanding, why can't fixtures be used here? I imagine there are some challenges with streaming, but I thought @jariy17 had a fix for this.

src/handlers/runtime/invoke/operation.ts
```diff
@@ -0,0 +1,14 @@
+import type { CoreOptions } from "../../../core/types";
```
Hweinstock, line 1: isn't this backwards? Shouldn't handlers be importing from core?

README.md
```diff
@@ -116,6 +126,42 @@ Global flags (declared at the root, available on every command):
...
+agentcore project invoke --runtime checkout "Check order 123." --json
+
+# Omit content to open the selected resource's interactive console.
+agentcore project invoke --runtime checkout
+
+# Omit content and selectors to choose from the project's Runtimes and Harnesses.
+agentcore project invoke
+
+# The project-aware shorthand has the same behavior inside a project.
+agentcore invoke
```
Hweinstock: thanks for updating the docs! 

Also, I like this idea of making it project aware, but I wonder if we should revisit this with all project commands together to make sure the behavior is consistent. If the team has already aligned this is where we're heading than I'm good to leave it in.

src/errors/errors.test.tsx
```diff
@@ -60,6 +61,20 @@ describe("AgentCoreCLIError", () => {
     expect(UserCancellationError.resolve(new Error("failed"))).toBeUndefined();
   });
 
+  test("RuntimeAgentResponseError is a visible service failure", () => {
```
Hweinstock: nit: could this be part of the test.each below? Its doing the same thing with one more check for a check for silent, which could be added as a parameter to the tabled test. 

ex. 
```
test.each([
...
[
  "RuntimeAgentResponseError", 
   new RuntimeAgentResponseError("Model access denied"),
   "service", 
   false" // not silent
]
...
```

src/core/project/manager.test.ts
```diff
@@ -484,6 +487,97 @@ describe("FsProjectManager.deploy", () => {
   });
 });
 
+describe("FsProjectManager.resolveDeployedResource", () => {
+  const targets: AwsDeploymentTarget[] = [
```
Hweinstock: q: what behavior are we testing here, that we don't get through the handlers?

src/handlers/runtime/invoke/promptResponse.ts
```diff
@@ -0,0 +1,125 @@
+import { parseAgentEvent } from "../../../core/project/agentEventParser";
```
Hweinstock: isn't this breaking the dependency inversion?

src/core/project/agentEventParser.ts
```diff
@@ -0,0 +1,60 @@
+export type AgentEvent =
```
Hweinstock: I'm kind of confused where this fits in. Isn't this assuming a certain response shape from the customer's runtime?

Hweinstock (commented):
thanks for making this change, I know its a big, but critical one. I honestly didn't get to all of it, but saw a few patterns I wanted to comment on: 
- not all of the invoke runtime logic lives in core, so we end up importing across handlers, and breaking dependency inversion boundary to share this functionality, but it feels like we're fighting the framework/architecture. Might be worth chatting w/ @AlexanderRichey to hear his vision on where this logic should live since he knows it best. 
- the tui being conditionally rendered rather than only on no flags adds a lot of complexity downstream with that launch context being passed around, and I'm wondering if this is required?

src/handlers/project/invoke/index.tsx
aidandaly24: Thought I had removed this. Will fix this.

src/handlers/project/invoke/index.tsx
aidandaly24: Yeah, I had the same concern. I extracted these operations because duplicating the Runtime and Harness invocation flows under project felt incorrect. I agree the dependency direction is unusual, though, and invokeRuntimeTarget in particular may belong behind CoreRuntimeClient.

The complication is that the current Runtime operation composes handler-owned request normalization, while the Harness operation folds service events into the handler-owned transcript model. I don’t think source resolution, terminal output, TUI state, or transcript presentation should move into Core, but there
may be a cleaner split where Core owns the Runtime lookup/normalization/invocation workflow and handlers retain presentation policy.

I’d definitely like to check with @AlexanderRichey before changing this too much so we follow the intended architecture

src/handlers/project/invoke/index.tsx
aidandaly24: Yeah this is legacy from the old CLI. Runtime and Harness invoke already treat session, qualifier, authentication, and target values as TUI launch configuration. Project invoke is consistent with those commands. 

I definitely agree with the concern about the pattern, like you mention. But I also agree with what you say where it keeps things simple. This makes it so that I am able to do things like pass a bearer token or an session id, before i open the TUI. If this wasn't an option we would need to provide a way for the user to do that within the TUI. Which we wanted to avoid for right now.

AlexanderRichey (dismissed):
How does `invoke` work today? Do customers have to specify `runtime` or `harness`?

src/core/project/backends/cdk.test.ts
```diff
@@ -122,6 +127,7 @@ type HarnessOptions = {
   bootstrapError?: Error;
   /** Whether CloudFormation still holds the target's stack. Defaults to present. */
   stackExists?: boolean;
```
Hweinstock: would it always be true that `stackExists` represents whether `stack` is present?

src/core/project/backends/cdk.ts
```diff
@@ -51,6 +73,7 @@ export type CdkBackendConfig = {
   stack?: StackProbe;
   resolveAccount?: AccountResolver;
   loadBootstrapTemplate?: BootstrapTemplateLoader;
+  describeStack?: StackDescriber;
```
Hweinstock, line 78: what exactly is the difference between `StackProbe` and `StackDescriber`?

src/handlers/harness/invoke/operation.ts
```diff
@@ -0,0 +1,48 @@
+import type { CoreOptions } from "../../../core/types";
```
Hweinstock, line 1: is this still breaking the dependency inversion by importing core from handlers? I see Alex approved, so perhaps this is an acceptable exception.

README.md
```diff
@@ -104,6 +104,17 @@ agentcore                          # interactive TUI
 │       ├── get                    # get an evaluator by id (type-agnostic)
 │       ├── list                   # list evaluators (server-side paginated)
 │       └── delete                 # delete an evaluator by id
+├── project                        # manage an AgentCore project
```
Hweinstock: thanks for adding all of this!

Hweinstock (commented):
Thanks for simplifying this one! Code looks good, few small things I noticed when playing around:
- the back doesn't seem to work on the invoke page. When i got to `project > invoke` in the TUI then press `esc`, the TUI exits rather than goes back to the project commands. 
- I'm able to see resources that aren't yet deployed. if I click on them, I see, which isn't intuitive. 
Image

src/core/project/backends/cdk.ts
aidandaly24: `StackProbe` just returns a boolean whereas `StackDescriber` returns the full stack. But you're right that it doesn't make much sense to have both of them. They should just share the existing underlying CFN reader/client.

aidandaly24:
Implemented the latest review feedback:

- `Esc` from the project invoke picker and its error/loading states now routes back to the project command menu.
- The picker resolves deployed resources in one bulk stack read and only lists Runtimes/Harnesses with deployed ID outputs.
- Configured-but-undeployed resources are no longer selectable. An undeployed project now shows one focused deployment error instead.
- Removed the separate `StackProbe`, `stackExists`, and duplicate CloudFormation read path. Teardown and invoke now use the same stack describer/client path.

Verified with the full source suite (`2560 pass`), focused project/CDK suites (`70 pass`), typecheck, lint, formatting, secrets, build, and live TUI runs for both deployed and undeployed projects.

Project invoke undeployed state after the fix

Hweinstock (approved):
LGTM, thanks for addressing comments. 

Noticed two small things, that could be future improvements. 

When going to invoke in the TUI with no deployed resoruces, I see 
```
No deployment targets are configured for project 'helloThere'. Add at least one to /local/home/hkobew/gh-repos/agentcore-cli/helloThere/agentcore/aws-targets.json, for example:
[{ "name": "default", "account": "111122223333", "region": "us-east-1" }]
```

I wonder if this should show something like "no invokable resources" rather than deploy error. 

Also, I don't think this is specific to this PR but the invoke "resolving project" hangs when used outside of a project. 

Image

PR #1660, opened 2026-06-29
fix(templates): retrieve SUMMARIZATION summaries with an actor-scoped namespace (#665)

aidandaly24:
## Description

**#665**

Agents created/extended with `--strategies SUMMARIZATION` write per-session conversation summaries correctly, but the generated session.py retrieves them with a per-session namespace, so a new session never sees prior sessions' summaries. Cross-session summary recall silently does nothing; the feature appears write-only to users.

## Fix

Re-apply PR #1299's one-line change in all three Strands session.py templates: change the SUMMARIZATION retrieval line at line 31 from f"/summaries/{actor_id}/{session_id}" to f"/summaries/{actor_id}". Run `npm run test:update-snapshots` to refresh the 3 affected entries in assets.snapshot.test.ts.snap (lines 2264, 3286, 6554). Update docs/memory.md:95 (manual-wiring snippet) and the strategy table at docs/memory.md:160 to /summaries/{actorId} for SUMMARIZATION. Leave the write-side namespaceTemplate (memory.ts:26) and EPISODIC ({session_id} kept intentionally; the issue is scoped to SUMMARIZATION) unchanged. Add a regression test asserting the generated SUMMARIZATION retrieval namespace contains no {session_id}, so a future squash-merge cannot silently revert it again. No SDK or CDK change needed (SDK prefix-retrieval is correct; CDK owns nothing in this path).

## Related Issue

Closes #665

## Documentation PR

N/A — bug fix; no devguide change required.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe):

## Testing

How have you tested the change? (full validation in PR comments after bug bash)

- [x] I ran `npm run test:unit` and `npm run test:integ`
- [x] I ran `npm run typecheck`
- [x] I ran `npm run lint`
- [x] If I modified `src/assets/`, I ran `npm run test:update-snapshots` and committed the updated snapshots

## Checklist

- [x] I have read the CONTRIBUTING document
- [x] I have added any necessary tests that prove my fix is effective or my feature works
- [ ] I have updated the documentation accordingly
- [x] I have added an appropriate example to the documentation to outline the feature, or no new docs are needed
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the
terms of your choice.

PR #2116, opened 2026-08-26
feat(templates): wire in memory to the runtime templates

Hweinstock:
Dependent on https://github.com/aws/agentcore-cli/pull/2099 (ignore this until that is merged)

### Problem 
Memory is currently hardcoded to none. The old CLI defaulted to a real memory, and allowed `none | longAndShort | short` options. 

### Solutions
- mirror old CLI arguments for memory with the same default. 
- wire that memory through to the template resolver. 
- refactor the fsTreeNode asset resolver to accept transformations and filters for added flexibility. 

### Testing 
created a project with the memory template then deployed, verified the project and my cdk stack included a memory. Also verified the memory code was included in the asset rendering. 

```
 > agentcore project create --name testP --template strands-python
 ...
 > ls testP/app/strands_agent
 README.md  main.py  mcp_client  memory  model  pyproject.toml  skills  uv.lock
 [ shows the memory folder which is conditionally rendered ] 

> cat testP/agentcore/agentcore.json | jq 

{
  "name": "testP",
  "version": 1,
  "managedBy": "CDK",
  "runtimes": [
    {
      "name": "strands_agent",
      "build": "CodeZip",
      "entrypoint": "main.py",
      "codeLocation": "app/strands_agent",
      "runtimeVersion": "PYTHON_3_14",
      "protocol": "HTTP"
    }
  ],
  "memories": [
    {
      "name": "strands_agentMemory",
      "eventExpiryDuration": 30,
      "strategies": [
        {
          "type": "SEMANTIC",
          "namespaceTemplates": [
            "/users/{actorId}/facts"
          ]
        },
        {
          "type": "USER_PREFERENCE",
          "namespaceTemplates": [
            "/users/{actorId}/preferences"
          ]
        },
        {
          "type": "SUMMARIZATION",
          "namespaceTemplates": [
            "/summaries/{actorId}/{sessionId}"
          ]
        },
        {
          "type": "EPISODIC",
          "namespaceTemplates": [
            "/episodes/{actorId}/{sessionId}"
          ],
          "reflectionNamespaceTemplates": [
            "/episodes/{actorId}"
          ]
        }
      ]
    }
  ]
}

[shows the memory config]
 
> agentcore project deploy
...
```
went to console and invoked it, and verified the memory was created. 
- unit tests!


src/handlers/project/shortcuts.ts
```diff
@@ -0,0 +1,89 @@
+import {
+  DEFAULT_EPISODIC_REFLECTION_NAMESPACE_TEMPLATES,
+  DEFAULT_STRATEGY_NAMESPACE_TEMPLATES,
+  type Memory,
+} from "../../projectSchemas/memory";
+import type { ScaffoldRuntimeInput } from "./types";
+
+export const MEMORY_SHORTCUTS = {
+  none: (_runtimeName: string) => undefined,
+  short: (runtimeName: string): Memory => ({
```
notgitika: just wanna point out that this renames shortTerm / longAndShortTerm that exist in CLI main, so existing CLI commands fail. was that intended?

src/handlers/project/shortcuts.ts
```diff
@@ -0,0 +1,89 @@
...
+    name: `${runtimeName}Memory`,
```
notgitika, line 13: I haven't fully confirmed this but just wanted to make sure that when we append "Memory" here we are staying under the character limit of the service API, which is 48 chars? can we make sure the runtimeName then stays 42 chars max? 

once bitten, twice shy 😊

notgitika (commented):
pretty straightforward we are mostly just porting over what already exists on main so I don't wanna comment about improvements in the templates themselves that we can bring in later.

LGTM just 2 questions

src/handlers/project/shortcuts.ts
Hweinstock: oh good catch, I actually thought this is what was on main, let me swap it to match.

src/handlers/project/shortcuts.ts
Hweinstock: good call! These are the types of edge cases that only someone who has implemented this before would know :)

notgitika (dismissed):
LGTM thanks for addressing comments :)

src/assets/templates/strands-http-python/memory/session.py
```diff
@@ -0,0 +1,47 @@
...
+{{#if (includes memoryStrategies "EPISODIC")}}
+        f"/episodes/{actor_id}/{session_id}": RetrievalConfig(top_k=5, relevance_score=0.5),
+{{/if}}
+{{#if (includes memoryStrategies "SUMMARIZATION")}}
+        f"/summaries/{actor_id}": RetrievalConfig(top_k=3, relevance_score=0.5),
+{{/if}}
+    }
+{{/if}}
+
+    return AgentCoreMemorySessionManager(
```
aidandaly24, line 37: This is fine for now since we said we don't want to make to many changes, and I know you said you would eventually like to improve the template. One thing we should change is to use the new AgentCoreMemoryManager and AgentCoreMemoryStore at some point. We should theoretically be using our own best practices.

src/assets/templates/strands-http-python/memory/session.py
```diff
@@ -0,0 +1,47 @@
...
+        f"/users/{actor_id}/facts": RetrievalConfig(top_k=3, relevance_score=0.5),
+{{/if}}
+{{#if (includes memoryStrategies "USER_PREFERENCE")}}
+        f"/users/{actor_id}/preferences": RetrievalConfig(top_k=3, relevance_score=0.5),
+{{/if}}
+{{#if (includes memoryStrategies "EPISODIC")}}
+        f"/episodes/{actor_id}/{session_id}": RetrievalConfig(top_k=5, relevance_score=0.5),
+{{/if}}
+{{#if (includes memoryStrategies "SUMMARIZATION")}}
+        f"/summaries/{actor_id}": RetrievalConfig(top_k=3, relevance_score=0.5),
```
aidandaly24, line 32: I think this retrieval namespace is different from the strategy being created. `MEMORY_SHORTCUTS` configures SUMMARIZATION as `/summaries/{actorId}/{sessionId}`, but the generated runtime queries `/summaries/{actor_id}`. That means summaries written under the configured session namespace will not be retrieved.


I think this should probablyt include `session_id`. I just looked and this mismatch also exists in the old template, but this PR makes that strategy part of the default memory so we may as well just make it right here.

src/core/project/templates/runtime.ts
```diff
@@ -61,11 +61,14 @@ const getTemplateResolvers = (assetSource: AssetSource, templateRenderer: Templa
...
+      { assetSource },
+      {
+        assetDir:
+          input.scaffoldRuntimeInput.build === "Container"
+            ? "templates/hello-world-python-container"
+            : "templates/hello-world-python",
+      },
+      { rootDirName: input.name },
     );
     return { tree, spec: { runtimes: [buildRuntimeSpec(input)] } };
```
aidandaly24, line 75: I think the custom hello-world path can silently ignore `--memory`. Selecting `--framework none` routes here, but even when `scaffoldRuntimeInput.memory` is set, this return only adds the Runtime and does not generate a `memories[]` entry in `agentcore.json` or the memory template files. Do you think we should reject `--memory` for this path for cleanliness in UX?

src/handlers/project/create/index.ts
```diff
@@ -85,21 +94,22 @@ export const createCreateProjectHandler = (config: CreateProjectHandlerConfig) =
...
           ? parseScaffoldRuntimeInput({
-              runtimeName: flags["runtime-name"] ?? flags["name"],
+              runtimeName,
               build: flags["build"],
               language: flags["language"],
               framework: flags["framework"],
               modelProvider: flags["model-provider"],
               apiKey,
-              memory: flags["memory"],
+              memory: MEMORY_SHORTCUTS[flags["memory"] ?? "longAndShortTerm"](runtimeName),
```
aidandaly24: Is memory supposed to be becoming the default here? Don't have a problem with it but like it is a breaking change. In the released CLI, omitting `--memory` does not create a Memory resource, but this defaults to `longAndShortTerm` and provisions four strategies.

src/assets/templates/strands-http-python/memory/session.py
Hweinstock: +1, good callout.

src/handlers/project/create/index.ts
Hweinstock: Yes, that was my intention. Since harness is also now defaulting to give customers a memory, I thought we should align and give them a memory by default as well. The runtime experience *should* be better with a memory attached, so I figure we should give them the best experience by default.

src/assets/templates/strands-http-python/memory/session.py
Hweinstock: I think this is actually intentional. The namespace in the `MEMORY_SHORTCUTS` is where the LTM records get written (session specific path), and then the agent retrieves those records across all sessions by dropping the sessionId on the retrieval path. 

I see a PR from main that fixes this exact behavior: https://github.com/aws/agentcore-cli/pull/1660.

src/handlers/project/create/index.ts
aidandaly24: That makes sense to me

src/assets/templates/strands-http-python/memory/session.py
aidandaly24: I see, I didn't realize this. This is definitely correct I was treating the retrieval namespace like an exact match, thanks!

src/core/project/templates/runtime.ts
Hweinstock: I think this is indirectly fixed in https://github.com/aws/agentcore-cli/pull/2130#top. Let me rebase and verify.

src/core/project/templates/fsTree.ts
```diff
@@ -71,18 +71,30 @@ export class FsTreeNode {
   }
 
   /**
-   * Expands the flat asset listing under assetDir into a nested tree of nodes.
+   * Builds a file tree from assets under `input.assetDir`.
```
jariy17, line 74: now thats a code comment!

src/core/project/templates/fsTree.ts
```diff
@@ -71,18 +71,30 @@ export class FsTreeNode {
...
+   * @param config - Asset source configuration.
+   * @param input - Asset directory to load.
+   * @param options - Optional root name, lazy content transform, and descendant filter. Rejecting a directory omits its subtree.
    */
   static async fromAssetSource(
-    src: AssetSource,
-    assetDir: string,
-    rootDirName?: string,
-    transform?: (content: string) => string,
+    config: { assetSource: AssetSource },
```
jariy17, line 81: I like how we are using object types here.

jariy17 (approved):

PR #2121, opened 2026-08-26
feat: project add/remove policy-engine and policy with natural-language generation

tejaskash:
## What

Adds project-based Policy support per the locked design in the workspace doc AddPolicy.md:

- `agentcore project add policy-engine --name <n> [--description] [--encryption-key-arn] [--tags] [--attach-to-gateways <names...>] [--attach-mode log-only|enforce]`
- `agentcore project add policy --engine <e> --name <n> --statement <inline|file://|-> [--validation-mode] [--enforcement-mode] [--authorization-phase]`
- `agentcore project remove policy --name <n> [--engine <e>]` (engine inferred when unambiguous)
- `agentcore project remove policy-engine --name <n>` (strips gateway `policyEngineConfiguration` references)

Supporting changes:

- Deployed-name rules live in `gatewayResourceName` / `policyEngineResourceName` next to the schemas, mirroring the L3 constructs and honoring the spec's `resourceName` override. The `add gateway` handler uses the shared rule.
- `ProjectSpecSchema` gained a referential check: a gateway `policyEngineConfiguration.policyEngineName` must name a declared engine (design decision 3 relies on schema validation of hand-edits).
- Attach mode defaults to ENFORCE, unknown gateway and engine names fail fast, and the policy duplicate check is project-wide because policy names are account-unique on the service.

**Per review, `--generate` (natural-language policy generation) and its `PolicyClient` core sub-client moved to the stacked follow-up #2127.**

## Not in this PR

- `--generate` and `PolicyClient`: #2127 (stacked on this branch).
- The README documents only the resource-inspection commands and has no `project` section, so no README change (per plan).
- L3 follow-ups tracked separately: mirror `enforcementMode` / `authorizationPhase` in the L3 `PolicySchema`, pass `EnforcementMode` to `CfnPolicy`, add policy construct tests.

## Testing

- Unit and integration: full suite green (`bun test src`), typecheck clean, per-slice and whole-branch simplify reviews applied.
- Live e2e in us-west-2 (account 887863153624), vended pin `@aws/agentcore-cdk@0.1.0-alpha.45`: project created, gateway/engine/policy added, deployed, engine `PolicyE2E_Guardrails` and policy `DenyAll` verified ACTIVE via the control plane, both removes exercised (including engine-inference without `--engine`), gateway reference stripped, stack deleted.

Service-side Cedar validation learnings recorded in the design doc: wildcard resources are rejected at CreatePolicy (constrain to `resource is AgentCore::Gateway`), and blanket forbids fail `FAIL_ON_ANY_FINDINGS` as Overly Restrictive, so deny-all policies need `--validation-mode ignore-all-findings`.


src/core/policy.test.ts
```diff
@@ -0,0 +1,128 @@
+import { describe, expect, test } from "bun:test";
+import {
```
jariy17: Golden test only in handlers

tejaskash:
Addressed all three harness review findings in fedf4c1b:

1. `add policy-engine` now fails fast when the deployed service name (`policyEngineResourceName`) exceeds the 48-char service limit, matching the gateway handler.
2. The `--generate` path pre-checks project-wide policy-name uniqueness before starting the multi-minute generation, so a duplicate name no longer surfaces only after the wait.
3. Exhausting the generation poll loop while still GENERATING now reports that the CLI's wait window elapsed and the generation may still complete, instead of `did not complete: GENERATING`.

Each fix landed with a test (32 tests across the three touched files, full suite 2146 green, typecheck clean).

tejaskash:
Rebased onto the latest `refactor` (now includes #2105/#2112 deployed-state work and the restructured `FsProjectManager.addResource`). The policy arms were adapted to the new direct-mutation idiom during the rebase: the `specPatch` mechanism is gone (gateway attachment now mutates `projectSpec.agentCoreGateways` in place inside the same validated write), and policy/policy-engine inserts go through the base's `parseResource` helper. Full suite green (2153 = base 2113 plus this branch's 40 tests), typecheck clean.

Note for a follow-up outside this PR: `--generate` resolves deployed engine/gateway by service-name matching because no deployed state existed when it was designed; with `deployed-state.json` now on the branch it could read ids from there instead.

src/core/policy.tsx
```diff
@@ -0,0 +1,137 @@
+import { setTimeout as sleep } from "node:timers/promises";
```
jariy17: Can we split up this PR into imperative based and project based? Also, I noticed that we are creating a seperate CorePolicyClient just for the projects add policy command.

jariy17 (approved):

jariy17 (changes requested):
Check review comments

src/core/policy.tsx
tejaskash: Everything in this PR is project-based: the two `project add` commands, the two `project remove` forms, and the spec/schema changes. There is no imperative `agentcore policy ...` surface here, so an imperative/project split would leave one of the PRs empty.

The piece that reads as imperative is `PolicyClient`, which makes control-plane calls, but it exists only to serve `project add policy --generate` (resolve the deployed engine and gateway, run `StartPolicyGeneration`, poll, read assets). If you'd like the PR smaller, the clean split is by capability rather than by surface: spec-editing commands (`--statement`, removes) in one PR and `--generate` plus `PolicyClient` in a second. That slice is one commit and extracts cleanly, so say the word and I'll split it.

On the dedicated `CorePolicyClient`: it follows the repo's consumer-owned-interface pattern; gateway, memory, and eval each pair a handler-side `Core*Client` interface with one core sub-client, and none of the existing sub-clients owns policy-generation calls. When an imperative `agentcore policy` surface lands later, it will consume this same client rather than needing a new one. A likely follow-up will also shrink it: now that `deployed-state.json` persistence is on `refactor` (#2105), the two list-and-match loops plus the `GetGateway` call can collapse into a state-file read.

src/core/policy.test.ts
tejaskash: The golden/flow coverage does live only in the handler tests: they drive the real root handler against a real temp project and assert on agentcore.json, with `TestPolicyClient` standing in for this client. `policy.test.ts` is the other altitude: a unit test faking the SDK at `.send()`, covering behavior the handler tests cannot reach through the fake, namely nextToken pagination, poll exhaustion, `GENERATE_FAILED` reasons, Dogwood `policy` definition members, and findings-only assets (the last two were caught live during the e2e). Same split as `gateway.tsx` / `eval.tsx` and their tests. If you meant something else by this, happy to adjust.

src/core/policy.tsx
tejaskash: Done: `--generate` and `PolicyClient` are extracted to #2127, stacked on this branch. This PR is now purely the spec-editing surface (add/remove for policy-engine and policy with `--statement`), no core client. Extraction was a pure move: the stacked branch tree is byte-identical to this branch's pre-split tip.

src/core/policy.test.ts
jariy17: Just follow what batch-evaluation does, it puts all the edge cases through a separate handler test file using regular mocking techniques. (batch-evaluation.test.ts)

src/projectSchemas/policy.ts
```diff
@@ -49,3 +49,10 @@ export const PolicyEngineSchema = z.object({
     ),
 });
 export type PolicyEngine = z.infer<typeof PolicyEngineSchema>;
+
+/**
+ The deployed service name of a policy engine; mirrors the L3 AgentCorePolicyEngine construct's rule.
+**/
+export function policyEngineResourceName(projectName: string, engineName: string): string {
```
jariy17: shift to handler since its only used there.

tejaskash: Done in f6586304: `policyEngineResourceName` now lives in the policy-engine handler. The generate follow-up (#2127) imports it from there when it gains a second consumer.

src/core/policy.test.ts
tejaskash: Done on #2127 in ef9f23ca, following the batch-evaluation shape: `src/core/policy.test.ts` is deleted and the edge cases (deployed-resource resolution, poll exhaustion, GENERATE_FAILED reasons, Dogwood definition members, findings-only assets) now run through the real root handler in `src/handlers/project/add/policy/generate.test.ts`, with the control client mocked at `.send()`.

src/projectSchemas/gateway.ts
```diff
@@ -433,6 +433,17 @@ export type GatewayPolicyEngineConfiguration = z.infer<
 >;
 export const GatewayProtocolTypeSchema = z.enum(["MCP", "None"]);
 export type GatewayProtocolType = z.infer<typeof GatewayProtocolTypeSchema>;
+
+/**
+ The deployed service name of a gateway; mirrors the L3 Gateway construct's rule.
+**/
+export function gatewayResourceName(
```
jariy17: Why are you adding helper functions here if its only used by the handler.

tejaskash: Done in 9065f01b: `gatewayResourceName` moved into the `add gateway` handler, same treatment as `policyEngineResourceName`. Both schema modules are back to pure shape definitions.

jariy17 (approved):

PR #2123, opened 2026-08-27
feat: create credential providers before synthesizing a deploy

notgitika:
Creates or updates a project's credential providers **before** synthesis, so the synthesized CDK app can read their ARNs out of `deployed-state.json`. Without this, deploying a project that declares any credential fails inside `cdk synth`.

## What it does

- Between the account preflight and synth, `CdkBackend.deploy` provisions each declared credential provider and records the ARNs via `updateTargetState`. Local prerequisites are checked first, so a setup error never mutates AWS.
- Created when absent, **updated when present**, so editing a secret and redeploying pushes the new value (as `main` does). A credential with no secret the CLI can see leaves an existing provider untouched instead of failing.
- Secrets come from `.env.local` (`AGENTCORE_CREDENTIAL_<NAME>`) or a Secrets Manager `secretRef`; **process-environment variables override the file**, so CI needs no secrets on disk.
- **Payment credentials are provisioned** (CoinbaseCDP, StripePrivy) and deleted on teardown. Deploy previously refused them, so `add credentials payment` could build a project that couldn't deploy.
- A failure part-way **deletes the providers that run created** — never a pre-existing one — before rethrowing.
- Runs through the existing `core.identity` client instead of a second SDK client; the target's credentials travel via `CoreOptions`.

## Notes

- **Behaviour change from this PR's first revision:** a redeploy now updates an existing provider, so it overwrites a secret rotated outside the CLI. The old "never update" behaviour meant an edited `.env.local` silently had no effect.
- `CoreClient`'s client cache is now keyed by credential identity — credentials are a function, which `JSON.stringify` drops, so different credentials in one region shared a cached client.
- Credential names ending in a field suffix (`_CLIENT_ID`, `_APP_SECRET`, …) are refused at `add` time; their variable would shadow another credential's field. `main` guarded this; the rewrite had lost it.
- Follow-up in `@aws/agentcore-cdk`: its API-key Gateway path doesn't grant `GetSecretValue` on an external secret ARN, so an API-key `secretRef` deploys and then fails at retrieval. The CLI already records the ARN.

## Tested

- Unit: create/update/reuse per credential type, environment override, missing-secret errors, legacy `_CLIENT_ID` fallback, payment vendors, teardown deletion, rollback.
- e2e (us-west-2), first revision: `add credentials api-key` → `deploy` created the provider before synth, wrote it to `deployed-state.json`, then merged `stackArn` into the same entry.


src/handlers/project/add/credentials/oauth/index.ts
```diff
@@ -95,7 +100,7 @@ export const createAddOauthCredentialHandler = (config: AddProjectResourceConfig
         ? []
         : [
             {
-              key: credentialEnvVarName(flags.name, "_CLIENT_SECRET"),
+              key: credentialEnvVarName(flags.name, CLIENT_SECRET_SUFFIX),
```
aidandaly24: I think collision detection needs to compare the final per-type environment keys, including this suffix. I added an API key named `foo_client_secret` and OAuth credential named `foo`. Both additions succeeded, and provisioning sent the API-key value as both the API key and OAuth `clientSecret` because both resolve to `AGENTCORE_CREDENTIAL_FOO_CLIENT_SECRET`. Could the shared validation derive the actual secret key for each credential type and reject collisions before writing or touching AWS?

src/core/project/backends/cdk.ts
```diff
@@ -106,6 +110,18 @@ export class CdkBackend implements ProjectBackend {
     // with the new stack ARN unrecorded because the post-deploy write can't parse it.
     await readDeployedState(this.json, project.rootPath);
 
+    // Credential providers aren't stack resources; the synthesized app reads their
+    // ARNs from deployed-state.json, so they must exist and be recorded before synth.
+    const provisioned = yield* this.provisionCredentials(project, {
```
aidandaly24, line 211: Could the local CDK prerequisite checks stay ahead of credential provisioning? I ran this with `agentcore/cdk/node_modules` missing: the provisioner ran and `deployed-state.json` was written, then deploy failed with the `npm install` guidance. Synthesis still has to happen after provisioning, but checking `npm` and dependencies separately first would avoid mutating AWS for a local setup error. Probably just good to have like for ux but not a blocker.

src/core/project/backends/cdk/credentials.ts
```diff
@@ -0,0 +1,330 @@
...
+
+    const env = await new EnvLocalFile(project.rootPath).read();
+    const client = await createClient(region, credentials);
+
+    const provisioned: DeployedCredentials = {};
+    for (const credential of declared) {
+      // Provider names are account-global, so a name already taken by another
+      // project in this account is adopted rather than recreated.
+      yield { message: `Preparing credential provider '${credential.name}'` };
+      provisioned[credential.name] = await provisionOne(client, credential, env, project.rootPath);
```
aidandaly24: I think this can leave partial AWS state when provisioning fails midway.

For example:

1. The first credential provider is created successfully.
2. Creating the second provider fails.
3. `provisionCredentials()` throws, so `updateTargetState()` is never reached.

The first provider now exists in AWS but is not recorded locally. Retrying with the same spec recovers it, but abandoning the deploy or removing that credential leaves the provider orphaned.

Could we validate all credential inputs before creating anything and roll back providers created by this invocation if a later creation fails?

src/core/project/backends/cdk/credentials.ts
```diff
@@ -0,0 +1,330 @@
...
+      : guidedCustomConfig(credential, secret),
+  });
+}
+
+/**
+ * Injects the secret into a complete, spec-supplied vendor config. The spec
+ * keeps provider configs secret-free, so the one vendor key it carries is the
+ * only place the secret can go.
+ */
+function vendorConfigWithSecret(
```
aidandaly24, line 585: I really like the breakdown of functionality in this file.

aidandaly24 (changes requested):
Mostly looks good to me just a few comments

src/core/project/backends/cdk/credentials.ts
```diff
@@ -0,0 +1,330 @@
...
+          clientSecretArn: response.apiKeySecretArn.secretArn,
+        }),
+      };
+    },
+    async createApiKeyProvider({ name, apiKey, secretRef }) {
+      const response = await client.send(
+        new CreateApiKeyCredentialProviderCommand({
+          name,
+          ...(apiKey !== undefined && { apiKey }),
+          ...(secretRef && { apiKeySecretConfig: secretRef, apiKeySecretSource: "EXTERNAL" }),
```
tejaskash: When this points to an external secret, the generated Gateway role cannot actually read it. We save the returned ARN as `clientSecretArn`, but `@aws/agentcore-cdk` 0.1.0-alpha.45 only uses that ARN for OAuth targets. Its API-key path grants `secretsmanager:GetSecretValue` only for the managed `bedrock-agentcore-identity!*` secrets. That means a Gateway target using an API-key credential with `secretRef` can deploy successfully and then fail when it tries to retrieve the key. Could we grant the recorded external ARN for API-key targets as well, either here or by updating the pinned construct?

aidandaly24: I think the change is on the refactor branch tho right @notgitika ?

src/handlers/project/add/credentials/oauth/index.ts
notgitika: good catch. I'll make the collision check derive each credential's actual secret key by type (api-key -> base, oauth -> base + _CLIENT_SECRET) and reject on the final key, before writing or touching AWS

src/core/project/backends/cdk.ts
notgitika: Agreed, a local setup error shouldn't mutate AWS. I'll hoist the npm/node_modules prerequisite check ahead of credential provisioning so we fail fast before creating anything. Synth still runs after provisioning, but the local-only checks don't need to.

src/core/project/backends/cdk/credentials.ts
notgitika: right, there's an orphan window. I'll validate all credential inputs up front (resolve every .env.local/secretRef before creating anything), which removes the most likely mid-loop failure before any AWS call. It's also self-healing on retry via reuse-by-name. I'm inclined to skip automatic rollback of already-created providers for now (extra delete calls + failure modes for an edge case retry already recovers), let me know if you'd rather I add it.

src/core/project/backends/cdk/credentials.ts
notgitika: I checked the l3 cdk constructs repo, it's not fixed on its refactor branch either. `Gateway.ts` there (same as main/alpha.45) grants the recorded external secret ARN for OAuth targets, but the API-key path grants GetSecretValue only on bedrock-agentcore-identity!* — so an API-key secretRef isn't covered. 

That's a construct fix (broaden the api-key Gateway grant to include the recorded external ARN, like the OAuth path), not something the CLI can do, it already records the ARN. 

Also worth noting the credential-provider CDK construct on that refactor branch is the "credentials via CDK" approach we're not going ahead with. I can file a follow-up on agentcore-l3-cdk-constructs and keep it out of scope for this PR

src/core/project/backends/cdk/credentials.ts
```diff
@@ -0,0 +1,356 @@
...
+      `Credential '${credential.name}' needs either a discoveryUrl or a providerConfig ` +
+        `to create its OAuth2 provider.`,
+    );
+  }
+  // `scopes` is deliberately not forwarded: provider creation has no scopes
+  // field, and the spec's scopes are consumed where the credential is used.
+  return {
+    customOauth2ProviderConfig: {
+      oauthDiscovery: { discoveryUrl: credential.discoveryUrl },
+      ...(credential.clientId !== undefined && { clientId: credential.clientId }),
```
aidandaly24: I think older OAuth projects need a compatibility fallback here. `v0.28.0` stored the client ID only in `AGENTCORE_CREDENTIAL_<NAME>_CLIENT_ID`, not in `agentcore.json`. If the provider needs to be recreated in another target/account or after deletion, this sends the discovery URL and secret without the client ID. Would it make sense to prefer `credential.clientId` and fall back to the legacy environment variable, with an upgrade test?

notgitika: good find, fixed!

src/core/project/backends/cdk/credentials.ts
```diff
@@ -0,0 +1,356 @@
...
+): Promise<Provision> {
+  // Provider names are account-global, so one already in this account is reused.
+  const existing = await client.getApiKeyProvider(credential.name);
+  if (existing) return { reuse: existing };
+
+  const input: ApiKeyProviderInput = credential.secretRef
+    ? { name: credential.name, secretRef: credential.secretRef }
+    : {
+        name: credential.name,
+        apiKey: requireEnvSecret(credential.name, env, rootPath, "secretRef"),
```
AlexanderRichey: I think we already have all the code we need to do this in the existing core package. Let's reuse that. We can discuss strategies for making that code reusable if it's not clear.

AlexanderRichey (dismissed):

aidandaly24 (dismissed):

notgitika:
## e2e run — account `998846730471`, us-west-2

Ran the full flow against a real account after the merge with `refactor`. Every scenario below passed; the account is clean afterwards (no providers, no stacks).

| # | Scenario | Result |
|---|---|---|
| 1 | First deploy creates the provider before synth | `Preparing credential provider 'e2ekey2'` precedes `Synthesizing CloudFormation templates`; provider ARN + secret ARN written to `deployed-state.json`, then `stackArn` merged into the same target entry |
| 2 | Rotate the secret in `.env.local`, redeploy | Provider **updated**: new `AWSCURRENT` secret version at 22:59:05 against a create at 22:57:22, and `GetSecretValue` returned the rotated value |
| 3 | Variable in the process environment | `.env.local` held one value, the environment another → the environment's value is what landed in AWS |
| 4 | No secret in the file or the environment | Deploy succeeded, no new secret version, stored value unchanged — the existing provider is left alone rather than failing the deploy |
| 5 | Second credential fails after the first is created | `Removing credential provider 'rbkeya' this deploy created`; `rbkeya` gone from AWS, pre-existing `e2ekey2` untouched, `deployed-state.json` unchanged, exit code 1 with the original service error |
| 6 | Payment credential (CoinbaseCDP) | `paymentcredentialprovider/e2epay` created and its ARN recorded. With two of three variables unset, one error named both: `missing the values its payment provider needs: ..._API_KEY_SECRET, ..._WALLET_SECRET` |
| 7 | Payment update, then teardown | Rotated `apiKeyId` reached AWS on redeploy. Teardown removed the stack, deleted `e2epay`, **kept** `e2ekey2` (account-global, may be shared), and emptied `targets` |

Two notes for anyone re-running this:

- **The service validates vendor key formats.** Placeholder payment secrets are rejected (`Invalid apiKeySecret format: Expected base64-encoded Ed25519 private key`, then `Invalid walletSecret format: Expected base64-encoded EC P-256 private key`). I used structurally valid throwaway keys, so create/update/delete and the request shape are verified — but not against real Coinbase credentials. Those service messages surface as-is rather than CLI-framed.
- **`lastUpdatedTime` on an API-key provider does not move on update**, so `get-api-key-credential-provider` alone can't confirm a rotation; Secrets Manager version history can.

Unrelated papercut noticed on the way: `project remove harness <name>` fails with "too many arguments" — it wants `--name <name>`.

src/handlers/project/add/credentials/shared.ts
```diff
@@ -6,6 +6,7 @@ import type { AddResourceInput } from "../../types";
 import {
```
Hweinstock, line 6: OOS here, but I think we should rename this file to avoid it becoming a dumping ground.

src/handlers/project/add/credentials/shared.ts
```diff
@@ -50,6 +51,21 @@ export async function addCredentialToProject(
     );
   }
 
+  // The collision check above only compares the fields credentials write today. A name
+  // ending in a field suffix would also shadow a field the CLI no longer writes but
+  // still reads (an OAuth client id in a pre-0.29 project) or one a later credential
```
Hweinstock: can we link to more context here? I'm not sure what a pre-0.29 project means

src/projectSchemas/credential.ts
```diff
@@ -114,9 +114,46 @@ export const CredentialSchema = z.discriminatedUnion("authorizerType", [
 ]);
 export type Credential = z.infer<typeof CredentialSchema>;
 
+/** The prefix every variable carrying credential material shares. */
+export const CREDENTIAL_ENV_PREFIX = "AGENTCORE_CREDENTIAL_";
```
Hweinstock, line 118: i'm not super familiar with these schemas, but I'm suprised to find helper functions in projectSchemas. I would have expected it to be strictly zod schemas or their related types/constants.

src/core/types.tsx
```diff
@@ -22,11 +29,12 @@ export interface CoreOptions {
 export interface ClientConfig {
```
Hweinstock, line 33: Should these two types be merged? The difference is less clear to me now.

src/core/types.tsx
```diff
@@ -7,13 +7,20 @@ import type {
   CloudFormationClientConfig,
 } from "@aws-sdk/client-cloudformation";
 
+// AwsCredentials is an explicit credential source for a call: either resolved
+// credentials or a provider that resolves them. Callers that rely on the SDK's own
+// default credential chain leave it unset.
+export type AwsCredentials = NonNullable<CloudFormationClientConfig["credentials"]>;
```
Hweinstock: q: is the credentials field in CloudFormationClientConfig valid for all aws sdk clients?

src/core/index.tsx
```diff
@@ -109,6 +109,9 @@ export class CoreClient implements AwsClients {
     this.projectManager = new FsProjectManager({
       logger: this.logger.child({ module: "projectManager" }),
       createCloudFormationClient: config.createCloudFormationClient,
+      // A project deploy provisions credential providers through the same Identity
```
Hweinstock: i feel like this comment is expressed by the code.

src/core/index.tsx
```diff
@@ -164,6 +167,27 @@ export class CoreClient implements AwsClients {
...
+//
+// `credentials` is a provider function or an object of resolved credentials, so it
+// cannot be serialized — JSON.stringify drops functions silently, which would map two
+// callers with different credentials in the same region onto one cached client. It is
+// keyed by identity instead.
 function cacheKey(config: ClientConfig): string {
-  return JSON.stringify(config);
+  const { credentials, ...serializable } = config;
+  const suffix = credentials ? `|credentials:${credentialsId(credentials)}` : "";
+  return JSON.stringify(serializable) + suffix;
```
Hweinstock: does this handle the case where the fields in the input config are ordered differently?

src/core/index.tsx
```diff
@@ -164,6 +167,27 @@ export class CoreClient implements AwsClients {
...
+}
+
+const credentialsIds = new WeakMap<object, number>();
+let nextCredentialsId = 0;
+
+// credentialsId assigns each credential source a stable id for the lifetime of the
+// object, so the same source reuses its client and a different one gets its own.
+function credentialsId(credentials: NonNullable<ClientConfig["credentials"]>): number {
```
Hweinstock: I'm not really following why we need this id?

src/core/identity.tsx
```diff
@@ -19,19 +23,33 @@ import {
...
 import { toClientConfig } from "./utils";
 
+// createIdentityClient builds an IdentityClient that owns its control-plane client,
+// for callers constructed outside CoreClient (which hands out its cached ones).
+export const createIdentityClient = (): IdentityClient =>
+  new IdentityClient({ control: createControlClient });
+
 export class IdentityClient implements CoreIdentityClient {
-  constructor(private readonly clients: AwsClients) {}
+  // Only the control plane is used, so the dependency is narrowed to it: CoreClient
```
Hweinstock: i feel like the code expresses this.

src/core/project/manager.tsx
```diff
@@ -121,6 +124,7 @@ export class FsProjectManager implements ProjectManager {
       CDK: new CdkBackend({
         logger: config.logger,
         createCloudFormationClient: config.createCloudFormationClient,
+        identity: config.identity ?? createIdentityClient(),
```
Hweinstock: whats the advantage to making this optional? It looks like we always pass it and I wonder if we it would simplify some of the core changes.

src/core/project/backends/cdk/credentials.ts
```diff
@@ -0,0 +1,579 @@
...
+    for (const { name } of payments) {
+      yield { message: `Removing credential provider '${name}'` };
+      try {
+        await identity.deletePaymentCredentialProvider(name, options);
+      } catch (error) {
+        if (error instanceof ResourceNotFoundException) continue;
+        yield {
+          message:
+            `Could not remove credential provider '${name}': ${(error as Error).message}. ` +
+            `Delete it with 'aws bedrock-agentcore-control delete-payment-credential-provider'.`,
```
Hweinstock, line 140: isn't there a way to remove these via the resource based commands in our cli? Or did we only do readonly?

src/core/project/backends/cdk/credentials.test.ts
```diff
@@ -0,0 +1,711 @@
+import { afterEach, describe, expect, test } from "bun:test";
```
Hweinstock, line 1: this testing setup feels extremely complex and coupled to the underlying implementation. Is there a way we can decouple it, and simplify?

Hweinstock (commented):
I wasn't able to full follow everything done here, so I left some random comments about where I was confused. 

I was kind of envisioning this as a single function executed on deploy to reconcile the local env with the deployed environment, but it feels like there's a lot more going on. Is there a way to descope this into something simpler as a start?

src/handlers/project/add/credentials/shared.ts
notgitika: Agreed. I'll leave it in this PR to keep the diff reviewable and rename in a follow-up, probably addFlow.ts or credentialInput.ts, since what's actually in here is the shared add-flow + flag parsing

src/handlers/project/add/credentials/shared.ts
notgitika: ah I meant pre v0.29.  fair, I'll reword to say what actually changed: `_CLIENT_ID` used to be written into `.env.local`, an OAuth client id now lives in agentcore.json, but deploy still reads the old variable so projects created before that move keep working. That's the field the collision check can't see, which is why the name is refused outright

src/projectSchemas/credential.ts
notgitika: ah, yes I found that the directory already does this, `runtime.ts`, `harness.ts`, etc. all export helpers alongside their schemas. these three derive their output from this schema's own field names (`credentialEnvironmentVariableNames` switches on `authorizerType`), so splitting them out would mean a second file that imports the schema and has to be kept in sync with it. 

happy to move them if you'd rather the directory be strictly schemas, but it'd be a broader cleanup than this PR intends

src/core/types.tsx
notgitika: thhey're deliberately different shapes. ClientConfig is spread straight into the SDK constructor, so its fields have to be the SDK's names: endpoint. 

CoreOptions is the handler-facing shape and uses endpointUrl, which is what the flag and context call it. toClientConfig is the translation. I'll make that explicit in the comment, since it's not visible from the type alone.

src/core/types.tsx
notgitika: yes!  every v3 client takes the same AwsCredentialIdentity | AwsCredentialIdentityProvider. but you're right that deriving it from the CFN client looks arbitrary. 

I can look into importing the type from @smithy/types directly

src/core/project/backends/cdk/credentials.ts
notgitika: agreed and I fixed it n the next commit. provisioning now goes through the existing core.identity client, so there's no second SDK client. 

2 things had to change in core to make it reusable that way:
1. CoreOptions/ClientConfig gained an optional credentials, since a deploy has to run against its target's creds instead of the default chain. 
2. CoreClient's client cache was keyed by JSON.stringify(config), which silently drops a credential provider function so two targets in the same region shared one cached client. now keyed by credential identity

src/core/index.tsx
notgitika: agreed lemme drop it

src/core/index.tsx
notgitika: no. today every config comes out of toClientConfig so the order is fixed, but it is latent. 

I'll build the key from named fields (region, endpoint) instead of stringifying the whole object.

src/core/index.tsx
notgitika: The bug it fixes: credentials is either a provider function or an object of resolved credentials, and JSON.stringify silently drops functions, so two deploys against different accounts in the same region got the same cache key and shared one client, i.e. the second one ran with the first one's credentials. 

It needs to be keyed by identity, not by value. but i can see how this is more complicated.

let me try switching to a 2 level cache: WeakMap<credentials, Map<string, Client>>, plus a plain Map for the no-credentials case. Same semantics, no global id, no string suffix

src/core/identity.tsx
notgitika: I feel like the Pick is visible, but why it's a Pick rather than the whole AwsClients isn't. I'll shorten it

src/core/project/manager.tsx
notgitika: ah I see what you mean. making it required and updating the call sites.

src/core/project/backends/cdk/credentials.ts
notgitika: Not for payment providers, there is CRUDL for api and oauth only

src/core/project/backends/cdk/credentials.test.ts
notgitika: I'll make the fake a small stateful in-memory Identity that mutates a Record<name, provider>, and assert the resulting store plus the returned DeployedCredentials instead of the call log. I'd keep call-order assertions in the few places where the order is the contract,  get-before-create, and rollback deleting only what the run created. What do you think?

src/core/project/backends/cdk.ts
```diff
@@ -258,6 +293,11 @@ export class CdkBackend implements ProjectBackend {
 
     yield { message: `Removing stack ${artifact.stackName}` };
     await this.cdk({ kind: "destroy", stackArtifactId: artifact.id }, options);
+    // After the stack, since a resource in it may still be using the provider.
+    yield* this.removePaymentCredentials(project, {
```
aidandaly24, line 333: I think teardown needs to use the payment providers recorded for the deployed target, not the current `project.spec.credentials`. `project remove all` writes `credentials: []` before this deploy, so the remover gets no names; `removeTargetState` then drops the only tracking while the providers remain in AWS. Could we persist the payment-provider names/type in deployed state and delete from that snapshot before removing the target entry, with a `remove all -> deploy` regression test?

src/core/project/backends/cdk/credentials.ts
```diff
@@ -0,0 +1,579 @@
...
+  rootPath: string,
+): Promise<Provision> {
+  const { name } = credential;
+  const existing = await undefinedWhenAbsent(() =>
+    identity.getPaymentCredentialProvider(name, options),
+  );
+
+  const fields = paymentFields(credential, env);
+  if ("missing" in fields) {
+    if (existing) return { reuse: paymentProvision(name, existing) };
```
aidandaly24, line 417: Should we verify `existing.credentialProviderVendor` before reusing this ARN? I used a `CoinbaseCDP` credential with an existing same-name `StripePrivy` provider and no local values; deploy accepted it and recorded the Stripe ARN as the Coinbase credential. The OAuth reuse branch above has the same issue, so maybe a shared mismatch check and tests for both provider types would help.

src/core/project/backends/cdk/credentials.ts
```diff
@@ -0,0 +1,579 @@
...
+    // rather than leaving one behind that nothing records.
+    const created: Credential[] = [];
+    try {
+      for (const { credential, provision } of plans) {
+        yield { message: `Preparing credential provider '${credential.name}'` };
+        if ("reuse" in provision) {
+          provisioned[credential.name] = provision.reuse;
+          continue;
+        }
+        provisioned[credential.name] = await provision.write();
```
aidandaly24: I think one rollback window remains here. `write()` can successfully create the provider and then throw while `requireArn()` validates a malformed response, but the credential is not added to `created` until the next line. I ran that response and got `get -> create` with no delete; could the create path record its side effect immediately after the service call returns and before response validation?

src/core/project/backends/cdk/credentials.ts
```diff
@@ -0,0 +1,579 @@
...
+ * Provisions the credential providers a project declares, before synthesis: the
+ * synthesized app reads their ARNs from `deployed-state.json`, so a project with
+ * credentials can't synthesize until they exist.
+ *
+ * A provider is created when absent and updated when present, so editing a secret
+ * in `.env.local` and redeploying pushes the new value. A provider whose secret the
+ * CLI cannot see — nothing in `.env.local` and no external reference — is left
+ * exactly as it is rather than failing the deploy.
+ */
+export function createCredentialProvisioner(
```
aidandaly24, line 157: I may be missing an intentional compatibility change, but released `v0.28.0` always configured the token-vault CMK before API-key provider creation (`GetTokenVault`, `CreateKey`, and `SetTokenVaultCMK` via `enableKmsEncryption: true`) and persisted `identityKmsKeyArn`. This path does none of that, so new refactor deployments no longer proactively get the CLI-managed CMK behavior. Should we preserve the released behavior here, or explicitly call out that encryption-default change?

src/core/project/backends/cdk.ts
notgitika: yes, fixed

src/core/project/backends/cdk/credentials.ts
notgitika: yes, fixed

src/core/project/backends/cdk/credentials.ts
notgitika: done!

src/core/project/backends/cdk/credentials.ts
notgitika: deferred to a follow up

AlexanderRichey:
Stale

aidandaly24 (approved):
Thanks for making these fixes, it's a great PR!

PR #2127, opened 2026-08-27
feat: generate Cedar policies from natural language in project add policy

tejaskash:
## What

Extracts `project add policy --generate` from #2121 per review, as a stacked PR (base: `feat/project-add-policy`). This PR re-adds exactly what #2121 dropped:

- `--generate <text>` and `--gateway <name>` on `project add policy`: generate the Cedar statement from a natural-language description against the deployed engine and gateway, stream progress, print the generated Cedar and every generation finding, then write the policy into the spec.
- `PolicyClient` core sub-client (`src/core/policy.tsx`): resolves the deployed engine and gateway by exact service name (via the shared `policyEngineResourceName` / `gatewayResourceName` rules that stay in #2121), starts the generation, polls with an injectable delay, surfaces `statusReasons` and findings on failure, accepts both `cedar` and Dogwood `policy` definition members.
- `CorePolicyClient` consumer-owned interface, `TestPolicyClient`, and the wiring through `Core` / `CoreClient` / handler configs.
- Handler-side fail-fast before the multi-minute generation call: engine exists, policy name unique, gateway resolved from the spec (single gateway inferred, multiple demand `--gateway`).

## Testing

- Unit: `PolicyClient` tested with a fake at the SDK `.send()` boundary (pagination, poll exhaustion, `GENERATE_FAILED` reasons, Dogwood members, findings-only assets); handler flows tested through the real root handler with `TestPolicyClient`.
- Live e2e in us-west-2 (run while this code was part of #2121): `--generate "forbid IAM principals from calling any tool on this gateway" --gateway tools` against a deployed stack produced real Cedar with a `DENY_ALL` finding printed; the generated policy deployed to ACTIVE and was removed cleanly.

## Follow-up

`deployed-state.json` (#2105) can replace the list-and-match resolution with a state-file read; kept out of this PR to keep the extraction a pure move.


tejaskash:
Note on the failing `e2e` check: it is `.github/workflows/e2e-tests.yml` running via `pull_request_target` from the default branch, which assumes main's npm layout (`package-lock.json`, `test-artifacts:*` scripts) that the bun-based refactor tree doesn't have. It misfires here only because this PR's base is a feature branch (stacked on #2121), so the refactor-base gating doesn't apply. The real e2e for this tree is the AWS CodeBuild `agentcore-e2e` check, which passes on #2121. Once #2121 merges and this PR retargets `refactor`, the proper gating applies.

tejaskash:
Same applies to the failing `pr-tarball` check: a default-branch `pull_request_target` workflow that expects main's npm lockfile. Both misfires resolve when this PR retargets `refactor` after #2121 merges; the refactor-tree CI (build, unit-test, check, CodeQL, CodeBuild e2e) is what gates this code.

src/core/index.tsx
```diff
@@ -7,6 +7,7 @@ import { GatewayClient } from "./gateway";
 import { HarnessClient } from "./harness";
```
jariy17, line 7: PAUSE FOR STANDUP ASK: 
* This is the first time where we have an add project command calling a core client. How should we introduce this? Also, don't you want to confirm the policy before adding to agentcore.json

src/handlers/project/add/policy/index.ts
```diff
@@ -55,16 +64,81 @@ export const createAddPolicyHandler = (config: AddProjectResourceConfig) =>
...
+          throw new InputValidationError(
+            gateways.length === 0
+              ? "--generate needs a deployed gateway; add one to this project and deploy it first"
+              : `this project declares multiple gateways: ${gateways
+                  .map((candidate) => candidate.name)
+                  .join(", ")}; pass --gateway to choose one`,
+          );
+        }
+
+        const generator = config.policy.generatePolicy(
```
jariy17, line 120: Standup discussion: 
* We should just make this imperative command for now.  The customers will have to call `agentcore policy generate` and then customers will copy and paste the cedar policy to `agentcore add policy --cedar policy`. It's two way door.

src/handlers/project/add/policy/index.ts
```diff
@@ -55,16 +64,81 @@ export const createAddPolicyHandler = (config: AddProjectResourceConfig) =>
       if (!flags.name) {
         throw new InputValidationError("required option '--name <name>' not specified");
       }
-      if (!flags.statement) {
-        throw new InputValidationError("required option '--statement <statement>' not specified");
+      const sources = [flags.statement, flags.generate].filter((value) => value !== undefined);
+      if (sources.length !== 1) {
+        throw new InputValidationError("specify exactly one of '--statement' or '--generate'");
```
AlexanderRichey, line 69: Should this be under `add`? Or should the be somewhere else? `agentcore policy generate`? Which could just print some Cedar?

AlexanderRichey (changes requested):

AlexanderRichey:
Stale

PR #2130, opened 2026-08-27
fix(scaffold): allow certain template values to be overriden

Hweinstock:
## Problem

Not all template fields that we'd expect to be able to overwrite are able to be overwritten. For example, in the create flow runtimeName is fixed since passing runtimeName + template is invalid.

## Solution

- allow all flags (except language and framework) to overwrite the template defaults.
- refactor shortcuts into their own file since it no includes non-type specific logic.(likely will conflict with https://github.com/aws/agentcore-cli/pull/2116 since it does the same thing). 

## Testing
- unit tests

```
> agentcore project create --template strands-python --runtimeName bob --name testP 

{
  "name": "testP",
  "version": 1,
  "managedBy": "CDK",
  "runtimes": [
    {
      "name": "bob",
      "build": "CodeZip",
      "entrypoint": "main.py",
      "codeLocation": "app/bob",
      "runtimeVersion": "PYTHON_3_14",
      "protocol": "HTTP"
    }
  ]
}

> agentcore project deploy 
...
```
went to console and verified the name is now generated from bob with name `testP_bob`.


Hweinstock:
> No new user-facing surface, so no new telemetry needed

I think we might need to adjust its prompt. I can take a look.

src/handlers/project/add/runtime/index.ts
```diff
@@ -108,19 +109,31 @@ export const createAddRuntimeHandler = (config: AddProjectResourceConfig) =>
...
+              build: flags.build,
+              runtimeVersion: flags.build === "CodeZip" ? "PYTHON_3_14" : undefined,
+            }),
+            ...(flags["model-provider"] !== undefined && {
+              modelProvider: flags["model-provider"],
+            }),
+            ...(apiKey !== undefined && { apiKey }),
+            ...(flags.memory !== undefined && { memory: flags.memory }),
+          })
         : isCustom
```
notgitika, line 137: OOS for your PR but would we see any value in making a shared component for the shared functionality b/w create and runtime?

src/handlers/project/add/runtime/index.test.ts
notgitika, line 1: wait why are adding the build override? shouldn't that also belong to the "locked" param?

for eg: strands-python template can be override to container but the generated files won't contain Dockerfile, while the agentcore.json config would reference one.

notgitika (commented):
Very easy to follow through PR I just added one comment about build also becoming a locked param imo lmk what you think

src/handlers/project/add/runtime/index.ts
Hweinstock: YES! I'm hoping to come back to this.

src/handlers/project/add/runtime/index.test.ts
Hweinstock: The build override should work since its passed as a parameter to the template. I haven't wired up container support for the strands-python one yet, so if its not rejecting that's a bug. 

Update: it is a bug, let me just fix that here.

notgitika (dismissed):
awesome, looks great to me!

src/handlers/project/create/index.ts
```diff
@@ -75,18 +74,33 @@ export const createCreateProjectHandler = (config: CreateProjectHandlerConfig) =
...
+            ...(flags["build"] !== undefined && {
+              build: flags["build"],
+              runtimeVersion: flags["build"] === "CodeZip" ? "PYTHON_3_14" : undefined,
+            }),
+            ...(flags["model-provider"] !== undefined && {
+              modelProvider: flags["model-provider"],
+            }),
+            ...(apiKey !== undefined && { apiKey }),
+            ...(flags["memory"] !== undefined && { memory: flags["memory"] }),
+          })
```
aidandaly24, line 103: Not saying necessarily to change this because it is correct, but this is kind of hard to read at first. To make this more readable we could move the override building logic into `resolveRuntimeTemplateShortcut`? Passing the optional flag values directly would make this easier to scan and keep the `build`/`runtimeVersion` relationship in one place
```
const scaffoldRuntimeInput = resolveRuntimeTemplateShortcut(flags["template"]!, {
  runtimeName: flags["runtime-name"],
  build: flags["build"],
  modelProvider: flags["model-provider"],
  apiKey,
  memory: flags["memory"],
});
```
The resolver could ignore undefined values and own the `build`/`runtimeVersion` relationship, which would make both callers easier to scan.

aidandaly24 (approved):
This looks good to me. Just had one readability comment.

src/handlers/project/create/index.ts
Hweinstock: I like that! I'll have to rebase https://github.com/aws/agentcore-cli/pull/2116 on top of this anyway, so I'm going to merge and address this there.

PR #2131, opened 2026-08-27
ci: bug-bash workflow — record CLI TUI to S3 on every PR

jariy17:
## This PR (`bug-bash.yml` + `record.mjs`) — shelved
The CI caller here assumes the shared **685197708687** E2E role and fetches secrets from **631957124172** — cross-account. That conflicts with the explore-only requirement, so it is **not for merge**; kept as reference. The runtime-bot path above supersedes it.

## Open reviewer findings (only if this CI path is revived)
- Pin the `private-tui-harness` clone to a full commit SHA / org-own it (supply-chain).
- Add an `AUTHORIZED_USERS` gate before assuming the AWS role (mirror `e2e-tests.yml`).

### DIFF
--- a/.github/harness/bug-bash/record.mjs
+++ b/.github/harness/bug-bash/record.mjs
@@ -8,19 +8,19 @@
  * drifts across versions, so the sequence is a calibration knob, not a hardcoded
  * assumption. Teams tune BUGBASH_KEYS/BUGBASH_ARGS per what the current TUI expects.
  */
-import { resolve } from 'node:path';
+import { resolve } from "node:path";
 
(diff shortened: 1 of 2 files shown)


jariy17:
Updated: this CI workflow is **shelved** (cross-account, conflicts with the explore-only requirement). The bug-bash now runs inside AgentCore Runtime in explore as a Strands agent + `run_bug_bash` tool (9/9, generates the HTML report → explore S3). Architecture + all code (clickable): https://artifactory.beta.harmony.a2z.com/view/a1fcbb5c-a839-466d-bf91-a5f34396c78f

AlexanderRichey:
Stale

PR #2135, opened 2026-08-27
feat(eval): ab-test target-based run (create)

jariy17:
Stacked on #2113 (config-based run). Adds the second A/B-test create mode.

## Command structure

```
agentcore eval ab-test target-based run
  -n, --name <name>              required   A/B test name
  -g, --gateway <id>             required   deployed gateway id
  --control <json>               required   {"gateway-target","online-eval"}  (inline · file:// · -)
  --treatment <json>             required   {"gateway-target","online-eval"}  (inline · file:// · -)
  --treatment-weight <1-99>                 control weight = 100 - this (default 50)
  --gateway-filter <json>                   GatewayFilter {"targetPaths":[...]}  (inline · file:// · -)
  --role-arn <arn>                          execution-role override (else auto-provisioned)
  --enable-on-create <true|false>           start immediately (default true; false = create PAUSED)
```

Splits gateway traffic between two **gateway targets**, each scored by its **own** online-eval. The target picks the runtime + endpoint, so those aren't passed separately.

## Actual input → output

**Input** (exploratory account, us-west-2):

```bash
agentcore eval ab-test target-based run \
  --name orders-v2-canary \
  --gateway orders-gateway-demo-000 \
  --control '{"gateway-target":"orders-prod-target","online-eval":"orders-prod-quality"}' \
  --treatment '{"gateway-target":"orders-v2-target","online-eval":"orders-v2-quality"}' \
  --treatment-weight 20
```

**Actual output** — resolves + validates the gateway before any create, so an unreachable gateway fails fast, creating nothing:

```
Error: User: …:assumed-role/Admin/… is not authorized to perform: bedrock-agentcore:GetGateway
```

**Output on success** (shape from `CreateABTest`):

```json
{
  "abTestId": "orders-v2-canary-abc123",
  "abTestArn": "arn:aws:bedrock-agentcore:us-west-2:…:ab-test/orders-v2-canary-abc123",
  "name": "orders-v2-canary",
  "status": "CREATING",
  "executionStatus": "RUNNING"
}
```

`--enable-on-create false` → `NOT_STARTED` (created PAUSED).

## How it works

- **Variants** use `variantConfiguration.target = { name }` (vs config-based's `configurationBundle`); **evaluation** is `perVariantOnlineEvaluationConfig: [{ name: "C", … }, { name: "T1", … }]` — one online-eval per variant.
- **Shared create seam:** extracted `EvalClient.createABTest` (GetGateway → account, provision role, `retryWhileRolePropagates` on assume/AccessDenied, rollback on failure). Both config-based and target-based `run` now go through it — removes the duplicated role/retry block.
- **`--control`/`--treatment`/`--gateway-filter`** via `SourceResolver` (inline / `file://` / `-`), same as config-based run.
- Validation is **server-side** (gateway READY, targets belong to it, each online-eval matches its target's endpoint) — the CLI resolves ids → the request and surfaces the service's 4xx.

## Tests

Consolidated `ab-test.test.tsx` gains target-based coverage: hierarchy (`ab-test → target-based → run`), required-flag matrix, mis-shaped JSON, identical-target rejection, and flag→request mapping (C/T1 targets, per-variant evals, weights). 52 ab-test tests pass; `typecheck` 0; `oxlint` clean.

## Deferred

- **Golden fixture** — recording needs a gateway wired to two runtime-backed targets + two online-evals (heavier than config-based's self-created bundle). The shared `createABTest` seam is already golden-covered by #2113's config-based run, so target-based rides the same recorded path. Follow-up.
- **TUI wizard** (the doc's interactive target-based flow) — follow-up.

## Review

Base = `feat/eval-ab-test-config-bundle-run` (stack). Review after #2113 merges, or view just this branch's diff.


src/core/eval.tsx
```diff
@@ -556,6 +531,99 @@ export class EvalClient implements CoreEvalClient {
...
+        const evalArn = (id: string) =>
+          `arn:aws:bedrock-agentcore:${options.region}:${accountId}:online-evaluation-config/${id}`;
+        return {
+          name: input.name,
+          gatewayArn,
+          variants: [
+            {
+              name: "C",
+              weight: 100 - treatmentWeight,
+              variantConfiguration: { target: { name: input.control.gatewayTarget } },
```
nborges-aws, line 613: Is target expecting a name or id here? `src/handlers/eval/ab-test/target-based/run/index.tsx#L21` documents gateway-target as `<id>`; I'd expect `input.control.gatewayTarget` is then the target ID. But this field looks to expect target name?

src/core/eval.tsx
```diff
@@ -556,6 +531,99 @@ export class EvalClient implements CoreEvalClient {
...
+            },
+            {
+              name: "T1",
+              weight: treatmentWeight,
+              variantConfiguration: { target: { name: input.treatment.gatewayTarget } },
```
nborges-aws, line 618: same question as above

nborges-aws (commented):
One clarifying question on target ID/name passing. Otherwise PR looks clean!

src/core/eval.tsx
jariy17: Good catch — it is the target **name**, not the id. The data-plane SDK `TargetRef` has a single field, `name` ("The name of the gateway target"), so `variantConfiguration.target` must carry the target name. The confusion was the flag doc: it said `<id>`, which reads like the `targetId` the `gateway target get/delete/update` commands use. Fixed in 2b7fd01 — the validation error and `--control`/`--treatment` help now show `{"gateway-target":"<name>","online-eval":"<id>"}`. The mapping itself is unchanged (correct).

src/core/eval.tsx
jariy17: Same as above — this is the gateway target **name** (SDK `TargetRef.name`), and the `<id>` doc was corrected to `<name>` in 2b7fd01.

nborges-aws (dismissed):
LGTM. Only nit would be if we need golden fixture tests for this?

notgitika (dismissed):

jariy17:
Thanks for the approve! Added the golden in 921d881 — no longer deferred.

It records the create seam **live** against 685197708687 (gateway `agentcore-cli-gateway-read-fixture-a-l6opkbe2kd`, its two READY targets `-a1`/`-a2`): two paused online-evals + a paused `ab-test target-based run`, replayed offline (fixture file = 8 pass). Mirrors the config-based golden, so both run modes now have one.

Honest caveat: like the config-based golden, it captures the **accepted synchronous** `CreateABTest` response (`executionStatus=NOT_STARTED`) — async provisioning isn't asserted. While recording I did confirm one robustness gap worth a follow-up (not in this PR): `ab-test delete` throws `ConflictException` while `status=CREATING` (it isn't waited/retried), so a just-created test can be briefly undeletable. Filing separately.

notgitika (approved):

PR #2138, opened 2026-08-28
feat(dev): supervise every runtime in headless mode

tejaskash:
## What

`--mode headless` without `--agent` now supervises every runtime in the project, streaming attributed `[name]` output to the terminal, the same behavior browser mode gets from the Inspector. Follows the discussion on #2086: multi-agent terminal supervision (useful for A2A testing) was the interim behavior on `refactor` between #2041 and #2086, and this restores it as a first-class mode instead of requiring `--agent`.

- `--mode headless` (no `--agent`): all runtimes start eagerly, output is `[name]`-attributed, one agent crashing leaves the others running, and the command exits non-zero only when nothing is left running and something failed.
- `--mode headless --agent <name>`: unchanged, the direct single-runtime path where a crash fails the command with a real exit code (scripts and CI rely on this).
- Browser mode: unchanged.

## Verification

- Handler suite covers the new path: attributed output and per-runtime OTEL env for two supervised runtimes, crash isolation, SIGINT exits 130 and closes the collector, all-agents-failed exits non-zero.
- End to end against a real 2-runtime project: `project dev --mode headless` started both agents eagerly on distinct ports with `[bob]`/`[alice]` attributed output, and SIGTERM tore both down.
- Full suite (2163 pass), typecheck, format clean. The two oxlint errors are pre-existing on `refactor` (DataTable.tsx, usePagedList.tsx).

### DIFF
--- a/src/handlers/project/dev/index.ts
+++ b/src/handlers/project/dev/index.ts
@@ -10,6 +10,7 @@ import type { ProjectRuntime } from "../../../projectSchemas/runtime";
 import {
   InputValidationError,
   ResourceNotFoundError,
+  SilentCLIError,
   UserCancellationError,
 } from "../../../errors";
 import type { AppIO, BrowserOpener, FileWatcher, PortChecker, startHttpServer } from "../../../io";
@@ -102,7 +103,7 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
       flag("traces", "disable local OTEL trace collection", z.boolean().default(true)),
       flag(
         "mode",
-        "how to run: browser (Agent Inspector web UI), headless (one agent in the terminal), or tui",
+        "how to run: browser (Agent Inspector web UI), headless (agents stream to the terminal), or tui",
         z.enum(["browser", "headless", "tui"]).default("browser"),
       ),
       flag(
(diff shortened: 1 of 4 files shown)


src/handlers/project/dev/index.ts
```diff
@@ -227,6 +222,23 @@ export const createDevProjectHandler = (config: DevProjectHandlerConfig) =>
           signal: controller.signal,
         });
 
+        if (flags.mode === "headless") {
+          const starts = Promise.allSettled(
+            runtimes.map((runtime) => supervisor.start(runtime.name)),
+          );
+          for await (const { agentName, event } of supervisor.events()) {
```
aidandaly24, line 227: I think normal Ctrl-C gets reported as a crash in this supervised headless path. The production process runner rejects with the child signal's `UserCancellationError`, so the supervisor emits `Agent 'orders' crashed: Operation cancelled by user` before the command exits 130. Could the supervisor treat an error from an already-aborted child signal as a normal stop? A test where the runner throws `input.signal.reason` on abort would cover the production behavior better than `stayingRunner`, which returns cleanly.

Hweinstock (commented):
lgtm, agree with Aidan's comment that we don't want to show user cancellations as errors

AlexanderRichey:
Stale

src/handlers/project/dev/index.ts
tejaskash: Good catch, fixed. The pump now treats an error from an already-aborted child signal as a stop, and the runner fakes throw the abort reason like the real runner so the SIGINT tests cover the production path. Verified end to end: Ctrl-C exits 130 with no crash line.

AlexanderRichey (approved):

PR #2142, opened 2026-08-28
feat(templates): support strands ts

Hweinstock:
### Problem 
The CLI is missing support for TS. 

### Solution 
- add strands TS template from upstream main https://github.com/aws/agentcore-cli/tree/main/src/assets/typescript/http/strands as is to establish parity. 
- add in TS support to dev. 
- handle `.ts` --> `.js` conversion for dev/deploy. 
- container support will be a follow-up. 
### Testing 
```
> agentcore project create --name hello --template strands-ts 
Creating project tree
Installing CDK dependencies with npm
Installing Node dependencies with npm
Initializing git repository
Created project 'hello' in ./hello
> agentcore project dev --mode headless --agent strands_agent
OTEL collector listening on port 43911; traces persist to [..]
[strands_agent] Starting development server
[strands_agent] (node:3374940) [FSTDEP023] FastifyDeprecation: disableRequestLogging option 
is deprecated. Use the logController option with disableRequestLogging or isLogDisabled over
ride instead. The disableRequestLogging top-level option will be removed in `fastify@6`.
[strands_agent] (Use `node --trace-warnings ...` to show where the warning was created)
[strands_agent] {"level":30,"time":1787947133429,"pid":3374940,"hostname":"dev-dsk-hkobew-1d
-51fde466.us-east-1.amazon.com","msg":"Server listening at http://169.254.253.253:8080"}
[strands_agent] {"level":30,"time":1787947133429,"pid":3374940,"hostname":"dev-dsk-hkobew-1d
-51fde466.us-east-1.amazon.com","msg":"Server listening at http://127.0.0.1:8080"}
[strands_agent] {"level":30,"time":1787947133429,"pid":3374940,"hostname":"dev-dsk-hkobew-1d
-51fde466.us-east-1.amazon.com","msg":"Server listening at http://10.39.201.239:8080"}
[strands_agent] {"level":30,"time":1787947133429,"pid":3374940,"hostname":"dev-dsk-hkobew-1d
-51fde466.us-east-1.amazon.com","msg":"Server listening on port 8080"}
...
[separate terminal]
>  curl -X POST http://127.0.0.1:8080/invocations \
    -H 'Content-Type: application/json' -H 'Accept: text/event-stream' \
    -H 'x-amzn-bedrock-agentcore-runtime-session-id: strands-ts-e2e-session-0000000000
02' \
    -d '{"prompt":"Reply with exactly: STRANDS_TS_DEV_OK"}'
data: "STRANDS_TS_DEV"

data: "_OK"

> agentcore project runtime add --template strands-ts --memory none --name bob 
Reading project spec file at '/[..]'
Scaffolding runtime in project
Installing Node dependencies with npm
Updating project spec file at '/local/home/hkobew/gh-repos/agentcore-cli/.claude/worktrees/s
trands-ts-template/hello/agentcore/agentcore.json'
added runtime 'bob' to 'hello'
> agentcore project deploy
..
[invoke both in the console]. 
```

### Notes
discovered in testing that the python sdk allows missing sessionId but ts sdk rejects it: https://github.com/aws/bedrock-agentcore-sdk-typescript/blob/v0.3.0/src/runtime/app.ts#L337-L341. Small papercut on the local dev experience.


src/handlers/project/types.ts
```diff
@@ -27,12 +26,11 @@ export const ScaffoldRuntimeInputSchema = z
   .object({
     runtimeName: AgentNameSchema,
     build: BuildTypeSchema,
-    language: z.enum(["Python"]),
+    language: z.enum(["Python", "TypeScript"]),
     framework: z.enum(["strands", "none"]),
     modelProvider: z.enum(["Bedrock"]),
     apiKey: z.string().min(1).optional(),
     memory: MemorySchema.optional(),
```
Hweinstock, line 36: this is now resolved within the template. We can bring it back if we choose to support BYO again.

src/assets/templates/strands-http-typescript/memory/memory.ts
```diff
@@ -0,0 +1,52 @@
...
+const memoryManagerCache = new Map<string, MemoryManager>();
+
+export function getOrCreateMemoryManager(sessionId: string, actorId: string): MemoryManager | null {
+  if (!MEMORY_ID) return null;
+
+  const key = `${actorId}:${sessionId}`;
+  let manager = memoryManagerCache.get(key);
+  if (manager) return manager;
+
+  const stores = createAgentCoreMemoryStores({
```
nborges-aws, line 26: IIRC we should only create memory stores if LTM strategies are included. This looks like it does so for STM as well (main.ts#L54 creates this whenever `hasMemory` is true, which is true for STM-only)

src/assets/templates/strands-http-typescript/memory/memory.ts
```diff
@@ -0,0 +1,52 @@
...
+import { createAgentCoreMemoryStores } from 'bedrock-agentcore/experimental/memory/strands';
+
+const MEMORY_ID = process.env.{{memoryEnvVarName}};
+
+const CUSTOM_ACTOR_ID_HEADER = 'x-amzn-bedrock-agentcore-runtime-custom-actor-id';
+
+export function getActorId(payload: any, context: any): string {
+  const raw =
+    context?.headers?.[CUSTOM_ACTOR_ID_HEADER] ||
+    payload?.userId ||
```
nborges-aws, line 12: `userId` isn't in the requestSchema in main.ts#L37. Zod may be stripping the userId field from the payload before it gets to this point

src/assets/templates/strands-http-typescript/memory/memory.ts
Hweinstock: good catch! It looks like the TS SDK diverges from the python SDK here and throws on missing namespace https://github.com/aws/bedrock-agentcore-sdk-typescript/blob/efef021a7a75aebefccb5111bc0427b5f4c1cbfb/src/memory/integrations/strands/factory.ts#L130-L133. Looks like an existing bug in the template. 

Was wondering why we didn't catch this with local dev, but its because the memory isn't deployed so it doesn't have the injected `MEMORY_ID` that guards the memory store inclusion. 

I reproduced it by deploying and invoking in the console and saw `event: error data: {"error":"createAgentCoreMemoryStores: at least one namespace is required"}.`

I'm going to explicitly reject memory without namespaces as incompatible with TS for now, and file a bug report for our TS SDK.

src/assets/templates/strands-http-typescript/memory/memory.ts
Hweinstock: good catch, looks like another bug in the template. This one is trickier to verify, but we should be able to invoke a deployed agent, and check the memory events for actorId. If its the passed in userId, we're good, but right now its still the session Id.

src/assets/templates/strands-http-typescript/memory/memory.ts
Hweinstock: issue in ts sdk: https://github.com/aws/bedrock-agentcore-sdk-typescript/issues/253

nborges-aws (approved):

notgitika (approved):
LGTM we might need to revisit using the exa mcp here.

P1 but could we explore making the MCP opt-in or non-fatal?
Strands connects during agent initialization and propagates connection failures, so exa downtime or VPC deployments without NAT fail before reaching Bedrock

PR #2144, opened 2026-08-28
feat(project): add `project add evaluator code-based`

jariy17:
## Command structure

```
agentcore project add evaluator          add a custom evaluator to the current project
├── llm-as-a-judge                       existing — LLM prompted to score a session
└── code-based                           NEW — a Lambda that scores a session
agentcore project remove evaluator --name <name>   NEW — enabled via the generic remove
```

`agentcore project add evaluator code-based --help`:

```
Usage: agentcore project add evaluator code-based [options]

add a code-based evaluator — a Lambda that scores a session. Pass a 3P metric,
an existing Lambda, or neither to scaffold an empty evaluator you fill in

Options:
  --name <name>                        the name of the evaluator
  --level <level>                      what to score: SESSION, TRACE, or TOOL_CALL
  --metric <metric>                    3P metric to scaffold as <library.Metric>,
                                       e.g. deepeval.FaithfulnessMetric or autoevals.Factuality
  --model <model>                      judge model for the 3P metric,
                                       e.g. bedrock/anthropic.claude-3-5-sonnet-20240620-v1:0
  --lambda-arn <lambda-arn>            ARN of an existing Lambda that scores a session
  --timeout-seconds <timeout-seconds>  Lambda timeout in seconds (1-300)
  --description <description>          a description of what this evaluator measures
  --kms-key-arn <kms-key-arn>          customer-managed KMS key ARN to encrypt the evaluator
  --tags <tags>                        tags to apply (JSON object of key/value strings)
  -h, --help                           display help for command
```

## Commits

1. `c4430c03` feat — the command + 3 scaffold templates + `remove evaluator`
2. `baec1630` fix — guard `app/<name>` collisions (up-front, no partial writes)
3. `e7bc3675` fix — validate `--metric` class + require a Bedrock `--model`
4. `9bd79980` fix — echo the inferred mode + caveats at add time
5. `32a10ef9` refactor — share `toPythonPackageName` via `fsUtils`; `DEFAULT_TIMEOUT` const
6. `853dcf86` refactor — move template knowledge into `templates/evaluator.ts` (runtime layering)

## Testing

- `bun run build` OK · `bun test src/handlers/project src/core/project` → **597 pass / 0 fail**.
- **Cloud bug bash** (5 parallel agents; 2 deployed to a non-prod account, us-west-2, then tore down): all 5 flows + a 13-case error matrix pass at the CLI/scaffold/synth layer. The managed evaluator **synthesizes correctly** into `AWS::BedrockAgentCore::Evaluator` + Lambda + role + permissions, and generated Python `ast.parse`s for deepeval + autoevals (bedrock + openai branches). Full report shared separately.

### DIFF
--- a/src/assets/evaluators/autoevals-lambda/execution-role-policy.json
+++ b/src/assets/evaluators/autoevals-lambda/execution-role-policy.json
@@ -1,0 +1,15 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": ["logs:CreateLogGroup", "logs:CreateLogStream", "logs:PutLogEvents"],
+      "Resource": "arn:*:logs:*:*:log-group:/aws/lambda/*"
+    },
+    {
+      "Effect": "Allow",
+      "Action": ["bedrock:InvokeModel"],
+      "Resource": "*"
+    }
+  ]
+}
--- a/src/assets/evaluators/autoevals-lambda/pyproject.toml
+++ b/src/assets/evaluators/autoevals-lambda/pyproject.toml
@@ -1,0 +1,22 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "{{ Name }}"
+version = "0.1.0"
+description = "AgentCore Code-Based Evaluator (Autoevals)"
+requires-python = ">=3.10"
+dependencies = [
+    "bedrock-agentcore[autoevals]",
+    "autoevals>=0.0.80,<1.0.0",
+{{#if ModelProviderBedrock}}
+    # autoevals grades via LiteLLMClient -> Bedrock (Converse); litellm replaces the openai judge
+    "litellm>=1.60,<1.85",
+{{else}}
+    "openai>=1.0.0",
+{{/if}}
+]
+
+[tool.hatch.build.targets.wheel]
+packages = ["."]
--- a/src/assets/evaluators/autoevals-lambda/pyproject.toml
+++ b/src/assets/evaluators/autoevals-lambda/pyproject.toml
@@ -8,8 +8,9 @@ version = "0.1.0"
 description = "AgentCore Code-Based Evaluator (Autoevals)"
 requires-python = ">=3.10"
 dependencies = [
-    "bedrock-agentcore[autoevals]>=1.6.0,<2.0.0",
-    "autoevals>=0.0.80,<1.0.0",
+    # 1.20.0 is the first release shipping third_party.autoevals.AutoEvalsAdapter;
+    # the extra owns the autoevals version, so pinning it here only conflicts.
+    "bedrock-agentcore[autoevals]>=1.20.0,<2.0.0",
 {{#if ModelProviderBedrock}}
     # autoevals grades via LiteLLMClient -> Bedrock (Converse); litellm replaces the openai judge
     "litellm>=1.60,<1.85",
--- a/src/assets/evaluators/autoevals-lambda/pyproject.toml
+++ b/src/assets/evaluators/autoevals-lambda/pyproject.toml
@@ -8,13 +8,13 @@ version = "0.1.0"
 description = "AgentCore Code-Based Evaluator (Autoevals)"
 requires-python = ">=3.10"
 dependencies = [
-    "bedrock-agentcore[autoevals]",
+    "bedrock-agentcore[autoevals]>=1.6.0,<2.0.0",
     "autoevals>=0.0.80,<1.0.0",
 {{#if ModelProviderBedrock}}
     # autoevals grades via LiteLLMClient -> Bedrock (Converse); litellm replaces the openai judge
     "litellm>=1.60,<1.85",
 {{else}}
-    "openai>=1.0.0",
+    "openai>=1.0.0,<2.0.0",
 {{/if}}
 ]
 
--- a/src/assets/evaluators/python-lambda/lambda_function.py
+++ b/src/assets/evaluators/python-lambda/lambda_function.py
@@ -1,0 +1,19 @@
+from bedrock_agentcore.evaluation.custom_code_based_evaluators import (
+    custom_code_based_evaluator,
+    EvaluatorInput,
+    EvaluatorOutput,
+)
+
+
+@custom_code_based_evaluator()
+def handler(input: EvaluatorInput, context) -> EvaluatorOutput:
+    """Evaluate agent behavior with custom logic.
+
+    Args:
+        input: Contains evaluation_level, session_spans, target_trace_id, target_span_id
+
+    Returns:
+        EvaluatorOutput with value/label for success, or errorCode/errorMessage for failure.
+    """
+    # TODO: Replace with your evaluation logic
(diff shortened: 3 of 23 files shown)


src/handlers/project/add/evaluator/code-based/index.ts
```diff
@@ -0,0 +1,193 @@
...
+        // managed: 3P metric, or empty stub when no metric is given.
+        if (flags["model"] && !hasMetric)
+          throw new InputValidationError("--model requires --metric");
+
+        let assetDir = EMPTY_ASSET_DIR;
+        let defaultTimeout = EMPTY_DEFAULT_TIMEOUT_SECONDS;
+        const context: Record<string, unknown> = { Name: toPythonPackageName(flags["name"]) };
+
+        if (hasMetric) {
+          const raw = flags["metric"]!;
```
jariy17: Can we use make this helper function and use zod here.

src/assets/evaluators/autoevals-lambda/execution-role-policy.json
```diff
@@ -0,0 +1,15 @@
+{
```
jariy17, line 1: I went back and forth on this, but decided not to abstract away common files like this one. It would add more complexity to our code for little benefit.

src/core/project/templates/evaluator.ts
```diff
@@ -0,0 +1,89 @@
+import { FsTreeNode } from "./fsTree";
+import type { AssetSource } from "../source";
+import type { Evaluator, EvaluationLevel } from "../../../projectSchemas/evaluator";
+import type { TemplateRenderer, TemplateResolver } from "./types";
+import { toPythonPackageName } from "../fsUtils";
+
+const DEFAULT_TIMEOUT = 60;
+
+export const EVALUATOR_LIBRARIES = {
```
jariy17: I considered the runtime templates' resolverKey map, but its only useful if multiple flags determines the asset. Here code-based evaluators pick the template from only one flag (--metric). If this changes, we can use resolverKey maps.

src/assets/evaluators/autoevals-lambda/execution-role-policy.json
Hweinstock: agree with this decision.

src/assets/evaluators/python-lambda/lambda_function.py
```diff
@@ -0,0 +1,19 @@
+from bedrock_agentcore.evaluation.custom_code_based_evaluators import (
```
Hweinstock, line 1: Should we scaffold a starter README into these assets like we do with the others? (could be a follow-up, as its just a nice to have)

src/assets/evaluators/autoevals-lambda/pyproject.toml
```diff
@@ -0,0 +1,22 @@
...
+description = "AgentCore Code-Based Evaluator (Autoevals)"
+requires-python = ">=3.10"
+dependencies = [
+    "bedrock-agentcore[autoevals]",
+    "autoevals>=0.0.80,<1.0.0",
+{{#if ModelProviderBedrock}}
+    # autoevals grades via LiteLLMClient -> Bedrock (Converse); litellm replaces the openai judge
+    "litellm>=1.60,<1.85",
+{{else}}
+    "openai>=1.0.0",
```
Hweinstock: could we use compatible releases or pin this above to avoid the templates breaking on version bumps?

https://packaging.python.org/en/latest/specifications/version-specifiers/#compatible-release

src/core/project/fsUtils.ts
Hweinstock, line 1: OOS, but I wonder if we should rename this to avoid it becoming a dumping ground.

src/handlers/project/add/evaluator/code-based/index.ts
```diff
@@ -0,0 +1,144 @@
...
+  EvaluatorSchema,
+  EvaluationLevelSchema,
+  isValidBedrockModelId,
+} from "../../../../../projectSchemas/evaluator";
+import { TagsSchema } from "../../../../../projectSchemas/tags";
+import {
+  EVALUATOR_LIBRARIES,
+  type EvaluatorLibrary,
+  type ManagedEvaluatorScaffoldInput,
+} from "../../../../../core/project/templates/evaluator";
```
Hweinstock: isn't this breaking the dependency inversion? The downside being that we're coupling the UI to the underlying core implementation, which I think the dependency inversion aims to avoid.

src/handlers/project/add/evaluator/code-based/index.ts
```diff
@@ -0,0 +1,144 @@
...
+        config.io.stderr.write(`added evaluator '${flags["name"]}' to '${project.name}'\n`);
+        return;
+      }
+
+      if (flags["model"] && !hasMetric) throw new InputValidationError("--model requires --metric");
+
+      const scaffold: ManagedEvaluatorScaffoldInput = {
+        ...base,
+        ...(hasMetric && { metric: parseMetric(flags["metric"]!) }),
+        ...(flags["model"] !== undefined && { model: resolveBedrockModel(flags["model"]) }),
```
Hweinstock, line 101: for my own understanding, do evals only supports bedrock?

Hweinstock (commented):
Looks good, two comments worth thinking about before merging related to the dependency inversion and version pinning.

src/core/project/fsUtils.ts
jariy17: Yeah, I was thinking that too, maybe we name this `ProjectFilePathUtils` and shift toPythonPackageName to it's own module.

src/handlers/project/add/evaluator/code-based/index.ts
jariy17: Right now, LLM-as-a-judge evaluators only support Amazon Bedrock. Code-based evaluators are just Lambda functions that return an output the service can read. This means a customer can write a Lambda that calls out to any model provider they want.

src/handlers/project/add/evaluator/code-based/index.ts
jariy17: Yeah i missed this completely. Ill fix this.

src/assets/evaluators/autoevals-lambda/pyproject.toml
jariy17: Sure thing.

src/assets/evaluators/python-lambda/lambda_function.py
jariy17: Yeah, that make sense. I'll add it for this PR.

Hweinstock (dismissed):
LGTM!

notgitika:
Blocking issue: the generated DeepEval scaffold is not currently installable/runnable.

- With `--model`, `uv lock` fails because the generated `bedrock-agentcore`, `deepeval`, `aiobotocore`, and `botocore` constraints are incompatible.
- Without `--model`, dependencies resolve, but importing `lambda_function` fails because the selected AgentCore SDK does not contain `third_party.deepeval.DeepEvalAdapter`.

Could we align the DeepEval/AgentCore dependency versions and add a scaffold smoke test that runs `uv lock` and imports the generated handler?

notgitika (approved):

PR #2147, opened 2026-08-31
feat: add observability abstractions and wire to runtime

nborges-aws:
## Description

This PR adds reusable observability infrastructure. The setup is built to apply generally across our primitives, while allowing for resource-specific customization where necessary. This PR wires runtime to the infrastructure. Remaining primitives wiring will be released as a follow pending alignment on the abstractions added here.

- Adds a reusable, logs-only `ObservabilityClient` as the shared API entry point
- Adds a source resolver registry, with runtime-specific log-group resolution
- Adds a primitive-agnostic CloudWatch source reader for search and live tail
- Adds a shared observability handler factory and mounts `logs` under Runtime
- Normalizes provider events into a generic `LogRecord`

## Architecture

Runtime handler → ObservabilityClient → RuntimeSourceResolver → CloudWatchSourceReader → LogRecord

The resolver owns resource-to-log-group translation. The source reader owns CloudWatch mechanics without ever needing knowledge of our resource types. The client is responsible for the orchestration of these layers.

## Commands 

Tail logs:

```sh
agentcore runtime logs --id <runtime-id> --tail
```

Search logs:
```
agentcore runtime logs \
    --id <runtime-id> \
    --since 1h \
    --until now \
    --level error \
    --query '"timed out"' \
    --limit 100
```
  --qualifier selects a non-default Runtime endpoint.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] `bun run test` (2322 pass, 0 fail)
- [x] I ran `npm run lint`

## Checklist

- [x]  I have added any necessary tests that prove my fix is effective or my feature works
- [x]  I have updated the documentation accordingly

---

terms of your choice.


src/core/observability/client.test.ts
```diff
@@ -0,0 +1,144 @@
+import { describe, expect, test } from "bun:test";
```
jariy17, line 1: We should use Golden tests here. Runtime logs handler tests should cover this.

src/core/observability/client.ts
```diff
@@ -0,0 +1,148 @@
...
+  LogSource,
+  ObservableResourceRef,
+  ObservabilitySourceResolver,
+  ObservabilitySourceResolverRegistry,
+  ResolvedObservabilityTarget,
+  ResolvedResourceIdentity,
+} from "./resolver";
+import type { LogSearchQuery, LogTailQuery, RawLogRecord, SourceReader } from "./sourceReader";
+
+export interface LogRecord {
```
jariy17: Since this is ONLY storing data, please declare it as a type.

src/core/observability/sourceReader.ts
```diff
@@ -0,0 +1,179 @@
...
+    signal: AbortSignal,
+  ): AsyncIterable<RawLogRecord>;
+}
+
+/**
+ * Executes CloudWatch operations against resolved descriptors. It has no
+ * knowledge of Runtime or any other AgentCore resource type.
+ */
+export class CloudWatchSourceReader implements SourceReader {
+  constructor(private readonly clients: Pick<AwsClients, "logs">) {}
```
jariy17, line 88: nit: Since we are only using the Logs Client, it should be logClient.

src/handlers/runtime/logs.test.tsx
```diff
@@ -0,0 +1,189 @@
+import { describe, expect, test } from "bun:test";
+import type { LogRecord } from "../../core/observability";
```
jariy17, line 2: Golden tests please.

src/core/observability/client.ts
AlexanderRichey: Is this importable from a package somewhere? Is this what all log records look like in AC?

src/core/observability/resolver.ts
```diff
@@ -0,0 +1,72 @@
...
+  provider: "cloudwatch";
+  logGroupName: string;
+};
+
+export interface ResolvedObservabilityTarget {
+  resource: ResolvedResourceIdentity;
+  logs: readonly LogSource[];
+}
+
+export interface ObservabilitySourceResolver<R extends ObservableResourceRef> {
```
AlexanderRichey: Do we really need all of these complex types?

src/handlers/observability/handlerFactory.ts
```diff
@@ -0,0 +1,140 @@
...
+  ),
+] as const;
+
+type LogFlagValues = ResourceFlagValues<typeof logFlags>;
+
+/**
+ * Builds reusable logs command behavior. Primitive routers contribute only
+ * identity flags and conversion to an ObservableResourceRef.
+ */
+export class ObservabilityHandlerFactory implements ObservabilityHandlerFactories {
```
AlexanderRichey: Interesting. Did you consider writing a function that returns a handler function? That might be simpler.

src/handlers/observability/time.ts
```diff
@@ -0,0 +1,32 @@
...
+  const relative = RELATIVE_DURATION_RE.exec(trimmed);
+  if (relative) {
+    return now() - parseInt(relative[1]!, 10) * UNIT_TO_MS[relative[2]!]!;
+  }
+  if (/^\d{13,}$/.test(trimmed)) return parseInt(trimmed, 10);
+
+  const timestamp = Date.parse(trimmed);
+  if (!Number.isNaN(timestamp)) return timestamp;
+
+  throw new InputValidationError(
```
AlexanderRichey, line 28: Any reason we're not using `date-fns`?

src/handlers/observability/types.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export interface ObservableResourceCommand<
+  K extends ObservableResourceRef["kind"],
+  F extends readonly Flag<string, unknown>[],
+> {
+  flags: F;
+  toResource(flags: ResourceFlagValues<F>): Extract<ObservableResourceRef, { kind: K }>;
+}
+
+export interface ObservabilityHandlerFactories {
```
AlexanderRichey: What's the upshot of an interface for this?

AlexanderRichey (changes requested):
There is some good stuff here but I think we can simplify a bit. I like the idea of creating a function that creates these handlers for us, but I think this can be done with a little less abstraction and more directly. It seems like what you want is something like:

```typescript
const createLogsHandler = (client: ObservabilityClient, io: AppIO) => createHandler(...)
```

Then use this for Runtime and Harness:

```typescript
export function createRuntimeHandler(...): Router {
    ...
    runtime.handler(createLogsHandler(obsC, io))
    ...
}
```

jariy17 (changes requested):
I would try to reduce the abstraction by getting rid of the factory. A static function like Alex suggested would reduce the complexity here. Otherwise, this is a very good abstraction that should expand to other AgentCore resources. Thanks for diving deep on this implementation.

src/core/observability/client.ts
nborges-aws: I looked and couldn't find any good source we could import this from. OpenTelemetry logRecord is the closest thing, but wouldn't support any cloudWatch metadata fields

src/core/observability/resolver.ts
nborges-aws: I was able to simplify a decent chunk of this. The resolver registry is now just a record keyed by the resource type, and got rid of `ResolvedResourceIdentity`. The logical flow stays the same but we now lookup directly with `resolvers[resuorce.kind]`. So no mapped Extract type, or generic resolver parameters.

src/handlers/observability/handlerFactory.ts
nborges-aws: Good call. I've refactored the handler as a static method

src/handlers/observability/time.ts
nborges-aws: No reason except it isn't currently in the project. Using it would save us the ISO parsing, but the rest of the logic would still be needed

src/handlers/observability/types.ts
nborges-aws: No reason to keep as interface. I've changed the handlers to come from a static function as suggested

src/handlers/runtime/logs.test.tsx
nborges-aws: Yep golden tests have been added

AlexanderRichey (changes requested):
This PR has so many changes across so many file that it's hard for me to tell what's going on. I think we need to rethink our approach here.

PR #2149, opened 2026-08-31
feat(feedback): port the feedback command to the refactor architecture

jariy17:
## What

Implements `agentcore feedback`  using a dedicated client and Handler. 

## Command surface

```
agentcore feedback <message> [--screenshot <path>] [--yes]
```

- `<message>` — required (max 1000 chars).
- `--screenshot <path>` — optional PNG/JPG, ≤100 MB.
- `--yes` — accept the AWS Customer Agreement and skip the consent prompt.
- `--json` — inherited global flag; envelope output.

**Screenshot** flow: presign `POST` → S3 `PUT` (SHA256 checksum + `scanstatus=NOT_SCANNED` tag) → form `POST`, referencing the object key parsed from the presigned URL (never fabricated).

## Files

New: `src/core/feedback.tsx` (FeedbackClient + `ApertureError` extending `AgentCoreCLIError` with `ERROR_SOURCE.SERVICE`), `src/handlers/feedback/{types,index,feedback.test}.tsx`.
Wired: `src/handlers/types.tsx` (`Core.feedback`), `src/core/index.tsx` (`CoreClient.feedback`, injected fetch), `src/handlers/index.tsx` (root registration), `src/testing/TestCoreClient.tsx` (`TestFeedbackClient`).

## Testing

- `bun test src/handlers/feedback` — 7/7 pass (consent y/n, non-TTY guard, empty/oversized message, screenshot 3-call flow with checksum/tag headers + object-key assertion). `tsc --noEmit` clean, oxlint clean.
- **3 real submissions to production Aperture** verified end to end (text ×2, text+screenshot): ids `f36eb155-…`, `2e32ed65-…`, `3525eaf4-…`, all `reference: agentcore-cli`.


src/core/feedback.tsx
```diff
@@ -0,0 +1,332 @@
...
+          category: FORM_CATEGORY,
+          name: FORM_NAME,
+          version: FORM_VERSION,
+          fileName: file.fileName,
+          fileSize: file.size,
+          uploadFileSHA256: file.sha256Base64,
+        },
+        userAgent,
+      );
+      await this.uploadFileToS3(
```
aidandaly24: I think the new invalid-presign handling runs too late. `uploadFileToS3()` receives the raw body before `objectKeyFromPresignedUrl()` validates it; with a 200 `not-a-url` response I still got `TypeError: fetch() URL is invalid`, so the intended `ApertureError` never runs. Parsing the object key before the PUT should fix the classification and avoid uploading before the reference is known to be usable.

src/handlers/feedback/feedback.fixture.test.tsx
```diff
@@ -0,0 +1,118 @@
...
+    );
+
+    matchGolden(FIXTURES, "submit-text.golden.json", stdout);
+    const result = JSON.parse(stdout);
+    expect(result.success).toBe(true);
+    expect(typeof result.id).toBe("string");
+    expect(result.reference).toBe("agentcore-cli");
+  }, 120_000);
+
+  test("submits feedback with a screenshot (presign → S3 PUT → form)", async () => {
```
aidandaly24, line 59: I do not think this golden test proves the request contract in its name. `fixtureFetch` keys only on method/path and ignores request headers and bodies; I replayed the PUT with no checksum/tagging headers and the form POST with `{}`, and both returned 200. I think we should keep the golden flow but restore a focused injected-fetch test for the checksum headers, `scanstatus=NOT_SCANNED`, and attachment object key.

src/handlers/feedback/feedback.fixture.test.tsx
```diff
@@ -0,0 +1,118 @@
...
+} from "../../testing";
+import { UserCancellationError } from "../../errors";
+
+// Record with: RECORD=1 bun test src/handlers/feedback/feedback.fixture.test.tsx
+//
+// The two "submits …" tests are WRITES: a record run posts a REAL feedback
+// submission to the Aperture public API (and, for the screenshot case, uploads
+// shot.png through a real presigned S3 PUT) — there is no undo, same as the
+// batch-evaluation evaluate/simulate fixtures that submit real jobs. After a
+// record run, strip the X-Amz-* query from the recorded presign Fetch fixture
```
aidandaly24: Can we avoid making presigned URL cleanup a manual recording step? `fixtureFetch` writes `response.text()` verbatim, so `RECORD=1` puts the live `X-Amz-*` query on disk before this cleanup can happen. A recording sanitizer could write the queryless URL while still returning the original URL to the live upload flow.

src/handlers/feedback/types.tsx
```diff
@@ -0,0 +1,26 @@
+import type { CoreOptions } from "../../core/types";
+
+export interface ScreenshotInput {
```
aidandaly24: Can `ScreenshotInput`, `SubmitFeedbackInput`, and `FeedbackSubmissionResult` be type aliases? These are concrete request/result data shapes, while `CoreFeedbackClient` is the behavioral interface. The neighboring handler type modules use interfaces only for `Core*Client` contracts and types for determinate data; the internal records in `core/feedback.tsx` should probably follow the same pattern.

src/handlers/feedback/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+import type { Core } from "../types.tsx";
+import type { AppIO } from "../../io";
+
+export const createFeedbackHandler = (core: Core, io: AppIO) =>
+  createHandler({
+    name: "feedback",
+    description: "Send feedback about the AgentCore CLI to the team.",
+    // Length/empty validation lives solely in core.submitFeedback so one code path
+    // guards every caller; the arg is unconstrained here beyond being a string.
+    arguments: [argument("message", "the feedback message to send", z.string())],
```
aidandaly24, line 15: Is the no-argument feedback wizard intentionally out of scope? Released `v0.28.1` uses optional `[message]` and opens `FeedbackScreen`; this makes `<message>` required, and I confirmed bare `agentcore feedback` now exits 2. If this is staged, I think the compatibility gap should be tracked or called out; otherwise this should preserve the optional route.

aidandaly24 (commented):
Left a few comments around compatibility and the request/fixture contracts.

src/core/feedback.tsx
```diff
@@ -0,0 +1,332 @@
...
+  name: string;
+  version: string;
+  locale: string;
+  reference: string;
+  location: string;
+  customerResponses: ApertureCustomerResponse[];
+  metadataList: { key: string; value: string }[];
+}
+
+export class FeedbackClient implements CoreFeedbackClient {
```
AlexanderRichey: Should this be a part of Core? I'm not so sure it should. Feedback seems like a different concern from doing stuff in the user's AWS account.

src/core/feedback.tsx
```diff
@@ -0,0 +1,332 @@
...
+    case 417:
+      return "Feedback service rejected the request content type (HTTP 417).";
+    case 500:
+      return "Feedback service returned an internal error (HTTP 500). Please try again later.";
+    default:
+      return `Feedback service returned HTTP ${status}.`;
+  }
+}
+
+async function readBody(response: Response): Promise<string> {
```
AlexanderRichey: What value is this adding?

src/handlers/feedback/index.tsx
AlexanderRichey: Let's not increase scope. Leaving out the TUI makes sense here.

src/core/feedback.tsx
jariy17: I viewed Core as the business-logic layer not only for AWS but for other external APIs. But I'll shift this into the handler.

src/core/feedback.tsx
jariy17: It cleans up the code  but it doesn't help much. I'll remove it.

src/handlers/feedback/feedback.fixture.test.tsx
jariy17: Sure we can do that

src/handlers/feedback/feedback.fixture.test.tsx
jariy17: Sure, I'll add a mock test for this.

AlexanderRichey (approved):

src/core/feedback.tsx
jariy17: Fixed

PR #2150, opened 2026-08-31
fix(create): restore harness-first project options

aidandaly24:
## Context

PR #2146 made harnesses the default `project create` path. Testing that merged behavior found three regressions in the new path. This PR fixes those regressions without changing the existing Bedrock runtime-scaffolding contract.

### 1. Harness model-provider options routed incorrectly

`project create` exposed harness flags for API keys, LiteLLM API base URLs, and provider-specific parameters, but it always wrote `provider: "bedrock"` into `harness.json`. The shared `--model-provider` flag also selected the runtime-scaffolding path by itself. As a result:

- `--api-base` and `--additional-params` always failed validation because they are valid only for LiteLLM.
- `--model-provider open_ai`, `gemini`, or `lite_llm` failed to create a harness.
- Provider-specific default model IDs from the original CLI were unavailable.

The fix keeps Bedrock as the default and restores these harness providers:

- `bedrock` defaults to `global.anthropic.claude-sonnet-4-6`.
- `open_ai` defaults to `gpt-5` and requires `--api-key-arn`.
- `gemini` defaults to `gemini-2.5-flash` and requires `--api-key-arn`.
- `lite_llm` defaults to `anthropic/claude-sonnet-4-5` and accepts `--api-base` and `--additional-params`.

`--model-provider` no longer selects the runtime path on its own. Explicit runtime flags such as `--template`, `--framework`, or `--build` still select runtime scaffolding, where Bedrock remains the only supported model provider.

### 2. A missing Dockerfile left a partial project

For `project create --container <Dockerfile>`, the project manager wrote the base project and CDK files before the harness template checked whether the source Dockerfile existed. A missing file therefore left a partial project directory. Retrying the command then failed because scaffolded files already existed.

The project manager now validates the harness template source before writing the base project tree. A missing Dockerfile returns the existing validation error and leaves no project directory behind.

### 3. `--no-harness-memory` bypassed mixed-flag validation

The flag was registered internally as `no-harness-memory`. Commander treats `--no-*` options as negations and stores their value under the positive attribute name. The router read both an explicit `--no-harness-memory` and an omitted flag as `false`, so this invalid combination created a runtime project:

```bash
agentcore project create \
  --name MyProject \
  --template hello-world-python \
  --no-harness-memory
```

The flag is now represented internally as the positive `harness-memory` boolean while the CLI continues to expose `--no-harness-memory`. The create handler now detects the explicit negation and rejects mixed harness/runtime flags before writing files.

### Automated

- `bun test src/handlers/project/project.test.ts src/core/project/manager.test.ts`
  - 129 passed, 0 failed
- `bun test src`
  - 2,556 passed, 0 failed across 185 files
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- `bun run build`

### Bundled CLI smoke tests

The bundled `dist/index.js` was exercised directly with isolated project directories:

- Explicit `bedrock`, `open_ai`, `gemini`, and `lite_llm` harness creation produced the expected provider-specific model configuration.
- LiteLLM preserved `apiBase` and `additionalParams` in `harness.json`.
- `--template strands-python --model-provider Bedrock` still created a runtime project.
- A non-Bedrock provider on the runtime path failed before writing a project.
- OpenAI without `--api-key-arn` failed before writing a project.
- A missing Dockerfile failed before writing a project.
- Mixing `--template` with `--no-harness-memory` failed before writing a project.

No AWS resources were required for these validation paths.

### DIFF
--- a/src/core/project/manager.test.ts
+++ b/src/core/project/manager.test.ts
@@ -1,4 +1,5 @@
 import { afterEach, describe, expect, test } from "bun:test";
+import { existsSync } from "node:fs";
 import { mkdir, mkdtemp, readdir, rm, writeFile } from "node:fs/promises";
 import { join, relative } from "node:path";
 import { tmpdir } from "node:os";
@@ -161,6 +162,26 @@ describe("FsProjectManager.create", () => {
 
     await runCreate(manager().manager, input);
     await expect(runCreate(manager().manager, input)).rejects.toBeInstanceOf(ProjectStateError);
+  });
+
+  test("validates a harness Dockerfile before writing the project tree", async () => {
+    const directory = await inTempDirectory();
+    const dockerfile = join(directory, "MissingDockerfile");
+
+    await expect(
+      runCreate(manager().manager, {
+        name: "example",
+        skipInstall: true,
+        skipGit: true,
+        scaffoldHarnessInput: {
+          name: "example",
+          model: { provider: "bedrock", modelId: "global.anthropic.claude-sonnet-4-6" },
+          dockerfile,
+        },
+      }),
+    ).rejects.toThrow(`dockerfile not found: '${dockerfile}'`);
+
+    expect(existsSync(join(directory, "example"))).toBe(false);
   });
 
   test("runs npm install, uv sync, and git init after scaffolding", async () => {
--- a/src/core/project/manager.tsx
+++ b/src/core/project/manager.tsx
@@ -26,7 +26,7 @@ import {
 } from "../../io";
 import { defaultSource, type AssetSource } from "./source";
 import { ENV_LOCAL_RELATIVE_PATH, EnvLocalFile } from "./envLocal";
-import { getHarnessTemplateResolver } from "./templates/harness";
+import { getHarnessTemplateResolver, validateHarnessTemplateSource } from "./templates/harness";
 import { createProjectTree } from "./templates/project";
 import { getRuntimeTemplateResolver } from "./templates/runtime";
 import {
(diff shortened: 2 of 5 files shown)


jariy17 (approved):
Its fixes bugs. OOS but there is too many flags for create now!
```
create a new AgentCore project

Options:
  --name <name>                                name of the project to create
  --defaults                                   create a harness project with default settings (this is the default) (default: false)
  --template <template>                        a preset of flags for scaffolding the runtime; compatible flags override preset values
  --build <build>                              build type for the scaffolded runtime code
  --language <language>                        target language for the scaffolded runtime code
  --framework <framework>                      agent framework for the scaffolded runtime code
  --model-provider <model-provider>            model provider: bedrock, open_ai, gemini, or lite_llm for harnesses; Bedrock for runtime code
  --api-key <api-key>                          API key for non-Bedrock providers: '-' for stdin, 'file://path' for file
  --memory <memory>                            memory option for the scaffolded runtime
  --runtime-name <runtime-name>                name of the scaffolded runtime
  --type <type>                                create scaffolds new agent code (the default); import wraps an existing Bedrock Agent
  --agent-id <agent-id>                        Bedrock Agent ID to import (requires --type import)
  --agent-alias-id <agent-alias-id>            Bedrock Agent Alias ID to import (requires --type import)
  --model-id <model-id>                        model ID for the created harness
  --api-key-arn <api-key-arn>                  API key credential ARN for the created harness's model provider
  --api-base <api-base>                        base URL for the harness model provider API endpoint (lite_llm)
  --additional-params <additional-params>      provider-specific harness model params as a JSON object (lite_llm)
  --no-harness-memory                          disable memory for the created harness (this is the default)
  --max-iterations <max-iterations>            max agent loop iterations per invocation (harness)
  --max-tokens <max-tokens>                    max total output tokens per invocation (harness)
  --timeout <timeout>                          max duration in seconds per invocation (harness)
  --truncation-strategy <truncation-strategy>  context truncation strategy for the harness
  --container <container>                      container image URI or Dockerfile path for the harness
  --skip-install                               skip installing dependencies (npm install, uv sync) (default: false)
  --skip-git                                   skip initializing a git repository (default: false)
  -h, --help                                   display help for command
```

Issue #2140, opened 2026-08-28
src/io/watchFile.test.ts flaky test on refactor branch.


Hweinstock:
```
src/io/watchFile.test.ts:
  24 |   watchFile(path, () => calls++, controller.signal);
  25 | 
  26 |   writeFileSync(path, '{"a":1}');
  27 |   writeFileSync(path, '{"a":2}');
  28 |   await Bun.sleep(250);
  29 |   expect(calls).toBe(1);
                       ^
  error: expect(received).toBe(expected)
  
  Expected: 1
  Received: 0
  
        at <anonymous> (/Users/runner/work/agentcore-cli/agentcore-cli/src/io/watchFile.test.ts:29:17)
  
  Error: Expected: 1
  Received: 0
  
        at <anonymous> (/Users/runner/work/agentcore-cli/agentcore-cli/src/io/watchFile.test.ts:29:17)
  (fail) debounces a burst of edits into a single callback and stops on abort [257.78ms]
  (pass) a missing file fails quietly rather than throwing [0.18ms]
```
Ex. https://github.com/aws/agentcore-cli/actions/runs/33117797255/job/98676836925 

Seen it fail twice on MacOS, retrying fixed.

PR #2151, opened 2026-08-31
feat(update): port CLI self-updater command to refactor architecture

jariy17:
## What

Ports the `agentcore update` command from the old CLI (`main`) into the refactored Handler/Router (Bun) architecture. `update` checks the npm registry for a newer `@aws/agentcore` and runs `npm install -g`; `update --check` reports availability without installing.

The `refactor` branch had resource-level `update` subcommands (gateway, harness, eval…) but no top-level self-updater — this adds it.

## Changes

- **`src/handlers/update/action.ts`** (new) — `fetchLatestVersion`, `compareVersions`, `installArgv`, `handleUpdate`. The install runs through an injectable `ProcessRunner` (defaults to the shared `runProcess` from `src/io/exec.ts`), so the install path is testable without spawning a real npm.
- **`src/handlers/update/index.tsx`** (new) — `createUpdateHandler`; always renders the `UpdateResult` as JSON (matching the resource-command convention); npm progress streams to `io.stderr` so stdout stays a clean, pipeable JSON result; `SilentCLIError` for a non-zero exit on failed install.
- **`src/handlers/index.tsx`** — mounts the handler at the root (2 lines).
- **`src/handlers/update/update.test.ts`** (new) — `compareVersions` table, `fetchLatestVersion` fetch spy (200/404), and `handleUpdate` branches (up-to-date / newer-local / update-available / updated / update-failed) via an injected fake runner.

## CLI surface

```
agentcore update           # check and install
agentcore update --check   # report availability only
```

## Verification

- `bun test src/handlers/update/` → **21 pass / 0 fail**
- `bunx tsc --noEmit` → clean
- Live: `update --help`, `update --check`, `update --check --json` all correct

## Notes

Faithful port of the old command's behavior (bare `update` installs; `--check` only checks). A separate adversarial bug bash surfaced pre-existing gaps carried over from the old CLI (dist-tag check/install channel mismatch, no CI/TTY install guard, unvalidated registry JSON) — not addressed here to keep this a clean port; happy to follow up.


src/handlers/update/action.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export function installArgv(): string[] {
+  return ["npm", "install", "-g", `${PACKAGE_NAME}@${distTag()}`];
+}
+
+export async function fetchLatestVersion(): Promise<string> {
+  const response = await fetch(`${REGISTRY_URL}/${PACKAGE_NAME}/latest`);
+  if (!response.ok) {
+    throw new Error(`Failed to fetch latest version: ${response.statusText}`);
```
AlexanderRichey: This should be an AgentCoreCLIError. I think we should create a NetworkingError to cover this and other cases.

src/handlers/update/action.ts
```diff
@@ -0,0 +1,106 @@
...
+      if (ln > cn) return 1;
+      if (ln < cn) return -1;
+    } else {
+      if (ls > cs) return 1;
+      if (ls < cs) return -1;
+    }
+  }
+
+  return 0;
+}
```
AlexanderRichey: This is looking a little gnarly! Can we not use `semver`?

src/handlers/update/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+    flags: [flag("check", "check for updates without installing", z.boolean().default(false))],
+    handle: async (ctx, flags) => {
+      const result = await handleUpdate(flags.check, {
+        onOutput: (chunk) => io.stderr.write(chunk),
+      });
+
+      ctx.require(JsonRendererKey).renderJson(result);
+
+      if (result.status === "update-failed") {
+        throw new SilentCLIError("failed to install update", { exitCode: 1 });
```
AlexanderRichey: Why should this error be silent?

src/handlers/update/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+import type { AppIO } from "../../io";
+import { handleUpdate } from "./action";
+
+export const createUpdateHandler = (io: AppIO) =>
+  createHandler({
+    name: "update",
+    description: "Check for and install CLI updates",
+    flags: [flag("check", "check for updates without installing", z.boolean().default(false))],
+    handle: async (ctx, flags) => {
+      const result = await handleUpdate(flags.check, {
```
AlexanderRichey, line 77: I'd probably just put all of the functionality here. I don't see an upshot of pulling this out. I think code could still be tested easily enough if it were inlined here.

AlexanderRichey (changes requested):

jariy17:
CI note: the only red check is `unit-test / Test (macOS)`, which fails solely on `src/io/watchFile.test.ts` → "debounces a burst of edits into a single callback and stops on abort" — a timing-sensitive debounce test unrelated to this PR (not in the diff; Linux + Windows run the same test and pass). Everything owned by this change is green: `check`, unit-test Linux/Windows, e2e, all builds, CodeQL, Analyze.

src/handlers/update/action.ts
jariy17: We already have a NetworkingError ([ref](https://github.com/aws/agentcore-cli/blob/df70b1b75d82bcf884dd274df273d38cbacfaa2f/src/errors/errors.tsx#L196)). I'll just use that.

src/handlers/update/action.ts
jariy17: I didn't realize it was a library. I'll use that instead.

src/handlers/update/index.tsx
jariy17: After using the semver library, we should shift the functionality here now.

src/handlers/update/index.tsx
jariy17: Because we are already streaming the error to stderr on line 15 but ill just pass the output into a regular CLI Error.

AlexanderRichey (approved):

Hweinstock:
> CI note: the only red check is unit-test / Test (macOS), which fails solely on src/io/watchFile.test.ts → "debounces a burst of edits into a single callback and stops on abort" — a timing-sensitive debounce test unrelated to this PR (not in the diff; Linux + Windows run the same test and pass). Everything owned by this change is green: check, unit-test Linux/Windows, e2e, all builds, CodeQL, Analyze.

known flaky test: https://github.com/aws/agentcore-cli/issues/2140 :(

src/handlers/update/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  if (!response.ok) {
+    throw new NetworkingError(`Failed to fetch latest version: ${response.statusText}`);
+  }
+  const data = (await response.json()) as { version: string };
+  return data.version;
+}
+
+export type UpdateStatus = "up-to-date" | "newer-local" | "update-available" | "updated";
+
+export interface UpdateResult {
```
Hweinstock, line 39: should this be a type since it represents a concrete grouping of data?

src/handlers/update/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  } catch (cause) {
+    throw new NetworkingError(
+      `Could not reach the npm registry: ${cause instanceof Error ? cause.message : String(cause)}`,
+      { cause },
+    );
+  }
+  if (!response.ok) {
+    throw new NetworkingError(`Failed to fetch latest version: ${response.statusText}`);
+  }
+  const data = (await response.json()) as { version: string };
```
Hweinstock, line 33: should these two lines lines also be wrapped in networking error? My understanding is that fetch doesn't consume the body until the `.json` call.

src/handlers/update/update.test.ts
```diff
@@ -0,0 +1,93 @@
...
+
+// No golden/fixture tests here: the repo's *.fixture.test.tsx harness records and
+// replays AWS SDK responses through CoreClient, but `update` makes no AWS calls —
+// it queries the npm registry (fetch) and shells out to `npm install -g`
+// (runProcess). There is nothing for that harness to record, so a fetch spy plus
+// an injected fake runner is the right, hermetic way to cover this command.
+
+describe("fetchLatestVersion", () => {
+  afterEach(() => {
+    spyOn(globalThis, "fetch").mockRestore();
```
Hweinstock, line 14: could we avoid the direct mocking by making the http fetch injectable?

Hweinstock (approved):

src/handlers/update/index.tsx
jariy17: Yes, ill fix this in a follow up

src/handlers/update/index.tsx
jariy17: Yes I'll fix this in a follow up

src/handlers/update/update.test.ts
jariy17: Yeah, I'll look into doing that in a follow up pr.

PR #2156, opened 2026-09-01
feat(telemetry): print a post-command notice on first run

Hweinstock:
### Problem 
We are missing the post-first run telemetry notice on the refactor branch. 

### Solution 
- avoid writing the installationId as a side effect within globalConfig. 
- at the top level, determine if its the firstRun by checking if installationId is there, otherwise attempt to persist one. 
- print a telemetry notice when installationId was not there, and telemetry is enabled

### Verification 
```
> rm ~/.agentcore/config.json
> agentcore config 
{
  "telemetry": {
    "enabled": true,
    "audit": false,
    "endpoint": "https://telemetry.agentcore.aws.dev"
  },
  "installationId": "e92dbb9b-cd4f-4b9f-94cc-ef69d95087cb"
}

The AgentCore CLI collects aggregated, anonymous usage analytics to help improve the tool.
To opt out:   agentcore config telemetry.enabled false
To audit:     agentcore config telemetry.audit true
> agentcore config 
{
  "telemetry": {
    "enabled": true,
    "audit": false,
    "endpoint": "https://telemetry.agentcore.aws.dev"
  },
  "installationId": "e92dbb9b-cd4f-4b9f-94cc-ef69d95087cb"
}
```
notice that it only printed the first time.

### DIFF
--- a/src/globalConfig/accessor.tsx
+++ b/src/globalConfig/accessor.tsx
@@ -45,8 +45,10 @@ export class DefaultGlobalConfigAccessor implements GlobalConfigAccessor {
 
     const configFileData = await this.readConfigFile();
 
-    // if no installationId is present, generate one and merge it into the file data
-    if (!configFileData.installationId) {
+    // a run with no persisted installationId is the first run on this machine
+    const isFirstRun = !configFileData.installationId;
+
+    if (isFirstRun) {
       configFileData.installationId = DEFAULT_GLOBAL_CONFIG.installationId;
       this.logger.info(`no installationId found, persisting one`);
 
@@ -61,7 +63,7 @@ export class DefaultGlobalConfigAccessor implements GlobalConfigAccessor {
       }
     }
 
-    this.cachedConfig = applyOverrides(DEFAULT_GLOBAL_CONFIG, configFileData);
+    this.cachedConfig = { ...applyOverrides(DEFAULT_GLOBAL_CONFIG, configFileData), isFirstRun };
     return this.cachedConfig;
   }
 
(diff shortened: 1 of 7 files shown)


src/globalConfig/accessor.tsx
```diff
@@ -45,22 +45,6 @@ export class DefaultGlobalConfigAccessor implements GlobalConfigAccessor {
 
     const configFileData = await this.readConfigFile();
 
-    // if no installationId is present, generate one and merge it into the file data
-    if (!configFileData.installationId) {
```
AlexanderRichey, line 49: Would it be simpler to return `isFirstRun` in the output here? Would that avoid having to make many of these other changes?

Hweinstock: I considered this approach, (and actually implemented it initially), but I didn't like the behavior of `get` writing to the config as a side effect since a caller would expect a strict-read here. 

However, I can see the argument that from the perspective of the caller, it is a strict-read still.

Let me swap to the simpler approach and if we see issues with the side-effect, we can revisit.

src/telemetry/notice.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Writes the telemetry-collection notice to the given stream on the first run of
+ * the CLI, unless telemetry is already disabled.
+ */
+export function printFirstRunNotice(
```
jariy17, line 5: Can we shift this function under the handler file since it's only used by the handler?

jariy17 (approved):

src/telemetry/notice.ts
Hweinstock: I was hoping to keep the root level entrypoint clean, and splitting it out made it easier/more natural to add unit tests, but I agree exporting this from telemetry module for a single consumer adds some indirection that shouldn't be necessary.

Hweinstock:
Merging since @AlexanderRichey comment is addressed and received another review. Can address any remaining comments as follow-ups.

PR #2158, opened 2026-09-01
feat(project): resolve all deployed resource types (prereq for `project status`)

jariy17:


jariy17:
## B11 live AWS e2e: **PASSED** (was previously listed as skipped)

Ran a real deploy in the EXPLORE sandbox (account `725476964917`, `us-west-2`) and exercised the widened resolver **through the real CLI** — not a scratch script — so the full DI + middleware chain (`withProject` → `ProjectKey` → `JsonRendererKey` → region resolution) is proven, which is exactly the call path `project status` will use.

### Method
Throwaway project `StatusE2E329bced7` with 5 deployed resources across 4 types, including an **underscore-bearing name** (`extra_mem`, `web_agent`) to exercise the sanitization path against the *real* `exportName`. The `status` stub was temporarily wired the way Plan 1 will wire it (factory takes `projectManager`, wrapped in `withProject`, reads `ProjectKey`, calls `resolveDeployedResources(project, { target, allowMissing: true })`, emits via `renderJson`) purely to drive the e2e. **That wiring was reverted — this PR stays scoped to Plan 0.**

### Real `agentcore project status --json` output
```json
{
  "projectName": "StatusE2E329bced7",
  "target": "default",
  "region": "us-west-2",
  "resources": [
    { "resourceType": "runtime",       "name": "web_agent",       "id": "StatusE2E329bced7_web_agent-YarMvkE6t3" },
    { "resourceType": "memory",        "name": "web_agentMemory", "id": "StatusE2E329bced7_web_agentMemory-rhPmfS5lun" },
    { "resourceType": "memory",        "name": "extra_mem",       "id": "StatusE2E329bced7_extra_mem-jnmCef3X4b" },
    { "resourceType": "gateway",       "name": "toolsgw",         "id": "statuse2e329bced7-toolsgw-atj9vtqicf" },
    { "resourceType": "policy-engine", "name": "polengine",       "id": "StatusE2E329bced7_polengine-6i3tmjl0q0" }
  ]
}
```

### Real stack Outputs vs resolver (per type)
Actual `describe-stacks` export names from the live stack — every one matched the `EXPORT_PARTS` label:

| Type | Real `ExportName` | Result |
|---|---|---|
| runtime | `AgentCore-StatusE2E329bced7-default-web-agent-RuntimeId` | ✅ |
| memory | `…-default-Memory-web-agentMemory-Id` | ✅ |
| memory (underscore) | `…-default-Memory-extra-mem-Id` | ✅ sanitization confirmed (`extra_mem` → `extra-mem`) |
| gateway | `…-default-Gateway-toolsgw-Id` | ✅ |
| policy-engine | `…-default-PolicyEngine-polengine-Id` | ✅ |

**5/5 resolved, 0 mismatches, nothing silently missing.** Ids returned by the CLI are byte-identical to the stack `OutputValue`s. The stack also carries `-Arn`/`-Url`/`-RoleArn` exports plus `StackNameOutput` (the only output with no `ExportName`) — correctly ignored.

**No bug found**, so no code change was needed from this run. The one deploy failure hit along the way was my own test fixture (a Cedar policy with a wildcard resource → `ValidationException`, then "Overly Permissive" `NotStabilized`), not the resolver; I dropped the policy and kept the policy-engine.

### Not covered live
`credential` / `payment` / `gateway-target` / `policy` / `knowledge-base` / `config-bundle` / `dataset` need external setup (OAuth providers, payment connectors, MCP endpoints, S3 sources) — they remain covered by the unit fixture, which asserts the same `ExportName`/`OutputKey`/state-map paths.

### Teardown (confirmed)
- `delete-stack` on `AgentCore-StatusE2E329bced7-default` → waiter completed; `describe-stacks` now returns `ValidationError: Stack ... does not exist`.
- `list-stacks` filtered on `StatusE2E` → no remaining non-`DELETE_COMPLETE` stacks.
- Temp project dir `/tmp/b11/StatusE2E329bced7` and the probe script deleted.
- Scratch `status` handler wiring reverted; branch diff contains only the 6 Plan 0 files. `tsc` clean, tests green after revert.

jariy17:
## Live-proof round 2: all 11 in-scope resource types now verified against real AWS

Follow-up to the [first e2e comment](https://github.com/aws/agentcore-cli/pull/2158#issuecomment-5498142476). Fixtures were not accepted as proof — a fixture I wrote encodes the same export-name assumption as the code I wrote, so it passes even when the assumption is wrong. Only a real stack can falsify it. So every type below was deployed for real in EXPLORE (`725476964917`, `us-west-2`) and read back through the **real CLI** (`agentcore project status --json`).

Also in this push: **`dataset` was removed from scope entirely** (commit `eb011115`) — dropped from the `DeployableResource` union, the `EXPORT_PARTS` table, the spec iteration, and the resolver test. The `satisfies Record<CfnOutputResource, …>` guard confirmed the union and table stayed in sync, which is exactly what it's for. (Corroborating signal: there is no `agentcore project add dataset` command.)

### Evidence table — 11 types, all live-proven

| # | Type | Path | Real `ExportName` / `OutputKey` | Real id returned | Result |
|---|---|---|---|---|---|
| 1 | runtime | ExportName | `…-default-web-agent-RuntimeId` | `StatusE2E405fab58_web_agent-0sy5bRGDa6` | ✅ |
| 2 | memory | ExportName | `…-default-Memory-web-agentMemory-Id` | `StatusE2E405fab58_web_agentMemory-GhrTKN7egi` | ✅ |
| 3 | memory (underscore) | ExportName | `…-default-Memory-extra-mem-Id` | `StatusE2E329bced7_extra_mem-jnmCef3X4b` | ✅ sanitization |
| 4 | knowledge-base | ExportName | `…-default-KnowledgeBase-kbdocs-Id` | `0FAGVD5SIP` | ✅ |
| 5 | **credential** | **deployed-state map** | *(not a CFN output — see note)* | `arn:aws:bedrock-agentcore:…:token-vault/default/apikeycredentialprovider/apikey_cred` | ✅ |
| 6 | gateway | ExportName | `…-default-Gateway-toolsgw-Id` | `statuse2e405fab58-toolsgw-eblile4r6g` | ✅ |
| 7 | **gateway-target** (nested) | ExportName + `parent` | `…-default-GatewayTarget-rtgt-Id` | `VRXCW7ECAW` (`parent: "toolsgw"`) | ✅ |
| 8 | policy-engine | ExportName | `…-default-PolicyEngine-polengine-Id` | `StatusE2E405fab58_polengine-12lvzke39n` | ✅ |
| 9 | **policy** (nested) | ExportName + `parent` | `…-default-Policy-polengine-scopedpol-Id` | `scopedpol-t7m7et_45d` (`parent: "polengine"`) | ✅ |
| 10 | config-bundle | ExportName | `…-default-ConfigBundle-cfgbundle-Id` | `StatusE2E405fab58cfgbundle-Rzxj0z37BS` | ✅ |
| 11 | **payment** | **`OutputKey`** (no ExportName) | `PaymentpaymgrManagerId` | `paymgr-p0nesptouc` | ✅ |
| — | capacity-provider | ExportName | *no CDK output yet* — `aws/agentcore-l3-cdk-constructs#336` | — | ⏳ label in place |
| — | ~~dataset~~ | — | **removed from scope** | — | n/a |

**10/10 resolved in one call, 0 mismatches.** Both custom (non-ExportName) paths and both nested-`parent` paths are live-proven. **No bug found, so no resolver fix was needed** from this round.

### Real `agentcore project status --json` (round-2 stack, 10 resources)
```json
{
  "projectName": "StatusE2E405fab58",
  "target": "default",
  "region": "us-west-2",
  "resources": [
    { "resourceType": "runtime",        "name": "web_agent",       "id": "StatusE2E405fab58_web_agent-0sy5bRGDa6" },
    { "resourceType": "memory",         "name": "web_agentMemory", "id": "StatusE2E405fab58_web_agentMemory-GhrTKN7egi" },
    { "resourceType": "knowledge-base", "name": "kbdocs",          "id": "0FAGVD5SIP" },
    { "resourceType": "credential",     "name": "apikey_cred",     "id": "arn:aws:bedrock-agentcore:us-west-2:725476964917:token-vault/default/apikeycredentialprovider/apikey_cred" },
    { "resourceType": "gateway",        "name": "toolsgw",         "id": "statuse2e405fab58-toolsgw-eblile4r6g" },
    { "resourceType": "gateway-target", "name": "rtgt", "parent": "toolsgw",   "id": "VRXCW7ECAW" },
    { "resourceType": "policy-engine",  "name": "polengine",       "id": "StatusE2E405fab58_polengine-12lvzke39n" },
    { "resourceType": "policy",         "name": "scopedpol", "parent": "polengine", "id": "scopedpol-t7m7et_45d" },
    { "resourceType": "config-bundle",  "name": "cfgbundle",       "id": "StatusE2E405fab58cfgbundle-Rzxj0z37BS" },
    { "resourceType": "payment",        "name": "paymgr",          "id": "paymgr-p0nesptouc" }
  ]
}
```

### Note on `credential` — a real gap found in the refactor branch (pre-existing, not from this PR)
The credential was declared in the spec and `project deploy` succeeded, but the credential did **not** appear in status, and `list-api-key-credential-providers` confirmed **no provider was ever created**. Root cause: on this branch the only `updateTargetState` call writes `{ stackArn }` (`cdk.ts:255`) — **nothing in the deploy path provisions spec-declared credentials or records them under `resources.credentials`**, even though the state schema models that map and the scaffolded CDK app reads it (`assets/cdk/bin/cdk.ts:131`). The resolver behaved correctly: no id ⇒ omitted.

To prove the read path against real data rather than a fixture, I created a **real** API-key credential provider via `agentcore identity api-key-credential-provider create` and recorded its real ARN in `deployed-state.json` (the same map the released CLI writes). Status then returned it correctly. So the branch's credential branch is verified against a real AWS ARN; the only synthetic step is *who wrote the map*, because on this branch nobody does yet. **Worth a separate issue** — `project deploy` should provision declared credentials.

### Fixture-vs-reality corrections this round caught
My earlier unit fixture guessed `Policy-<engine>-<name>-Id` and `GatewayTarget-<name>-Id`; the real stack confirms both. It also guessed `Dataset-<name>-Id`, now moot. The real `payment` output has **no** `ExportName` at all, confirming the `OutputKey` branch is required rather than a nicety.

### Getting a policy to deploy
My first two attempts failed on real service validation (worth recording, since the CDK can't catch these):
- `permit(principal, action, resource);` → `ValidationException: a wildcard resource was detected…`
- `permit(principal, action, resource is AgentCore::Gateway);` → `NotStabilized: Overly Permissive…`
- `…action == AgentCore::Action::"InvokeTool", resource is AgentCore::Gateway` → `ValidationException: a constrained action scope was encountered, please constrain the resource to a specific AgentCore::Gateway resource`

Accepted form (validated in seconds against `create-policy` directly, then deployed): `permit(principal == AgentCore::IamEntity::"arn:aws:iam::…:role/Admin", action, resource is AgentCore::Gateway);` → reached `ACTIVE`.

### Teardown (enumerated, all confirmed)
- CloudFormation stacks `AgentCore-StatusE2E329bced7-default` and `AgentCore-StatusE2E405fab58-default` → deleted, waiters completed, `describe-stacks` returns `ValidationError: … does not exist`.
- Credential provider `apikey_cred` → deleted (`list-api-key-credential-providers` now empty for it).
- Probe policy `p2-mqdn7mqbkz` + policy engine `statuse2e_probe_engine-ch31xhf14w` (used for fast Cedar validation) → deleted.
- S3 bucket `statuse2e-kb-405fab58` → emptied and deleted.
- Temp dirs `/tmp/b11`, `/tmp/b12` and the dummy API-key file → deleted.
- Sweeps for leftovers (stacks / buckets / credential providers / policy engines matching `StatusE2E`) → all empty.

### Branch hygiene
Scratch `status` stub wiring reverted (the stub is back to `NotImplementedError`); branch diff is the **6 Plan 0 files only**; `tsc --noEmit` clean; 746 tests pass in the affected suites. One unrelated flake seen in a full run (`src/io/exec.test.ts` `streamProcess` 5s timeout under load) passes 17/17 in isolation and is untouched by this PR.

jariy17:
## Rebased onto #2160 + folded id resolution into one exhaustive switch

Rebased onto `refactor` @ `8dc1051c` (the #2160 merge, which consolidated `DeployedProjectResource` + `ResolvedDeployedResource` into a single `target`-bearing type). Head is now `4d647515`, 3 commits.

### Rebase resolution

Two conflicted files, 5 hunks:

| File | Resolution |
|---|---|
| `handlers/project/types.ts` | deleted `DeployedProjectResource`; moved its `resourceType: DeployableResource` + `parent?` onto the merged `ResolvedDeployedResource`, keeping `name` + `target`. **`ResolveDeployedResourceInput.resourceType` deliberately stays `ProjectInvokableResource`** — only the *returned* type widens, so `invoke` still cannot ask for a gateway's id. |
| `backends/cdk.ts` (4 hunks) | imports, the `findDeployedResourceId` param, the return type + `allowMissing` destructure, and the `flatMap` (now emits `{ ...r, id, target }`). |

`backends/types.ts` and `invoke/screen.tsx` auto-merged cleanly — `screen.tsx`'s `.filter()` narrowing to runtime/harness is what keeps the widened type compiling in the picker.

### ⚠️ The one edit git got silently wrong

`backends/cdk.test.ts` auto-merged with **no conflict marker** and was broken: the new `"resolves every deployed resource type"` assertion had 14 rows with no `target` key, which the merged type now requires. Confirmed by typecheck before fixing — **15 × `TS2769` at `cdk.test.ts:760`** — then `target: TARGET` added to every row. Nothing but `typecheck` catches this, so it is worth knowing for any other branch rebasing over #2160.

### Refactor: `EXPORT_PARTS` → `resolveResourceId`

`findDeployedResourceId` only understood CloudFormation exports, so its two exceptions leaked outward: `payment` matched by `OutputKey` in an early return, and `credential` — never a stack output at all — was branched on by the caller. "Where does this id come from" lived in three places.

Now one switch over all 14 types, taking both sources (`stack` + deployed-state `credentialArns`). Deleted: `EXPORT_PARTS`, the `CfnOutputResource` type, the `Exclude<DeployableResource, "credential">` param, the `if (payment)` early return, and the caller's ternary. The caller is one line.

Renamed off `find*` because it no longer only searches the stack.

**The `never` default was verified to bite:** adding a 15th `DeployableResource` member produced `cdk.ts(127,13): error TS2322: Type '"sentinel-probe"' is not assignable to type 'never'`, then reverted. So a new resource type is a compile error rather than one that silently vanishes from `project status`.

Export-name literals are **unchanged** — each was verified against a real CloudFormation stack, so they are deliberately not derived from the `resourceType` via PascalCase; a wrong label fails silently.

### Gates

`tsc --noEmit` clean · `bun test` **2603 pass / 0 fail** (188 files, 12,981 expects) · `oxlint` clean · `prettier --check` clean.

One unrelated flake surfaced mid-run (`core/dev/inspector/invocations.test.ts`, A2A 502-vs-403) — passes 26/26 in isolation across 3 runs, has zero references to this change, and did not recur on the final full run.

jariy17:
## Inlined deployed-id resolution into its only caller

`d68ac22f` — follow-up to the exhaustive-switch commit, pure refactor, no behavior change.

`resolveResourceId` was module-level with **one** call site, and it took a two-source parameter object:

```ts
function resolveResourceId(
  sources: { stack: Stack; credentialArns: Record<string, { credentialProviderArn: string }> },
  input: { resourceType: DeployableResource; name: string; parent?: string },
): string | undefined
```

Both of those sources were already locals inside `resolveDeployedResources`, so the parameter object existed only to carry them back out and in again. It is now a local `idOf` (plus `byExportName`) closing over `stack` and `credentialArns` directly — nothing to pass, nothing to thread.

**Net −12 lines** in `cdk.ts` (422 → 410): the `sources` type, the function boilerplate, and the argument at the call site all go away. The call site reduces to `const id = idOf(r);`.

Also dropped the now-unused `import type { Stack } from "@aws-sdk/client-cloudformation"` — its only use was that deleted signature.

### Unchanged on purpose
- Every `ExportName` string literal is byte-identical; each was verified against a real CloudFormation stack, so they are not derived from `resourceType` (a wrong label fails *silently*).
- The `never` default stays. **Re-verified it still bites after the inline**: adding a 15th member to `DeployableResource` produced `cdk.ts(344,17): error TS2322: Type '"sentinel-probe"' is not assignable to type 'never'`, then reverted clean.

### Gates
`tsc --noEmit` exit 0 · `bun test` **2603 pass / 0 fail** (188 files, 12,981 expects) · `oxlint` exit 0 · `prettier --check` clean.

jariy17:
## `identifier` now resolves ARNs, not bare ids

`project status` reports these to customers, where the ARN is the useful identifier. Switched every case in `idOf` from the CDK's `-Id` export to its `-Arn` export:

| type | export |
|---|---|
| runtime | `<stack>-<name>-RuntimeArn` |
| harness / memory / knowledge-base / evaluator / online-eval / gateway / policy-engine / config-bundle / capacity-provider | `<stack>-<Type>-<name>-Arn` |
| policy | `<stack>-Policy-<engine>-<name>-Arn` |
| payment | `OutputKey` `Payment<Mgr>ManagerArn` (verified present at `assets/cdk/lib/cdk-stack.ts:228`, alongside the `…ManagerId` it replaces) |
| credential | unchanged — `credentialProviderArn` from deployed-state was already an ARN |

### One exception: `gateway-target`

`AgentCoreMcp` emits `GatewayTarget-<name>-Id` and **no** `-Arn`, so this is the only resource still reported by id. Marked with a `TODO(cdk)` — once the construct exports an Arn, that case joins the others.

Verified against the installed `@aws/agentcore-cdk` dist by enumerating every `exportName(stack.stackName, …)` call; all the labels above exist, and `GatewayTarget` genuinely has only `Id`.

Fixtures in `cdk.test.ts` now carry `-Arn` export names and assert ARN-shaped values, with `gateway-target` asserting a bare id so the exception is pinned rather than assumed.

`tsc` clean · `bun test` 2603 pass / 0 fail · oxlint clean · prettier clean.

jariy17:
Closing in favour of #2165, which contains every commit from this branch (`f52730c4` is an ancestor of `aea9cf1c`). Merging #2165 lands both the widened `resolveDeployedResources` and `project status` together.

PR #2160, opened 2026-09-01
refactor(project): merge DeployedProjectResource into ResolvedDeployedResource

jariy17:
## Summary

`DeployedProjectResource` and `ResolvedDeployedResource`  share the same responsibility which is described the state of the deployed resource. Therefore, I have merged them.  No behavior change should occur. 

## Verification

- `tsc --noEmit` — 0 errors
- `bun test` — **2598 pass / 0 fail** (188 files)
- `oxlint` — clean


jariy17:
## Bug bash: one real defect found and fixed (`2998f2f8`)

Two independent planners — a devil's-advocate pass and a Munger-inversion pass — were run over this diff. They converged on **the same single defect** (filed as INV-1 / B1 respectively). Everything else they raised was either already correct, inert, or a note for later.

### The defect: target authority silently moved from the manager to the backend

The merge changed `resolveDeployedResource` from

```ts
return { id: resource.id, target: resolved.target };   // manager-authoritative
```

to `return resource;` — which hands back **whatever target the backend stamped on the item**. That is not a no-op: `resolveExistingTarget` resolves the target from `aws-targets.json`, and both invoke handlers pin the AWS region off this value (`invoke/runtime.tsx`, `invoke/harness.tsx`: `ctx.withValue(RegionKey, deployed.target.region)`). A backend echoing a stale or wrong target could therefore redirect a customer's invoke to the wrong region — **impossible before this PR**, because the manager's own resolved target always won.

Latent rather than live today (`CdkBackend` stamps the same object it was handed), but the merge created the degree of freedom, and neither the compiler nor the suite checked it.

**Fix** — the declared target overrides the item's copy:

```ts
if (resource) return { ...resource, target: resolved.target };
```

### Doc comments deleted rather than reworded

Three comments misstated ownership and have been removed:

- the two blocks above `ResolvedDeployedResource` / `ResolvedDeployedResources` — one claimed a caller holding a single item needs no envelope, but the envelope still exists and `invoke/screen.tsx` is what actually reads it; the other restated what the types already show;
- the `cdk.ts` comment claiming *"The backend already resolved `target`"* — the **manager** resolved it and passed it in.

### New test coverage

`manager.test.ts` had **no `describe` for either method** — the changed line was covered only transitively. Added `FsProjectManager.resolveDeployedResource(s)` with five cases:

| Case | Asserts |
|---|---|
| divergence guard | backend claims `us-east-1` while `aws-targets.json` declares `eu-west-1` → singular returns `eu-west-1` |
| identity | a runtime and a harness sharing one `name` resolve to their own ids; `resourceType`/`name` echo the request |
| empty stack | zero resources still yields `{ resources: [], target }` |
| throw path | `ProjectStateError` matching `/is not deployed to target 'default'.*project deploy --target default/` |
| invariant | every item's `target` is the same reference the envelope reports |

**The divergence guard was verified in both directions: it FAILS on `42138bcb` and PASSES with the fix.**

### Mutation checks (each reverted)

| # | Mutation | Result |
|---|---|---|
| M1 | `manager.tsx` → `return resolved.resources[0]!` | 🔴 RED (2 fail) |
| M2 | `cdk.ts` → drop `target` from the item | 🔴 RED — `tsc` TS2322 **and** tests; proves `target` is structurally required |
| M3 | `cdk.ts` → stamp `region: "eu-central-1"` | 🔴 RED (`cdk.test.ts`) |
| M4 | `invoke/index.test.tsx` stub → drop `target: input.target` | 🔴 `tsc` TS2322 |

M3 is worth a note: my first attempt mutated to `us-east-1`, which is **exactly** `cdk.test.ts`'s `TARGET.region`, so the mutant object was identical to the original and the suite stayed green. That was an invalid mutation, not a coverage gap — re-running with `eu-central-1` turns it red. Flagging it because a fixture whose value collides with a mutant is an easy way to fool yourself into trusting a green run.

### Gates

`tsc --noEmit` clean · `bun test` **2603 pass / 0 fail** · `oxlint` clean · `prettier --check` clean.

### Still outstanding

Live-AWS (`agentcore project deploy`) and TUI-picker evidence are being produced separately against this commit and will be posted in a follow-up comment. Staying in **draft** until then.

Hweinstock (approved):
thanks for cleaning this up!

PR #2163, opened 2026-09-01
feat(project): live step-based progress UI for deploy and build

AlexanderRichey:
## Summary

Replaces the flat line-per-message stderr output of `project deploy` and `project build` with a live, inline Ink step list: a spinner on the running step, a ~5-line scrolling tail of that step's real output (synth output, CDK Toolkit per-resource CloudFormation events) behind a muted `│` gutter, and ✓/✕ as steps finish. Renders in normal scrollback — no alternate screen — so the completed step list and success line remain in the terminal afterwards. Non-TTY and `--json` invocations keep plain line-per-step output, and stdout stays machine-clean: nothing is written to it on a plain deploy. With `--json`, both commands always emit one JSON document on stdout — `{ message, ... }` on success (deploy includes its outputs and `tornDown`) and `{ error }` on failure, with exit codes unchanged.

```
✓ Verifying AWS account 111122223333
✓ Synthesizing CloudFormation templates
⠸ Deploying AgentCore-orders-default-0
  │ AgentCore-orders-default-0 | 4/12 | CREATE_IN_PROGRESS | AWS::IAM::Role | RuntimeRole
  │ AgentCore-orders-default-0 | 5/12 | CREATE_COMPLETE    | AWS::IAM::Role | RuntimeRole
```

Built as reusable layers so other long-running commands can adopt the same treatment:

- **Structured events** — `ProjectEvent` becomes a discriminated union of `step` and `output` events (a step completes when the next begins; the last when the generator returns). All producers (manager, CDK backend) and consumers (add/create/export handlers, create wizard) updated.
- **Output plumbing** — `AsyncChannel` + `createLineSplitter` (`src/io/channel.ts`) and `withOutputEvents` (`src/core/project/events.ts`) bridge push-style output callbacks into the pull-based event generators. Synth chunks are line-buffered into `output` events; `createCdkRunner` builds a per-operation `IIoHost` that both debug-logs (unchanged) and feeds the operation's `onOutput` sink. The backend keeps the last 20 Toolkit lines and appends them to terse Toolkit errors, so failures are debuggable from the terminal alone; `ProcessFailedError` still carries full subprocess output untruncated.
- **Component + driver** — `TaskList` (`src/components/ui/task-list`) renders tasks with the shared Spinner/✓/✕ vocabulary; `runWithProgress` (`src/tui/progress.tsx`) mounts it inline on stderr, resolves with the generator's return value, and on failure marks the step ✕, keeps its tail in scrollback, and rethrows unchanged for the existing exit-code handling.
- **Teardown confirmation moves in front of the generator** — the deploy handler prompts up front when the spec declares none of the resources `removeAllResources` clears, using a new read-only `ProjectManager.resolveTarget` to name the target's account and region. The decision reaches the backend pre-resolved, so nothing downstream blocks on interactive input; the backend's post-synth zero-resource count remains the non-interactive backstop.

## Testing

- `bun test` (2,624 pass), `bun run typecheck`, `bun run lint:check`, `bun run format:check` all clean. New coverage: channel/splitter, event bridge, `TaskList` (ink-testing-library), driver (interactive + plain + error paths), backend output plumbing, and the resequenced confirmation flows.
- Eyeballed under a pty: paced demo, real `project build` (success and a sabotaged failing build showing ✕ + full untruncated error), and a full live AWS round trip — fresh `project create` → `deploy` to us-east-2 streaming 2,700+ genuine CloudFormation resource events through bootstrap and deploy, then `remove all --yes` + `deploy --yes` rendering the teardown and deleting the stack.

https://claude.ai/code/session_017KhMu3YVxNSVCst1LU877y

### DIFF
--- a/src/handlers/project/add/gateway/index.ts
+++ b/src/handlers/project/add/gateway/index.ts
@@ -129,7 +129,7 @@ export const createAddGatewayHandler = (config: AddProjectResourceConfig) =>
         resourceType: "gateway",
         resourceConfig: gateway,
       })) {
-        config.io.stderr.write(`${event.message}\n`);
+        if (event.type === "step") config.io.stderr.write(`${event.message}\n`);
       }
       config.io.stderr.write(`added Gateway '${flags.name}' to '${project.name}'\n`);
     },
--- a/src/handlers/utils.tsx
+++ b/src/handlers/utils.tsx
@@ -1,9 +1,10 @@
 import type { Context } from "../router";
 import type z from "zod";
 import type { CoreOptions } from "../core/types";
-import { InputValidationError } from "../errors";
+import { AgentCoreCLIError, InputValidationError, SilentCLIError } from "../errors";
 import { formatZodError } from "../router/schema";
 import { EndpointKey, RegionKey } from "./keys";
+import { JsonRendererKey } from "../tui";
 
 // coreOptsFromCtx builds the standard CoreOptions handed to Core operations from
 // the values pinned on the context: the resolved region (always present, see the
@@ -122,4 +123,15 @@ export function parseTags(values: string[] | undefined): Record<string, string> 
     result[entry.slice(0, eqIndex)] = entry.slice(eqIndex + 1);
   }
   return result;
+}
+
+// renderJsonError reports a command failure as a JSON document in --json mode,
+// so scripted callers can read the outcome from stdout instead of parsing the
+// human-oriented `Error: ...` line the exit-code handler prints to stderr.
+// Callers rethrow afterwards; exit codes are untouched. Errors marked silent
+// (e.g. a user cancellation) stay silent here too.
+export function renderJsonError(ctx: Context, error: unknown): void {
+  const cliError = AgentCoreCLIError.fromError(error);
+  if (cliError instanceof SilentCLIError) return;
+  ctx.require(JsonRendererKey).renderJson({ error: cliError.message });
 }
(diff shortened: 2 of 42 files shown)


notgitika (approved):
Looks good to me!

src/handlers/utils.tsx
```diff
@@ -123,3 +124,14 @@ export function parseTags(values: string[] | undefined): Record<string, string>
   }
   return result;
 }
+
+// renderJsonError reports a command failure as a JSON document in --json mode,
+// so scripted callers can read the outcome from stdout instead of parsing the
+// human-oriented `Error: ...` line the exit-code handler prints to stderr.
+// Callers rethrow afterwards; exit codes are untouched. Errors marked silent
+// (e.g. a user cancellation) stay silent here too.
+export function renderJsonError(ctx: Context, error: unknown): void {
```
notgitika, line 133: does this cover middleware failures? `withProject` can fail before the build/deploy handlers call this helper, leaving  the --json output empty. i would think it's fine but we could bring it further up in the pipeline if we want those covered

jariy17 (commented):
Pretty clean and minimal code change that introduces a gorgeous UI

src/handlers/project/add/gateway/index.ts
```diff
@@ -129,7 +129,7 @@ export const createAddGatewayHandler = (config: AddProjectResourceConfig) =>
         resourceType: "gateway",
         resourceConfig: gateway,
       })) {
-        config.io.stderr.write(`${event.message}\n`);
+        if (event.type === "step") config.io.stderr.write(`${event.message}\n`);
```
jariy17, line 132: Nit: If ProjectManager.addResource emits output events, they are silently dropped instead of being written to stderr.

PR #2165, opened 2026-09-02
feat(project): implement `project status`

jariy17:
The status handler calls `ProjectManager.resolveProjectResources` to build its report.

`resolveProjectResources` returns
every resource the project declares, by ARN, reading `CfnOutput` values from `DescribeStacks`.
A resource the target's stack has not deployed is reported `local-only` rather than omitted,
and children such as gateway targets and policies are nested under the resource that owns them.

Given an `agentcore.json` like this:

```json
{
  "name": "orders",
  "version": 1,
  "runtimes": [{ "name": "assistant" }],
  "memories": [{ "name": "user_mem" }],
  "agentCoreGateways": [
    { "name": "tools", "targets": [{ "name": "search" }] }
  ],
  "policyEngines": [
    { "name": "guards", "policies": [{ "name": "no_pii" }] }
  ]
}
```

`agentcore project status` produces:

```json
{
  "projectName": "orders",
  "target": "prod",
  "region": "us-west-2",
  "resources": [
    {
      "resourceType": "runtime",
      "name": "assistant",
      "deploymentState": "deployed",
      "id": "arn:aws:bedrock-agentcore:us-west-2:111122223333:runtime/assistant-AbCdEf1234"
    },
    {
      "resourceType": "memory",
      "name": "user_mem",
      "deploymentState": "deployed",
      "id": "arn:aws:bedrock-agentcore:us-west-2:111122223333:memory/user_mem-K3l4M5n6O7"
    },
    {
      "resourceType": "gateway",
      "name": "tools",
      "deploymentState": "local-only",
      "children": [
        {
          "resourceType": "gateway-target",
          "name": "search",
          "deploymentState": "local-only"
        }
      ]
    },
    {
      "resourceType": "policy-engine",
      "name": "guards",
      "deploymentState": "deployed",
      "id": "arn:aws:bedrock-agentcore:us-west-2:111122223333:policy-engine/guards-1a2b3c",
      "children": [
        {
          "resourceType": "policy",
          "name": "no_pii",
          "deploymentState": "deployed",
          "id": "arn:aws:bedrock-agentcore:us-west-2:111122223333:policy/no_pii-7d8e9f"
        }
      ]
    }
  ]
}
```

If the target isn't declared in `aws-targets.json`, it throws `ProjectStateError`:
`No deployment targets are configured for project 'orders'.`

### DIFF
--- a/src/core/project/backends/cdk.test.ts
+++ b/src/core/project/backends/cdk.test.ts
@@ -16,6 +16,7 @@ import type {
 } from "./cdk/credentials";
 import { DEPLOYED_STATE_RELATIVE_PATH, updateTargetState } from "./cdk/deployedState";
 import type { DeployBackendInput } from "./types";
+import type { ResolvedProjectResource } from "../../../handlers/project/types";
 import type { BootstrapState } from "./cdk/environment";
 import type { CdkCredentialProvider, CdkOperation, CdkOutputs, CdkRunOptions } from "./cdk/toolkit";
 
@@ -931,5 +932,149 @@ describe("CdkBackend.resolveDeployedResources", () => {
       subject.backend.resolveDeployedResources(input, { target: TARGET }),
     ).rejects.toThrow(/expects AWS account 111122223333.*999900001111/s);
     expect(subject.stackReads).toEqual([]);
+  });
+});
+
+describe("CdkBackend.resolveProjectResources", () => {
+  const out = (ExportName: string, OutputValue: string) => ({ ExportName, OutputValue });
+  const key = (OutputKey: string, OutputValue: string) => ({ OutputKey, OutputValue });
+  const S = "AgentCore-example-default";
+
+  test("resolves every declared type: exports, payment OutputKeys, credential from state, nested parents, underscores", async () => {
+    const input = await project();
+    input.spec = {
+      ...input.spec,
+      runtimes: [{ name: "web" }],
+      harnesses: [{ name: "chat" }],
+      memories: [{ name: "user_mem" }], // an underscore becomes a dash in the export
+      knowledgeBases: [{ name: "kb" }],
+      credentials: [{ name: "cred" }], // read from deployed state, not the stack
+      evaluators: [{ name: "ev" }],
+      onlineEvalConfigs: [{ name: "oe" }],
+      agentCoreGateways: [{ name: "gw", targets: [{ name: "tgt" }] }],
+      policyEngines: [{ name: "pe", policies: [{ name: "pol" }] }],
+      configBundles: [{ name: "cb" }],
+      payments: [{ name: "pay", connectors: [{ name: "wallet_one" }] }],
+    } as unknown as typeof input.spec;
+    await updateTargetState(json, input.rootPath, TARGET.name, {
+      stackArn: STACK_ARN,
+      resources: { credentials: { cred: { credentialProviderArn: "arn:aws:cred/cred" } } },
+    });
+    const subject = harness({
+      describedStack: {
+        StackName: S,
+        CreationTime: new Date(0),
+        StackStatus: "CREATE_COMPLETE",
+        Outputs: [
+          out(`${S}-web-RuntimeArn`, "arn:runtime/web-1"),
+          out(`${S}-Harness-chat-Arn`, "arn:harness/chat-1"),
+          out(`${S}-Memory-user-mem-Arn`, "arn:memory/mem-1"),
+          out(`${S}-KnowledgeBase-kb-Arn`, "arn:kb/kb-1"),
+          out(`${S}-Evaluator-ev-Arn`, "arn:evaluator/ev-1"),
+          out(`${S}-OnlineEval-oe-Arn`, "arn:online-eval/oe-1"),
+          out(`${S}-Gateway-gw-Arn`, "arn:gateway/gw-1"),
+          out(`${S}-GatewayTarget-tgt-Id`, "tgt-1"),
+          out(`${S}-PolicyEngine-pe-Arn`, "arn:policy-engine/pe-1"),
+          out(`${S}-Policy-pe-pol-Arn`, "arn:policy/pol-1"),
+          out(`${S}-ConfigBundle-cb-Arn`, "arn:config-bundle/cb-1"),
+          key("PaymentpayManagerArn", "arn:payment-manager/pay-1"),
+          key("PaymentpaywalletoneConnectorId", "conn-1"),
+        ],
+      },
+    });
+
+    const resources = await subject.backend.resolveProjectResources(input, { target: TARGET });
+
+    // [type, name, arn, [children...]] so a child under the wrong owner fails here
+    const shape = (resource: ResolvedProjectResource): unknown => [
+      resource.resourceType,
+      resource.name,
+      resource.deploymentState === "deployed" ? resource.id : undefined,
+      ...(resource.children ? [resource.children.map(shape)] : []),
+    ];
+    expect(resources.map(shape)).toEqual([
+      ["runtime", "web", "arn:runtime/web-1"],
+      ["harness", "chat", "arn:harness/chat-1"],
+      ["memory", "user_mem", "arn:memory/mem-1"],
+      ["knowledge-base", "kb", "arn:kb/kb-1"],
+      ["credential", "cred", "arn:aws:cred/cred"],
+      ["evaluator", "ev", "arn:evaluator/ev-1"],
+      ["online-eval", "oe", "arn:online-eval/oe-1"],
+      ["gateway", "gw", "arn:gateway/gw-1", [["gateway-target", "tgt", "tgt-1"]]],
+      ["policy-engine", "pe", "arn:policy-engine/pe-1", [["policy", "pol", "arn:policy/pol-1"]]],
+      ["config-bundle", "cb", "arn:config-bundle/cb-1"],
+      [
+        "payment-manager",
+        "pay",
+        "arn:payment-manager/pay-1",
+        [["payment-connector", "wallet_one", "conn-1"]],
+      ],
+    ]);
+    expect(subject.stackReads).toHaveLength(1);
+  });
+
+  test("reports a declared resource the stack does not publish as local-only", async () => {
+    const input = await project();
+    input.spec = {
+      ...input.spec,
+      memories: [{ name: "shortTerm" }, { name: "longTerm" }],
+    } as unknown as typeof input.spec;
+    await updateTargetState(json, input.rootPath, TARGET.name, { stackArn: STACK_ARN });
+    const subject = harness({
+      describedStack: {
+        StackName: S,
+        CreationTime: new Date(0),
+        StackStatus: "CREATE_COMPLETE",
+        Outputs: [out(`${S}-Memory-shortTerm-Arn`, "arn:memory/short-1")],
+      },
+    });
+
+    const resources = await subject.backend.resolveProjectResources(input, { target: TARGET });
+
+    expect(resources).toEqual([
+      {
+        resourceType: "memory",
+        name: "shortTerm",
+        deploymentState: "deployed",
+        id: "arn:memory/short-1",
+      },
+      { resourceType: "memory", name: "longTerm", deploymentState: "local-only" },
+    ]);
+  });
+
+  test("reports local-only without reading AWS when the target has no recorded stack", async () => {
+    const input = await project();
+    input.spec = {
+      ...input.spec,
+      memories: [{ name: "mem" }],
+    } as unknown as typeof input.spec;
+    const subject = harness({ describedStack: null });
+
+    const resources = await subject.backend.resolveProjectResources(input, { target: TARGET });
+
+    expect(resources).toEqual([
+      { resourceType: "memory", name: "mem", deploymentState: "local-only" },
+    ]);
+    expect(subject.stackReads).toEqual([]);
+    expect(subject.accountCredentials).toEqual([]);
+  });
+
+  test("nests a gateway's targets under the gateway", async () => {
+    const input = await project();
+    input.spec = {
+      ...input.spec,
+      agentCoreGateways: [{ name: "gw", targets: [{ name: "owned" }, { name: "second" }] }],
+    } as unknown as typeof input.spec;
+    const subject = harness({ describedStack: null });
+
+    const resources = await subject.backend.resolveProjectResources(input, { target: TARGET });
+
+    expect(
+      resources.map(({ resourceType, name, children }) => [
+        resourceType,
+        name,
+        children?.map((child) => child.name),
+      ]),
+    ).toEqual([["gateway", "gw", ["owned", "second"]]]);
   });
 });
--- a/src/core/project/backends/cdk.ts
+++ b/src/core/project/backends/cdk.ts
@@ -6,7 +6,9 @@ import type {
   DeployResult,
   Project,
   ProjectEvent,
+  DeployableResource,
   ResolvedDeployedResource,
+  ResolvedProjectResource,
 } from "../../../handlers/project/types";
 import {
   createLineSplitter,
@@ -23,6 +25,7 @@ import type {
   DeployBackendInput,
   ProjectBackend,
   ResolveDeployedResourcesBackendInput,
+  ResolveProjectResourcesBackendInput,
 } from "./types";
 import { createCloudFormationClient } from "../../factories";
 import type { CreateCloudFormationClient } from "../../types";
@@ -68,6 +71,11 @@ type StackDescriber = typeof describeStack;
  * the cap streamProcess uses for a failed subprocess's captured output.
  */
 const MAX_ERROR_OUTPUT_LINES = 20;
+
+// Payment logical ids drop underscores the same way the template's toCdkId does.
+function cdkId(name: string): string {
+  return name.replace(/_/g, "");
+}
 
 function findDeployedResourceId(
   stack: Stack,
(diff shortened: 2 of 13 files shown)


jariy17:
```json
{ "projectName": "bbStatus30455", "target": "default", "region": "us-west-2",
  "resources": [
    { "resourceType": "harness", "name": "bbStatus30455", "deploymentState": "deployed",
      "identifier": "arn:aws:bedrock-agentcore:us-west-2:725476964917:harness/bbStatus30455_bbStatus30455-R7ttcnrOG8" },
    { "resourceType": "memory", "name": "user_mem", "deploymentState": "deployed",
      "identifier": "arn:aws:bedrock-agentcore:us-west-2:725476964917:memory/bbStatus30455_user_mem-9ndWxKF91x" },
    { "resourceType": "gateway", "name": "tools", "deploymentState": "deployed",
      "identifier": "arn:aws:bedrock-agentcore:us-west-2:725476964917:gateway/bbstatus30455-tools-hwyi9qiqkn" },
    { "resourceType": "policy-engine", "name": "guards", "deploymentState": "deployed",
      "identifier": "arn:aws:bedrock-agentcore:us-west-2:725476964917:policy-engine/bbStatus30455_guards-qjcpxplfak",
      "children": [
        { "resourceType": "policy", "name": "no_pii", "deploymentState": "deployed",
          "identifier": "arn:aws:bedrock-agentcore:us-west-2:725476964917:policy-engine/bbStatus30455_guards-qjcpxplfak/policy/no_pii-i9vsi_xtqh" } ] } ] }
```

src/handlers/project/status/index.ts
AlexanderRichey, line 1: Why is this a binary file?

AlexanderRichey (dismissed):
This doesn't look like it implements the approach we discussed of calling DescribeStacks. What motivated changing direction here?

jariy17:
### First-run fix: `project status` on a freshly created project

`project create` leaves `agentcore/aws-targets.json` as `[]`, and only `project deploy` provisions the `default` target. `status` resolved its target through a path that demanded the target already exist, so the most likely first command after `create` returned a config error instead of reporting that nothing is deployed yet. `allowMissing` did not cover it — that flag handles a *declared* target with no stack, not an *undeclared* target.

**Mechanism.** `resolveExistingTarget` became `resolveTarget(project, name, { undeclaredDefaultIsUndeployed })`, returning `AwsDeploymentTarget | undefined`. `resolveDeployedResources` passes `undeclaredDefaultIsUndeployed: input.allowMissing === true` and short-circuits to `{ resources: [] }` when no target is declared, so `ResolvedDeployedResources.target` is now optional. The status envelope falls back to the requested name and the CLI's effective region (`RegionKey`); a declared target still reports its own region.

Chosen over the alternatives because it keeps `status` read-only — no `aws-targets.json` write, no `provisionDefaultTarget`, no STS lookup — and expresses the absence in the type rather than via caught exceptions or a fabricated account. Only `default` is treated as implicitly-present, mirroring the guard on the deploy side and for the same reason: inventing a *named* target would turn a typo'd `--target` into a report about somewhere unintended.

**Live, against a fresh project (no deploy):**

Before — exit 1:
```
Error: No deployment targets are configured for project 'bbFirstRun12141'. Add at least one to .../agentcore/aws-targets.json, for example:
[{ "name": "default", "account": "111122223333", "region": "us-east-1" }]
```

After — exit 0:
```json
{
  "projectName": "bbFirstRun12141",
  "target": "default",
  "region": "us-west-2",
  "resources": [
    { "resourceType": "harness", "name": "bbFirstRun12141", "deploymentState": "local-only" },
    { "resourceType": "memory", "name": "shortTerm", "deploymentState": "local-only" }
  ]
}
```

Unchanged, verified live with targets declared:
- `--target typo` → exit 1, `Project '...' has no deployment target named 'typo'. ... defines: default, eu.`
- `--target eu` → reports `"target": "eu", "region": "eu-west-1"`, not the ambient region.

One handler test covers it; it fails without the fix and passes with it (verified by mutation).

### Separately: the handler was a binary file to git

`status/index.ts` joined its resource-key parts with literal NUL bytes, so git classified the file as binary and its contents did not render in the diff — the handler was unreviewable in this PR. The separator is now `/`, which cannot appear in a resource name (`^[a-zA-Z][a-zA-Z0-9_]{0,47}$`), so collision safety is unchanged. The committed blob is now ASCII text.

Code comments have also been dropped from the code added by this PR and by #2158.

`tsc` clean · `bun test` 2608 pass / 0 fail · oxlint clean · prettier clean.

jariy17:
Removed `allowMissing` from the deployed-resource resolution path, and dropped the `unassignedTargets` row from `project status` (`7754e358`, one commit).

**`allowMissing`** — `resolveDeployedResources` and the CDK backend now always throw `ProjectStateError` when the target is undeclared, has no recorded stack ARN, or its stack is gone. It is gone from `ResolveDeployedResourcesInput`, `ResolveDeployedResourcesBackendInput`, and the `undeclaredDefaultIsUndeployed` mechanism; `resolveExistingTarget` is restored to its `refactor` form, so that function's diff is now empty. `ResolvedDeployedResources.target` is required again, which let the `?? flags.target` / `?? RegionKey` fallbacks in the status handler go too (and `RegionKey`'s now-unused import).

Consequence, accepted: `agentcore project status` on a freshly created project exits 1 with `No deployment targets are configured for project '<name>'. Add at least one to <path>/agentcore/aws-targets.json` — `project create` writes `aws-targets.json` as `[]` and only `project deploy` provisions the `default` target.

**`unassignedTargets`** — removed from `describe()`. Nothing in the CLI can create an unassigned gateway target: `AddResourceInput` makes `gatewayName` mandatory for `gateway-target`, and the only assignment to `spec.unassignedTargets` anywhere sets it to `undefined`. The row was unreachable except from a hand-edited `agentcore.json`. The resolver in `backends/cdk.ts` keeps its entry — out of scope here.

**Tests** — dropped the three `allowMissing` cases in `cdk.test.ts` (wrong-account stays covered at two other sites) and inverted the first-run status test to assert the throw. No status fixture declared `unassignedTargets`, so no assertion needed updating.

tsc / oxlint / prettier clean, 2605 pass / 0 fail.

jariy17:
| This doesn't look like it implements the approach we discussed of calling DescribeStacks. What motivated changing direction here?

So ProjectManager.resolveDeployedResource resolves arns for harness and runtime using DescribeStacks. Therefore, I modifying that function to return all project resources arns. The only addition I added was children parameter where resources like Gateway has targets as children.

jariy17:
## Fixed a `project invoke` regression, plus a rename and a deployment-state union

### The regression

Widening the resolver to cover every declared resource also switched its identifiers from bare IDs to ARNs. That broke `project invoke` for both runtime and harness: `GetAgentRuntime` and `GetHarness` each take an ID, and the service builds its IAM policy resource from whatever identifier it is handed. An ARN therefore produced a resource matching no policy statement, so the call failed as `AccessDeniedException` rather than a validation error — which is why the failure initially read as a permissions gap.

Confirmed with an A/B using identical credentials: bare ID succeeds, ARN returns `AccessDeniedException`.

Note `runtimeIdSchema` already rejects ARNs, but only on the standalone `runtime invoke --id` path, so it never guarded this one.

### The fix

The backend input takes an identifier kind. `resolveProjectResource` (the invoke path) asks for `"id"`; `resolveProjectResources` (status) keeps ARNs.

```ts
export type ResolveProjectResourcesBackendInput = {
  target: AwsDeploymentTarget;
  identifier?: "arn" | "id";
};
```

The CDK backend picks the export suffix from it, so one line covers every resource type:

```ts
const suffix = identifier === "arn" ? "Arn" : "Id";
```

### Renames

`resolveDeployedResource(s)` → `resolveProjectResource(s)`, and `ResolvedDeployedResource` → `ResolvedProjectResource`. The plural now returns everything the project declares, not just what is deployed, so "deployed" was the wrong word in the name.

### Deployment state as a union

```ts
export type ResolvedProjectResource = {
  resourceType: DeployableResource;
  name: string;
  parent?: string;
  target: AwsDeploymentTarget;
} & ({ deploymentState: "deployed"; id: string } | { deploymentState: "local-only" });

export type DeployedProjectResource = Extract<ResolvedProjectResource, { deploymentState: "deployed" }>;
```

`id` exists only where it is meaningful, so nothing can read an identifier off a resource that has none. The singular resolver returns `DeployedProjectResource`, keeping its callers unchanged.

Resources with no matching stack output now come back as `local-only` rows instead of being dropped, which is what lets `project status` list them. The invoke picker filters to deployed rows.

### Live verification (EXPLORE account, us-west-2)

Project declaring a runtime, a harness, and two memories:

```
$ agentcore project status --json     # pendingMem added to the spec, not yet deployed
{
  "projectName": "bbInv29570",
  "target": "default",
  "region": "us-west-2",
  "resources": [
    { "resourceType": "runtime", "name": "inv_agent", "deploymentState": "deployed",
      "identifier": "arn:aws:bedrock-agentcore:us-west-2:...:runtime/bbInv29570_inv_agent-Pdmb8BChEa" },
    { "resourceType": "harness", "name": "bbInv29570", "deploymentState": "deployed",
      "identifier": "arn:aws:bedrock-agentcore:us-west-2:...:harness/bbInv29570_bbInv29570-eRORVSlFKe" },
    { "resourceType": "memory", "name": "inv_agentMemory", "deploymentState": "deployed",
      "identifier": "arn:aws:bedrock-agentcore:us-west-2:...:memory/bbInv29570_inv_agentMemory-1XFzmh2LoC" },
    { "resourceType": "memory", "name": "pendingMem", "deploymentState": "local-only" }
  ]
}
```

After `project deploy`, `pendingMem` flips to `deployed` with its ARN.

Both invoke paths succeed against the same stack, including while a `local-only` row is present in the resolved list:

- `project invoke runtime --name inv_agent` → `status=200`, streamed response
- `project invoke harness --name bbInv29570 --prompt ...` → `stopReason: end_turn`

Stack torn down afterwards.

### Gates

`typecheck`, `bun test` (2606 pass), `lint:check`, `format:check` — all clean.

src/handlers/project/status/index.ts
jariy17: My agent add NUL bytes to this file which made github make it into a binary that is now fixed.

jariy17:
| This doesn't look like it implements the approach we discussed of calling DescribeStacks. What motivated changing direction here?

`ProjectManager.resolveProjectResources` already calls DescribeStacks so I just expanded to all project resources. The handler will call this function. The only change from yesterday is that I nested resources that are children of a resource i.e GatewayTargets to Gateway.

src/handlers/project/invoke/screen.tsx
```diff
@@ -69,7 +69,7 @@ export function ProjectInvokePickerScreen({ ctx, core }: ScreenProps) {
     if (!project) return;
     let active = true;
     void core.projectManager
-      .resolveDeployedResources(project, { target: "default" })
+      .resolveProjectResources(project, { target: "default" })
```
tejaskash: Regression from this PR. The manager never forwards an identifier here, so the resolver defaults to ARNs and `select()` passes them as harnessId/agentRuntimeId. Picking a deployed harness in the TUI fails with `Value 'arn:aws:...:harness/...' at 'harnessId' failed to satisfy constraint`. Headless invoke was fixed with `identifier: "id"`, this path was not. Returning both id and arn on each row would remove the toggle entirely.

src/handlers/project/status/index.ts
```diff
@@ -1,11 +1,71 @@
...
+  deploymentState: resource.deploymentState,
+  ...(resource.deploymentState === "deployed" ? { identifier: resource.id } : {}),
+});
+
+function nest(resources: ResolvedProjectResource[]): ResourceStatus[] {
+  const entries = resources.map((resource) => ({ resource, row: toResourceStatus(resource) }));
+  const top: ResourceStatus[] = [];
+  for (const { resource, row } of entries) {
+    const owner = resource.parent
+      ? entries.find(({ resource: other }) => !other.parent && other.name === resource.parent)
```
tejaskash: Parent is matched by name only. Names are unique per collection, not across them, so a gateway and a policy engine can both be `guards`. Reproduced live: the deployed policy showed under the gateway and the engine had no children. Match on the expected parent type too.

src/handlers/project/status/index.ts
```diff
@@ -1,11 +1,71 @@
...
+        target: flags.target,
+      });
+
+      const status: ProjectStatus = {
+        projectName: project.name,
+        target: resolved.target.name,
+        region: resolved.target.region,
+        resources: nest(resolved.resources),
+      };
+      ctx.require(JsonRendererKey).renderJson(status);
```
tejaskash, line 41: deploy, build, dev and export in this router check `JsonKey` and print a human line otherwise. This always dumps JSON, so a bare `agentcore project status` in a terminal gets a raw blob.

src/handlers/project/status/index.test.ts
```diff
@@ -0,0 +1,236 @@
...
+      target: "default",
+      region: "us-east-1",
+      resources: [
+        HARNESS_ROW,
+        { resourceType: "memory", name: "shortTerm", deploymentState: "local-only" },
+      ],
+    });
+  });
+
+  test("rejects a freshly created project that declares no targets", async () => {
```
tejaskash: `project create` leaves aws-targets.json empty and `deploy` fills it, so right after create `status` tells the user to hand-write the file with an example account id. Commit bfef4ea fixed this and `drop allowMissing` undid it. The "all local-only" test above only passes because the fake backend never throws, the real one always does here.

src/core/project/backends/cdk.ts
```diff
@@ -370,11 +363,11 @@ export class CdkBackend implements ProjectBackend {
...
     project: Project,
-    input: ResolveDeployedResourcesBackendInput,
-  ): Promise<ResolvedDeployedResource[]> {
-    const { target } = input;
+    input: ResolveProjectResourcesBackendInput,
+  ): Promise<ResolvedProjectResource[]> {
+    const { target, identifier = "arn" } = input;
     const deployedState = await readDeployedState(this.json, project.rootPath);
     const stackArn = deployedState.targets[target.name]?.stackArn;
     if (!stackArn) {
```
tejaskash, line 388: For status, a missing stack just means nothing is deployed yet, which is exactly what `local-only` describes. Deploy and remove can keep the hard failure.

src/core/project/backends/cdk.ts
```diff
@@ -393,13 +386,100 @@ export class CdkBackend implements ProjectBackend {
...
+          return deployedState.targets[target.name]?.resources?.credentials?.[name]
+            ?.credentialProviderArn;
+        default: {
+          const unhandled: never = resourceType;
+          return unhandled;
+        }
+      }
+    };
+
+    const declared: Declared[] = [
```
tejaskash: The spec also has `datasets` and `mcpRuntimeTools`, and `declaresNothingDeployable` in deploy counts both. They are missing here, so a deployed dataset gets no row at all and deploy and status disagree on what a resource is.

src/core/project/backends/cdk.ts
```diff
@@ -393,13 +386,100 @@ export class CdkBackend implements ProjectBackend {
...
+        case "gateway-target":
+          return byExportName("GatewayTarget", name, "Id");
+        case "policy-engine":
+          return byExportName("PolicyEngine", name, suffix);
+        case "policy":
+          // ExportName: <StackName>-Policy-<engineName>-<policyName>-Arn
+          return byExportName("Policy", parent ?? "", name, suffix);
+        case "config-bundle":
+          return byExportName("ConfigBundle", name, suffix);
+        case "capacity-provider":
```
tejaskash: No `capacityProviders` field exists anywhere in the spec, so this case and the union member are unreachable, and the `never` default forces every future switch to keep handling it.

src/core/project/backends/cdk.ts
```diff
@@ -393,13 +386,100 @@ export class CdkBackend implements ProjectBackend {
...
+          return byExportName("CapacityProvider", name, suffix);
+        case "payment":
+          // Payments doesn't set a ExportName so we search for OutputKey
```
tejaskash: This links to the L3 `AgentCorePayments` construct, but the generated project never uses it. The `Payment<name>ManagerArn` output comes from `toCdkId` in the CLI's cdk-stack.ts template. Also the workspace rule is no line comments, so these should go.

tejaskash (commented):
Built it and drove it against a real account in us-west-2. Status resolves every resource correctly and headless invoke works. Two blockers before merge: the interactive invoke picker is broken (ARNs sent where ids are expected) and child nesting matches parents by name only. Details inline.

src/handlers/project/types.ts
```diff
@@ -160,25 +160,46 @@ export type ResolveTargetInput = {
...
+  | "knowledge-base"
+  | "credential"
+  | "evaluator"
+  | "online-eval"
+  | "gateway"
+  | "gateway-target"
+  | "policy-engine"
+  | "policy"
+  | "config-bundle"
+  | "payment"
```
tejaskash: `project add` and `project remove` call this `payment-manager` (see `ProjectResource` below), status calls it `payment`. Deriving this from `ProjectResource` keeps the two from drifting.

src/core/project/backends/cdk.test.ts
```diff
@@ -810,8 +811,8 @@ describe("CdkBackend.deploy", () => {
   });
 });
 
-describe("CdkBackend.resolveDeployedResources", () => {
-  test("describes the stack once and returns only resources with deployed ID outputs", async () => {
+describe("CdkBackend.resolveProjectResources", () => {
```
tejaskash: The base branch tested `-RuntimeId` and `-Harness-<name>-Id`. Now nothing here asks for `identifier: "id"`, so a typo in the suffix logic would make headless invoke report a deployed runtime as not deployed with CI green. One id-mode case would cover it.

src/core/project/backends/cdk.ts
jariy17: Yeah that makes sense. I'll let the handlers throw the ProjectStateError if they want too.

src/core/project/backends/cdk.ts
jariy17: I was thinking we will add capacityProvider in the new cli but I can remove it for now.

src/core/project/backends/cdk.ts
jariy17: `Datasets` are being deprecated in favor of the imperative based commands. I think `mcpRuntimeTools` is also deprecated because I don't see any add command for it. `declaresNothingDeployable` doesn't exist.

src/handlers/project/status/index.ts
jariy17: Nice catch!

src/handlers/project/status/index.ts
jariy17: agentcore project status will just print out the json output for now. If customers want a human readable, Alex is implementing the tui experience after this pr gets merged.

src/core/project/backends/cdk.ts
jariy17: But I like these comments. It help developers understand why the credentials and Payments are special cases.

src/handlers/project/invoke/screen.tsx
```diff
@@ -83,29 +83,41 @@ export function ProjectInvokePickerScreen({ ctx, core }: ScreenProps) {
...
+            r,
+          ): r is typeof r & {
+            resourceType: "runtime" | "harness";
+            deploymentState: "deployed";
+            id: string;
+          } =>
+            (r.resourceType === "runtime" || r.resourceType === "harness") &&
+            r.deploymentState === "deployed",
+        )
+        .map((resource) => {
```
AlexanderRichey: Why is this code changing in this PR? I thought the scope was the status command.

jariy17: The invoke commands/tui uses ProjectManager.resolvedProjectResources to figure out what deployed harness/runtimes it can invoke.

src/handlers/project/invoke/runtime.tsx
```diff
@@ -57,11 +57,12 @@ export const createProjectInvokeRuntimeHandler = (
     handle: async (ctx, flags) => {
       const project = ctx.require(ProjectKey);
       const name = selectProjectResource(project, "runtime", flags.name);
-      const deployed = await core.projectManager.resolveDeployedResource(project, {
+      const deployed = await core.projectManager.resolveProjectResource(project, {
```
AlexanderRichey: There doesn't seem to be anything pressing that necessitates this name change. It's causing a lot of cascading changes that make this code hard to review and increase the risk of introducing errors. Let's omit this part of the PR and focus on just the status functionality.

jariy17: I renamed resolveDeployResource to resolveProjectResources because it now returns whether a resource was deployed or not. I'll just make resolveProjectResource it's own function  in the next revision. Later we can merge the functions.

src/handlers/project/invoke/index.test.tsx
```diff
@@ -95,6 +95,9 @@ function backend() {
         })),
       ];
     },
+    async resolveProjectResources() {
```
jariy17, line 98: This was meant to prevent invoke handlers from using this resolvedProjectResources because they should use resolveDeployedResource. We will merge these functions later.

tejaskash (approved):

PR #2166, opened 2026-09-02
feat(project): add TUI wizards for four `project add` resources

notgitika:
## What

Selecting `add` from the project menu now opens a resource menu, and four of the fifteen resources get a wizard: **memory**, **gateway**, **policy-engine**, **config-bundle**. The other eleven keep the not-implemented screen and are unchanged on the command line.

```
 agentcore → project → add → memory → adds a memory to the current project
──────────────────────────────────────────────────────────────────────────
 ✓ name ── ● strategies ── ○ retention ── ○ description ── ○ review
──────────────────────────────────────────────────────────────────────────
 what should be extracted from raw events?
 selecting none keeps short-term events only
 ╭────────────────────────────────────────────────────────────────────────╮
 │ ❯ [✓] SEMANTIC         durable facts extracted from conversations      │
 │   [ ] SUMMARIZATION    running summaries of each session               │
 │   [ ] USER_PREFERENCE  preferences the user states or implies          │
 │   [ ] EPISODIC         past episodes, plus reflections over them       │
 ╰────────────────────────────────────────────────────────────────────────╯
──────────────────────────────────────────────────────────────────────────
[↑↓] move  [space] toggle  [enter] continue  [esc] back  [ctl+c] quit
```

| resource | steps (*italic* = conditional) |
|---|---|
| `memory` | name → strategies → retention → description → review |
| `gateway` | name → protocol → authorizer → *jwt config* → *search* → description → review |
| `policy-engine` | name → description → *attach* → *mode* → review |
| `config-bundle` | name → components → description → branch → commit → review |

## Design

Two pieces, each with one job.

**1. A shared `<Wizard>` shell** (`components/wizard/`, 277 lines). It owns the step list, position, key handling, and the `form → running → success | error` phases. A screen supplies only the questions, as `<Step>` children, and branches with a plain conditional:

```tsx
{isCustomJwt && (
  <Step name="authorizer-configuration" question="paste the authorizerConfiguration">
    <TextAreaField required json schema={GatewayAuthorizerConfigurationInputSchema} … />
  </Step>
)}
```

`React.Children.toArray` drops the `false` a closed branch produces, so an inapplicable step is absent from the flow and the stepper alike. Position is keyed by step **name**, not index, so a step appearing or vanishing never moves the user. Duplicate names throw at render. Each field owns its own `useInput`, so the shell has no focus concept — **one field per step** is the invariant; a step that needs two related inputs gets a compound field, not shell-level focus management.

**2. One input builder per resource**, exported from the handler: `toAddGatewayInput(project, input)`, `toAddMemoryInput(input)`, `toAddConfigBundleInput(input)`, `toAddPolicyEngineInput(project, input)`. Each takes a typed `XxxInput` (already-parsed JSON, already-split tags) and is the *only* place defaults, cross-field rules, and resource assembly live. The flag handler's `handle` reduces to flags → `XxxInput`; the screen's `onSubmit` reduces to form values → `XxxInput`. Neither knows what a Gateway is.

```
flags ──(resolve file://, parse JSON, pair flags)──┐
                                                    ├──► toAddGatewayInput(project, input) ──► projectManager.addResource
form  ──(trim, parse, drop hidden-step answers)────┘
```

This is what stops the two paths drifting. It already caught one: the flag path validated `--authorizer-configuration` with `GatewayAuthorizerConfigSchema.strict()`, the first cut of the wizard used the non-strict schema, so the SDK's `customJWTAuthorizer` casing was rejected by flags but accepted by the wizard — and then failed at submit as a missing `customJwtAuthorizer`. Both now share `GatewayAuthorizerConfigurationInputSchema`; a screen test pins it.

The wizard convention that makes the builders' cross-field errors unreachable from the TUI: pass only the answers to steps the user saw. A JWT config typed before switching the authorizer to `NONE` is dropped, not sent.

## Decisions worth a reviewer's attention

- **Submit goes through `projectManager.addResource`, not `handler.handle`.** Handlers write to a `config.io.stderr` captured at wiring time, which Ink's alternate screen swallows.
- **Flags stay `.optional()`.** Required-ness is a field prop; a non-optional schema lets Commander reject a bare command before the TUI middleware can open a screen (`69aa0c19`).
- **`ProjectGate`** resolves `ProjectKey` for screens the user navigated to (middleware never ran) and reports the CLI's own not-found guidance.
- **`Form*` components** omit label/help rows when passed empty strings, so a field under a `<Step>` renders just the control. Existing callers pass non-empty strings and are unaffected.

## Testing

`bun test`: **2645 pass, 0 fail**. `tsc --noEmit`, `oxlint`, `prettier --check` clean.

38 new tests. Handler flag tests are unchanged and now exercise the shared builders; screen tests exercise the same builders from the form side. `add.screen.test.tsx` reads its case list off the compiled Commander tree, so a resource routed later without a screen fails the suite.

## Not in scope

- A bare `agentcore project add memory` still runs the flag path. Opening the wizard needs a per-command launcher, not the router-wide `withTuiOnEmptyFlagsAndArgs`.
- Migrating `project create`, `HarnessWizard`, `EndpointWizard` onto the shell. `create`'s forks are expressible as `{cond && <Step/>}`; that's the natural next PR.
- The remaining eleven `add` resources. Each follows the same recipe: export `XxxInput` + `toAddXxxInput` from the handler, reduce `handle` to flag conversion, write a screen of `<Step>`s.

### DIFF
--- a/src/components/wizard/fields.tsx
+++ b/src/components/wizard/fields.tsx
@@ -1,0 +1,397 @@
+import { useState } from "react";
+import { Box, Text, useInput } from "ink";
+import type z from "zod";
+import { FormTextInput } from "../FormTextInput";
+import { FormRadioGroup } from "../FormRadioGroup";
+import { FormCheckboxMultiSelect } from "../FormCheckboxMultiSelect";
+import { FormTextArea } from "../FormTextArea";
+import { KeyValueTable } from "../KeyValueTable";
+import { darkTheme } from "../ui/_core.js";
+import { useKeyHints, useWizard } from "./context";
+
+const theme = darkTheme;
+
+// Every field owns the key handling for its own step — esc goes back, enter
+// advances, arrows move — so a screen never writes that boilerplate again.
+
+// firstIssue renders the schema's own message, so the wizard rejects exactly
+// what the flag-driven path rejects and says the same thing about it. The issue
+// path is prefixed when there is one: for a nested value — a component inside a
+// components map, say — "expected object, received string" alone does not say
+// which key is wrong.
+function firstIssue(schema: z.ZodType, value: unknown): string | undefined {
+  const parsed = schema.safeParse(value);
+  if (parsed.success) return undefined;
+  const issue = parsed.error.issues[0];
+  if (!issue) return "invalid value";
+  const path = issue.path.join(".");
+  return path === "" ? issue.message : `${path}: ${issue.message}`;
+}
+
+interface ValidateOptions {
+  label: string;
+  required: boolean;
+  schema?: z.ZodType;
+  // json parses the value before the schema sees it, so a malformed blob is
+  // reported as bad JSON rather than as a shape the schema cannot read.
+  json?: boolean;
+}
+
+// validateEntry returns the message that should block the step, or undefined to
+// let it advance. Shared by the single-line and multi-line fields so both refuse
+// the same input for the same stated reason.
+function validateEntry(
+  value: string,
+  { label, required, schema, json = false }: ValidateOptions,
+): string | undefined {
+  const trimmed = value.trim();
+  if (trimmed === "") return required ? `${label} is required` : undefined;
+
+  let parsed: unknown = trimmed;
+  if (json) {
+    try {
+      parsed = JSON.parse(trimmed);
+    } catch (cause) {
+      return `${label} is not valid JSON: ${(cause as Error).message}`;
+    }
+  }
+  return schema ? firstIssue(schema, parsed) : undefined;
+}
+
+export interface TextFieldProps {
+  // label names the value in validation messages ("<label> is required").
+  // It is not rendered: the <Step> already asks the question, and repeating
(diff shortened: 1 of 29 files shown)


src/components/wizard/fields.tsx
```diff
@@ -0,0 +1,397 @@
...
+    const issue = validateEntry(value, { label, required, schema, json });
+    if (issue !== undefined) {
+      setError(issue);
+      return;
+    }
+    setError(undefined);
+    advance();
+  });
+
+  return (
```
AlexanderRichey, line 348: Are we moving stuff around? I think we already have components for all of this. Let's reuse and not duplicate.

AlexanderRichey (commented):
A lot of this looks good, but I'm concerned about potential duplication with existing components/patterns. Can you take a pass over this to make sure we're not duplicating code/components/logic we already have?

notgitika:
temporarily closing this

PR #2167, opened 2026-09-02
feat(project): add TUI wizards for six more `project add` resources

notgitika:
> **Stacked on #2166.** Review only the last commit (`64b98a77`); the rest is #2166. Once that merges, this diff collapses to it.

## What

Six more `project add` resources get a wizard, on the shell #2166 introduced: **policy**, **gateway-connector**, **online-eval**, **online-insight**, **evaluator llm-as-a-judge**, **payment-manager**. That makes ten of fifteen. The remaining five (`harness`, `runtime`, `credentials`, `gateway-target`, `payment-connector`) keep the not-implemented screen.

| resource | steps (*italic* = conditional) |
|---|---|
| `policy` | engine → name → statement → phase → validation → enforcement → description → review |
| `gateway-connector` | gateway → connector → *knowledge base* → *knowledge base ID* → name → review |
| `online-eval` | name → *source* → *agent* → *endpoint* / *log groups* → *services* → *project evaluators* → evaluators → sampling → enabled → description → review |
| `online-insight` | name → *source* → … → insights → clustering → sampling → enabled → description → review |
| `evaluator` | a menu; `llm-as-a-judge`: name → level → model → instructions → scale → *custom scale* → description → review |
| `payment-manager` | name → authorizer → *issuer* → *clients* → *audience* → *scopes* → auto-payment → spend limit → description → review |

Where a project resource can answer a question, it is offered: `policy` picks from the project's Policy Engines, `gateway-connector` from its Gateways and Knowledge Bases, `online-eval` from its runtimes and evaluators. Where the project has none and the wizard cannot proceed without one, a `Prerequisite` screen names what is missing and the command that adds it.

## Design

Each resource follows the recipe #2166 established, unchanged:

```
flags ──(resolve file://, parse JSON, pair flags)──┐
                                                    ├──► toAddXxxInput(input) ──► projectManager.addResource
form  ──(trim, split lists, drop hidden-step answers)┘
```

The handler exports `XxxInput` + `toAddXxxInput`; `handle` reduces to flag conversion; the screen's `onSubmit` reduces to form conversion. Every existing flag test now exercises the shared builder, and the screen tests exercise the same builder from the form side — so the two paths write byte-identical `agentcore.json` entries, which the tests assert with `toEqual` against the full object.

Two things moved *into* builders that had been handler-body-only: online-insight's `Builtin.Insight.*`/ARN check (now `InsightIdSchema`, also used by the wizard step), and llm-as-a-judge's model-ID check (now `JudgeModelSchema`, likewise). gateway-connector's builder takes a discriminated `target` — shortcut or complete configuration — so the JSON path goes through it too rather than around it.

## Shell additions

Both small, both driven by these six:

- **`Prerequisite`** — for a wizard whose first question has no possible answer. esc returns to the add menu.
- **`values.ts`** — `blankToUndefined`, `splitList`, `numberSchema`. Text answers → the typed values builders take. memory's retention step now uses `numberSchema` instead of its own copy.

`online-eval` and `online-insight` share their source/sampling/enable steps through `monitoring-steps.tsx` — a hook returning arrays of `<Step>`s, which `React.Children.toArray` flattens into the wizard's step list. The two wizards ask the same questions in the same words and produce the same shapes.

`evaluator` is a command group, so `agentcore/project/add/evaluator` is a `RouterScreen` menu like `add` itself, reading its kinds off the Commander tree.

## Decisions worth a reviewer's attention

- **Curation.** The wizards don't ask everything the flags accept. Skipped: `--tags`, `--kms-key-arn`, `--encryption-key-arn` (all six); payment-manager's `--tool-allowlist` and `--network-preferences`; gateway-connector's `--connector-configuration` JSON path. All remain available on the command line and pass through the same builder.
- **Builtin IDs are typed, not picked.** There's no list of `Builtin.*` evaluators or insights in the codebase, so those steps are comma-separated text. Project evaluators are offered as a multi-select first, and the text step is `required` only when nothing was picked there — the schema's "at least one evaluator" rule is never reached at submit.
- **`LogGroupNamesSchema`** is extracted from `OnlineEvalConfigSchema` (same definition, now named) so the log-groups step validates against the schema's own 1–5 bound.
- **Warnings carry over.** payment-manager's post-success stderr warnings (auto-payment enabled; runtimes need SDK config) are shown in the wizard's success hint from the same exported strings.

## Testing

`bun test`: **2656 pass, 0 fail**. `tsc --noEmit`, `oxlint` (1.74 and 1.80), `prettier --check` clean.

17 new screen tests. Each resource has a "same as the flag path" test asserting the written spec with `toEqual`, plus its branch behaviour (prerequisite screens, conditional steps, picker vs. text) and one field-validation case using the schema's own message.

### DIFF
--- a/src/components/FormCheckboxMultiSelect.tsx
+++ b/src/components/FormCheckboxMultiSelect.tsx
@@ -31,10 +31,14 @@ export function FormCheckboxMultiSelect({
 
   return (
     <Box flexDirection="column">
-      <Box flexDirection="column">
-        <Text color={theme.colors.text}>{name}</Text>
-        <Text color={theme.colors.muted}>{helpText}</Text>
-      </Box>
+      {/* Either row is omitted when empty, so a caller whose surrounding
+          context already asks the question renders just the checkboxes. */}
+      {(name !== "" || helpText !== "") && (
+        <Box flexDirection="column">
+          {name !== "" && <Text color={theme.colors.text}>{name}</Text>}
+          {helpText !== "" && <Text color={theme.colors.muted}>{helpText}</Text>}
+        </Box>
+      )}
       <Box
         flexDirection="column"
         paddingX={1}
--- a/src/components/FormRadioGroup.tsx
+++ b/src/components/FormRadioGroup.tsx
@@ -22,10 +22,14 @@ export function FormRadioGroup({ name, helpText, options, selectedIndex }: FormR
 
   return (
     <Box flexDirection="column">
-      <Box flexDirection="column">
-        <Text color={theme.colors.text}>{name}</Text>
-        <Text color={theme.colors.muted}>{helpText}</Text>
-      </Box>
+      {/* Either row is omitted when empty, so a caller whose surrounding
+          context already asks the question renders just the options. */}
+      {(name !== "" || helpText !== "") && (
+        <Box flexDirection="column">
+          {name !== "" && <Text color={theme.colors.text}>{name}</Text>}
+          {helpText !== "" && <Text color={theme.colors.muted}>{helpText}</Text>}
+        </Box>
+      )}
       <Box
         flexDirection="column"
         paddingX={1}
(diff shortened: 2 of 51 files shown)


notgitika:
temporarily closing this

PR #2171, opened 2026-09-02
feat: gateway policy generate 

tejaskash:
## What

- Adds `agentcore gateway policy generate --gateway-id <id|arn> --prompt <text>` which generates a Cedar policy for a deployed Gateway from a natural-language prompt and prints only the Cedar to stdout, ready to pipe into `project add policy --statement -`.
- `--policy-engine-id` defaults to the Gateway's attached engine, `--name` names the generation request, `--json` prints the structured result including the service findings.
- New `PolicyClient` Core sub-client (`core.policy`) behind a consumer-owned `CorePolicyClient` interface.
- Replaces the `project add policy --generate` approach from #2127 per review.

## Testing

- Golden tests in `src/handlers/gateway/gateway.policy.test.tsx` recorded against a deployed project (human output, `--json` by ARN with an explicit engine, no engine attached, nonexistent engine, untranslatable prompt). Waiter timeout and `GENERATE_FAILED` are faked at `.send()` in `gateway.test.tsx` alongside the flag validation rows.
- Live in us-west-2: deployed a project with a Gateway and attached Policy Engine, generated by id and by ARN, added the Cedar with `project add policy`, redeployed, confirmed the policy `ACTIVE`, tore down.

https://github.com/user-attachments/assets/e12790cc-4e72-4cd8-a14a-046691173f0f


src/core/policy.test.ts
```diff
@@ -0,0 +1,184 @@
+import { describe, expect, test } from "bun:test";
```
jariy17: Golden Tests only please. If there is a test case that can't use Golden Tests like `Throttling Error Handling`, throw it into gateway.test.tsx.

src/core/policy.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+export type PolicyGenerationWait = {
+  maxWaitTime: number;
+  minDelay: number;
+  maxDelay: number;
+};
+
+const DEFAULT_WAIT: PolicyGenerationWait = { maxWaitTime: 60, minDelay: 2, maxDelay: 5 };
+
+function resourceIdFromArn(value: string): string {
```
jariy17, line 29: nit: OOS but there must be a shared util function for this.

src/core/policy.tsx
```diff
@@ -0,0 +1,136 @@
...
+}
+
+export class PolicyClient implements CorePolicyClient {
+  constructor(
+    private readonly clients: AwsClients,
+    private readonly logger: Logger,
+    private readonly wait: PolicyGenerationWait = DEFAULT_WAIT,
+  ) {}
+
+  async *generatePolicy(
```
jariy17, line 40: OOS but a lot of our coreClient function's don't yield helpful messages like in here. We should work on this later.

src/core/policy.tsx
```diff
@@ -0,0 +1,136 @@
...
+    input: GeneratePolicyInput,
+    options: CoreOptions,
+  ): AsyncGenerator<ProgressEvent, PolicyGenerationResult> {
+    const control = this.clients.control(toClientConfig(options));
+    const gatewayId = resourceIdFromArn(input.gatewayId);
+
+    yield { type: "step", message: `Resolving gateway ${gatewayId}` };
+    const gateway = await control.send(new GetGatewayCommand({ gatewayIdentifier: gatewayId }));
+    if (!gateway.gatewayArn) {
```
jariy17: It's not possible for GetGateway to return no gatewayArn. Do you know why ur agent is being this overprotective. It's probably because aws-sdk-js make every response parameter optional.

jariy17 (commented):
Good just fix the tests.

src/core/policy.tsx
tejaskash: Agreed

src/core/policy.tsx
tejaskash: Agreed

src/core/policy.test.ts
tejaskash: Done. The Core tests are now golden fixtures recorded against a deployed project in `gateway.policy.test.tsx`. The two waiter outcomes that cannot be recorded (timeout, GENERATE_FAILED) live in `gateway.test.tsx` with the control client faked at `.send()`.

src/core/policy.tsx
tejaskash: Agreed and removed, along with the same guard on `policyGenerationId`. Both now use a non-null assertion like the rest of Core. Added a CLAUDE.md guideline so this pattern does not come back.

src/handlers/gateway/gateway.test.tsx
```diff
@@ -155,3 +171,50 @@ describe("gateway validation", () => {
     expect(core.gateway.calls).toEqual([]);
   });
 });
+
+/**
+ The waiter outcomes below cannot be recorded against the live service, so the
+ control plane is faked at .send() while the real PolicyClient and waiter run.
+**/
+describe("gateway policy generate against a faked control plane", () => {
+  function coreWith(status: string, statusReasons?: string[]): Core {
```
jariy17, line 180: Follow Up:  Can't you use the settle() function in fixtures.tsx? Settle() will perform a wait while recording the Golden tests.

jariy17 (approved):
LGTM, just a follow up on the tests.

AlexanderRichey (approved):

PR #2172, opened 2026-09-02
feat(project): build and deploy TUI screens render the CLI's own progress steps

notgitika:
## What

`project build` and `project deploy` open from the TUI menu instead of reporting "not implemented", and show **exactly the progress the command shows** — same steps, same ✓/✕/spinner glyphs, same output tail under the running step.

```
 agentcore → project → deploy → deploy the project to AWS
──────────────────────────────────────────────────────────────
 ╭────────────────────────────────────────────────────────────╮
 │ project  orders                                            │
 │ target   default                                           │
 ╰────────────────────────────────────────────────────────────╯
 ✓ Verifying AWS account 111122223333
 ⠋ Deploying AgentCore-orders-default
   │ CREATE_IN_PROGRESS | AWS::IAM::Role
──────────────────────────────────────────────────────────────
[ctl+c] quit
```

**build** starts as soon as the screen opens (nothing to confirm — it only writes inside the project). Done, it suggests `agentcore project deploy`; enter returns to the project menu.

**deploy** behaves as the command does: it deploys at once, and asks only in the one case the command asks — when `agentcore.json` declares nothing deployable, so deploying would remove the stack. Then the confirmation *is* the CLI's teardown question, and "y" is the pre-answered `--yes`. With several targets in `aws-targets.json` it first asks which (the TUI's stand-in for `--target`); with one or none it uses that or `default`, provisioning it on first deploy exactly as the command does.

## How 

Everything on screen is a CLI piece wrapped in the TUI frame:

| the command uses | the screen uses |
|---|---|
| `projectManager.build(project)` / `.deploy(project, …)` | the same generator, verbatim: the screen's action is `yield* core.projectManager.build(project)` |
| `TaskList`, rendered inline by `runWithProgress` | the same `TaskList`, rendered by `ConfirmAction` while the generator streams |
| the generator drain + event→task fold inside `runWithProgress` | extracted to `driveProgress` (over pure `applyProgressEvent` / `settleProgress`) in `src/tui/progress.tsx`; **both** renderers call it and supply only how to draw the tasks |
| `ConfirmAction` (the confirm → run → success \| error body behind `harness delete`) | the whole screen. It learns to accept an `AsyncGenerator<ProgressEvent, …>` as its `action` and shows the `TaskList` in place of the bare spinner. `message`/`title`/`rows` become optional (no question → run when ready; no header → no box). Existing callers unchanged |
| `Confirm` (y/N) | the teardown confirmation |
| the readline teardown question | `teardownQuestion()`, exported from the handler and used by both; `declaresNothingDeployable()` (now exported) decides whether it is asked |
| `Built project '…'` / `Deployed project '…' to target '…'` | `builtMessage()` / `deployedMessage()`, exported and used by both. The deploy outcome is read from `result.tornDown`, as the command does — not from the preflight heuristic, which synthesis can disagree with |
| `DataTable` (the invoke picker) | the target picker |
| `KeyValueTable` | `ConfirmAction`'s header and result rows (replacing its fixed-width renderer) |
| `withProject`'s not-found message | `projectNotFoundMessage()`, exported; `useProject` / `ProjectGate` show it for a screen the user navigated to, esc back to the menu |

Stack outputs are not listed on success, matching the command, which prints them only with `--json`. Bare `agentcore project deploy` on the command line is unchanged.

## Also

- `ProjectManager.listTargets(project)` — the declared targets, `[]` with no file; `resolveTarget` delegates to it.
- `ProjectInvokePickerScreen` drops its inline project-resolution effect and message copy for `useProject`.
- `ConfirmAction` doesn't advertise `esc` while the action runs — nothing listens for it mid-flight, by design — and gains `doneLabel` / `nextSteps` / `onCancel`.
- `flatFrame` / `waitForFlatText` test helpers.

## Testing

`bun test`: **2754 pass, 0 fail**. `tsc --noEmit`, `oxlint`, `prettier --check` clean.

9 screen tests stub the backend exactly as the handler tests do, so both paths are asserted against identical events: build happy (enter → menu) / failure (✕ step with tail kept, esc → menu) / outside a project (esc → menu); deploy one target / several targets (picker) / fresh project with no `aws-targets.json` (default provisioned, step streamed) / empty-project teardown confirm / result-vs-heuristic disagreement / failure. 6 unit tests for the extracted fold and driver.

Driven by hand against a real project as well.

### DIFF
--- a/src/components/ConfirmAction.tsx
+++ b/src/components/ConfirmAction.tsx
@@ -5,8 +5,9 @@ import { Layout } from "./Layout";
 import { Spinner } from "./ui/spinner";
 import { Confirm } from "./ui/confirm";
 import { TaskList, type Task } from "./ui/task-list";
+import { KeyValueTable } from "./KeyValueTable";
 import { darkTheme } from "./ui/_core.js";
-import { applyProgressEvent, settleProgress, type ProgressEvent } from "../tui/progress";
+import { driveProgress, type ProgressEvent } from "../tui/progress";
 
 const theme = darkTheme;
 
@@ -14,6 +15,8 @@ export interface SummaryRow {
   label: string;
   value: string;
 }
+
+export type ActionResult = SummaryRow[] | { title: string; rows: SummaryRow[] };
 
 export interface ConfirmActionProps {
   // breadcrumb labels the screen.
@@ -30,13 +33,16 @@ export interface ConfirmActionProps {
   // isPending / error reflect the summary fetch backing the overlay.
   isPending: boolean;
   error: Error | null;
-  // action performs the confirmed operation and resolves to result rows shown
-  // on the success panel. A long-running operation may instead return a
-  // progress generator — the same AsyncGenerator<ProgressEvent> runWithProgress
-  // drives for the headless command — and its steps render as a live task list
-  // while it runs, exactly as they do on the command line.
-  action: () => Promise<SummaryRow[]> | AsyncGenerator<ProgressEvent, SummaryRow[]>;
-  // successTitle heads the success panel (e.g. "Harness deleted").
+  // action performs the confirmed operation and resolves to what the success
+  // panel shows: result rows, optionally under a title that replaces
+  // successTitle — for an outcome only known once the action has run. A
+  // long-running operation may instead return a progress generator — the same
+  // AsyncGenerator<ProgressEvent> runWithProgress drives for the headless
+  // command — and its steps render as a live task list while it runs, exactly
+  // as they do on the command line.
+  action: () => Promise<ActionResult> | AsyncGenerator<ProgressEvent, ActionResult>;
+  // successTitle heads the success panel (e.g. "Harness deleted") unless the
+  // action's result carries its own.
   successTitle: string;
   // runningLabel is the spinner label while the action runs, shown until the
   // action's first progress step arrives (or throughout, for a plain promise).
@@ -51,7 +57,7 @@ export interface ConfirmActionProps {
 type Phase =
   | { kind: "confirm" }
   | { kind: "running" }
-  | { kind: "success"; rows: SummaryRow[] }
+  | { kind: "success"; title: string; rows: SummaryRow[] }
   | { kind: "error"; message: string };
 
 // ConfirmAction is the shared destructive-action screen body: a summary overlay
@@ -84,22 +90,14 @@ export function ConfirmAction({
     setTasks([]);
     try {
       const result = action();
-      let rows: SummaryRow[];
-      if (isProgressGenerator(result)) {
-        let next = await result.next();
-        while (!next.done) {
-          const event = next.value;
-          setTasks((current) => applyProgressEvent(current, event));
-          next = await result.next();
-        }
-        rows = next.value;
-      } else {
-        rows = await result;
-      }
-      setTasks((current) => settleProgress(current, "done"));
-      setPhase({ kind: "success", rows });
+      const outcome = isProgressGenerator(result)
+        ? await driveProgress(result, setTasks)
+        : await result;
+      const { title, rows } = Array.isArray(outcome)
+        ? { title: successTitle, rows: outcome }
+        : outcome;
+      setPhase({ kind: "success", title, rows });
     } catch (err) {
-      setTasks((current) => settleProgress(current, "failed"));
       setPhase({ kind: "error", message: err instanceof Error ? err.message : String(err) });
     }
   };
@@ -113,10 +111,14 @@ export function ConfirmAction({
         ]
       : phase.kind === "success"
         ? [{ key: "enter", label: "continue" }]
-        : [
-            { key: "esc", label: "back" },
-            { key: "ctl+c", label: "quit" },
-          ];
+        : phase.kind === "running"
+          ? // Nothing listens for esc mid-action: an operation in flight is
+            // not abandoned by leaving the screen.
+            [{ key: "ctl+c", label: "quit" }]
+          : [
+              { key: "esc", label: "back" },
+              { key: "ctl+c", label: "quit" },
+            ];
 
   return (
     <Layout breadcrumb={breadcrumb} description={description} keyHints={hints}>
@@ -134,7 +136,7 @@ export function ConfirmAction({
             marginBottom={1}
           >
             <Text bold>{title}</Text>
-            <SummaryRows rows={rows} />
+            <KeyValueTable items={toItems(rows)} />
           </Box>
 
           {phase.kind === "confirm" && (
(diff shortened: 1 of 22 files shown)


notgitika:
Demo of the new `project build` and `project deploy` TUI screens (2:21, narrated): the TUI renders the same progress steps, glyphs and messages the plain CLI prints

- `project build` on the plain CLI, then the same build in the TUI 
- `project deploy` with the target picker: several targets → it asks, one target → it just deploys 

Demo: 
https://github.com/user-attachments/assets/65cf4ead-fd01-42c6-a13c-da30cc3aafc1

tejaskash (commented):
Verified locally: `tsc --noEmit` clean, touched tests pass (202/202). No blockers, a few low-severity notes inline.

src/handlers/project/deploy/screen.tsx
```diff
@@ -0,0 +1,179 @@
...
+}) {
+  const navigate = useNavigate();
+  const [chosen, setChosen] = useState<string>();
+
+  // A fresh project has no aws-targets.json yet: the list is empty and deploy
+  // provisions `default` on first run.
+  const targets = useQuery({
+    queryKey: ["project-targets", project.rootPath],
+    queryFn: () => core.projectManager.listTargets(project),
+  });
```
tejaskash, line 63: With the default `gcTime` this list is served from cache on a second visit, so `DeployConfirm` mounts and `immediate` fires `run()` before the refetch lands. If `aws-targets.json` gained a second target in between, `targetName` flips to `undefined`, the picker renders, and `DeployConfirm` unmounts while the deploy keeps running detached.

Since this is a local file the CLI reads fresh every time, `gcTime: 0` here (or gating on `!targets.isFetching`) would avoid it.

src/handlers/project/deploy/screen.tsx
```diff
@@ -0,0 +1,179 @@
...
+        query={targets}
+        loadingLabel="reading deployment targets…"
+        onBack={() => navigate(PROJECT_MENU)}
+      />
+    );
+  }
+
+  const declared = targets.data;
+  const targetName =
+    chosen ?? (declared.length <= 1 ? (declared[0]?.name ?? DEFAULT_TARGET_NAME) : undefined);
```
tejaskash, line 79: One declared target that is not named `default` gets deployed to here, whereas the CLI with no `--target` would provision and deploy `default`. The PR description says this is intended, flagging only so it stays a conscious choice.

src/handlers/project/ProjectGate.tsx
```diff
@@ -0,0 +1,112 @@
...
+  return useQuery({
+    queryKey: ["project", from],
+    queryFn: async () => {
+      const project = await core.projectManager.resolve({ filePath: from });
+      if (!project) throw new ProjectStateError(projectNotFoundMessage(from));
+      return project;
+    },
+    // A seeded project is authoritative — it is what the launching command ran
+    // against — so it is never refetched from the cwd.
+    ...(seed && { initialData: seed, staleTime: Infinity }),
```
tejaskash, line 30: Same cache shape as the targets query in the deploy screen: without a seed, a revisit reads `project.spec` from cache and `declaresNothingDeployable` and the `action` closure act on it before the refetch. Milder here since the backend re-reads from disk, but `gcTime: 0` would make both behave the same way.

src/components/ConfirmAction.tsx
```diff
@@ -13,62 +16,109 @@ export interface SummaryRow {
...
+  nextSteps,
   onDone,
+  doneLabel = "continue",
+  onCancel,
 }: ConfirmActionProps) {
   const navigate = useNavigate();
-  const [phase, setPhase] = useState<Phase>({ kind: "confirm" });
+  const cancel = onCancel ?? (() => navigate(-1));
+  const [phase, setPhase] = useState<Phase>({
+    kind: trigger.kind === "confirm" ? "confirm" : "idle",
```
tejaskash: Phase is fixed from `trigger.kind` once at mount. A caller that mounts with `isPending: true` and later switches the trigger from `immediate` to `confirm` would stay `idle` and run the action with no question asked. No current caller does this, but it is easy to trip on in a component meant to guard destructive actions. Either document that `trigger.kind` must be stable or resolve the phase when `isPending` flips false.

src/components/ConfirmAction.tsx
```diff
@@ -77,65 +127,99 @@ export function ConfirmAction({
...
+  result: Promise<ActionResult> | AsyncGenerator<ProgressEvent, ActionResult>,
+): result is AsyncGenerator<ProgressEvent, ActionResult> {
+  return (
+    typeof (result as AsyncGenerator<ProgressEvent, ActionResult>)[Symbol.asyncIterator] ===
+    "function"
+  );
+}
+
+function toItems(rows: SummaryRow[]): Record<string, string> {
+  return Object.fromEntries(rows.map((row) => [row.label, row.value]));
```
tejaskash: Converting to a `Record` drops what `SummaryRow[]` promises: duplicate labels collapse to the last value and integer-like labels get reordered by object key ordering. Nothing hits this today. If `KeyValueTable` is the renderer, taking `Record<string, string>` at the prop boundary would make the type match what is shown.

src/core/project/manager.tsx
```diff
@@ -919,13 +919,17 @@ export class FsProjectManager implements ProjectManager {
   // A read-only lookup, so callers (e.g. the deploy handler's up-front teardown
   // confirmation) can name the target's account and region without triggering
   // the default-target provisioning deploy performs.
+  public async listTargets(project: Project): Promise<AwsDeploymentTarget[]> {
```
tejaskash, line 922: `deploy` further up still reads `aws-targets.json` inline. It also needs the exists check for its error branches, so not a one-line swap, but the read itself could go through `listTargets` now that it owns that path.

src/components/Root.tsx
```diff
@@ -116,7 +118,7 @@ import type { Context } from "../router";
 // PROJECT_COMMANDS are the `agentcore project` subcommands that are listed in
 // the menu but have no screen of their own yet (`create` has the wizard). Each
```
tejaskash, line 119: Stale now: `invoke`, `build`, and `deploy` have screens too.

notgitika: this one is gone in the next PR

src/handlers/project/deploy/screen.tsx
notgitika: Agreed. I fixed it and added a regression test

src/handlers/project/deploy/screen.tsx
notgitika: yes, this was intended

src/handlers/project/ProjectGate.tsx
notgitika: fixed!

src/components/ConfirmAction.tsx
notgitika: Fixed by keeping the action in a waiting phase until loading completes, then snapshotting the latest trigger and confirmation message.

src/components/ConfirmAction.tsx
notgitika: fixed

src/core/project/manager.tsx
notgitika: done!

jariy17 (approved):

tejaskash (approved):

PR #2173, opened 2026-09-02
feat: interactive screen for gateway policy generate

tejaskash:
## What

- Interactive screen for `agentcore gateway policy generate`: pick a Gateway, see its attached Policy Engine, type a prompt, watch the generation steps, read the Cedar and findings. `e` edits the prompt and runs again.
- `policy` appears in the `agentcore gateway` menu. A bare `gateway policy generate` opens the picker, `--gateway-id X` without `--prompt` opens the form on that Gateway. `--json` and any run with `--prompt` stay headless.
- A Gateway with no Policy Engine attached gets an explanation instead of a prompt.
- `ErrorPanel` and `EventLog` promoted from private copies to `src/components` and reused.

## Testing

- `bun test src`: 2794 pass. Four screen tests through `renderScreen` (picker to form, no engine, happy path with exact Core input and rendered Cedar plus the `e` retry, error phase back to the form), the gateway menu test lists `policy`, one deep-link handler test with an injected renderer.
- Live drive in us-west-2 through the tui-harness against a deployed project with an attached and a bare Gateway: menu, picker, no-engine message, real generation with a `DENY_ALL` finding, `e` retry, esc while running, deep link, headless `--json` still errors on the missing prompt. Screenshot below.


src/handlers/gateway/policy/screen.tsx
```diff
@@ -0,0 +1,183 @@
...
+      if (aliveRef.current) setPhase({ kind: "result", result: next.value });
+    } catch (error) {
+      if (aliveRef.current) setPhase({ kind: "error", message: (error as Error).message });
+    }
+  };
+
+  useInput(
+    (input, key) => {
+      if (key.escape) {
+        navigate(-1);
```
notgitika: This is a no-op when launched directly through --gateway-id, because the MemoryRouter contains no previous history entry. Please navigate explicitly to /agentcore/gateway/policy/generate.

src/handlers/gateway/policy/screen.tsx
```diff
@@ -0,0 +1,183 @@
...
+    setEvents([]);
+    setPhase({ kind: "running" });
+    try {
+      const generation = core.policy.generatePolicy(
+        { gatewayId, prompt, name: `cli_generation_${Date.now()}` },
+        opts,
+      );
+      let next = await generation.next();
+      while (!next.done) {
+        if (!aliveRef.current) return;
```
notgitika: this prevents state updates but does not cancel or close the generator. The AWS waiter may continue polling after the user leaves. can we abort the operation on Esc/unmount and close the iterator, threading an AbortSignal through generatePolicy?

src/components/EventLog.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { darkTheme } from "./ui/_core.js";
+
+const theme = darkTheme;
+
+export function EventLog({ events }: { events: string[] }) {
+  return (
+    <Box flexDirection="column">
+      {events.map((message, index) => (
+        <Text key={`${index}-${message}`} color={theme.colors.muted}>
+          ✓ {message}
```
notgitika: A yielded progress step is the currently running step, not a completed one. This can display ✓ Waiting for generation to complete while it is still waiting. 

Also, can we reuse the existing TaskList task-state model instead?

src/handlers/gateway/policy/screen.tsx
tejaskash: Fixed in 9e3946ef: esc now navigates to `/agentcore/gateway/policy/generate` explicitly, verified live from a `--gateway-id` deep link.

src/handlers/gateway/policy/screen.tsx
tejaskash: Done in 9e3946ef: `generatePolicy` takes an optional `AbortSignal`, passed to every `send` and to the waiter's `abortSignal`. The screen aborts the controller and calls `return()` on the iterator on esc and on unmount. A screen test asserts the signal is aborted on esc while running, and a Core test asserts an aborted signal rejects with `UserCancellationError`.

src/components/EventLog.tsx
tejaskash: Agreed and done in 9e3946ef: the screen now uses `TaskList` with running/done/failed states, the same model `runWithProgress` uses. `EventLog` is no longer promoted, the create screen keeps its private copy untouched.

notgitika (approved):

PR #2174, opened 2026-09-02
feat(project): show create progress like build and deploy

aidandaly24:
`project create` printed a plain line per step while `build` and `deploy` rendered a live
step list. This runs create under the same driver, so all three look the same.

```
✓ Creating project tree
⠋ Installing CDK dependencies with npm
  │ added 320 packages, and audited 341 packages in 6s
  Syncing Python dependencies with uv
```

- `create` now goes through `runWithProgress`, the same driver `build` and `deploy` use.
- The four subprocess steps (npm install, uv sync, lockfile generation) stream their output
  as `output` events, so the tail under the running step shows real progress instead of
  sitting empty.

Non-TTY and `--json` keep the previous plain line-per-step output, byte for byte.

## Testing

- `bun test src` — 2789 pass, 0 fail
- `bun run typecheck`, `bunx oxlint src`, `bunx prettier --check src` clean
- Drove a real `project create` in a TTY: spinner, ✓ marks, and the output tail all render;
  measured no event-loop stall over 150ms across the run
- Verified `--json` and piped stderr produce output identical to before, and that a failing
  step renders `✕` with its tail preserved


src/core/project/manager.tsx
```diff
@@ -1086,6 +1088,27 @@ export class FsProjectManager implements ProjectManager {
...
+  private async *runStreaming(
+    command: string[],
+    cwd: string,
+  ): AsyncGenerator<ProjectEvent, void, unknown> {
+    yield* withOutputEvents((emit) => {
+      const lines = createLineSplitter(emit);
+      return this.runner(command, {
+        cwd,
+        onOutput: (chunk) => {
+          this.logger.debug(chunk);
```
Hweinstock, line 1126: OOS here since I think its existing behavior, but this is dumping raw output (with ANSI codes) into the logs which are not human readable. 

Ex. 
```
{
  "level": "debug",
  "msg": "\u001b[32m+\u001b[39m \u001b[1ms3transfer\u001b[0m\u001b[2m==0.19.2\u001b[0m\n \u001b[32m+\u001b[39m \u001b[1msix\u001b[0m\u001b[2m==1.17.0\u001b[0m\n \u001b[32m+",
  "time": 1788392133886,
  "cliSessionId": "cebc9a0c-9985-400f-b5e7-28c9221ce9cd",
  "version": "1.0.0",
  "module": "projectManager"
}
{
  "level": "debug",
  "msg": "\u001b[39m \u001b[1msse-starlette\u001b[0m\u001b[2m==3.4.8\u001b[0m",
  "time": 1788392133887,
  "cliSessionId": "cebc9a0c-9985-400f-b5e7-28c9221ce9cd",
  "version": "1.0.0",
  "module": "projectManager"
}
{
  "level": "debug",
  "msg": "\n ",
  "time": 1788392133887,
  "cliSessionId": "cebc9a0c-9985-400f-b5e7-28c9221ce9cd",
  "version": "1.0.0",
  "module": "projectManager"
}
{
  "level": "debug",
  "msg": "\u001b[32m",
  "time": 1788392133887,
  "cliSessionId": "cebc9a0c-9985-400f-b5e7-28c9221ce9cd",
  "version": "1.0.0",
  "module": "projectManager"
}
```

src/handlers/project/create/index.ts
```diff
@@ -300,9 +301,12 @@ export const createCreateProjectHandler = (config: CreateProjectHandlerConfig) =
         );
       }
 
-      for await (const event of config.projectManager.create(createInput)) {
-        if (event.type === "step") config.io.stderr.write(`${event.message}\n`);
-      }
+      // Same driver as build and deploy: a live step list in a TTY, and the previous plain
```
Hweinstock, line 304: nit: i don't feel like this comment is relevant.

Hweinstock (approved):
Thanks for fixing this!

tejaskash (approved):

src/core/project/manager.tsx
aidandaly24: Ahhhh this is a good call out. I can fix this in a follow up. I don't think that this was pre existing.

PR #2175, opened 2026-09-03
feat(templates): wire up non-bedrock model providers to templates

Hweinstock:
### Problem 
non-bedrock model providers are not wired up with the existing templates. 

### Solution
- Update the flag schemas to accept all model providers accepted by the CLI on mainline (anthropic, openai, gemini, liteLLM). We also must handle the mismatch of accepted providers in `project create` between runtime and harness flows. 
- route those fields and their api-keys to the template code. 
- write entries to `.env.local` for those values so that local dev works. 
- update templates to include necessary env entries to ensure credentials are created as part of deploy. 

### Verification
Starting with a gemini key in my local .env file:
```
$ cat .env | grep GEMINI_API_KEY | cut -d'=' -f2 > gemini.key
...
$ agentcore project create --name GEMINI --template agent-python-strands --model-provider gemini --api-key file://gemini.key --skip-git
... 
$ cd GEMINI
...
$ agentcore project dev --mode headless & 
...
$ curl -X POST localhost:8081/invocations -d '{ "prompt": "what model are you?" }'
data: {"event": {"messageStart": {"role": "assistant"}}}

data: {"event": {"contentBlockDelta": {"delta": {"text": "I am a large language model, trained by Google.\n"}}}}

data: {"event": {"contentBlockStop": {}}}

data: {"event": {"messageStop": {"stopReason": "end_turn"}}}

data: {"event": {"metadata": {"usage": {"inputTokens": 477, "outputTokens": 38, "totalTokens": 515}, "metrics": {"latencyMs": 0}}}}
$ agentcore project deploy 
...
$ agentcore project invoke runtime --name agent_python_strands --payload '{ "prompt": "what is 2+2 and what is your name" }'
data: {"event": {"messageStart": {"role": "assistant"}}}

data: {"event": {"contentBlockStart": {"start": {"toolUse": {"name": "add_numbers", "toolUseId": "add_numbers"}}}}}

data: {"event": {"contentBlockDelta": {"delta": {"toolUse": {"input": "{\"b\": 2, \"a\": 2}"}}}}}

data: {"event": {"contentBlockStop": {}}}

data: {"event": {"contentBlockStop": {}}}

data: {"event": {"messageStop": {"stopReason": "tool_use"}}}

data: {"event": {"metadata": {"usage": {"inputTokens": 450, "outputTokens": 101, "totalTokens": 551}, "metrics": {"latencyMs": 0}}}}

data: {"event": {"messageStart": {"role": "assistant"}}}

data: {"event": {"contentBlockDelta": {"delta": {"text": "I"}}}}

data: {"event": {"contentBlockDelta": {"delta": {"text": " am a large language model, and 2+2 = 4."}}}}

data: {"event": {"contentBlockStop": {}}}

data: {"event": {"messageStop": {"stopReason": "end_turn"}}}

data: {"event": {"metadata": {"usage": {"inputTokens": 490, "outputTokens": 16, "totalTokens": 506}, "metrics": {"latencyMs": 0}}}}

status=200 content-type=text/event-stream; charset=utf-8 runtime-session-id=0bc3ea8a-13f9-43a2-af64-c13b121dd48d mcp-session-id=- mcp-protocol-version=- trace-id=- trace-parent=- trace-state=- baggage=- complete=true bytes=1023
```
- had my agent verify the other ones + with container build. 

### Notes
Non container python runtimes, and all TS runtimes require a user id (`X-Amzn-Bedrock-AgentCore-Runtime-User-Id`) header since otherwise the workload access token used to fetch the api key is not injected. 

The python non-container handles this with a special case in the [SDK](https://github.com/aws/bedrock-agentcore-sdk-python/blob/82eddf32c7134e549c630f1a29bb59697b7dd596/src/bedrock_agentcore/identity/auth.py#L300). 

The typescript does not. [here](https://github.com/aws/bedrock-agentcore-sdk-typescript/blob/v0.3.0/src/identity/wrappers.ts#L94-L110)

This is never experienced through the old CLI since we default this value https://github.com/aws/agentcore-cli/blob/543a6da3925bfc25468292f731e37883a3378303/src/cli/aws/agentcore.ts#L57. 

However, I'm not confident this is the right behavior, so I left it out here. 

Right now it requires doing this:
```
agentcore project invoke runtime --name agent_typescript_strands \
    --payload '{ "prompt": "hello world" }' \
    --accept text/event-stream \
    --user-id tester-1
```

i.e. providing the user id, and an explicit accept flag. 

The accept flag is only required in TS SDK due to a check here: https://github.com/aws/bedrock-agentcore-sdk-typescript/blob/efef021a7a75aebefccb5111bc0427b5f4c1cbfb/src/runtime/app.ts#L371-L382, whereas Python ignores it completely and forces `text/event-stream`. 

Python code that ignores it https://github.com/aws/bedrock-agentcore-sdk-python/blob/82eddf32c7134e549c630f1a29bb59697b7dd596/src/bedrock_agentcore/runtime/app.py#L573-L579. 

### Future Work
- After this PR, I plan to slim down the templates significantly. 
- two rough edges to address as follows: determine how to fix the userId issue, determine how to fix the required explicit `accept` param for ts only.


notgitika (approved):
Good work! LGTM with the rough edges defined in the description let's have that as a follow up

nborges-aws (approved):
LGTM no findings!

PR #2177, opened 2026-09-03
feat(project): command-line-only screens for project commands without a TUI

notgitika:
## What

Every `project` command without a TUI screen now opens a **command-line-only** screen instead of tearing the TUI down with an error. It says so, and shows the command's help — usage, options, arguments, and the "Parameter details" that `--help` appends — read from the same Commander help the CLI prints. Groups (`add`, `export`) open their menu, so every resource beneath them is reachable the same way. Esc returns to the parent menu.

The project menu lists screen-backed commands first, then a `command line only` divider, then the rest:

```
 ❯ create   create a new AgentCore project
   deploy   deploy the project to AWS
   invoke   invoke a Runtime or Harness from the current project
   build    build the project's deployable artifacts
── command line only ───────────────────────────────────────────
   add      add project resources
   export   convert project resources into editable code you own
   remove   remove a resource from the project
   dev      run the project locally for development
   status   show the status of the project's deployed resources
```

## How

- `project` Router declares its screen-backed children via the existing `supportedTuiCommands(...)` (what `eval`/`identity` use); `compile()` propagates it to nested commands.
- `RouterScreen` partitions on `isTuiCommandSupported` instead of filtering, behind a `showCliOnly` prop. Other menus are unchanged.
- New `CliOnlyScreen` (`Layout` + `KeyValueTable` + `ScrollView`, over `command.createHelp()`), and one wildcard route `agentcore/project/*` replacing Root's hand-written list and `ProjectCommandNotImplementedScreen`.
- `KeyValueTable` wraps values under themselves and caps the key column at 50%.

`dev` is CLI-only here; a TUI screen for it is a follow-up.

Formerly stacked on #2172; rebased onto `refactor` now that it has merged, so this PR is the single feature commit plus one small fix:

- `fix(project): satisfy ProjectBackend in the build/deploy screen test fake` — `buildDeploy.screen.test.tsx` on `refactor` predates `resolveProjectResources` landing on the `ProjectBackend` interface, so `tsc --noEmit` fails on the base branch today. The fake gets the stub the interface requires.

Also resolved the CodeQL `js/bad-tag-filter` finding on the test regex (it is a terminal-layout assertion, not an HTML filter; made it case-insensitive so the rule is satisfied).

## Testing

`bun test` 2848 pass; `tsc`, `oxlint`, `prettier` clean. Tests read the CLI-only commands off the Commander tree and assert each shows its usage and every option; plus divider order, `add → gateway` drill-down, scrolling to the parameter details, a small-terminal sweep, and resize-after-scroll.

### DIFF
--- a/src/components/CliOnlyScreen.tsx
+++ b/src/components/CliOnlyScreen.tsx
@@ -1,0 +1,149 @@
+import { useRef } from "react";
+import { Box, Text, useInput, useWindowSize } from "ink";
+import { ScrollView, type ScrollViewRef } from "ink-scroll-view";
+import type { Command } from "commander";
+import { useLocation, useNavigate } from "react-router";
+import { CommandKey, commandParameterDetails } from "../router";
+import type { ScreenProps } from "../handlers/types";
+import { Layout } from "./Layout";
+import { KeyValueTable } from "./KeyValueTable";
+import { RouterScreen, resolveCommand } from "./RouterScreen";
+import { darkTheme } from "./ui/_core.js";
+
+const theme = darkTheme;
+
+export interface CliOnlyScreenProps extends ScreenProps {
+  // path is the command's path, e.g. ["agentcore", "project", "dev"].
+  path: string[];
+}
+
+// CliOnlyScreen stands in for a command that has no screen of its own: it says
+// so, and shows the command's help — usage, arguments, options, parameter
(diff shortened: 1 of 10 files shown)


AlexanderRichey (approved):

notgitika:
flaky unit test i just reran it and it passed

tejaskash (approved):

PR #2181, opened 2026-09-03
perf(build): cut CLI startup time 

tejaskash:
## Summary

Profiling `agentcore --version` showed ~10 ms of actual work inside ~620 ms of wall time. This PR removes the three largest self-inflicted costs in the build.

1. **Dedupe `@smithy/core`.** package.json pinned it at `3.29.3` while every `@aws-sdk/client-*` wanted `^3.33.3`, so bun nested 51 private copies and the bundler inlined all of them (34 copies of `httpAuthSchemeMiddleware` in the bundle). Bumped to `^3.33.3`. Bundle: 9.2 MB → 5.6 MB.
2. **Production React.** `NODE_ENV` was never defined at bundle time, so both the npm bundle and the compiled binary carried React's development build (`runWithFiberInDEV` etc.). Both builds now `define` it as `production`.
3. **V8 compile cache.** `dist/index.js` is now a four-line loader that calls `module.enableCompileCache()` and imports the bundle at `dist/main.js`. V8 only caches modules loaded *after* the call, so a single-file bundle cannot cache its own compilation. No-op on Node < 22.1. Bun inlines relative imports even when marked `external`, so the build script writes the loader directly.

`bin`, `main`, `node dist/index.js`, and the Makefile/README references are unchanged: `dist/index.js` is still the entry, it just delegates.

## Measurements

Artifact size:

| Artifact | Before | After |
|---|---|---|
| npm bundle (`dist/main.js`) | 9.24 MB | 5.58 MB (−40%) |
| darwin-arm64 binary | 85.1 MB | 76.0 MB (−11%) |

`agentcore --version`, median of 7 runs, telemetry enabled:

| Runtime | Before | After |
|---|---|---|
| Node 20.20 | 620 ms | 475 ms |
| Node 24.15 | 478 ms | 407 ms |
| darwin-arm64 binary | 652 ms | 442 ms |

With `AGENTCORE_TELEMETRY_DISABLED=1` (isolates startup from the exit flush): Node 24 311 ms → 246 ms, binary → 300 ms.

## Not in this PR

- **Telemetry exit flush** (140–180 ms, network-bound HTTPS POST in `OtelHistogramSink.shutdown`) is now the largest remaining cost. Needs a design decision (spool-and-forward vs. detached flusher); follow-up.
- **Bun bytecode for the binary** is not possible: it requires CJS output, and `ink` and `yoga-layout` use top-level await.

## Verification

- `bun test`: 2803 pass, 0 fail
- `bun run typecheck`, `bun run lint:check`, `prettier --check`: clean
- `bun run build` and `bun run compile:darwin-arm64`: both artifacts run `--version` and render the interactive TUI (root menu, harness submenu) with the production React build
- Compile cache confirmed populated under `os.tmpdir()/node-compile-cache` on Node 24

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -19,7 +19,7 @@
         "@opentelemetry/otlp-transformer": "0.213.0",
         "@opentelemetry/resources": "^2.10.0",
         "@opentelemetry/sdk-metrics": "^2.10.0",
-        "@smithy/core": "3.29.3",
+        "@smithy/core": "^3.33.3",
         "@tanstack/react-query": "^5.101.2",
         "cli-truncate": "^6.1.1",
         "commander": "^15.0.0",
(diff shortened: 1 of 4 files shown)


AlexanderRichey (approved):
Nice finds!

PR #2186, opened 2026-09-03
fix(project): deploy payments through the published AgentCorePayments construct

aidandaly24:
## What

Retires the Payment orchestration the generated CDK app has been carrying, and makes `project deploy` print a Quick Create connector's authorization URL while the service makes one available.

## Why

`#2146` restored a pre-Quick-Create adapter into the generated app to unblock scaffolding while `AgentCorePayments` was unpublished. That construct now ships in `@aws/agentcore-cdk@0.1.0-alpha.51`, so the vendored loops are obsolete. While they remain:

- `bin/cdk.ts` requires every connector to resolve a `credentialName` to a deployed ARN. A Quick Create connector deliberately has neither, so synthesis throws before constructing the stack.
- `cdk-stack.ts` concatenates underscore-stripped manager and connector names into construct IDs, allowing distinct names to collide.

Quick Create is also a two-phase operation: deployment creates the connector, then someone must follow its short-lived authorization URL. A successful deploy that does not print that URL leaves the connector unusable.

## How

`bin/cdk.ts` passes the raw spec and target credentials through to `AgentCorePayments`. The published construct owns connector variant mapping, credential resolution, runtime wiring, IAM, and outputs instead of the frozen generated scaffold.

After a successful deploy, the CLI:

1. Lists the resources in that project's CloudFormation stack.
2. Reads the manager and connector IDs from each Payment Connector physical ARN.
3. Calls `GetPaymentConnector`.
4. Prints `authorizationUrl` when the response contains one.

That is the complete post-deploy behavior. It does not implement connector status reporting or recovery guidance. A project without Quick Create connectors performs no additional reads, and a read failure never turns a completed deployment into a failure.

## Verification

Verified the service contract directly against account `603141041947` in `us-east-1`:

- Created a disposable CoinbaseCDP Quick Create connector.
- `CreatePaymentConnector` returned `PENDING_AUTHENTICATION` with an authorization URL.
- A separate `GetPaymentConnector` returned the same status and URL.
- An existing `READY` connector returned no authorization URL.
- Deleted the disposable connector and confirmed it was gone.

Code validation:

- Focused backend and authorization URL tests: **43 passed**
- Generated CDK tests: **2 passed**
- `bun run lint`
- `bun run format:check`
- `bun run build`

The earlier full lifecycle verification still covers:

- Quick Create authoring, synthesis, deployment, and authorization to `READY`
- Manual credential provisioning and connector deployment to `READY`
- Idempotent redeploy
- Connector removal preserving reusable credentials
- `remove all` deleting the stack, manager, connector, and credential provider

## Follow-ups

- `aws/agentcore-l3-cdk-constructs#348` is merged but unreleased. It removes ephemeral `AuthorizationUrl` and `ConnectorStatus` stack outputs that can break later stack updates. This PR still pins alpha.51; the pin moves to alpha.52 once published. The CLI requires no behavioral change because it reads the URL from `GetPaymentConnector`.
- The released CLI's separate vendored Payment implementation emits the same ephemeral outputs and needs the equivalent removal.


aidandaly24:
Rebased onto `2e30e1f2`.

The `typecheck` job fails on this base for a reason outside this PR: `refactor`'s tip does not typecheck on its own.

```
src/handlers/project/buildDeploy.screen.test.tsx(40,9): error TS2741:
  Property 'resolveProjectResources' is missing in type '{ build(); deploy(); resolveDeployedResources(); }'
  but required in type 'ProjectBackend'.
```

`11fb8b5d` added `resolveProjectResources` to `ProjectBackend`, and #2172's new `buildDeploy.screen.test.tsx` stub does not implement it. Reproduced in a clean worktree at `origin/refactor` with none of this branch's code present, and neither implicated file is touched here. It is test-only, so the CLI still builds and deploys.

That is the only typecheck error in the tree. Everything else on this branch is green: `lint`, `format:check`, `bun test src` (2853 passed, 211 files), `bun run build`, and the generated package's own `npm test` against the pinned L3.

Verified end to end against a real account since the last push, including the paths that only real AWS can exercise:

- Quick Create deploy prints the live authorization link and its expiry window
- completing the provider flow moves the connector to `READY`
- an expired connector is reported with recovery guidance, and recovery by recreating the connector mints a fresh link (confirmed: neither `UpdatePaymentConnector` nor a repeat deploy mints one)
- manual connector reaches `READY` through its pre-synth provisioned credential
- redeploy idempotent; removing a connector preserves its reusable credential; `remove all` deletes stack, manager, connector and credential provider

src/core/project/backends/cdk/paymentConnectorAuthorizationUrls.ts
```diff
@@ -0,0 +1,93 @@
...
+            resource.ResourceType !== PAYMENT_CONNECTOR_RESOURCE_TYPE ||
+            !resource.PhysicalResourceId
+          ) {
+            continue;
+          }
+
+          const ids = paymentConnectorIds(resource.PhysicalResourceId);
+          if (!ids) continue;
+
+          const connector = await paymentsClient.send(
```
notgitika: Could we catch `GetPaymentConnector` failures per resource and continue processing?

Currently, one stale or throttled connector lookup stops the entire loop, preventing authorization URLs from being reported for later Quick Create connectors. This can also happen if an unrelated manual connector lookup fails.

Please add a test where the first lookup fails and a subsequent connector still reports its authorization URL.

aidandaly24: Updated in `1bf77c64`. The catch now wraps only the individual `GetPaymentConnector` SDK request, emits a connector-scoped warning on failure, and continues processing later resources. I added the requested regression where the first lookup throws and the second still reports its authorization URL. Focused backend tests pass 43/43; lint, format, and build also pass.

notgitika (approved):

src/assets/cdk/package.json
```diff
@@ -23,7 +23,7 @@
     "typescript": "~5.9.3"
   },
   "dependencies": {
-    "@aws/agentcore-cdk": "0.1.0-alpha.45",
+    "@aws/agentcore-cdk": "0.1.0-alpha.51",
```
AlexanderRichey, line 26: I presume this is the crucial bit. Do we know how these changes will impact existing projects? There's a lot of changes here.

aidandaly24: The changes should only be additive. We haven't made any breaking changes to the CDK. The main version of the CLI also already uses up to `0.1.0-alpha.50`.

aidandaly24:
I had thought this had been approved by two people 😬 can revert the merge if needed.

PR #2194, opened 2026-09-03
feat(tui): command-line-only screens for every command without a TUI

notgitika:
Follow-up to #2177, for everything outside `project`.

Every menu now lists its subcommands without a screen below a `command line only` divider, and selecting one opens its help via the existing `CliOnlyScreen`. Previously these were either hidden from the menu (`create`/`update`/`delete`, `runtime logs`/`traces`, most `eval` mutators — 67 commands) or listed but exited the TUI with a help dump (`feedback`, `config`, `update`, `eval ondemand`/`recommendation`, `batch-evaluation evaluate`/`simulate`).

- `RouterScreen` always partitions; the `showCliOnly` prop is gone.
- One `*` route → `CommandFallbackScreen` replaces the `project/*` wildcard and `HelpScreen` (deleted).
- `root`, `eval`, `batch-evaluation` declare their screen-backed children via `supportedTuiCommands`, like the other routers already do, so the menu labels them correctly. `feedback`/`config`/`update` never used the TUI middleware, so no CLI behaviour changes.
- `menuEntries` test helper reads a menu frame into `{screens, cliOnly}`; the existing "offers only read-only commands" tests now assert placement instead of absence.
- New `CliOnlyScreen.test.tsx` walks the compiled Commander tree and checks every command-line-only command (currently 94) opens its menu or help and `esc` returns to the parent.
- `KeyValueTable`'s parent is now a column box so long option terms don't wrap on wide terminals.

`bun test` 2944 pass; `tsc`, `oxlint`, `prettier` clean.


src/components/Root.tsx
```diff
@@ -772,15 +772,9 @@ export function Root({ path, ctx, core, queryClient }: RootProps) {
...
           />
-          {/* Every other project command: a group opens its menu, a leaf its
-              help, so a command added later needs no route here. */}
-          <Route
-            path="agentcore/project/*"
-            element={
-              <CommandFallbackScreen ctx={ctx} core={core} basePath={["agentcore", "project"]} />
-            }
-          />
-          <Route path="*" element={<HelpScreen ctx={ctx} core={core} />} />
```
AlexanderRichey, line 783: Are we sure we want to remove the HelpScreen fallback?

AlexanderRichey (changes requested):
Looks good! But we hav some merge conflicts and I'm not sure we want to remove the HelpScreen fallback.

src/components/Root.tsx
notgitika: Hmm good question. I think we should preserve it because if an unknown or a new route is introduced, we should have that as a fallback. Let me revert it

notgitika: added help screen back

AlexanderRichey (approved):

nborges-aws (approved):

PR #2198, opened 2026-09-03
feat(runtime): add interactive shell

aidandaly24:
## Summary

Adds `agentcore runtime shell`, a native interactive PTY session for AgentCore
Runtimes using the currently published TypeScript SDK (`bedrock-agentcore`
`0.4.3`).

- Bare `runtime shell` opens Runtime and endpoint pickers.
- `--id` skips the Runtime picker.
- `--id` with `--qualifier` connects directly.
- `--session-id` opens the shell in a specific Runtime session.
- IAM and CUSTOM_JWT Runtimes are supported.
- Raw input, `Ctrl+C`, `Ctrl+D`, terminal resize, ANSI/full-screen
  applications, transparent reconnect, and remote exit status are preserved.
- Shells launched from the TUI return to their originating Runtime picker,
  endpoint picker, or Runtime details screen when the native terminal ends.

The ownership boundaries are intentional:

- **Handler:** target/auth policy and orchestration.
- **Core:** remote shell connection.
- **IO:** local terminal ownership.

The Runtime Shell screen uses Ink's `suspendTerminal()` API so Ink releases raw
input and the alternate screen before IO takes terminal ownership. When the
shell ends, Ink restores the same TUI instance and the screen navigates back to
the route that launched it.

## Command

```text
agentcore runtime shell
agentcore runtime shell --id <runtimeId>
agentcore runtime shell --id <runtimeId> --qualifier DEFAULT
agentcore runtime shell \
  --id <runtimeId> \
  --qualifier DEFAULT \
  --session-id <runtimeSessionId>
```

Additional behavior:

- `--bearer-token` accepts inline or `file://` CUSTOM_JWT tokens.
- stdin token sources are rejected because the PTY owns stdin.
- `--json` and `--endpoint-url` are rejected.
- Running `exit` or sending `Ctrl+D` terminates the remote shell.

## Validation

- `CI=true bun test --coverage --coverage-reporter=lcov`
- Focused Runtime Shell/TUI suite
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- `bun run secrets:check`
- `bun run build`

Live TUI Harness validation against a READY IAM Runtime in `us-west-2` covered:

- direct and picker-based launch
- raw command input and exact output
- working-directory and environment persistence
- `Ctrl+C` interrupt without disconnecting
- ANSI clear-screen output
- clean `exit` with exit code 0
- return to the Runtime picker, endpoint picker, and Runtime details origin

## Stack

This PR is independently mergeable and does not require a TypeScript SDK
release. Follow-up PR #2199 adds detach and reattach after the corresponding SDK
behavior is published.


src/io/interactiveTerminal.ts
```diff
@@ -0,0 +1,112 @@
...
+    };
+    const close = async () => {
+      if (closed) return;
+      closed = true;
+      await peer.close();
+    };
+    this.stopCurrent = close;
+
+    const onData = (chunk: Buffer | string) => {
+      const bytes = Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk, "binary");
```
notgitika: ink leaves stdin utf-8 encoded, so input can arrive as a string. using `"binary"` corrupts non-ascii characters. can we consume buffers directly or encode strings as utf-8, with a unicode test?

src/handlers/runtime/shell/operation.ts
```diff
@@ -0,0 +1,64 @@
...
+  if (options.endpointUrl !== undefined) {
+    throw new InputValidationError("runtime shell does not support --endpoint-url");
+  }
+  const detail = await core.runtime.getRuntime(runtimeId, options);
+  const request = normalizeRuntimeShellRequest(detail, {
+    qualifier,
+    runtimeSessionId: launchContext?.runtimeSessionId,
+    bearerToken: launchContext?.bearerToken,
+  });
+  request.onReconnect = () => io.stderr.write("\r\nReconnected to shell.\r\n");
```
notgitika: the sdk passes whether the existing shell was actually reattached. when false, a fresh shell was created and the prior state was lost. can we preserve this boolean and display an accurate message?

src/tui/index.tsx
```diff
@@ -56,14 +57,22 @@ export async function renderTuiAt(
...
   // takes over the screen and the prior scrollback is restored on exit (like Vim).
-  const { waitUntilExit } = render(<Root path={path} ctx={ctx} core={core} />, {
-    stdin: io.stdin,
-    stdout: io.stdout,
-    stderr: io.stderr,
-    alternateScreen: true,
-    incrementalRendering: true,
-  });
-  await waitUntilExit();
+  let nextPath: string | undefined = path;
```
notgitika: can we use ink’s existing `suspendTerminal()` api here? it already handles terminal handoff and restoration, allowing us to remove the custom controller and remount loop.

src/core/runtimeShell.ts
```diff
@@ -0,0 +1,152 @@
+import { RuntimeClient as AgentCoreRuntimeClient } from "bedrock-agentcore/runtime";
+import type { AwsCredentialIdentityProvider } from "@smithy/types";
+import { Buffer } from "node:buffer";
+import type { RuntimeShellFrame, RuntimeShellSession } from "../handlers/runtime/types";
+import type { OpenRuntimeShell } from "./runtime";
+import type { CoreOptions } from "./types";
+
+export type RuntimeShellSdkOpenInput = {
```
notgitika: can these interfaces and protocol constants be derived from the sdk’s exported types and values? mirroring them locally risks drift—the reconnect callback signature has already lost its boolean argument.

src/io/interactiveTerminal.ts
aidandaly24: Fixed. `InteractiveTerminal` now encodes string chunks as UTF-8 while preserving `Buffer` chunks byte-for-byte. I added a Unicode regression test using `café 🙂`.

src/handlers/runtime/shell/operation.ts
aidandaly24: Fixed. The Core callback now preserves the SDK boolean and async signature. The CLI reports `Reattached to existing shell.` when state was preserved and `Previous shell unavailable; started a new shell.` when the SDK created a fresh PTY, with command-path tests for both outcomes.

src/tui/index.tsx
aidandaly24: Updated to use `useApp().suspendTerminal()`. The shell screen now hands Ink’s stdin/stdout/stderr to the native terminal callback, resumes the same TUI instance, and navigates back to the picker or Runtime details. The custom handoff controller and render/remount loop were removed, while direct errors still propagate through `exit(error)`.

src/core/runtimeShell.ts
aidandaly24: Updated. The adapter now derives its input, frame, and session surface from the SDK’s exported `OpenShellParams`, `ShellFrame`, and `ShellSession`, and uses exported `ShellChannel` and `MAX_FRAME_SIZE` values. The build plugin exposes only the SDK client and protocol modules so packaged binaries do not pull in the Runtime server dependency tree. The CLI Core interface remains the ownership boundary, with the reconnect boolean preserved.

notgitika (approved):
LGTM

AlexanderRichey (approved):
A lot going on in this PR! I presume a lot of this is a copy/paste job from the old CLI?

aidandaly24:
@AlexanderRichey that is mostly the case the features are all pretty much identical, the code was mainly just rearchitected into the new repo's system.

PR #2201, opened 2026-09-03
feat(project): add agent-python-strands-container template

tejaskash:
## Summary

The TUI create wizard has no build question, so a Container build of the Strands agent was unreachable from it. This adds an `agent-python-strands-container` template preset: `agent-python-strands` with `build: Container`, which keeps the `Dockerfile` and `.dockerignore` in the scaffold.

- `shortcuts.ts`: the preset. The `--template` enum is derived from the table, so the flag accepts it automatically.
- `create/screen.tsx`: the wizard option, placed after the recommended Strands entry. The three memory-step gates that hardcoded `"agent-python-strands"` are now one `asksMemory` helper keyed on the preset's framework.
- README: one example line.

On the CLI this is a second spelling of `--template agent-python-strands --build Container`. The preset exists for the TUI, and the parameterized test asserts both spellings scaffold identically.

**Behavior change:** because the memory gate keys off framework, `a2a-python-strands` now also gets the memory question in the wizard. Its resolver supports memory and its preset already defaults to long and short-term, matching the flag path.

## Testing

- `bun test`: 2854 pass. `tsc --noEmit` clean.
- Flags: built `dist/index.js`, ran `project create --template agent-python-strands-container`. Dockerfile and `.dockerignore` present, CMD rendered with `opentelemetry-instrument`, spec has `build: Container` and no `runtimeVersion`, memory with four strategies, `uv.lock` generated.
- TUI: drove the wizard through the TUI harness. New option renders, selecting it shows the memory step, review shows the template, project created. The TUI scaffold and the flag scaffold diff identically apart from the project name.
- Live deploy to us-east-1: runtime reached READY with an ECR container URI, `project invoke runtime` returned the expected reply over the event stream. Torn down with `project remove all --yes` then `project deploy --yes`. Stack and ECR repository gone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,8 @@ agentcore harness invoke --id <id from the deploy outputs> --prompt "hello"
 
 # Scaffold runtime code instead (pass a template or framework flags).
 agentcore project create --name MyAgent --template agent-python-strands
+# The same Strands agent built as a container image, with a Dockerfile.
+agentcore project create --name MyAgent --template agent-python-strands-container
 
 # Translate an existing Amazon Bedrock Agent version into editable runtime code.
 # The selected alias identifies the immutable source version; generated code
--- a/src/handlers/project/create/create.screen.test.tsx
+++ b/src/handlers/project/create/create.screen.test.tsx
@@ -364,38 +364,45 @@ describe("project create wizard", () => {
     r.unmount();
   }, 10000);
 
-  test("template flow: the container strands template asks about memory", async () => {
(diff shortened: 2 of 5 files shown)


tejaskash:
Addressed in 57110acc: the CodeZip-override fallback now reads `LANGUAGE_VERSION_DEFAULTS[template.language]`, and the container wizard test is a `test.each` that also drives `a2a-python-strands` through the memory step.

Hweinstock (approved):

PR #2202, opened 2026-09-03
fix: allow arrow key navigation while filtering

nborges-aws:
## Description

The current `DataTable.tsx` used by pagination does not allow arrow key usage while filtering. The user has to click enter and then interact with the filtered items. The UI is confusing from the user perspective, as the first item in the table stays in focus while filtering. This gives the impression that the cursor is currently on the first list item. 

## Example before fix
* my cursor is in the "filter" input, but first harness still focused in the UI
* arrow key navigation is a no-op

image

## Example after fix: 
* table items lose focus when input is active
* User can click enter or down arrow to navigate to first table item, at which point item gains focus styling
* Filter input loses focus styling once enter or down arrow is clicked
* User can click "/" or scroll to top with up arrow to return to filter menu, at which point focus styles reset

**After example:**

https://github.com/user-attachments/assets/562c4bc7-bccb-45dd-b466-4a28f981cd2b

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] `bun run test` (x pass, 0 fail)
- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.

### DIFF
--- a/src/components/PaginatedTablePicker.test.tsx
+++ b/src/components/PaginatedTablePicker.test.tsx
@@ -303,6 +303,91 @@ describe("paginated table picker contract", () => {
     expect(r.lastFrame()).toContain("agentcore → harness → get → alpha-1");
   });
 
+  test("moves focus from the Runtime filter into its matching rows", async () => {
+    const core = new TestCoreClient();
+    core.runtime.setListResponse({
+      agentRuntimes: [
+        runtime({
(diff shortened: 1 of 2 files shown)


notgitika (approved):
LGTM thanks for the neat solution and the demo attached!

PR #2206, opened 2026-09-03
ci: add release-prepare and release-publish workflows for npm releases

tejaskash:
## Summary

Two workflows give the refactor branch a PR-gated npm release process for `@aws/agentcore`, in the shape of the Python SDK's release-prepare and release-publish pair.

- **`release-prepare.yml`**, `workflow_dispatch` with `bump` (major, minor, patch) and `channel` (rc, stable). Bumps package.json with `bun pm version`, refreshes the vended `@aws/agentcore-cdk` pin in the template, force-pushes `release/v<version>` and opens the PR as the automation App. The PR is the only release gate.
  - Version arithmetic: `0.28.1` → `1.0.0-rc.1` → `1.0.0-rc.2` → `1.0.0` → `1.1.0-rc.1`. `bump` is ignored while an rc series is open. Series boundaries set the version explicitly: Bun's prerelease counter starts at 0 and its `pm version major` on `1.0.0-rc.3` yields `2.0.0`.
- **`release-publish.yml`**, `pull_request: closed` into refactor for merged `release/v*` PRs. Runs check, build and unit-test on the merge commit, then on `ubuntu-latest` builds, packs, cross-compiles six binaries, `npm publish --provenance` under `rc` for prereleases or `latest` otherwise, and creates the GitHub release with generated notes, tarball and binaries. Publish is skipped if the version is already on npm, so a rerun after a failed release step completes. 
- **`scripts/sync-vended-cdk.ts`**, Bun port of #2118's pin sync. Dependency-free because it runs before `bun install`.
- **package.json** renamed to `@aws/agentcore`, set to the last published version `0.28.1`, with the `repository` field provenance requires.
- **Telemetry fix found by the first dry run**: `src/telemetry/shapes.tsx` only accepted `X.Y.Z`, so a binary versioned `1.0.0-rc.1` printed its version and exited 1 at startup. Now accepts a prerelease suffix, length-capped like the other attributes, with a `test.each` for accepted and rejected shapes.

No GitHub environments and no concurrency groups. Publish runs on a GitHub-hosted runner because npm refuses provenance from self-hosted ones.

## Verification

`workflow_dispatch` only lists workflow files that exist on main, and these two do not yet, so neither could be dispatched from this branch. Verified instead:

- `actionlint` with shellcheck clean on both workflows.
- Version bump shell simulated locally for all five transitions listed above, each producing the expected version.
- `scripts/sync-vended-cdk.ts` run with node_modules absent: bumps `alpha.45` → `alpha.51`, is a no-op on the second run, exits 1 on an unknown dist-tag.
- The build, pack, cross-compile and artifact steps are unchanged from the earlier single-file version, which ran end to end in dry-run mode: https://github.com/aws/agentcore-cli/actions/runs/33810416086 (artifact held `aws-agentcore-1.0.0-rc.0.tgz` with the stamped version and six binaries). The release PR path also ran end to end then: https://github.com/aws/agentcore-cli/actions/runs/33810398072 opened #2203, and https://github.com/aws/agentcore-cli/actions/runs/33816972577 re-dispatched against an existing PR without failing.
- After the telemetry fix, a local bundle stamped `1.0.0-rc.1` runs `--version` and exits 0.

## Review fixes after the rebuild

- Release notes only consider this workflow's own tags (`vX.Y.Z` and `vX.Y.Z-rc.N`). Without the filter, the first rc's notes would have started at main's `v1.0.0-preview.29`, which sorts just below `1.0.0-rc.1`. Verified against the real tag list: `1.0.0-rc.1` and `1.0.0` start at `v0.28.1`, `1.0.0-rc.1` at `v1.0.0-rc.0`, `1.1.0-rc.0` at `v1.0.0`.
- `gh release create` is rerun-safe: if the release already exists the step re-uploads assets with `--clobber` instead of failing on the tag.
- The package rename is followed through: `pr-automation.yml`'s PR tarball step and the README looked for `agentcore-*.tgz`, Bun names a scoped tarball `aws-agentcore-*.tgz`.

## Per @AlexanderRichey's thread

`check.yml`, `build.yml` and `unit-test.yml` are merged into one reusable `verify.yml` with one job per platform: install Bun and deps once, then static checks (Linux only, they are platform-independent), bundle, package, compile, smoke test and unit tests. Three jobs and three installs instead of seven. ci.yml and release-publish.yml each call it once. Check names become `verify / Linux`, `verify / Windows`, `verify / macOS`. No ruleset requires status checks by name. The `workflow_dispatch` on release-publish is gone too, since it existed only for dry-run verification and cannot be used until the file is on main.

## Further review findings addressed

- First rc is `rc.1`, per Alex.
- Bun is pinned to 1.4.0 through `packageManager` in package.json, which setup-bun reads in every workflow. The release scripts depend on `bun pm version`, `bun pm view` and `bun pm pack --quiet` output shapes.
- verify's static checks all report even when one fails (`!cancelled()`), matching the old `check.yml`. Build and tests run only when they pass.
- `agentcore update` now tracks the dist-tag named by its own prerelease identifier (`rc`, `preview`) and compares against that tag, instead of mapping every prerelease to `preview` and comparing against `latest`. Without this, an rc user running `update` after `1.0.0` reached `latest` would have installed main's preview build. Covered by a `test.each`.
- README documents recovery when publish fails after the release PR merges: rerun `release-publish`, never re-dispatch `release-prepare`.
- verify's `bun pm pack` skips `prepublishOnly`, which would have rebuilt right after `bun run build`.

## Before the first real release

1. Add `NPM_TOKEN` as a repo secret (granular token with publish rights on `@aws/agentcore`). The Trusted Publisher stays bound to `release-main-and-preview.yml` on main.
2. `release-prepare.yml` becomes dispatchable once refactor lands on main, or once the file exists there. `release-publish.yml` has no dispatch and is unaffected.
3. `latest` contention with main needs a decision before the first stable: `release-main-and-preview.yml` on main also publishes `--tag latest`, so after refactor publishes `1.0.0`, the next `0.28.x` from main would move `latest` back down. Either main stops publishing `latest` once `1.0.0` is out, or its releases move to a different dist-tag.
4. Runner allowlisting for a dedicated release runner, if we want one, with @aidandaly24.

## Follow-ups, not in this PR

- `src/handlers/update/index.tsx` maps any prerelease to dist-tag `preview` and compares against `latest`, so rc users would not be offered a newer rc by `agentcore update`.
- A telemetry attribute validation failure takes the whole CLI down. Telemetry should degrade, not crash.
- The vended CDK pin bump lands in the release PR, so any template incompatibility with a newer `@aws/agentcore-cdk` surfaces as red CI on that PR.


package.json
```diff
@@ -1,6 +1,10 @@
 {
-  "name": "agentcore",
-  "version": "1.0.0",
+  "name": "@aws/agentcore",
+  "version": "0.28.1",
```
Hweinstock, line 3: are we releasing under 1.0.0-rc or the previous version number?

.github/workflows/release.yml
```diff
@@ -0,0 +1,208 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      bump:
+        description: Semver component to bump from the last stable release
+        required: true
+        type: choice
+        options: [major, minor, patch]
+      channel:
+        description: rc publishes after approval, stable opens a release PR
```
Hweinstock: would it be simpler to always open a PR? That also allows us to see the releases in the commit history and gives a consistent way to approve releases.

.github/workflows/release.yml
```diff
@@ -0,0 +1,208 @@
...
+    branches: [refactor]
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
+jobs:
+  version:
+    if: github.event_name == 'workflow_dispatch' || (github.event.pull_request.merged && startsWith(github.head_ref, 'release/'))
+    runs-on: codebuild-agentcore-e2e-${{ github.run_id }}-${{ github.run_attempt }}
```
Hweinstock: don't we need to do some allowlisting for this to work on the codebuild runners? I remember @aidandaly24 mentioning something recently.

Hweinstock (commented):
I kind of like how the python sdk does this with two workflows. In this case, it might look something like:
- `prepare-release.yml` takes in the versionBumpType and releaseChannel, and prepares a PR for the release on a branch with `release/<channel>/<version>` or some pattern so that we can extract it below. 
- `release.yml` triggers on closed PRs that follow release/<channel>/<version> from the GH bot. This could just run the build/check and release. 

The main difference is we always get a PR to gate releases consistently, and the workflow triggers are much simpler.

.github/workflows/release.yml
jariy17: I agree about this. This pr would also  update the cdk-contruct version on the project's agentcore.json using this pr like [here](https://github.com/aws/agentcore-cli/pull/2118)

.github/workflows/README.md
```diff
@@ -19,22 +19,20 @@ care what triggered it.
...
 
-```
-release.yml
-  ├── unit-test.yml
-  ├── build.yml
-  └── publish-npm.yml
+release-publish.yml (npm publish and GitHub release when a release PR merges)
+  |-- check.yml
+  |-- build.yml
+  `-- unit-test.yml
```
AlexanderRichey: Can we take the opportunity to put all of these in one. That will be much more efficient and will save time.

AlexanderRichey (changes requested):

package.json
AlexanderRichey: I think we want to do 1.0.0-rc.1

.github/workflows/README.md
tejaskash: Harrison and TJ seemed to think that the release workflow
From the SDKs made a lot of sense. So I restructured it to be like those, it was originally one file. 

Let's discuss

.github/workflows/release-publish.yml
```diff
@@ -0,0 +1,124 @@
+# Publishes @aws/agentcore to npm and creates the GitHub release when a release PR from
+# release-prepare.yml merges. The version is whatever the merged package.json says.
+name: release-publish
+on:
+  pull_request:
+    types: [closed]
+    # TODO: switch to main once the refactor lands there.
+    branches: [refactor]
+  workflow_dispatch:
```
Hweinstock: why do we need workflow dispatch here?

.github/workflows/README.md
Hweinstock: I think alex's comment is in reference to the check/build/unit-test workflows which are always run together, so I think the suggestion is to make it a single workflow containing check, build, and unit-test as jobs. 

My understanding is that the runtime behavior would be the same, but we no longer need to manually wire in each job as a separate call. Instead, we call a single workflow that triggers all three, so it'd simplify some of the changes here.

package.json
tejaskash: `1.0.0-rc.1`. We will do release channel -> rc and bump type -> major that should give us 1.0.0-rc.1

.github/workflows/release-publish.yml
tejaskash: This is only for the dryrun verification path

.github/workflows/verify.yml
```diff
@@ -0,0 +1,111 @@
...
+      - run: bun run format:check
+        if: always()
+      - run: bun run typecheck
+        if: always()
+      - run: bun audit
+        if: always()
+      - run: bun run secrets:check
+        if: always()
+
+  build:
```
AlexanderRichey: I meant to bring all three of these into one job to avoid installing bun and deps three times. We can just install bum, install deps, typecheck, build, and run tests all in one.

jariy17 (approved):
Looks good.

AlexanderRichey (approved):

PR #2208, opened 2026-09-03
fix: windows bug bash fixes for the node bundle, spawns, conhost glyphs, and the create wizard

tejaskash:
Fixes the CLI issues from the 2026-09-03 Windows bug bash. Write-ups and verification are in the workspace tracker under `.scratch/windows-bugbash/`.

- **Node bundle**: no more `Bun.*` in production code, so the config-file region, local eval datasets, and `eval online create` work under Node. An oxlint rule guards it.
- **Windows spawns**: `shell: true` is gone. `.cmd` wrappers go through `cmd.exe` with quoted arguments, real executables spawn directly. Fixes the DEP0190 warning and the broken OTEL lookup.
- **Python dev agents** get `PYTHONUTF8` and `PYTHONUNBUFFERED`, so emoji no longer kill the stream.
- **Messages**: no `cd X && ...` hints, no internal retry command, `ctrl+c` spelling, unknown subcommands named instead of "too many arguments", `project remove` ends its line and honours `--json`.
- **TUI**: ASCII glyph fallback for terminals without Unicode fonts, create wizard keeps its input after a failed create (`esc` goes back, `r` retries when nothing was written yet).
- **Long paths**: `project create` refuses a Windows destination that would push `node_modules` past MAX_PATH unless `--skip-install` is passed.
- README gains a "Windows notes" section.

Behavior changes worth a changelog line:
- `Bun.hash` became sha256, so online-eval policy and long role names change. Roles created by earlier binaries are still recognised as managed (truncated names match on their prefix) and get refreshed on a data-source move; the superseded policy written under the legacy name cannot be derived, so it stays attached and the existing stale-scope warning now says so. Fixtures were rekeyed.
- A failed create in the TUI wizard no longer exits the process nonzero; the wizard stays open.
- `project remove` prints its message on stderr, stdout carries only the `--json` object.

Verified on a Windows Server 2022 box under Node 24 (scripted over SSH plus the TUI over a pseudo-terminal). Issue 11 (paths with spaces) is in the L3 repo and needs its own PR.


src/handlers/project/create/screen.tsx
```diff
@@ -293,6 +296,7 @@ export function ProjectCreateScreen({ core }: ScreenProps) {
   const submit = async () => {
     let input: CreateProjectInput;
     try {
+      assertProjectPathFits(values.name, ctx.require(PlatformKey));
```
aidandaly24, line 279: I think the wizard has no way past this refusal. The message ends with "or pass `--skip-install` and install the CDK dependencies yourself", but the wizard hardcodes `skipInstall: false` (lines 188 and 204) and never asks about it, so a Windows user in a deep directory only gets a dead end and `ctrl+c`. I reproduced the exact text at a 187 character destination. Could the wizard either offer install as a step, or raise a wizard-specific message that just asks for a shorter directory?

src/components/ui/_core.ts
```diff
@@ -90,12 +87,74 @@ export const borderStyles = {
...
+    env.TERM_PROGRAM === "vscode" ||
+    env.TERM_PROGRAM === "Terminus-Sublime" ||
+    env.TERM === "xterm-256color" ||
+    env.TERM === "alacritty" ||
+    env.TERM?.startsWith("rxvt-unicode") ||
+    env.TERMINAL_EMULATOR === "JetBrains-JediTerm",
+  );
+}
+
+const unicode = unicodeSupported(process.env, process.platform);
```
aidandaly24, line 110: I may be missing a reason to keep this at module scope, but it decides the Windows branch from `process.platform` right after `withPlatform` was added so handlers would stop doing that. The cost is coverage: `core.test.ts` covers the predicate, but with `glyphs` frozen at import no component test can render the ASCII fallback across 27 call sites, and `renderScreen` pins `process.platform` too, so the wizard's new win32 refusal has no screen-level test. Would it make sense to resolve glyphs from the context, or at least let `renderScreen` accept a platform the way `createRootHandler` now does?

src/core/onlineEvalExecutionRole.tsx
```diff
@@ -36,10 +41,7 @@ export function onlineEvalExecutionRoleName(configName: string): string {
   const full = `${ROLE_NAME_PREFIX}${configName}`;
   if (full.length <= ROLE_NAME_MAX) return full;
 
-  const hash = Bun.hash(configName)
-    .toString(16)
-    .padStart(NAME_HASH_LENGTH, "0")
-    .slice(-NAME_HASH_LENGTH);
+  const hash = fingerprint(configName);
```
aidandaly24: I think this changes resource identity for configs created by released binaries, and the worse case fails silently rather than loudly. The prefix is 20 characters, so any config name over 44 gets a hashed role name, and for `my-production-online-evaluation-config-for-the-support-assistant` the suffix moves from `20487c61` to `63113c5d`. `eval.tsx:1232` decides a role is CLI-managed with `roleArn.endsWith(onlineEvalExecutionRoleName(configName))`, so after an upgrade that check fails, `refreshManagedRole` goes false, and a data-source move quietly stops granting the new log group instead of erroring. Shorter names keep their role but move `scopePolicyName` (`ff687d46` to `0a7a30a7`), and `revokeOnlineEvalScope` swallows `NoSuchEntity`, so the superseded policy stays attached with no warning. Since the goal here is Node compatibility rather than a naming change, would it make sense to fall back to the legacy name before treating the role as unmanaged?

aidandaly24 (commented):
Left three inline comments. Also, this is showing as conflicting with the current `refactor` tip on `src/components/EndpointWizard.tsx` (#2211 touched it after `ErrorPanel` moved out), so it will need a rebase before merge.

src/handlers/project/create/screen.tsx
tejaskash: Fixed in aeb1431d. The way out is now caller-supplied: the CLI message mentions --skip-install, the wizard's message only asks for a shorter directory. The refusal happens before anything is written, so the wizard offers r to retry after moving directories. Added a screen-level test that drives the wizard with platform win32 and a deep cwd and checks the message, the retry hint, and that the directory stays empty.

src/components/ui/_core.ts
tejaskash: renderScreen now takes a platform option (aeb1431d), and the wizard's win32 refusal has a screen-level test through it. I kept glyphs at module scope deliberately: it is a capability of the process's terminal rather than a per-command decision, it sits next to darkTheme, which is also a module constant every component takes as a default prop, and resolving it from the context means threading a prop through 27 leaf components for a fallback that only matters on legacy conhost. core.test.ts covers the predicate on every branch. If a second env-dependent rendering decision appears I would move both onto the theme object at that point. Happy to do it now if you feel strongly.

src/core/onlineEvalExecutionRole.tsx
tejaskash: Done in aeb1431d. isManagedOnlineEvalRole matches a truncated role on its prefix (AgentCoreOnlineEval-<truncated config name>-), so a role with a legacy hash suffix is still treated as managed, and grantOnlineEvalScope takes the role name from the config's ARN so the refresh lands on the existing role instead of creating a second one. The superseded policy written under the legacy name cannot be derived, so revokeOnlineEvalScope now returns false on NoSuchEntity and the update path surfaces the existing stale-scope warning instead of staying silent. Unit tests cover detection for short names, truncated names with the current suffix, and truncated names with a legacy suffix.

AlexanderRichey (dismissed):

notgitika (dismissed):
LGTM

src/components/CliOnlyScreen.tsx
```diff
@@ -77,7 +77,7 @@ export function CliOnlyScreen({ ctx, path }: CliOnlyScreenProps) {
       keyHints={[
         { key: "↑↓", label: "scroll" },
         { key: "esc", label: "back" },
-        { key: "ctl+c", label: "quit" },
+        { key: "ctrl+c", label: "quit" },
```
Hweinstock, line 80: I wonder if we should standardize/centralize this constant string to ensure its consistent across the TUI.

src/components/ui/_core.ts
```diff
@@ -90,12 +87,74 @@ export const borderStyles = {
...
+
+const unicodeGlyphs = {
```
Hweinstock, line 112: nit: the variable name `unicode` makes me think its a specific symbol. Would `unicodeSupported` convey this better and maybe `supportsUnicode` or `isUnicodeSupported` as the function name?

src/components/ui/_core.ts
```diff
@@ -1,6 +1,3 @@
-// Auto-generated by sync-registry.ts
```
Hweinstock, line 1: Was this already stale? I don't see this script in the `refactor` branch or this pr.

Also do we know why this file is `_core`? Is this intended to signify that it should not be exported?

aidandaly24 (dismissed):

aidandaly24 (approved):

src/handlers/index.tsx
```diff
@@ -22,6 +28,8 @@ export interface RootHandlerConfig {
   io: AppIO;
   logger: Logger;
   globalConfigAccessor: GlobalConfigAccessor;
+  /** Host platform, defaults to `process.platform`. Tests pass "win32" to exercise Windows paths. */
```
Hweinstock, line 31: nit: I don't think we need to mention how the tests use this here.

src/handlers/index.tsx
```diff
@@ -53,6 +61,9 @@ export function createRootHandler(core: Core, config: RootHandlerConfig): Router
   // Pin the global config accessor on the context for any handler that needs it.
   root.use(withGlobalConfigAccessor(config.globalConfigAccessor));
 
+  // Pin the host platform so Windows-specific behavior is decided from the context.
```
Hweinstock, line 64: Should the TUI code that decides which glyphs to use leverage this same host platform from the context? 

Right now my understanding is that its determined at import time.

src/handlers/utils.tsx
```diff
@@ -135,3 +136,10 @@ export function renderJsonError(ctx: Context, error: unknown): void {
   if (cliError instanceof SilentCLIError) return;
   ctx.require(JsonRendererKey).renderJson({ error: cliError.message });
 }
+
+// reportMessage is the success-side twin: the human line goes to stderr and,
+// under --json, the same message becomes the machine-readable result on stdout.
```
Hweinstock, line 141: nit: I feel this comment is explained more concisely in the code itself. 

I also wonder if the name should match the convention above. Maybe like `renderError` above and `renderMessage` here? Or both use report?

.oxlintrc.json
```diff
@@ -4,6 +4,20 @@
   "categories": {
     "correctness": "error"
   },
```
Hweinstock, line 6: thanks for adding this!

Hweinstock (commented):
Few things I didn't follow from the description. 

> Write-ups and verification are in the workspace tracker under .scratch/windows-bugbash/.

Is this a local file?

> issue 11 (paths with spaces) is in the L3 repo and needs its own PR.

what is issue 11? or 1-10?

Otherwise no blockers. I do think this could have been easier to review as separate PRs as some of the changes seemed pretty unrelated.

PR #2209, opened 2026-09-03
feat(project): support local Runtime invocation

aidandaly24:
## Summary

- add `--local` and `--port` to `agentcore project invoke runtime`
- invoke an already-running headless project dev server through `127.0.0.1:<port>/invocations`
- bypass project resource selection, deployment targets, CloudFormation, and Runtime lookup for local invocations
- preserve the existing streaming, JSON, binary, and output-file response handling
- reject deployed-only request options and report an actionable error when the local server is unavailable

## Testing

- `bun test src/handlers/project/invoke/index.test.tsx` (`8` passed)
- `bun test` (`2962` passed)
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- `bun run build`
- manually created an `agent-python` project and invoked its real headless `project dev` server successfully on the default port `8080`
- repeated the manual invocation with explicit `--port 8081`

## Notes

- `--local` is intentionally headless and requires `--payload`
- the existing bare deployed Runtime invoke TUI remains unchanged

### DIFF
--- a/src/core/dev/localInvoke.ts
+++ b/src/core/dev/localInvoke.ts
@@ -1,0 +1,50 @@
+import { InvalidEnvironmentError } from "../../errors";
+import { abortable } from "../abortable";
+import type { RuntimeInvokeResponse } from "../invokeRuntime";
+
+export type LocalRuntimeInvokeRequest = {
+  port: number;
+  payload: Uint8Array;
+};
+
+async function* emptyBody(): AsyncGenerator<Uint8Array> {}
+
+export async function invokeLocalRuntime(
+  request: LocalRuntimeInvokeRequest,
+  signal?: AbortSignal,
+): Promise<RuntimeInvokeResponse> {
+  let response: Response;
+  try {
+    response = await fetch(`http://127.0.0.1:${request.port}/invocations`, {
+      method: "POST",
+      redirect: "manual",
+      headers: { "Content-Type": "application/json" },
+      body: request.payload as RequestInit["body"],
+      signal,
+    });
+  } catch (error) {
+    if (signal?.aborted) throw signal.reason ?? error;
+    throw new InvalidEnvironmentError(
+      `Local dev server is not running on port ${request.port}. Start it with: ` +
+        `agentcore project dev --mode headless --agent <name> --port ${request.port}`,
+      { cause: error },
+    );
+  }
+
+  const body = (response.body as AsyncIterable<Uint8Array> | null) ?? emptyBody();
+  return {
+    statusCode: response.status,
+    contentType: response.headers.get("content-type") ?? "",
+    runtimeSessionId:
+      response.headers.get("x-amzn-bedrock-agentcore-runtime-session-id") ??
+      response.headers.get("x-session-id") ??
+      undefined,
+    mcpSessionId: response.headers.get("mcp-session-id") ?? undefined,
+    mcpProtocolVersion: response.headers.get("mcp-protocol-version") ?? undefined,
+    traceId: response.headers.get("x-amzn-trace-id") ?? undefined,
+    traceParent: response.headers.get("traceparent") ?? undefined,
+    traceState: response.headers.get("tracestate") ?? undefined,
+    baggage: response.headers.get("baggage") ?? undefined,
+    body: signal ? abortable(body, signal) : body,
+  };
+}
--- a/src/core/dev/localInvoke.ts
+++ b/src/core/dev/localInvoke.ts
@@ -1,10 +1,20 @@
-import { InvalidEnvironmentError } from "../../errors";
+import { randomUUID } from "node:crypto";
+import { InputValidationError, InvalidEnvironmentError } from "../../errors";
 import { abortable } from "../abortable";
 import type { RuntimeInvokeResponse } from "../invokeRuntime";
 
 export type LocalRuntimeInvokeRequest = {
   port: number;
   payload: Uint8Array;
+  contentType?: string;
+  accept?: string;
+  runtimeSessionId?: string;
+  runtimeUserId?: string;
+  applicationHeaders?: [string, string][];
+  traceId?: string;
+  traceParent?: string;
+  traceState?: string;
+  baggage?: string;
 };
 
 async function* emptyBody(): AsyncGenerator<Uint8Array> {}
@@ -13,19 +23,40 @@ export async function invokeLocalRuntime(
   request: LocalRuntimeInvokeRequest,
   signal?: AbortSignal,
 ): Promise<RuntimeInvokeResponse> {
+  const runtimeSessionId = request.runtimeSessionId ?? randomUUID();
+  let headers: Headers;
+  try {
+    headers = new Headers(request.applicationHeaders);
+    for (const [name, value] of [
+      ["Content-Type", request.contentType ?? "application/json"],
+      ["Accept", request.accept ?? "text/event-stream"],
+      ["X-Amzn-Bedrock-AgentCore-Runtime-Session-Id", runtimeSessionId],
+      ["X-Amzn-Bedrock-AgentCore-Runtime-User-Id", request.runtimeUserId ?? "default"],
+      ["X-Amzn-Trace-Id", request.traceId],
+      ["traceparent", request.traceParent],
+      ["tracestate", request.traceState],
+      ["baggage", request.baggage],
+    ] as const) {
+      if (value !== undefined) headers.set(name, value);
+    }
+  } catch {
+    throw new InputValidationError("Invalid local Runtime request header");
+  }
+
   let response: Response;
   try {
     response = await fetch(`http://127.0.0.1:${request.port}/invocations`, {
       method: "POST",
       redirect: "manual",
-      headers: { "Content-Type": "application/json" },
+      headers,
       body: request.payload as RequestInit["body"],
       signal,
     });
   } catch (error) {
     if (signal?.aborted) throw signal.reason ?? error;
+    const detail = error instanceof Error ? error.message : String(error);
     throw new InvalidEnvironmentError(
-      `Local dev server is not running on port ${request.port}. Start it with: ` +
+      `Could not reach local dev server on port ${request.port} (${detail}). Start it with: ` +
         `agentcore project dev --mode headless --agent <name> --port ${request.port}`,
       { cause: error },
     );
@@ -36,9 +67,7 @@ export async function invokeLocalRuntime(
     statusCode: response.status,
     contentType: response.headers.get("content-type") ?? "",
     runtimeSessionId:
-      response.headers.get("x-amzn-bedrock-agentcore-runtime-session-id") ??
-      response.headers.get("x-session-id") ??
-      undefined,
+      response.headers.get("x-amzn-bedrock-agentcore-runtime-session-id") ?? runtimeSessionId,
     mcpSessionId: response.headers.get("mcp-session-id") ?? undefined,
     mcpProtocolVersion: response.headers.get("mcp-protocol-version") ?? undefined,
     traceId: response.headers.get("x-amzn-trace-id") ?? undefined,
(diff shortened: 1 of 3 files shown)


src/handlers/project/invoke/runtime.tsx
```diff
@@ -56,9 +64,70 @@ export const createProjectInvokeRuntimeHandler = (
...
+        if (jsonOutput && flags["output-file"] !== undefined) {
+          throw new InputValidationError("--json cannot be used with --output-file");
+        }
+        const unsupportedFlag = Object.entries({
+          name: flags.name,
+          target: flags.target,
+          qualifier: flags.qualifier,
+          "content-type": flags["content-type"],
+          accept: flags.accept,
+          "session-id": flags["session-id"],
```
tejaskash: Rejecting `--session-id` here means there's no way to hold a multi-turn conversation against the local server. Both deployed paths always send `X-Amzn-Bedrock-AgentCore-Runtime-Session-Id` (falling back to `randomUUID()` in `invokeRuntime.ts`), and the Inspector's local proxy does the same, so memory-backed agents will treat every `--local` call as a brand-new session.

Since the local server is plain HTTP, this flag maps directly to a request header. Suggest allowing it and generating a UUID when absent, matching the other paths. The same argument applies to `--user-id`, `--accept`, `--content-type`, and `--header`, which could all be forwarded rather than rejected.

src/core/dev/localInvoke.ts
```diff
@@ -0,0 +1,50 @@
...
+export async function invokeLocalRuntime(
+  request: LocalRuntimeInvokeRequest,
+  signal?: AbortSignal,
+): Promise<RuntimeInvokeResponse> {
+  let response: Response;
+  try {
+    response = await fetch(`http://127.0.0.1:${request.port}/invocations`, {
+      method: "POST",
+      redirect: "manual",
+      headers: { "Content-Type": "application/json" },
```
tejaskash: No session header is sent here. If `--session-id` is allowed (see comment in `runtime.tsx`), this is where it would go:

```ts
headers: {
  "Content-Type": "application/json",
  "X-Amzn-Bedrock-AgentCore-Runtime-Session-Id": request.sessionId ?? randomUUID(),
},
```

src/core/dev/localInvoke.ts
```diff
@@ -0,0 +1,50 @@
...
+    );
+  }
+
+  const body = (response.body as AsyncIterable<Uint8Array> | null) ?? emptyBody();
+  return {
+    statusCode: response.status,
+    contentType: response.headers.get("content-type") ?? "",
+    runtimeSessionId:
+      response.headers.get("x-amzn-bedrock-agentcore-runtime-session-id") ??
+      response.headers.get("x-session-id") ??
```
tejaskash: This `x-session-id` fallback is only ever set by the Inspector proxy (`core/dev/inspector/invocations.ts`), which runs on a different port than the dev server we're calling here, so it will never match. Safe to drop, or keep if you plan to point `--local` at the proxy later.

src/handlers/project/invoke/runtime.tsx
```diff
@@ -56,9 +64,70 @@ export const createProjectInvokeRuntimeHandler = (
...
+        }
+
+        await withUserCancellation(async (signal) => {
+          const sources = await resolveRuntimeInvokeSources(
+            { payload: flags.payload! },
+            io.stdin,
+            signal,
+          );
+          const response = await invokeLocalRuntime(
+            { port: flags.port ?? DEV_PORTS.HTTP, payload: sources.payload },
```
tejaskash: The 8080 default only holds for HTTP-protocol runtimes. `project dev` picks its port from the runtime protocol (`port.ts`): MCP defaults to 8000 and A2A to 9000, and MCP serves `/mcp` rather than `/invocations`. Because `--local` rejects `--name`, the handler can't look up the protocol, so for an MCP/A2A project the default hits the wrong port and the error hint suggests `--port 8080`, which still won't work.

Two options: (a) note "HTTP-protocol Runtimes only" in the `--local` flag description, or (b) resolve the runtime via `selectProjectResource` and pick the protocol-appropriate default. (a) is fine for this PR if (b) is out of scope.

src/handlers/project/invoke/runtime.tsx
```diff
@@ -27,7 +29,13 @@ export const createProjectInvokeRuntimeHandler = (
     description: "invoke a Runtime from the current project",
     flags: [
       flag("name", "the logical project Runtime name", z.string().optional()),
-      flag("target", "project deployment target", z.string().default("default")),
+      flag("local", "invoke a local development server", z.boolean()),
+      flag(
+        "port",
+        "local development server port (default: 8080)",
```
tejaskash: Nit: if the HTTP-only limitation stands, this is the place to say so, e.g. `"local development server port (default: 8080; HTTP-protocol Runtimes only)"`.

tejaskash (commented):
A few things worth actioning, mostly around parity between `--local` and the deployed invoke path. Details inline.

src/handlers/project/invoke/index.test.tsx
```diff
@@ -212,13 +289,15 @@ describe("project invoke", () => {
...
+        name: undefined,
+        local: false,
+        port: undefined,
+        target: undefined,
         payload: undefined,
         qualifier: undefined,
         "content-type": undefined,
         accept: undefined,
-        "session-id": "project-session",
+        "session-id": undefined,
```
tejaskash: This test previously passed `"session-id": "project-session"` and asserted it came through as `runtimeSessionId` in the launch context. Nothing in the TUI path changed in this PR, so dropping it looks accidental and loses the only coverage of session-id passthrough into the TUI. Only `local: false, port: undefined` needed adding to the flags object.

src/handlers/project/invoke/index.test.tsx
```diff
@@ -239,7 +318,6 @@ describe("project invoke", () => {
     expect(launches[0]!.context.require(RegionKey)).toBe(TARGET.region);
     expect(launches[0]!.context.require(RuntimeInvokeLaunchContextKey)).toMatchObject({
       runtimeId: RUNTIME_ID,
```
tejaskash, line 479: Related to the above: restore `runtimeSessionId: "project-session"` here once the flag is passed again.

src/handlers/project/invoke/runtime.tsx
```diff
@@ -56,9 +64,70 @@ export const createProjectInvokeRuntimeHandler = (
...
+            signal,
+          );
+          await writeRuntimeInvokeResponse(response, {
+            stdout: io.stdout,
+            stderr: io.stderr,
+            outputFile: flags["output-file"],
+            json: jsonOutput,
+            signal,
+          });
+          if (response.statusCode < 200 || response.statusCode >= 300) {
```
tejaskash, line 126: FYI, not necessarily a change request: on a non-2xx the full body is written (stdout / `--output-file` / JSON) before throwing, whereas the deployed CUSTOM_JWT path in `invokeRuntime.ts` cancels the body and throws without writing. Streaming the error body is arguably more useful in local dev, so this is probably intentional. One side effect: with `--output-file`, a 500 body overwrites the target file before the command exits non-zero. If that's fine, consider a one-line comment here so the divergence is visibly deliberate.

src/core/dev/localInvoke.ts
```diff
@@ -0,0 +1,50 @@
...
+    response = await fetch(`http://127.0.0.1:${request.port}/invocations`, {
+      method: "POST",
+      redirect: "manual",
+      headers: { "Content-Type": "application/json" },
+      body: request.payload as RequestInit["body"],
+      signal,
+    });
+  } catch (error) {
+    if (signal?.aborted) throw signal.reason ?? error;
+    throw new InvalidEnvironmentError(
```
tejaskash, line 58: Every non-abort fetch failure is reported as "server is not running". Connection refused is the common case and the message fits, but an `ECONNRESET` mid-request (agent crashed while streaming) gets the same wording, and the attached `cause` isn't printed by the root error handler (`runnable/index.tsx` only prints `error.message`). Consider checking for `ECONNREFUSED` before using this message, and falling back to something generic that includes the cause otherwise.

src/handlers/project/invoke/runtime.tsx
```diff
@@ -27,7 +29,13 @@ export const createProjectInvokeRuntimeHandler = (
...
     flags: [
       flag("name", "the logical project Runtime name", z.string().optional()),
-      flag("target", "project deployment target", z.string().default("default")),
+      flag("local", "invoke a local development server", z.boolean()),
+      flag(
+        "port",
+        "local development server port (default: 8080)",
+        z.coerce.number().int().min(1).max(65535).optional(),
+      ),
+      flag("target", "project deployment target", z.string().optional()),
```
tejaskash: Nit: moving from `.default("default")` to `.optional()` + `?? "default"` at the call site is behaviorally identical, but Commander's `--help` will no longer show the default value for `--target`. Could keep it visible via the description: `"project deployment target (default: default)"`.

src/handlers/project/invoke/runtime.tsx
aidandaly24: Fixed in 83dd3ae9. Local HTTP invokes now forward content type, accept, session/user IDs, application headers, and trace/baggage headers. Missing session IDs are generated and user IDs default to `default`.

src/core/dev/localInvoke.ts
aidandaly24: Added in 04a6ef4e. The local transport now sends `X-Amzn-Bedrock-AgentCore-Runtime-Session-Id` using the supplied value or a generated UUID.

src/core/dev/localInvoke.ts
aidandaly24: Dropped the Inspector-only `x-session-id` fallback. The direct transport now uses the Runtime session response header and otherwise retains the request session ID for response metadata.

src/handlers/project/invoke/runtime.tsx
aidandaly24: Kept this PR explicitly HTTP-only and updated the `--local` and `--port` help text. MCP and A2A require protocol-specific paths and request handling beyond choosing a different default port.

src/handlers/project/invoke/runtime.tsx
aidandaly24: Updated the option wording to `local HTTP Runtime development server port (default: 8080)`.

src/handlers/project/invoke/index.test.tsx
aidandaly24: Restored the `project-session` passthrough assertion while retaining separate coverage of the genuinely bare Runtime TUI launch.

src/handlers/project/invoke/index.test.tsx
aidandaly24: Restored the `runtimeSessionId: "project-session"` launch-context assertion in the updated TUI regression.

src/handlers/project/invoke/runtime.tsx
aidandaly24: Kept the body-first local behavior intentionally. Added a concise comment plus a regression proving a 500 body is written before the command exits nonzero.

src/core/dev/localInvoke.ts
aidandaly24: Addressed the misleading wording without `ECONNREFUSED`-specific branching. Bun 1.4 reports a refused fetch as a generic `TypeError` without a stable error code/cause, so the message is now `Could not reach...` and includes the visible underlying detail plus the recovery command. Mid-stream failures remain handled by the response writer.

src/handlers/project/invoke/runtime.tsx
aidandaly24: Kept the schema optional so local mode can detect an explicitly supplied target, and restored the visible default in the description as `project deployment target (default: default)`.

tejaskash (approved):

tejaskash (approved):

AlexanderRichey (approved):

PR #2213, opened 2026-09-03
chore(tui): align TUI copy with the CLI help descriptions

notgitika:
Follow-up to #2207: ran the same description audit across the TUI by rendering all 163 routes and checking every user-visible string against the same rules.

- added the 16 missing header descriptions (the list screens and `project create`), each verbatim the CLI description for that command
- resource nouns match the CLI — `harness` lowercase, AgentCore/Runtime/Gateway/Memory/Target/Rule capitalized; `ID`/`ARN`/`URL`/`MCP`/`API` initialisms in table headers and wizard fields
- transient status text starts lowercase and uses a real ellipsis
- key hints use one spelling per action (`navigate`, `open`, `back`, `↑↓/jk`)
- `HarnessWizard`'s provider copy now matches the newer `project create` wizard

Picker steps keep their own wording (`choose a Gateway to list Rules for`) rather than the terminal command's description, since the step is a picker.

### DIFF
--- a/src/components/EndpointWizard.tsx
+++ b/src/components/EndpointWizard.tsx
@@ -161,7 +161,7 @@ export function EndpointWizard({
         )}
 
         {phase.kind === "submitting" && (
-          <Spinner label={mode === "create" ? "Creating endpoint…" : "Updating endpoint…"} />
+          <Spinner label={mode === "create" ? "creating endpoint…" : "updating endpoint…"} />
         )}
         {phase.kind === "success" && (
           <SuccessPanel
@@ -316,7 +316,7 @@ function VersionStep({
       {versions.isPending ? (
         <>
           <Question text="which harness version should this endpoint serve?" />
-          <Spinner label="Loading versions…" />
+          <Spinner label="loading versions…" />
         </>
       ) : versions.isError ? (
         <>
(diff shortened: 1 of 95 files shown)


tejaskash (dismissed):

nborges-aws (dismissed):

src/components/EndpointWizard.tsx
```diff
@@ -316,7 +328,7 @@ function VersionStep({
       {versions.isPending ? (
         <>
           <Question text="which harness version should this endpoint serve?" />
-          <Spinner label="loading versions…" />
+          <Spinner label="Loading versions…" />
```
AlexanderRichey: I thought we were going to do all lowercase? Otherwise all looks good. Approving.

AlexanderRichey (dismissed):

Hweinstock (dismissed):
aligned with our fellow gen z :) 

Thanks for taking the time to address these!

src/components/EndpointWizard.tsx
notgitika: fixed the spinner labels as well good catch!

nborges-aws (approved):

PR #2218, opened 2026-09-04
feat(cli): add consistent JSON output to project mutations

notgitika:
Uses the existing global `JsonKey` / `JsonRendererKey` path with a shared, type-checked project mutation envelope for create, add, and remove. All project add leaves now use one `runWithProgress` wrapper, nested help shows inherited global options, and command guidance is returned in structured `notes` under `--json`.

Human-readable success and guidance output is preserved when `--json` is not used.

### DIFF
--- a/src/handlers/project/add/credentials/shared.ts
+++ b/src/handlers/project/add/credentials/shared.ts
@@ -72,13 +72,16 @@ export async function addCredentialToProject(
   await addProjectResource(
     ctx,
     config,
+    project,
     {
       resourceType: "credential",
       ...input,
     },
-    `added credential '${input.resourceConfig.name}' to '${project.name}'\n`,
+    `added credential '${input.resourceConfig.name}' to '${project.name}'`,
+    {
+      notes: (input.envEntries ?? [])
+        .filter((entry) => entry.value === undefined)
+        .map((entry) => `Set ${entry.key} in agentcore/.env.local before you deploy.`),
+    },
   );
-  for (const entry of (input.envEntries ?? []).filter((e) => e.value === undefined)) {
-    config.io.stderr.write(`Set ${entry.key} in agentcore/.env.local before you deploy.\n`);
-  }
 }
--- a/src/handlers/project/add/credentials/shared.ts
+++ b/src/handlers/project/add/credentials/shared.ts
@@ -8,6 +8,7 @@ import {
   credentialEnvVarName,
   credentialNameFieldSuffix,
 } from "../../../../projectSchemas/credential";
+import { addProjectResource } from "../shared";
 
 export { credentialEnvVarName };
 
@@ -68,14 +69,15 @@ export async function addCredentialToProject(
     );
   }
 
-  for await (const event of config.projectManager.addResource(project, {
-    resourceType: "credential",
-    ...input,
-  })) {
-    if (event.type === "step") config.io.stderr.write(`${event.message}\n`);
-  }
-
-  config.io.stderr.write(`added credential '${input.resourceConfig.name}' to '${project.name}'\n`);
+  await addProjectResource(
+    ctx,
+    config,
+    {
+      resourceType: "credential",
+      ...input,
+    },
+    `added credential '${input.resourceConfig.name}' to '${project.name}'\n`,
+  );
   for (const entry of (input.envEntries ?? []).filter((e) => e.value === undefined)) {
     config.io.stderr.write(`Set ${entry.key} in agentcore/.env.local before you deploy.\n`);
   }
--- a/src/handlers/project/add/shared.ts
+++ b/src/handlers/project/add/shared.ts
@@ -1,9 +1,10 @@
-import { ProjectKey, type Context } from "../../../router";
+import type { Context } from "../../../router";
 import { runWithProgress } from "../../../tui/progress";
 import { renderResult } from "../../utils";
 import {
   projectMutationResource,
   projectReference,
+  type ProjectMutationResult,
   type ProjectMutationResourceType,
 } from "../output";
 import type { AddResourceInput, Project } from "../types";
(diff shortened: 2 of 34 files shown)


src/handlers/project/output.ts
```diff
@@ -0,0 +1,70 @@
...
+  const parent = parentFor(input);
+  return parent ? { type, name, parent } : { type, name };
+}
+
+export function renderProjectMutationResult(
+  ctx: Context,
+  result: ProjectMutationResult,
+  renderHuman: () => void,
+): void {
+  renderResult(ctx, result, renderHuman);
```
AlexanderRichey: heh this doesn't seem to add anything

AlexanderRichey (dismissed):

src/handlers/project/output.ts
notgitika: removed it :)

src/handlers/project/project.test.ts
```diff
@@ -67,6 +67,26 @@ async function inProject(name = "TestProject"): Promise<string> {
...
+      "--json",
+    ]);
+    const projectRoot = join(directory, "JsonProject");
+
+    expect(JSON.parse(io.stdout())).toEqual({
+      operation: "create",
+      project: { name: "JsonProject", path: projectRoot },
+    });
+    expect(io.stderr()).not.toContain("Created project");
+    expect(io.stderr()).not.toContain("To deploy it");
```
tejaskash: The create handler never emits "To deploy it", so this assertion passes unconditionally. Use `not.toContain("Next steps")` to actually check that the second stderr line is suppressed under `--json`.

src/handlers/project/output.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+function parentFor(
+  input: AddResourceInput | RemoveResourceInput,
+): ProjectMutationParent | undefined {
+  switch (input.resourceType) {
+    case "gateway-target":
+      return { type: "gateway", name: input.gatewayName };
+    case "policy":
+      return input.engineName ? { type: "policy-engine", name: input.engineName } : undefined;
```
tejaskash, line 59: `parent` here depends on whether `--engine` was passed, not on whether a parent exists. When the manager resolves the owning engine itself (`manager.tsx` `owner`), the JSON omits `parent` even though one was found. Every other `parent` case reflects the real parent, so a script cannot rely on `resource.parent` for policy removals. Either return the resolved engine name from `removeResource`, or document that this field echoes caller input.

src/handlers/utils.tsx
```diff
@@ -151,6 +151,14 @@ export function renderJsonError(ctx: Context, error: unknown): void {
   ctx.require(JsonRendererKey).renderJson({ error: cliError.message });
 }
 
+export function renderResult(ctx: Context, result: unknown, renderHuman: () => void): void {
```
tejaskash: `result: unknown` means the envelope literals in `create`, `remove`, and `add/shared.ts` are unchecked. `ProjectMutationResult` in `output.ts` is exported but never used. Either make this generic and have project callers pass `ProjectMutationResult`, or type the literals at the call sites. Otherwise the union is documentation only.

src/handlers/project/add/shared.ts
```diff
@@ -0,0 +1,46 @@
...
+};
+
+export async function addProjectResource(
+  ctx: Context,
+  config: AddProjectResourceConfig,
+  input: AddResourceInput,
+  humanSuccessMessage: string,
+  options: AddProjectResourceResultOptions = {},
+): Promise<Project> {
+  const project = ctx.require(ProjectKey);
```
tejaskash: Every caller already does `ctx.require(ProjectKey)` for its message. Passing `project` in removes the duplicate lookup and makes the helper's inputs explicit.

src/handlers/project/add/shared.ts
```diff
@@ -0,0 +1,46 @@
...
+    {
+      operation: "add",
+      project: projectReference(updatedProject),
+      resource: projectMutationResource(
+        options.resourceType ?? input.resourceType,
+        input.resourceConfig.name,
+        input,
+      ),
+    },
+    () => config.io.stderr.write(humanSuccessMessage),
```
tejaskash: All 17 callers append `\n` to `humanSuccessMessage`. Having this write own the newline removes that repetition.

src/handlers/project/add/credentials/shared.ts
```diff
@@ -68,14 +69,15 @@ export async function addCredentialToProject(
...
+    ctx,
+    config,
+    {
+      resourceType: "credential",
+      ...input,
+    },
+    `added credential '${input.resourceConfig.name}' to '${project.name}'\n`,
+  );
   for (const entry of (input.envEntries ?? []).filter((e) => e.value === undefined)) {
     config.io.stderr.write(`Set ${entry.key} in agentcore/.env.local before you deploy.\n`);
```
tejaskash: Still writes to stderr under `--json`. Same for `policy-engine/index.ts:74`, `payment-manager/index.ts:102`, and `evaluator/code-based/index.ts:119`. Stdout stays clean so this is not a bug, but a script consumer will not see these notes. If they matter, a `notes: string[]` field on the envelope is the place for them.

tejaskash (commented):
Verified locally on 67f5981d: tsc, oxlint, and 668 tests under src/handlers/project and src/router pass. No behavior-breaking bugs. Inline comments below.

One note not tied to a line: the PR body says human-readable output is unchanged, but `remove ... --json` previously emitted `{ message: "removed ..." }` and a stderr success line. It now emits the new envelope only and the stderr line is suppressed. That is a fine outcome, but the description should state it since this was the one pre-existing project mutation JSON contract.

src/handlers/project/project.test.ts
notgitika: changed it to check the actual "Next steps" output

src/handlers/project/output.ts
notgitika: manager now returns the resolved engine, so JSON includes the parent even without `--engine`

src/handlers/utils.tsx
notgitika: made renderResult generic and explicitly validates ProjectMutationResult

src/handlers/project/add/shared.ts
notgitika: done

src/handlers/project/add/shared.ts
notgitika: added that to the shared helper

src/handlers/project/add/credentials/shared.ts
notgitika: i added structured notes to JSON for the ones you pointed out. human output stays the same

tejaskash (approved):

tejaskash (approved):

PR #2221, opened 2026-09-04
feat: add an environment-backed feature flag client wired through the context

AlexanderRichey:
## Summary

Adds a small feature-flag facility so experimental behavior can ship behind an explicit switch. This is the first commit of #2216, split out on its own so the wiring can land ahead of the imperative deploy that consumes it. The only flag it declares is `AGENTCORE_CLI_EXPERIMENTAL_IMPERATIVE_DEPLOY`, which nothing reads yet; there is no user-visible behavior change beyond one debug log line.

## How it works

- `src/featureFlags/` declares `FEATURE_FLAGS` (code name → environment variable), the consumer-facing `FeatureFlags` interface (`isEnabled`, `enabled`), and `EnvFeatureFlags`. The class is constructed with an injected `processEnv` and never reads `process.env` itself; values are read once at construction so a flag cannot flip mid-command. The contract is deliberately narrow: a flag is on only when the variable's trimmed value is exactly `1`. Any other value (`0`, `true`, `yes`, empty) is off.
- `FeatureFlagsKey` joins the other root context keys in `src/router/router.tsx`; `withFeatureFlags` pins an instance the same way `withGlobalConfigAccessor` does. `createRootHandler` installs it for every command (CLI and TUI), defaulting to an instance with nothing enabled so the existing test call sites need no change. `src/index.ts` is the one place that passes `process.env`.
- `withLogging` writes a single debug line, `experimental feature flags enabled`, naming the enabled flags, only when at least one is on, so unflagged runs keep their exact log shape.
- `TestFeatureFlags` (exported from `src/testing`) and a `featureFlags` option on `renderScreen` let handler and screen tests turn an experiment on.
- README gains an "Experimental features" subsection documenting the `=1` contract and that flagged features may change or vanish.

## Verification

- `bun run lint:check`, `bun run format:check`, `bun run typecheck`, `bun test` (2983 tests, 0 failures at this commit), and `bun run build` all pass.
- Live: `bun run src/index.ts --help` works. With `AGENTCORE_CLI_EXPERIMENTAL_IMPERATIVE_DEPLOY=1`, a `harness list --region us-east-1 --json` run wrote `{"msg":"experimental feature flags enabled","featureFlags":["imperativeDeploy"],…}` to `~/.agentcore/logs/output-*.log`; with the variable set to `0` or unset, no such line was written and the command still listed harnesses, so the middleware chain is intact.

## Reviewer guide

`src/featureFlags/env.ts`, `src/middleware/withFeatureFlags.tsx`, `src/handlers/index.tsx`, `src/middleware/withLogging.tsx`, `src/testing/featureFlags.tsx`.

https://claude.ai/code/session_014D5SZ5sApMQHrwd87VjDar

### DIFF
--- a/src/featureFlags/env.ts
+++ b/src/featureFlags/env.ts
@@ -1,0 +1,36 @@
+import { FEATURE_FLAGS, type FeatureFlag, type FeatureFlags } from "./types";
+
+/** The one value that switches a flag on. Anything else — including "true" — leaves it off. */
+const ENABLED_VALUE = "1";
+
+/**
+ * Feature flags read from environment variables.
+ *
+ * Every flag maps to one variable (see {@link FEATURE_FLAGS}) and is enabled only
+ * when that variable's trimmed value is exactly "1". The contract is deliberately
+ * narrow — no truthiness rules — so an experiment cannot be turned on by accident
+ * and the README can state it in one line.
+ *
+ * The environment is read once at construction, not on every call: a flag that
+ * flipped mid-command could leave a deploy half in one mode and half in another.
+ * `src/index.ts` passes `process.env`; nothing here reaches for it.
+ */
+export class EnvFeatureFlags implements FeatureFlags {
+  private readonly enabledFlags: ReadonlySet<FeatureFlag>;
+
+  constructor(processEnv: Record<string, string | undefined>) {
+    const enabled = new Set<FeatureFlag>();
+    for (const [flag, variable] of Object.entries(FEATURE_FLAGS) as [FeatureFlag, string][]) {
+      if (processEnv[variable]?.trim() === ENABLED_VALUE) enabled.add(flag);
+    }
+    this.enabledFlags = enabled;
+  }
+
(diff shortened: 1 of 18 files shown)


src/featureFlags/env.ts
```diff
@@ -0,0 +1,36 @@
...
+ * and the README can state it in one line.
+ *
+ * The environment is read once at construction, not on every call: a flag that
+ * flipped mid-command could leave a deploy half in one mode and half in another.
+ * `src/index.ts` passes `process.env`; nothing here reaches for it.
+ */
+export class EnvFeatureFlags implements FeatureFlags {
+  private readonly enabledFlags: ReadonlySet<FeatureFlag>;
+
+  constructor(processEnv: Record<string, string | undefined>) {
```
Hweinstock, line 21: what do you think about leveraging the global config for this? This could allow customers could do something like:

```
agentcore config experiments.imperativeDeploy true
```
and 
```
agentcore config experiments
[insert JSON here with all experiment settings]
```

We could still allow an env var override to take precedence too.

PR #2228, opened 2026-09-04
fix(tui): reorder project creation options

notgitika:
## Summary

- list `agent code` before `harness` in the project create TUI
- remove the `(recommended)` label from `harness`
- keep harness as the existing default selection
- update wizard tests for the new ordering

## Verification

- `bun test` — 2,987 pass, 0 fail
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`


tejaskash (approved):

Hweinstock:
I think even though agent code is on the top, its still not selected by default:
image

When I ran through a fresh TUI harness was selected by default. 

I also think we should remove the reccomended here:
image

PR #2241, opened 2026-09-04
fix(project): refuse status outside the project's deployed region

AlexanderRichey:
## What

`project status` now refuses to describe a project whose target is deployed in a region other than the one the command runs in. The CLI throws a `ProjectStateError` naming both regions; the bare TUI shows the same message in place of the Linked Resources tree and keeps `esc` as the way back. Rerun with `--region <target region>` to see the report.

## Why

A bare `project status` links each deployed row to its detail page with `?region=<target>`, but the actions those pages open in turn (endpoints, events, detail JSON, …) run in the ambient region, so a project deployed anywhere else broke one step past the tree. #2239 tried to carry the region through every flow instead; this is the smaller guard. The headless report has the same problem one step later: the ids it prints only work with the same `--region`.

## How

- `src/handlers/project/status/index.ts`: compare the resolved target's region with the context's `RegionKey` after resolving resources and before rendering.
- `src/handlers/project/status/screen.tsx`: `ProjectStatusView` takes the context, and renders the mismatch message instead of the `resources` heading and tree.

## Tests

- Headless (`index.test.ts`): the suite pins `AWS_REGION` to the default target's region (restored after each test), since `withRegion` would otherwise read the developer's shell. The `--target staging` case now passes a matching `--region`. New case: a mismatched target rejects with a `ProjectStateError` that names both regions, an explicit `--region` takes part in the same comparison, and nothing is written to stdout.
- Screen (`status.screen.test.tsx`): `renderStatus` pins `RegionKey` to the target's region so the existing tree and navigation tests still apply. New case: with the base context's `us-east-1` against a `eu-west-1` target, the message renders, no agent group or runtime row does, `enter` stays put, and `esc` returns to the project menu.
- Both new cases fail against the previous source and pass with the change.

## Checks

- `bun test src` — 3167 pass. The 11 failures are all in `src/io/exec.test.ts`, which spawns `node`, not installed in this environment (same as #2239's run).
- `bun run typecheck`, `bun run lint:check`, `bun run format:check` — clean for the repository's own files.

https://claude.ai/code/session_016KnHdxcPYTnQiDY7Y1PY6s

### DIFF
--- a/src/handlers/project/status/screen.tsx
+++ b/src/handlers/project/status/screen.tsx
@@ -15,6 +15,7 @@ import { ProjectKey } from "../../../router";
 import type { ScreenProps } from "../../types";
 import type { DeployableResource, Project, ResolvedProjectResource } from "../types";
 import { LoadingFrame, ProjectGate } from "../ProjectGate";
+import { RegionKey } from "../../keys";
 
 const theme = darkTheme;
 
@@ -168,15 +169,16 @@ export function ProjectStatusScreen({ ctx, core }: ScreenProps) {
       seed={ctx.value(ProjectKey)}
       onBack={() => navigate(PROJECT_MENU)}
     >
-      {(project) => <ProjectStatusView core={core} project={project} />}
+      {(project) => <ProjectStatusView core={core} ctx={ctx} project={project} />}
     </ProjectGate>
   );
 }
 
 function ProjectStatusView({
   core,
+  ctx,
   project,
-}: Pick<ScreenProps, "core"> & {
+}: ScreenProps & {
   project: Project;
 }) {
   const navigate = useNavigate();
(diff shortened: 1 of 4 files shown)


tejaskash (approved):

Hweinstock (approved):

notgitika (approved):

src/handlers/project/status/screen.tsx
```diff
@@ -237,16 +245,26 @@ function ProjectStatusView({
...
-          ) : (
-            <TreeView nodes={nodes} onSelect={select} showIcons={false} focusMarker />
-          )}
-        </Box>
+        {isWrongRegion && (
+          <Text color="red">
+            This project is deployed to {status.data.target.region}, not {region}
+          </Text>
+        )}
+        {!isWrongRegion && (
```
notgitika, line 253: maybe we also throw ProjectStateError here

PR #2248, opened 2026-09-08
fix: fallback to legacy stack name in deployed state

nborges-aws:
## Description

The refactored CLI updated deployedState.ts to track `stackArn`, where our legacy/main deployedState.ts instead tracked `stackName`. Therefore, inspecting a legacy project with the new CLI resulted in deployed resources showing as 'local-only'.

This PR addresses this concern by adding `stackName` as a fallback in our resource resolution logic. This change allows legacy projects to be inspected correctly through the new CLI, without the need for redeploying resources.

## Example
This example shows a legacy project with two deployed harnesses in the project. The recording shows the following commands:
* `agentcore` -> this is the currently released RC of our refactor branch installed via npm. 
* `acr-dev` -> alias pointing to my local clone and dev workspace of the CLI. Built from refactor branch including the changes present in this PR

https://github.com/user-attachments/assets/d7df777a-e0c0-4153-9ce4-c76c9113d992

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Manually verified as shown in screen recording above. Additionally, added a unit test asserting that a legacy stack name is resolved correctly when the target does not have a stack ARN. Focused unit test passed.

- [x] `bun run test` (3180 pass, 0 fail)
- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.

### DIFF
--- a/src/core/project/backends/cdk.test.ts
+++ b/src/core/project/backends/cdk.test.ts
@@ -872,7 +872,43 @@ describe("CdkBackend.resolveDeployedResources", () => {
     expect(subject.stackReads).toHaveLength(1);
   });
 
-  test("fails without reading AWS when the target has no deployed stack ARN", async () => {
+  test("resolves resources from a legacy stack name when the target has no stack ARN", async () => {
+    const input = await project();
+    input.spec = ProjectSpecSchema.parse({
+      ...input.spec,
+      harnesses: [{ name: "support", path: "app/support" }],
+    });
+    await updateTargetState(json, input.rootPath, TARGET.name, {
+      resources: { stackName: "AgentCore-example-default" },
+    });
+    const subject = harness({
+      describedStack: {
+        StackName: "AgentCore-example-default",
+        CreationTime: new Date(0),
+        StackStatus: "CREATE_COMPLETE",
+        Outputs: [
+          {
+            ExportName: "AgentCore-example-default-Harness-support-Id",
+            OutputValue: "support-AbCdEf1234",
+          },
+        ],
+      },
+    });
+
+    await expect(
+      subject.backend.resolveDeployedResources(input, { target: TARGET }),
+    ).resolves.toEqual([
+      {
+        resourceType: "harness",
+        name: "support",
+        id: "support-AbCdEf1234",
+        target: TARGET,
+      },
+    ]);
+    expect(subject.stackReads[0]?.stackName).toBe("AgentCore-example-default");
+  });
+
+  test("fails without reading AWS when the target has no recorded stack", async () => {
     const input = await project();
     const subject = harness({ describedStack: null });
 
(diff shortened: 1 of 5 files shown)


src/core/project/backends/cdk.ts
```diff
@@ -437,14 +438,16 @@ export class CdkBackend implements ProjectBackend {
     const { spec } = project;
     const deployedState = await readDeployedState(this.json, project.rootPath);
     const recorded = deployedState.targets[target.name];
+    const stackReference = recorded?.stackArn ?? recorded?.resources?.stackName;
```
aidandaly24: nit: Would it make sense to pull this fallback into `deployedState.ts` next to the field it reads, something like `stackReferenceOf(recorded)`? The same expression now appears here and in `resolveProjectResources`, and the reason it exists is only explained by the schema comment on `stackName`. Keeping the legacy rule beside the schema means the next reader of the stack binding, like the delete-and-recreate check the `stackArn` comment describes, gets the fallback without having to know it exists.

aidandaly24 (dismissed):
One small nit but looks good to me

src/core/project/backends/cdk.test.ts
```diff
@@ -872,7 +872,43 @@ describe("CdkBackend.resolveDeployedResources", () => {
     expect(subject.stackReads).toHaveLength(1);
   });
 
-  test("fails without reading AWS when the target has no deployed stack ARN", async () => {
+  test("resolves resources from a legacy stack name when the target has no stack ARN", async () => {
```
Hweinstock, line 875: These tests appear to be asserting on implementations rather than behaviors. Would it be possible to test at the handler level, such that we create a project, insert the legacy stackName field into deployedState, then ensure the status command works as expected?

src/core/project/backends/cdk.ts
Hweinstock: +1

src/handlers/project/status/index.test.ts
```diff
@@ -48,23 +49,26 @@ function fakeBackend(deployed: ResolvedProjectResource[]) {
...
   return {
-    ...fake,
     io,
     json: () => JSON.parse(io.stdout()),
     run: (args: string[] = []) => root.route(["node", "agentcore", "project", "status", ...args]),
     create: (args: string[]) => root.route(["node", "agentcore", "project", ...args]),
   };
 }
 
+function testStatusCommand(deployed: ResolvedProjectResource[] = [], io = testIO()) {
```
Hweinstock, line 67: nit: are both `testStatusCommand` and `statusCommand` necessary? Feels like a lot of indirection and the names aren't clearly conveying the difference to me.

src/handlers/project/status/index.test.ts
```diff
@@ -133,6 +138,63 @@ const HARNESS_ROW = localOnly("harness", "orders");
 const memory = (name: string) => ({ name, eventExpiryDuration: 30 });
 const policy = (name: string) => ({ name, statement: "permit(principal, action, resource);" });
 describe("project status handler", () => {
+  test("reports resources deployed by a legacy CLI using resources.stackName", async () => {
```
Hweinstock, line 141: nice! I like this test, it gives me very high confidence this won't regress and works as expected.

Hweinstock (approved):

aidandaly24 (approved):

PR #2250, opened 2026-09-09
fix: mark logs and traces as cli only

nborges-aws:
## Description

In our TUI, `logs` and `traces` are both appearing under harness as standard TUI items, though both are CLI-only commands. 

This PR updates the harness handler to exclude the above operations as TUI supported, adding the CLI-only line and styling for unsupported commands. 

## Example
`agentcore` -> running RC project installed via npm
`acr-dev` -> running project from my dev environment

https://github.com/user-attachments/assets/774e1cbb-3c03-49e0-bbb7-b7f8739f309c

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] `bun run test` (3179 pass, 0 fail)
- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.

### DIFF
--- a/src/handlers/harness/index.tsx
+++ b/src/handlers/harness/index.tsx
@@ -22,6 +22,17 @@ export function createHarnessHandler(core: Core, io: AppIO): Router {
   harness.use(withTuiOnEmptyFlagsAndArgs(core, io));
   // Open the TUI at this root, i.e., `agentcore harness`
   harness.default(renderTui(core, io));
+  harness.supportedTuiCommands(
+    "create",
+    "get",
+    "list",
+    "update",
+    "delete",
+    "invoke",
+    "exec",
+    "endpoint",
+    "version",
+  );
 
   // Register handlers
   harness.handler(createCreateHarnessHandler(core));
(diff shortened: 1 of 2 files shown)


jariy17 (approved):

src/handlers/harness/index.tsx
```diff
@@ -22,6 +22,17 @@ export function createHarnessHandler(core: Core, io: AppIO): Router {
   harness.use(withTuiOnEmptyFlagsAndArgs(core, io));
   // Open the TUI at this root, i.e., `agentcore harness`
   harness.default(renderTui(core, io));
+  harness.supportedTuiCommands(
```
Hweinstock, line 25: i too was bitten by the default behavior here being that all commands are tui supported :/

Hweinstock (approved):

notgitika (approved):

src/handlers/harness/index.tsx
notgitika: Hmm should it be the other way around? I can make that change if we need it

nborges-aws: Maybe we can do an include / exclude set up. Callers can choose the more sensible of the two

PR #2252, opened 2026-09-09
fix(security): harden Windows subprocess argument handling

tejaskash:
## Summary

On Windows, the subprocess helpers spawned with `shell: true`, which let the shell interpret the contents of arguments passed to a child process. This hardens that path.

The five subprocess helpers now delegate spawning to [`cross-spawn`](https://github.com/moxystudio/node-cross-spawn), which resolves Windows `.cmd`/`.bat` wrappers and escapes arguments so their contents reach the child process intact rather than being interpreted by a shell. Nothing spawns through a shell, and the `shell` option is removed from `SubprocessOptions`.

cross-spawn was already present transitively (via eslint/vite); this promotes it to a direct dependency and adds `@types/cross-spawn`. Delegating also removes the hand-rolled Windows quoting and PATHEXT resolution.

## Changes

- Promote `cross-spawn` to a direct dependency; add `@types/cross-spawn`
- `runSubprocess`, `checkSubprocess`, `runSubprocessCapture`, `runSubprocessCaptureSync`, `checkSubprocessSync` call `crossSpawn` / `crossSpawn.sync`
- Drop `shell` from `SubprocessOptions` and the now-redundant `shell: false` in `initGitRepo`
- Behavioral test: an argument containing shell metacharacters arrives at the child as a single argv element

## Testing

- `tsc --noEmit` clean, eslint clean, 18 subprocess unit tests pass
- Follow-up before merge: re-run the Windows verification against this build, since the fix now depends on cross-spawn's escaping

## Out of scope (follow-up)

Other callers (`codezip-dev-server.ts`, `container.ts`, `update/action.ts`, `import/phase2-import.ts`, and others) still spawn directly and bypass these hardened helpers. Routing them through the helpers closes the same class of issue everywhere.


tejaskash:
### CodeQL alerts 47 & 48 dismissed as false positive

Both `js/shell-command-injection-from-environment` alerts on the `crossSpawn` / `crossSpawn.sync` calls have been dismissed. Reasoning:

- **No shell is involved.** cross-spawn spawns the named command directly (`shell: false`). Neither the command nor any argument is interpreted by `cmd.exe` or `/bin/sh`. This is precisely the property this PR set out to guarantee.
- **The alert is about the command *path*, not metacharacters.** The rule fires because the command resolves from an environment value (a `PATH` lookup of known tools such as `git`, `uv`, `npm`). That is standard subprocess usage, not the shell-metacharacter injection this PR fixes.
- **Not a new exposure.** An actor who can change the resolved binary path already controls `PATH` or the project config, and therefore already has local code execution regardless of this change.
- **Zero of these alerts exist on `main`.** They appear only because CodeQL models cross-spawn's first argument as a command-execution sink. The prior hand-rolled `cmd.exe` routing escaped the rule only because the tainted path sat in the args array rather than the command position, so the safer library trips a rule the riskier code slipped past.

If the security reviewers prefer, the alternative is to allowlist/validate the command at each caller so the taint flow breaks, but that is a larger change for no real reduction in risk.

jariy17 (approved):

PR #2255, opened 2026-09-09
refactor(project): simplify code-based evaluator scaffolding

jariy17:
## Summary

- remove --metric and --model from project add evaluator code-based
- always scaffold the generic Python evaluator for managed code-based evaluators
- remove the bundled DeepEval and AutoEvals templates
- describe --timeout-seconds as the evaluator timeout

## Compatibility

- existing evaluator project specs are unchanged
- scripts using the removed flags now receive an unknown-option error

## Testing

- bun test (3171 passed)
- bun run typecheck
- bun run lint:check
- bun run format:check
- bun run build

## Live verification

Verified with a temporary evaluator-only project in an exploratory AWS account in us-west-2:

- scaffolded a managed code-based evaluator with --timeout-seconds 123
- synthesized CloudFormation with Lambda Timeout and evaluator LambdaTimeoutInSeconds both set to 123
- deployed successfully; the evaluator reached ACTIVE status
- confirmed the deployed Lambda was Active with Timeout 123
- invoked the packaged evaluator Lambda with a representative evaluator event and received a Pass response with HTTP 200
- removed the project and confirmed the stack reached DELETE_COMPLETE and both resources no longer exist


src/core/project/templates/evaluator.ts
```diff
@@ -44,14 +29,7 @@ function buildManagedEvaluatorSpec(input: ManagedEvaluatorScaffoldInput): Evalua
 }
 
 function buildRenderContext(input: ManagedEvaluatorScaffoldInput): Record<string, unknown> {
```
Hweinstock: is this function still needed or could it be inlined?

Hweinstock (dismissed):

src/core/project/templates/evaluator.ts
jariy17: This could be inlined. I'll remove it

jariy17: Inlined the render context in 52787db6. Focused evaluator tests, type-check, lint, formatting, and build all pass.

Hweinstock (approved):

notgitika (approved):

notgitika (approved):

PR #2256, opened 2026-09-09
refactor(project): clean up project handler tests to use common utils

Hweinstock:
### Problem 
Each project test redefined their own version of the same utilities. 

Examples:
- https://github.com/aws/agentcore-cli/blob/b18bb4bbba8a9363e0f8fa6a5c24fae5904f11c2/src/handlers/project/add/runtime/index.test.ts#L43
- https://github.com/aws/agentcore-cli/blob/refactor/src/handlers/project/project.test.ts#L58 

### Solution
- standardize the temp directory helper, and project initiation in `src/testing/projects.ts`.

### Future Work
- Hoping to also migrate to a more consistent model where handler tests live next to the handler rather than project.test.ts

### DIFF
--- a/src/testing/fs.ts
+++ b/src/testing/fs.ts
@@ -1,0 +1,20 @@
+import { mkdtemp, rm } from "node:fs/promises";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+
+/** A temp directory and a handler that restores the cwd and removes it. */
+export type TempDirectory = { path: string; cleanup: () => Promise<void> };
+
+/** Creates a temp directory, cds into it, and returns its realpath plus a cleanup handler. */
+export async function inTempDirectory(prefix = "agentcore-project-"): Promise<TempDirectory> {
+  const originalCwd = process.cwd();
+  const directory = await mkdtemp(join(tmpdir(), prefix));
+  process.chdir(directory);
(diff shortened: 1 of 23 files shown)


src/testing/fs.ts
```diff
@@ -0,0 +1,20 @@
...
+/** Creates a temp directory, cds into it, and returns its realpath plus a cleanup handler. */
+export async function inTempDirectory(prefix = "agentcore-project-"): Promise<TempDirectory> {
+  const originalCwd = process.cwd();
+  const directory = await mkdtemp(join(tmpdir(), prefix));
+  process.chdir(directory);
+  return {
+    path: process.cwd(),
+    cleanup: async () => {
+      process.chdir(originalCwd);
+      await rm(directory, { recursive: true, force: true });
```
notgitika: Could we make cleanup safe when multiple temp directories are created? Each call captures the cwd, so running cleanups together can leave cwd pointing to a deleted temp directory, causing later file operations to fail

notgitika (dismissed):
Added a small comment for a bug that my agent found otherwise LGTM!

notgitika (approved):

nborges-aws (approved):
This is great. Nice work!

PR #2263, opened 2026-09-09
fix(handlers): standardize mutual flag parsing behavior

Hweinstock:
### Problem
CLI handlers hand-roll "mutually exclusive" and "specify exactly one of" flag checks with at least three message formats, so identical failures read differently across commands.  

### Solution
- Replace `assertMutuallyExclusiveInputs` with `assertMutuallyExclusiveFlags` that is general enough to handle all of the mutually exclusive checks we make today. 
- Migrate handlers with mutually-exclusive checks onto the shared utility.

### Verification
- All handler tests green; changed lines have explicit coverage.

manually tested a migrated case:
```
$ agentcore  identity api-key-credential-provider create --name x --region us-west-2
Error: specify exactly one of --api-key, --api-key-secret-reference
```


nborges-aws (approved):
LGTM clean PR

jariy17 (approved):

PR #2265, opened 2026-09-09
feat(batch-evaluation): add --output-config and rename --qualifier to --endpoint

jariy17:
Renamed `--qualifier` to `--endpoint` in **both** `agentcore eval batch-evaluation simulate` and `agentcore eval ondemand simulate` to match the other commands' runtime-endpoint flag. This is a breaking rename on both commands. We also introduce `--output-config` to `agentcore eval batch-evaluation simulate` and `agentcore eval batch-evaluation evaluate` so customers can control where their results go ([docs](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/results-and-output.html)). 

### `agentcore eval batch-evaluation evaluate`

```text
Usage: agentcore eval batch-evaluation evaluate [options]

evaluate existing sessions service-side (async; returns a job ID)

Configuration:
  --name <name>                              batch evaluation name (must be unique in the account)
  --description <description>                optional description
  --kms-key-arn <kms-key-arn>                KMS key to encrypt evaluation data at rest

Session source (choose exactly one):
  --agent <agent>                            harness ID or Runtime ID whose sessions to use
  --online-eval <online-eval>                use sessions an online-eval config already sampled
  --data-source-config <data-source-config>  the traces to read (JSON DataSourceConfig); escape hatch

Source filters:
  --endpoint <endpoint>                      Runtime endpoint qualifier (default DEFAULT; only with --agent)
  --start-time <start-time>                  window start (ISO-8601, with --end-time)
  --end-time <end-time>                      window end (ISO-8601, with --start-time)
  --session-ids <session-ids...>             specific session IDs (only with --agent)

Evaluation:
  --evaluators <evaluators...>               evaluator ID(s) to apply
  --ground-truth <ground-truth>              expected answers for the sessions (JSON SessionMetadataShape[])

Result output:
  --output-config <output-config>            where results and metrics are written (JSON OutputConfig)

Other options:
  -h, --help                                 display help for command

Global Options:
  --region <region>                          AWS region
  --debug                                    debug logging (default: false)
  --json                                     JSON output (default: false)
  --endpoint-url <endpoint-url>              endpoint URL override

Parameter details:

  --data-source-config (JSON: tagged union object)
      Where sessions and traces are read from, for sources the --agent and
      --online-eval convenience flags cannot express. Only top-level key:
      cloudWatchLogs.

      Accepts inline JSON, file://<path>, or - to read stdin.

      JSON syntax:
        {
          "cloudWatchLogs": {
            "logGroupNames": ["string", ...],  // [required] groups holding the traces
            "serviceNames": ["string", ...],   // e.g. "my_agent.DEFAULT"
            "filterConfig": {
              "sessionIds": ["string", ...],
              "timeRange": {
                "startTime": "timestamp",
                "endTime": "timestamp"
              }
            }
          }
        }

      API reference:
        https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_DataSourceConfig.html

      Example:
        --data-source-config '{"cloudWatchLogs":{"logGroupNames":["/aws/bedrock-agentcore/runtimes/support_agent-AbC123XyZ9-DEFAULT"],"serviceNames":["support_agent.DEFAULT"],"filterConfig":{"sessionIds":["session-123"]}}}'

  --ground-truth (JSON: list of objects)
      Expected answers for the sessions being evaluated, so an evaluator can score a
      response against a reference instead of judging it on its own. Each entry names
      one session; omit an entry for a session that has no reference answer.

      Accepts inline JSON, file://<path>, or - to read stdin.

      JSON syntax:
        [
          {
            "sessionId": "string",              // [required] the session this applies to
            "testScenarioId": "string",         // groups sessions replaying one scenario
            "groundTruth": {                    // exactly one key; only inline today
              "inline": {
                "turns": [                      // the expected exchange, in order
                  {
                    "input": { "prompt": "string" },
                    "expectedResponse": { "text": "string" }
                  },
                  ...
                ],
                "assertions": [                 // statements the response must satisfy
                  { "text": "string" },
                  ...
                ],
                "expectedTrajectory": {
                  "toolNames": ["string", ...]  // tools the agent should have called
                }
              }
            },
            "metadata": { "string": "string", ... }
          },
          ...
        ]

      Example:
        --ground-truth '[{"sessionId":"session-123","groundTruth":{"inline":{"turns":[{"input":{"prompt":"Where is my order?"},"expectedResponse":{"text":"It shipped on Tuesday."}}]}}}]'

        --ground-truth file://ground-truth.json

  --output-config (JSON: tagged union object)
      Where evaluation results and metrics are written. Omit it and results go to the
      service-managed default location. Only top-level key: cloudWatchConfig.

      Accepts inline JSON, file://<path>, or - to read stdin.

      JSON syntax:
        {
          "cloudWatchConfig": {
            "logGroupName": "string",      // result log group; omit for
                                           // SOURCE_LOG_GROUP, and it cannot sit
                                           // under /aws/bedrock-agentcore/evaluations/
            "logStreamName": "string",     // result log stream
            "metricsNamespace": "string",  // defaults to
                                           // Bedrock-AgentCore/Evaluations; cannot
                                           // begin with "AWS/"
            "resultDestination": "DEDICATED_LOG_GROUP" | "SOURCE_LOG_GROUP"
                                           // DEDICATED_LOG_GROUP (default) writes to
                                           // a dedicated result group;
                                           // SOURCE_LOG_GROUP writes back to the log
                                           // group the traces were read from
          }
        }

      API reference:
        https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_OutputConfig.html

      Example:
        --output-config '{"cloudWatchConfig":{"logGroupName":"/company/agent-evaluations","metricsNamespace":"Company/AgentEvaluations","resultDestination":"DEDICATED_LOG_GROUP"}}'

```

<details>
<summary><h3><code>agentcore eval batch-evaluation simulate</code></h3></summary>

```text
Usage: agentcore eval batch-evaluation simulate [options]

replay a dataset against a Runtime, then batch-evaluate the resulting sessions

Runtime invocation:
  --runtime-id <runtime-id>                Runtime ID to invoke per scenario
  --endpoint <endpoint>                    Runtime endpoint qualifier (default DEFAULT)
  --payload-template <payload-template>    request body per example (JSON object); {input} is replaced with the input
  --header <header...>                     an ordered application header (repeatable)
  --bearer-token <bearer-token>            CUSTOM_JWT bearer token (for JWT-auth Runtimes)
  --user-id <user-id>                      Runtime user ID

Dataset:
  --dataset <dataset>                      dataset source: local JSONL path or a dataset ID
  --dataset-version <dataset-version>      dataset version (with a dataset ID)
  --ingestion-wait-ms <ingestion-wait-ms>  ms to wait for span ingestion before grading (default 180000; 0 to skip)

Configuration:
  --name <name>                            batch evaluation name (unique in the account)
  --description <description>              description for the batch evaluation
  --kms-key-arn <kms-key-arn>              KMS key to encrypt evaluation data at rest

Evaluation:
  --evaluators <evaluators...>             evaluator ID(s) to apply

Result output:
  --output-config <output-config>          where results and metrics are written (JSON OutputConfig)

Other options:
  -h, --help                               display help for command

Global Options:
  --region <region>                        AWS region
  --debug                                  debug logging (default: false)
  --json                                   JSON output (default: false)
  --endpoint-url <endpoint-url>            endpoint URL override

Parameter details:

  --payload-template (JSON object)
      The request body sent to the Runtime for each dataset example. Every occurrence
      of {input} is replaced with that example's input, so the template describes the
      shape your agent expects and {input} marks where the prompt goes.

      Example:
        --payload-template '{"prompt":"{input}"}'

        --payload-template '{"messages":[{"role":"user","content":"{input}"}],"stream":false}'

  --output-config (JSON: tagged union object)
      Where evaluation results and metrics are written. Omit it and results go to the
      service-managed default location. Only top-level key: cloudWatchConfig.

      Accepts inline JSON, file://<path>, or - to read stdin.

      JSON syntax:
        {
          "cloudWatchConfig": {
            "logGroupName": "string",      // result log group; omit for
                                           // SOURCE_LOG_GROUP, and it cannot sit
                                           // under /aws/bedrock-agentcore/evaluations/
            "logStreamName": "string",     // result log stream
            "metricsNamespace": "string",  // defaults to
                                           // Bedrock-AgentCore/Evaluations; cannot
                                           // begin with "AWS/"
            "resultDestination": "DEDICATED_LOG_GROUP" | "SOURCE_LOG_GROUP"
                                           // DEDICATED_LOG_GROUP (default) writes to
                                           // a dedicated result group;
                                           // SOURCE_LOG_GROUP writes back to the log
                                           // group the traces were read from
          }
        }

      API reference:
        https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_OutputConfig.html

      Example:
        --output-config '{"cloudWatchConfig":{"logGroupName":"/company/agent-evaluations","metricsNamespace":"Company/AgentEvaluations","resultDestination":"DEDICATED_LOG_GROUP"}}'

```

</details>


src/core/eval.test.ts
```diff
@@ -0,0 +1,87 @@
...
+  type BedrockAgentCoreClient,
+} from "@aws-sdk/client-bedrock-agentcore";
+import type { BedrockAgentCoreControlClient } from "@aws-sdk/client-bedrock-agentcore-control";
+import type { CloudWatchLogsClient } from "@aws-sdk/client-cloudwatch-logs";
+import type { IAMClient } from "@aws-sdk/client-iam";
+import { CoreClient } from "./index";
+import { createSilentLogger } from "../testing";
+import type { OutputConfig } from "@aws-sdk/client-bedrock-agentcore";
+
+// The TestCoreClient suites assert what a handler hands to Core; the fixture
```
jariy17, line 13: Use Golden tests.

src/handlers/eval/batch-evaluation/simulate/index.tsx
```diff
@@ -97,6 +99,8 @@ export const createSimulateBatchEvaluationHandler = (core: Core, _io: AppIO) =>
 
       // Ctrl-C aborts the run (invokes, the ingestion wait, the dataset download).
       // TODO(#1986): swap for the shared SIGINT/abort helper once it merges.
```
jariy17, line 100: I'll fix this TODO later in a follow up PR.

src/handlers/eval/batch-evaluation/outputConfig.tsx
```diff
@@ -0,0 +1,56 @@
...
+    }
+  }
+
+API reference:
+  https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_OutputConfig.html
+
+Example:
+  --output-config '{"cloudWatchConfig":{"logGroupName":"/company/agent-evaluations","metricsNamespace":"Company/AgentEvaluations","resultDestination":"DEDICATED_LOG_GROUP"}}'`;
+
+const RESULT_OUTPUT = "Result output:";
```
jariy17: We don't need this

src/handlers/eval/batch-evaluation/batch-evaluation.test.tsx
```diff
@@ -262,3 +266,146 @@ describe("eval batch-evaluation simulate", () => {
     expect(JSON.parse(stdout).failures).toEqual([{ exampleId: "bad", error: "HTTP 500" }]);
   });
 });
+
+describe("eval batch-evaluation evaluate --output-config", () => {
```
jariy17: Use golden tests.

src/handlers/eval/types.tsx
```diff
@@ -289,6 +290,10 @@ export type StartBatchEvaluationInput = {
   // Already-parsed --ground-truth (SessionMetadataShape[]) → evaluationMetadata.
   groundTruth?: SessionMetadataShape[];
   kmsKeyArn?: string;
+  // Already-parsed --output-config, forwarded to the request untouched. Left
```
jariy17: remove this code comment

src/handlers/eval/batch-evaluation/outputConfig.tsx
```diff
@@ -0,0 +1,54 @@
...
+    flag(
+      "output-config",
+      "where results and metrics are written (JSON OutputConfig)",
+      z.string().optional(),
+      { group: "Result output:", help: outputConfigHelp },
+    ),
+  ] as const;
+
+  static async resolve(value: string | undefined, io: AppIO): Promise<OutputConfig | undefined> {
+    const resolver = new SourceResolver({ stdin: io.stdin });
```
notgitika: `SourceResolver` owns the one-stdin-option guard, but this creates a new instance separate from the resolvers used by `SessionSource` and `--ground-truth` in `evaluate`. With `--ground-truth - --output-config -` (or `--data-source-config - --output-config -`), the first option drains stdin and this one receives an empty string instead of reporting the conflict. Could we share one resolver or prevalidate that only one option uses stdin, and add a regression test?

bun.lock
```diff
@@ -95,9 +95,9 @@
...
 
-    "@aws-sdk/client-bedrock-agentcore": ["@aws-sdk/client-bedrock-agentcore@3.1131.0", "", { "dependencies": { "@aws-sdk/core": "^3.978.0", "@aws-sdk/credential-provider-node": "^3.972.83", "@aws-sdk/types": "^3.974.5", "@smithy/core": "^3.33.3", "@smithy/fetch-http-handler": "^5.7.2", "@smithy/node-http-handler": "^4.11.3", "@smithy/types": "^4.17.2", "tslib": "^2.6.2" } }, "sha512-xyPxp33VylkesNriC2WOJMkYmj0BX7borAUMojWg/sFcoTyEU0KVs0U6bElZkq7X46/1w6w3hhSGrMVEti7RcA=="],
+    "@aws-sdk/client-bedrock-agentcore": ["@aws-sdk/client-bedrock-agentcore@3.1130.0", "", { "dependencies": { "@aws-sdk/core": "^3.978.0", "@aws-sdk/credential-provider-node": "^3.972.83", "@aws-sdk/types": "^3.974.5", "@smithy/core": "^3.33.3", "@smithy/fetch-http-handler": "^5.7.2", "@smithy/node-http-handler": "^4.11.3", "@smithy/types": "^4.17.2", "tslib": "^2.6.2" } }, "sha512-HgHR24kDJwRN9NNrtWNeuyrJZgl0nXOikeeD2HBAprsRXFl3tvZQw6hHOADHR41sjT4z3+SmqIqN5U7JiaF40Q=="],
```
notgitika: This downgrades `@aws-sdk/client-bedrock-agentcore` from 3.1131.0 to 3.1130.0, and the control client is downgraded on the next entry as well, while `package.json` is unchanged. Could we regenerate the lockfile or otherwise preserve the parent branch's 3.1131.0 resolutions to avoid an unrelated dependency regression?

src/handlers/eval/ondemand/simulate/index.tsx
```diff
@@ -16,7 +16,7 @@ export const createSimulateOnDemandHandler = (core: Core, _io: AppIO) =>
     description: "replay a dataset against a Runtime, then evaluate the sessions client-side",
     flags: [
       flag("runtime-id", "Runtime ID to invoke per scenario", z.string().optional()),
-      flag("qualifier", "Runtime endpoint qualifier (default DEFAULT)", z.string().optional()),
+      flag("endpoint", "Runtime endpoint qualifier (default DEFAULT)", z.string().optional()),
```
notgitika, line 19: This also makes the breaking `--qualifier` → `--endpoint` rename for on-demand `simulate`, while the PR description only calls out `batch-evaluation simulate`. Please document both affected commands. Should we retain `--qualifier` as a deprecated alias to avoid breaking existing scripts?

notgitika (commented):
Looks solid! my agent found some bugs that I added in the comments.

apart from this, the PR title still mentions requiring timestamp timezones, but the current diff and commit msg say that change was dropped. lets update the PR title

src/handlers/eval/ondemand/simulate/index.tsx
jariy17: I'll update the PR description and we haven't released the CLI to the public so it's okay to make breaking changes :)

bun.lock
jariy17: fixed

src/handlers/eval/batch-evaluation/outputConfig.tsx
jariy17: We will share one resolved then and I'll add a regression test!

src/handlers/eval/ondemand/simulate/index.tsx
notgitika: Yeah makes sense to me I think we wanted to use `--endpoint` anyway AFAIK but I wanted to just point it out

notgitika (approved):
Thanks for addressing the comments LGTM!

nborges-aws (approved):
PR LGTM after fixes. Thanks!

PR #2266, opened 2026-09-09
feat(online-eval): add --output-config and --tags, --description on update

jariy17:
Added `--output-config` to `agentcore eval online-eval create/update`. Added `--tags` and `--description` to `agentcore eval online-eval update`. Rerecorded fixtures because we added `--output-config/--tags/--description` to the tests. 

## Live verification — an exploratory account, us-west-2

**Create** with a custom dedicated group and tags → the config came back with `logGroupName: /company/pr4-online-eval-results`, and the CLI-provisioned role's policy granted exactly the intended scope:

```json
{
  "Sid": "WriteEvaluationResults",
  "Action": ["logs:CreateLogGroup", "logs:CreateLogStream", "logs:DescribeLogStreams", "logs:PutLogEvents"],
  "Resource": [
    "arn:aws:logs:us-west-2:123456789012:log-group:/aws/bedrock-agentcore/evaluations/*",
    "arn:aws:logs:us-west-2:123456789012:log-group:/company/pr4-online-eval-results*"
  ]
}
```

## Tests

- **`online-eval.flags.test.tsx` (new, 13 tests)** — output config from inline / stdin reaching Core unchanged on both commands; tags parsed to a map; a non-string tag value rejected; omitted flags staying `undefined`; malformed `--output-config` or `--tags` failing with an **empty Core call log**, which is the assertion that no IAM role was provisioned for a config that was never created; `--description` reaching Core; and, on each command, a second option reading from stdin rejected before any SDK call.
- **`onlineEvalExecutionRole.test.ts`** — the bare-string invariant, the custom dedicated group, `SOURCE_LOG_GROUP` scoping to the source prefix, a reserved-namespace destination adding nothing, and a destination change producing a different policy name.

## Verification

- `bun test` — 3242 pass, 0 fail (228 files)
- `bun run typecheck`, `bun run lint:check`, `bun run format:check` — clean
- Existing online-eval fixture suite passes unchanged, which is the backward-compatibility evidence for the policy document


src/core/onlineEvalExecutionRole.tsx
```diff
@@ -83,10 +82,35 @@ function runtimeLogGroupPrefix(logGroupName: string): string {
...
+  outputConfig: OnlineEvalResultDestination | undefined,
+): string | string[] {
+  const arns = [`${logs}:${SERVICE_RESULT_PREFIX}*`];
+  const cloudWatch = outputConfig?.cloudWatchConfig;
+
+  if (cloudWatch?.resultDestination === "SOURCE_LOG_GROUP") {
+    arns.push(...sampledArns);
+  } else if (
+    cloudWatch?.logGroupName &&
+    !cloudWatch.logGroupName.startsWith(SERVICE_RESULT_PREFIX)
```
jariy17, line 99: Do we need to add permissions to write back in a service log?

src/core/eval.tsx
```diff
@@ -2201,6 +2226,18 @@ function logGroupNamesOf(dataSourceConfig: DataSourceConfig): string[] {
     : [];
 }
 
+function destinationLogGroupNames(
```
jariy17: this is a private function.

nborges-aws (approved):
LGTM clean

notgitika (approved):
LGTM

PR #2271, opened 2026-09-10
feat: support project aware logging in runtime

nborges-aws:
## Description

Adds project-scoped Runtime logging commands.

```sh
agentcore project log runtime
agentcore project log runtime --name FooRuntime --target production
agentcore project log runtime --name FooRuntime --since 1h --level error
```

The command resolves a runtime from the current project into its deployed id and target region. The rest is delegated to the existing observability log search/tail implementation. Projects with only one configured runtime can omit `--name`; projects with multiple runtimes are required to select one explicitly. 

This PR also makes the following updates to observability:
- Generalize the shared logs handler to support both imperative and project commands
- Move runtime log-group derivation into core/observability (previous lived in runtime directory)
- Moves `CoreObservabilityClient` from runtime handler types into observability 
- Generalize project resource selection for reuse by invoke and logging commands.
- Remove the legacy observability client
- updates documentation with project commands

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] `bun run test` (3169 pass, 0 fail)
- [x] I ran `npm run lint`

## Checklist

---

terms of your choice.


README.md
```diff
@@ -251,7 +269,7 @@ agentcore runtime version list --id <runtimeId> --max-results 20
 agentcore runtime endpoint get --id <runtimeId> --qualifier DEFAULT
 agentcore runtime endpoint list --id <runtimeId> --max-results 20
 
-# Follow a Runtime's logs live (Ctrl+C to stop); inside a project --id is optional
+# Follow a Runtime's logs live by physical ID (Ctrl+C to stop)
```
Hweinstock: nit: physical ID feels like strange language? Isn't this the resource id or more specifically the runtime's id?

src/handlers/project/log/runtime.test.tsx
```diff
@@ -0,0 +1,199 @@
...
+afterEach(async () => {
+  process.chdir(originalCwd);
+  await Promise.all(
+    temporaryDirectories
+      .splice(0)
+      .map((directory) => rm(directory, { recursive: true, force: true })),
+  );
+});
+
+async function inProject(
```
Hweinstock, line 46: should we just call create to scaffold the project? If create behavior changes, this will start failing and be unclear why.

Hweinstock (commented):
one testing comment, otherwise lgtm.

Hweinstock (dismissed):

src/core/observability/runtime.ts
```diff
@@ -0,0 +1,5 @@
+export const DEFAULT_ENDPOINT_QUALIFIER = "DEFAULT";
+
+export function runtimeLogGroup(runtimeId: string, endpoint: string): string {
```
aidandaly24: nit: I feel like we can rename this file to something more understandable for the time being. Something just like `runtimeLogGroup.ts`. Or just inline this since it's doing something super basic.

src/core/observability/runtime.test.ts
```diff
@@ -0,0 +1,8 @@
+import { expect, test } from "bun:test";
+import { runtimeLogGroup } from "./runtime";
+
+test("runtimeLogGroup derives the fixed per-runtime endpoint path", () => {
```
aidandaly24: nit: Feel like this test is kinda unnecessary

src/handlers/project/log/runtime.tsx
```diff
@@ -0,0 +1,51 @@
...
+      const name = selectProjectResource(project, "runtime", flags.name, "inspect logs for");
+      const deployed = await core.projectManager.resolveDeployedResource(project, {
+        target: flags.target,
+        resourceType: "runtime",
+        name,
+      });
+      const options = {
+        ...coreOptsFromCtx(ctx),
+        region: deployed.target.region,
+      };
```
aidandaly24, line 33: When environment credentials belong to account A but AWS_PROFILE selects account B, CDK resolves the runtime in A while CloudWatch uses B. I reproduced this through the new command with dummy credentials and fake AWS transport. Both steps should use the same credential provider. This repeats an existing project-invoke issue, rather than resulting from the helper deletion.

aidandaly24 (changes requested):

src/handlers/project/log/runtime.tsx
nborges-aws: Looked into this and this bug finding is valid. Given its a pre-existing bug in invoke paths and not introduced in this PR, I will fix in a follow up asap. It's a mid-sized change; I'd rather not pollute this PR with an OOS fix.

src/core/observability/index.ts
```diff
@@ -1,5 +1,10 @@
 export { CloudWatchClient } from "./cloudWatchClient";
 export { ObservabilityClient } from "./client";
+export const DEFAULT_ENDPOINT_QUALIFIER = "DEFAULT";
```
jariy17, line 3: Do we have this in a common util file?

src/core/observability/index.ts
```diff
@@ -1,5 +1,10 @@
...
+
+export function runtimeLogGroup(runtimeId: string, endpoint: string): string {
```
jariy17, line 5: Do we have this in a common util file?

jariy17 (approved):
Good left some follow up PRs

aidandaly24 (approved):
Thanks for making the fixes

PR #2283, opened 2026-09-11
fix(router): allow required flags in handlers

Hweinstock:
### Problem 
Flags are not able to be marked as required in the flag schema because it broke the default handler behavior on the TUI. 

Ex. if a flag was marked required, the default handler to open the TUI would never trigger, since commander rejected the input first. 

### Solution 
- Avoid marking flags as required at the commander level and instead rely on zod parsing for this enforcement. 
- move flag parsing to a middleware layer, such that we can redirect to TUI before flag validation. 
- use harness get as an example of required flags now working e2e. 

Note: this does change the behavior of custom middleware such it is executed BEFORE flags/args are coerced and validated. Since flags/args are typed as any in middleware, this is unlikely to have a significant effect. 

### Testing/Verification
Migrated the get harness handler and tested a few cases:
```
$ agentcore harness get 
--> opens TUI
$ agentcore harness get --id <HARNESS_ID>
--> returns JSON
$ agentcore harness get --region us-west-2
--> opens TUI
```

Help page:

```
$ agentcore harness get --help                                                       
Usage: agentcore harness get [options]

get a harness

Options:
  --id <id>                      the ID of the harness (required)
  -h, --help                     display help for command

Global Options:
  --region <region>              AWS region
  --debug                        debug logging (default: false)
  --json                         JSON output (default: false)
  --endpoint-url <endpoint-url>  endpoint URL override

```
note that it now shows required. 
### Alternatives Considered
- add a special case in the routing layer to skip parsing when empty flags/args => smaller change, but leads to lots of edge cases + requires router to know about application logic (tui opening). 
- move TUI into routing layer => couples the core framework to a business logic decision. 
- move validation middleware out of the routing layer => might be more explicit, but leads to commander specific logic outside of the routing layer, which breaks the current abstraction. 

### Future Work
- we can migrate all the handlers to use required flags, not manual checks. 
- we can clean up the error message as a follow-up.


Hweinstock:
Harness reviewer called out two valid consequences of this change that I think we can address as follow-ups:
> Missing-required-flag error message regressed

This is actually good, because we now control the error message, and the error type so we not get `InputValidationError` instead of `CommanderError`. We can fix up this message as we see fit. 

> --help no longer marks required options

By marking flags as required in commander, any missing required flags will halt the execution before any of our code runs. We can wire this ourselves as part of the description. Note: this is the existing behavior since we didn't have any required flags.

src/handlers/runtime/logs/logs.test.tsx
```diff
@@ -107,7 +107,7 @@ describe("runtime logs", () => {
     process.chdir(root);
     try {
       await expect(route(["runtime", "logs", "--since", `${SINCE_MS}`])).rejects.toThrow(
-        "required option '--id <id>' not specified",
+        "Invalid value for option '--id': Invalid input: expected string, received undefined",
```
notgitika: are we sure we want this error message instead?

src/router/flags.tsx
```diff
@@ -23,15 +22,14 @@ export function toOption(flag: Flag): Option {
     token = `${long} <${flag.name}>`;
   }
 
-  const option = new Option(token, flag.description);
+  const description =
+    info.required && !info.boolean ? `${flag.description} (required)` : flag.description;
+  const option = new Option(token, description);
```
notgitika, line 27: what do we think about using commander's `helpGroup`? ref: https://github.com/tj/commander.js/blob/master/examples/help-groups.js

output would look like: 
```
  Required Options:
    --id <id>          the ID of the harness

  Options:
    -h, --help         display help
```

this would render required flags in a dedicated section while keeping presentation separate from validation. also, we should avoid makeOptionMandatory(), since it rejects missing flags before the TUI middleware can run

notgitika (commented):
added some comments + I think harness reviewer comments are good as well!

src/handlers/runtime/logs/logs.test.tsx
Hweinstock: This is the existing error message for when a flag does not match the schema, which this now fallsthrough to instead of hitting the commander error. I was originally going to update this and some other error messages as a follow-up but I can pull it in here.

src/router/flags.tsx
Hweinstock: I like that idea! I think @jariy17 is working on something similar about grouping flags, so I think we can revisit this once that lands.

nborges-aws (dismissed):

src/handlers/project/index.ts
```diff
@@ -47,25 +47,14 @@ export function createProjectHandler({ core, io }: ProjectHandlerConfig): Router
...
-      isInteractive()
-        ? createProjectWithWizard.handle(ctx, flags, args)
-        : createProject.handle(ctx, flags, args),
-  };
-  project.handler(createProjectDispatch);
+  project.handler(
+    createCreateProjectHandler({
+      projectManager,
+      io,
+      middlewares: isInteractive() ? [withTuiOnEmptyFlagsAndArgs(core, io)] : [],
```
jariy17: OOS: Is there we can pass this in the root handler [here](https://github.com/aws/agentcore-cli/blob/48ef5eb246075a6b3680a6e045a8329f5e970115/src/handlers/index.tsx#L62)? So we don't need to pass in every handler?

src/router/handler.tsx
```diff
@@ -102,6 +103,7 @@ type CreateHandlerInput<
   arguments?: A;
   handle?: HandleFn<F, A>;
   children?: Handler[];
+  middlewares?: Middleware[];
```
jariy17, line 114: Only the router should know about the middleware, not all the handlers.

jariy17 (dismissed):
Pretty good, left some oos comments

src/handlers/project/index.ts
Hweinstock: yea, discussed offline and this approach should work. I'd like to test and verify this doesn't affect the CLI-only logic so will leave it OOS here.

src/router/handler.tsx
Hweinstock: yeah this is only needed because we have handlers that want middleware that isn't on its parent router (ex. some project commands support tui, some don't). 

However, if we move the tui middleware to the root handler this should be able to go away.

jariy17 (approved):

PR #2284, opened 2026-09-11
feat(wizard): shared wizard shell, used by project create and add runtime

notgitika:
First of a few PRs. This one lands the shared wizard shell and puts two commands (agentcore project create and agetncore project add runtime) on it; the remaining `project add` wizards follow separately.

## What

`src/components/wizard/`:

- `Wizard`: derives the step list from its `<Step>` children, owns position, key handling and the form → running → success | error phases, and renders the standard Layout + Stepper frame. Position is keyed by step name, so a `{condition && <Step/>}` branch appearing or disappearing does not move the user. A submit that streams `ProgressEvent`s is rendered with the same `driveProgress` + `TaskList` as create, build and deploy.
- `Step`: one question, one field.
- `fields`: `TextField`, `ChoiceField`, `Summary`. Each field owns its step's `useInput` and publishes its own key hints, so a screen writes questions and nothing else.

## Callers

- `project create` moved onto it, unchanged in behavior. Its own controller (phases, step components, success panel, hint plumbing) is deleted; the two-level model step stays as a compound field, since the shell has no notion of focus within a step.
- `project add runtime` is new: name → template → description → review. `toAddRuntimeInput` is extracted from the handler so the flag path and the wizard validate the same way.

The runtime wizard scaffolds from a template. The JSON infrastructure flags, `--model-provider`/`--api-key` (a secret read from stdin or `file://`) and the Bedrock Agent import stay on the flag path.

## Shared components

The fields render existing components rather than new ones: `FormTextInput`, `FormRadioGroup`, `KeyValueTable`, `ErrorPanel`, `Layout`, `ui/{stepper,divider,spinner,task-list}`. Two fixes at the source:

- `KeyValueTable` grows into the space it is given, so its percentage key cap has a parent width to resolve against when the table is laid out as a row item.
- `FormTextInput`/`FormRadioGroup` render bare when given no name or help text — the step already asks the question.

## Testing

`bun test` 3218 pass / 0 fail; `tsc --noEmit`, oxlint and prettier clean. New: `wizard.test.tsx` for the shell, `runtime.screen.test.tsx` for the wizard (writes a real project and asserts the spec and scaffolded files), `add.screen.test.tsx` for the add menu. Three help-viewport tests in `project.screen.test.tsx` moved to `add payment-manager`, since `add runtime` now opens a wizard instead of its help.

## Next

1. Migrate `HarnessWizard`/`EndpointWizard` onto the shell (serves `harness create/update`, `endpoint create/update`, and gives `project add harness` a screen).
2. The remaining add wizards


src/middleware/withTuiOnEmptyFlagsAndArgs.tsx
```diff
@@ -54,3 +54,27 @@ export function withTuiOnEmptyFlagsAndArgs(core: Core, io: AppIO): Middleware {
     },
   });
 }
+
+// withTuiWhenInteractive is withTuiOnEmptyFlagsAndArgs behind a TTY gate: a bare
+// invocation opens the TUI only in an interactive session. The gate sits here
+// rather than inside renderTui so that a piped or CI run stays headless and
+// reports a missing required flag as the usage error it is, instead of
+// renderTui's "interactive mode requires a TTY".
+export function withTuiWhenInteractive(core: Core, io: AppIO): Middleware {
```
notgitika, line 63: `withTuiOnEmptyFlagsAndArgs` above opens the TUI even without a TTY, so a piped/CI add runtime would fail with "requires a TTY" instead of the missing --name error. This is that middleware behind a TTY check. create and status each had it hand-rolled; now they share it.

src/components/wizard/context.tsx
```diff
@@ -0,0 +1,51 @@
...
+  // back steps to the previous step, or cancels out of the wizard on the first.
+  back: () => void;
+  // isLast reports whether the active step is the final one, so a field can
+  // label its enter hint "submit" rather than "continue".
+  isLast: boolean;
+  // setHints replaces the footer's action hints. Fields call it via useKeyHints.
+  setHints: (hints: KeyHint[]) => void;
+}
+
+const WizardContext = React.createContext<WizardControls | null>(null);
```
jariy17: Please import just createContext, not the whole React library Object.

notgitika: done!

src/handlers/project/add/runtime/screen.tsx
```diff
@@ -0,0 +1,145 @@
...
+  });
+  const set = (update: Partial<RuntimeFormValues>) =>
+    setValues((current) => ({ ...current, ...update }));
+
+  return (
+    <Wizard
+      breadcrumb={BREADCRUMB}
+      description={DESCRIPTION}
+      onCancel={() => navigate(ADD_MENU)}
+      onSubmit={() =>
```
tejaskash: After the add succeeds the react-query project cache still holds the old spec, so status, invoke, deploy and build in the same TUI session don't see the new runtime. Call setQueryData with the returned project, or drop staleTime Infinity on the seed.

src/handlers/project/add/index.ts
```diff
@@ -17,9 +18,23 @@ import type { AddProjectResourceConfig } from "./types";
...
-  const projectAdd = new Router("add", "add project resources");
-  projectAdd.use(withProject({ projectManager: config.projectManager, cwd: process.cwd() }));
+export function createAddProjectResourceHandler(
+  config: AddProjectResourceConfig,
+  core: Core,
+): Router {
+  // The resources with a wizard of their own. Every other resource is listed in
+  // the add menu as command line only and opens its help instead (see
+  // CliOnlyScreen).
+  const projectAdd = new Router("add", "add project resources").supportedTuiCommands("runtime");
```
tejaskash, line 29: Bare `agentcore project add` prints Commander help and exits 1 because this router has no `.default(renderTui(core, io))`. Navigating here from `agentcore project` works, so the two entry points disagree.

src/components/KeyValueTable.tsx
```diff
@@ -23,7 +26,7 @@ export function KeyValueTable({ items }: KeyValueTableProps) {
   // Two boxes rather than one padded string, so a value that wraps continues
   // under itself, not under the key.
   return (
-    <Box flexDirection="column">
+    <Box flexDirection="column" flexGrow={1}>
```
tejaskash: flexGrow here makes the table take all free height in any growing column parent. The gateway policy generate screen now has 14 blank lines between the table and the input. Put flexDirection="column" on the two bordered wrapper Boxes instead and leave the shared table alone.

src/components/wizard/Wizard.tsx
```diff
@@ -0,0 +1,303 @@
...
+  return (
+    <Layout
+      breadcrumb={breadcrumb}
+      description={description}
+      keyHints={footerHints(phase, hints, retryable, doneLabel)}
+    >
+      <Box flexDirection="column">
+        {phase.kind === "form" && (
+          <>
+            <Box paddingX={1}>
```
tejaskash: The old create screen had flexShrink={0} on this Box. Without it the stepper paints over the header Divider on short terminals (80x14 and 80x18).

src/projectSchemas/runtime.ts
```diff
@@ -21,6 +21,18 @@ export const AgentNameSchema = z
...
   );
+// RUNTIME_NAME_MAX_LENGTH is six characters short of the 48 a name may hold,
+// because a runtime scaffolds a memory called `${runtimeName}Memory` (see
+// getDefaultMemorySpec) and MemoryNameSchema caps that at 48.
+export const RUNTIME_NAME_MAX_LENGTH = 42;
+
+// RuntimeNameSchema is the one runtime-name rule: the `--name` flag and the
+// interactive wizard both validate against it, so neither accepts a name the
+// other rejects.
+export const RuntimeNameSchema = AgentNameSchema.max(
```
tejaskash: AgentNameSchema already says max 48 in its message, so the first error the user sees contradicts the max 42 help line. The 42 only matters for the memory suffix, which most templates never create, and RuntimeResourceConfigSchema doesn't enforce it. Truncate in getDefaultMemorySpec and validate with AgentNameSchema instead.

src/components/wizard/fields.tsx
```diff
@@ -0,0 +1,255 @@
...
+}
+
+// ChoiceField is a single-choice step. The arrow arithmetic here replaces the
+// copy of it in each of the hand-written wizards. It takes no `label`: one of
+// the choices is always selected, so there is no validation message to name.
+export function ChoiceField<T>({ help = "", choices, value, onChange }: ChoiceFieldProps<T>) {
+  const { advance, back, isLast } = useWizard();
+
+  useKeyHints([
+    { key: "↑↓", label: "choose" },
```
tejaskash: Every other arrow hint in src, including the model step in this same wizard, says "navigate".

src/components/wizard/Wizard.tsx
```diff
@@ -0,0 +1,303 @@
...
+    return [
+      ...(retryable ? [{ key: "r", label: "retry" }] : []),
+      { key: "esc", label: "back" },
+      { key: "ctrl+c", label: "quit" },
+    ];
+  }
+  return [...fieldHints, { key: "esc", label: "back" }, { key: "ctrl+c", label: "quit" }];
+}
+
+function SuccessPanel({
```
tejaskash: This is the same block as ConfirmAction's SuccessBody, and isProgressStream is the same check as isProgressGenerator can we export and re-use from one place

tejaskash (commented):
Few comments

src/components/wizard/fields.tsx
```diff
@@ -0,0 +1,255 @@
...
+  // here the same way the handlers state it in their own bodies.
+  required?: boolean;
+  // schema validates the value as typed before advancing. Pass the schema the
+  // flag declares.
+  schema?: z.ZodType;
+  // json parses the value before validating it. A JSON flag belongs here rather
+  // than in TextAreaField unless the value is genuinely prose: JSON is valid on
+  // one line, and this field can be edited in place.
+  json?: boolean;
+  // example is a dimmed line under `help`, kept on screen while the user types.
```
tejaskash: I don't see any other caller setting `json` or `example` on TextField, and both screens pass `onError="retry"`, so the `"exit"` mode and `ExitOnError` never run. Remove `json`, `example`, `onError` and `ExitOnError`. Add them back when a screen needs them.

src/components/wizard/fields.tsx
```diff
@@ -0,0 +1,255 @@
...
+// firstIssue renders the schema's own message, so the wizard rejects exactly
+// what the flag-driven path rejects and says the same thing about it. The issue
+// path is prefixed when there is one: for a nested value — a component inside a
+// components map, say — "expected object, received string" alone does not say
+// which key is wrong.
+function firstIssue(schema: z.ZodType, value: unknown): string | undefined {
+  const parsed = schema.safeParse(value);
+  if (parsed.success) return undefined;
+  const issue = parsed.error.issues[0];
+  if (!issue) return "invalid value";
```
tejaskash: Zod always returns at least one issue on failure. Line 200 in this file already uses the non-null assertion, lets do the same here.

src/components/wizard/fields.tsx
```diff
@@ -0,0 +1,255 @@
...
+  json?: boolean;
+}
+
+// validateEntry returns the message that should block the step, or undefined to
+// let it advance. It sits outside TextField so that a further field collecting
+// text — a multi-line one, say — refuses the same input for the same stated
+// reason rather than growing its own rules.
+function validateEntry(
+  value: string,
+  { label, required, schema, json = false }: ValidateOptions,
```
tejaskash: The new files carry multi-line comment - can we audit and keep the comment only when the code is genuinely hard to follow?

src/components/wizard/Step.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { Box, Text } from "ink";
+import { darkTheme } from "../ui/_core.js";
+
+const theme = darkTheme;
+
+export interface StepProps {
+  // name is the step's stable key. Position is tracked by key rather than by
+  // index because branches have different lengths: a conditional step that
+  // appears or disappears must not shift the user to a different question.
+  name: string;
```
jariy17: Can we call this key then?

src/components/wizard/Step.tsx
```diff
@@ -0,0 +1,43 @@
...
+  // title labels the step in the Stepper; defaults to `name`.
+  title?: string;
+  // question is the one-line prompt shown under the Stepper. The Stepper
+  // already names the step, so the body opens with the question itself.
+  question?: string;
```
jariy17: Prompt/Request is better here

src/components/wizard/Wizard.tsx
```diff
@@ -0,0 +1,303 @@
...
+    const list = stepElements.map((element) => ({
+      key: element.props.name,
+      title: element.props.title ?? element.props.name,
+    }));
+    // Position is keyed by name, so two steps sharing one would make advance()
+    // land on the first of them forever. Catch that at render time, where the
+    // author sees it, instead of as a wizard that quietly cannot move on.
+    const seen = new Set<string>();
+    for (const step of list) {
+      if (seen.has(step.key)) throw new Error(`duplicate <Step name="${step.key}">`);
```
jariy17: Throw a cli error that's a internal error so we can make sure this doesnt happen.

jariy17: Throw a cli error that's a internal error so we can detect it in telemetry and fix it asap or can we catch this during build time?

src/components/wizard/Wizard.tsx
```diff
@@ -0,0 +1,303 @@
...
+  result: WizardSubmitResult,
+): result is AsyncGenerator<ProgressEvent, unknown> {
+  return (
+    result !== null &&
+    typeof result === "object" &&
+    typeof (result as AsyncIterable<ProgressEvent>)[Symbol.asyncIterator] === "function"
+  );
+}
+
+function toError(error: unknown): Error {
```
jariy17: Whats the point of this function? Just throw the error

src/components/wizard/Wizard.tsx
```diff
@@ -0,0 +1,303 @@
...
+        </Box>
+      )}
+    </Box>
+  );
+}
+
+// ExitOnError tears the TUI down through exit(error): that rejects the
+// waitUntilExit() renderTuiAt awaits, so the failure is reported by the normal
+// CLI error path instead of as a React stack trace.
+function ExitOnError({ error }: { error: Error }) {
```
jariy17: Isn't this an issue everywhere then? Like there must be a middleware component that handles errors throw by the tui

src/handlers/project/index.ts
```diff
@@ -27,46 +27,29 @@ type ProjectHandlerConfig = {
...
-    handle: (ctx, flags, args) =>
-      isInteractive()
-        ? createProjectWithWizard.handle(ctx, flags, args)
-        : createProject.handle(ctx, flags, args),
-  };
-  project.handler(createProjectDispatch);
-  project.handler(createAddProjectResourceHandler(config));
+  // create wizard; any user-supplied flag, --json, or a non-TTY invocation keeps
+  // the headless handler (see withTuiWhenInteractive).
+  const tuiWhenInteractive = withTuiWhenInteractive(core, io);
```
jariy17, line 50: I thought the middleware handles when jo flags are passed. Do we still need this?

jariy17 (dismissed):
Really good left some follow up nits

src/components/wizard/Step.tsx
notgitika: react resserves `key`

notgitika: renamed this to stepKey

src/components/wizard/Wizard.tsx
notgitika: Updated this to throw an AgentCoreCLIError named DuplicateWizardStepError, classified as an internal error. build-time validation isn’t reliable here because conditional steps are resolved at render time, so this catches the error as soon as the wizard renders and makes it identifiable in telemetry

src/handlers/project/add/runtime/screen.tsx
notgitika: done

src/handlers/project/add/index.ts
notgitika: Added the default TUI handler and test

src/components/KeyValueTable.tsx
notgitika: removed flexgrow

src/components/wizard/Wizard.tsx
notgitika: restored

src/projectSchemas/runtime.ts
notgitika: Standardized on 48 and truncated generated memory names

src/components/wizard/fields.tsx
notgitika: changed

src/components/wizard/Wizard.tsx
notgitika: extracted

src/components/wizard/fields.tsx
notgitika: removed this

src/components/wizard/fields.tsx
notgitika: used the non-null assertion

src/components/wizard/fields.tsx
notgitika: removed :)

src/components/wizard/Step.tsx
notgitika: done

src/components/wizard/Wizard.tsx
notgitika: done

src/components/wizard/Wizard.tsx
notgitika: Removed the special exit-error path; failures now render through ErrorPanel

src/handlers/project/index.ts
notgitika: Yes, the middleware is what handles the no-flags case. This line just configures it once so it can be applied to both create and status; it isn’t an additional check. 

we still need to apply it explicitly to each handler, and the interactive wrapper also ensures non-TTY invocations remain headless. I can rename the local variable if that would make its purpose clearer

notgitika: You’re right that the existing middleware handles the no-flags check. This wrapper only adds the TTY gate: interactive invocations open the TUI, while non-TTY invocations use the normal headless handler. 

If this is being centralized elsewhere, we can remove it here. I can make this change in a follow up PR by adding it into the existing middleware. What do you think?

src/components/wizard/Wizard.tsx
```diff
@@ -119,48 +82,41 @@ export function Wizard({
...
-      // its output, ✓ once it settles.
-      if (isProgressStream(result)) await driveProgress(result, setTasks);
+      if (isProgressGenerator(result)) await driveProgress(result, setTasks);
       else await result;
       setPhase({ kind: "success" });
     } catch (error) {
-      setPhase({ kind: "error", error: toError(error) });
+      setPhase({
+        kind: "error",
+        message: error instanceof Error ? error.message : String(error),
```
jariy17: Follow up: Should this be an AgentCoreCLIError

jariy17 (dismissed):
Left one nit

src/components/wizard/Wizard.tsx
notgitika: Done — the wizard now normalizes caught values with `AgentCoreCLIError.fromError(error)`, stores the CLI error in its error phase, and passes its message to `ErrorPanel`.

notgitika: nice catch done!

jariy17 (approved):

PR #2286, opened 2026-09-11
fix(errors): use resource not found over input validation for missing resources

Hweinstock:
### Problem

  The refactor branch distinguishes malformed input (`InputValidationError`) from well-formed input naming an absent resource (`ResourceNotFoundError`):

Ex. already throws `ResourceNotFoundError` for lookup misses (e.g. https://github.com/aws/agentcore-cli/blob/0caec7fe932a0d4f8f8ce2d4821fee0664ab77cf/src/core/project/manager.tsx#L573 ,
  https://github.com/aws/agentcore-cli/blob/0caec7fe932a0d4f8f8ce2d4821fee0664ab77cf/src/core/project/manager.tsx#L586 ), but several other paths still throw plain `InputValidationError` for the same "named
  entity absent" case
  
  The language is also inconsistent across multiple places. 

### Solution 
- migrate existing cases. 

### Verification 
- test-pass, messages updated

### DIFF
--- a/src/core/dev/codezip.test.ts
+++ b/src/core/dev/codezip.test.ts
@@ -102,7 +102,7 @@ describe("CodeZipDevRunner", () => {
     tempDirectories.push(root);
 
     await expect(collect(harness().runner.run(input(root, runtime())))).rejects.toThrow(
-      /runtime code directory not found/,
+      /no runtime code directory exists at/,
     );
   });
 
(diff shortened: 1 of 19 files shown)


notgitika (approved):
makes me think if we should invest in something that would help us enforce this.

also, I'm seeing that `class ResourceNotFoundError extends InputValidationError` I'm not sure I understand this. Can we make the heirarchy better like so:

```
class UserError extends AgentCoreCLIError
class InputValidationError extends UserError
class ResourceNotFoundError extends UserError
```

Hweinstock:
Yeah fully agree that is confusing. I think we can rework the structure of the errors to be more normalized.

jariy17 (approved):

PR #2288, opened 2026-09-11
feat(wizard): interactive project add config-bundle

notgitika:
Stacked on #2284 — review the last commit only (`feat(wizard): interactive project add config-bundle`). The earlier commits in this diff are #2284 and go away once it merges.

Adds the wizard behind a bare `agentcore project add config-bundle`: name → components → description → branch → commit → review. Enter on the success panel goes back to the add menu.

Bundle assembly moves out of the flag handler into `toAddConfigBundleInput`, which both paths call, so the wizard cannot accept a bundle the flags would reject. `--components` and the wizard's components step share one `ComponentsSchema`, and `mainline` is now a named default rather than a literal in the flag declaration.

No behaviour change on the flag path. Every other resource under `project add` stays command line only.

### DIFF
--- a/src/components/FormRadioGroup.tsx
+++ b/src/components/FormRadioGroup.tsx
@@ -21,7 +21,7 @@ export interface FormRadioGroupProps {
 // FormRadioGroup renders a column of radio rows. It is fully controlled: the
 // parent owns the focused index and the key handling that moves it.
 export function FormRadioGroup({
-  name,
+  name = "",
   helpText,
   options,
   focusedIndex,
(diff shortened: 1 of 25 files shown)


notgitika:
Thanks — going through these.

**Missing `--components`.** The guard is still there. `handle` keeps both early checks, `!flags.name` and `!flags.components`, and the second one throws `required option '--components <components>' not specified` before `resolveText` or `parseJsonFlag` runs; only the parse and schema step moved into the shared builder. I checked it against the built CLI, and there is now a dispatch test asserting the exact string for `--name foo` with no `--components`, so it can't regress silently.

**Runtime name schema.** Not a breaking change. `ProjectRuntimeSchema.shape.name` is `AgentNameSchema` (`/^[a-zA-Z][a-zA-Z0-9_]{0,47}$/`), so a hyphenated name was already rejected downstream — the flag just says so earlier now, with the same message. The old `max(42)` cap is preserved by `RuntimeNameSchema`.

**`onError`.** Agreed, and applied to both add screens: a failed add now reports itself and hands the form back, so esc returns to the review step with every answer intact instead of exiting. `project create` keeps `"exit"` — that one owns the terminal it was launched in. Covered by a test that adds a duplicate runtime name and drives the form back.

One unrelated change in the latest push: the components step is now a textarea (enter inserts a newline, ctrl+d continues), since a components map is usually pasted pretty-printed, and the description step is gone — `--description` still sets one.

PR #2289, opened 2026-09-11
fix(import): strip reserved aws: tags from imported resources

jariy17:
## Problem

`agentcore import <resource>` copies every tag from the source resource into `agentcore.json`. CloudFormation stamps `aws:cloudformation:{stack-id,stack-name,logical-id}` on every resource it provisions, and a resource retained after its stack is deleted keeps these as orphans. Because `aws:`-prefixed keys are reserved and not customer-writable, they fail `TagKeySchema` during config validation — blocking the import before any AWS call:

```
runtimes[0].tags.aws:cloudformation:stack-name: Invalid key in record
runtimes[0].tags.aws:cloudformation:logical-id: Invalid key in record
runtimes[0].tags.aws:cloudformation:stack-id: Invalid key in record
```

## Fix

Add a shared `stripReservedTags()` helper in `import-utils.ts` and apply it at all four import read-paths (runtime, memory, gateway, evaluator). User tags are preserved; `aws:`-prefixed tags are dropped from the local config and stay untouched on the resource in AWS.

The `aws:` refine in `TagKeySchema` is intentionally left in place — it is correct for user-authored tags. Filtering happens at the import read-path instead.

## Testing

- New `strip-reserved-tags.test.ts`: reserved keys dropped, user tags kept, `undefined` when only reserved remain, and output passes `TagsSchema.parse`.
- Full import suite green (373 tests); typecheck, lint, build clean.
- Verified end-to-end against a real CloudFormation-provisioned runtime carrying `aws:cloudformation:*` tags: pre-fix reproduces the validation failure above; post-fix the import writes `agentcore.json` successfully.
- Also updates the `vi.mock('../import-utils')` factory in `import-runtime-handler.test.ts` to spread the real module, so new utils are no longer silently undefined in that test.

### DIFF
--- a/src/cli/commands/import/import-utils.ts
+++ b/src/cli/commands/import/import-utils.ts
@@ -15,6 +15,20 @@ import * as path from 'node:path';
 
 const { green, reset } = ANSI;
 
+/**
+ * Drop AWS-managed reserved tags (aws:* keys) when importing an existing resource.
+ * CloudFormation stamps aws:cloudformation:{stack-id,stack-name,logical-id} on every
+ * resource it provisions; a runtime retained after its stack is deleted keeps them as
+ * orphans. They are not customer-writable, so persisting them into agentcore.json would
+ * fail TagKeySchema ("aws:" is reserved) and block the import.
+ * Returns undefined when nothing user-owned remains, so callers can omit the tags field.
+ */
+export function stripReservedTags(tags: Record<string, string> | undefined): Record<string, string> | undefined {
+  if (!tags) return undefined;
+  const userTags = Object.fromEntries(Object.entries(tags).filter(([key]) => !key.startsWith('aws:')));
+  return Object.keys(userTags).length > 0 ? userTags : undefined;
+}
+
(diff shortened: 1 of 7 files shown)


src/cli/commands/import/import-utils.ts
```diff
@@ -15,6 +15,20 @@ import * as path from 'node:path';
 
 const { green, reset } = ANSI;
 
+/**
+ * Drop AWS-managed reserved tags (aws:* keys) when importing an existing resource.
+ * CloudFormation stamps aws:cloudformation:{stack-id,stack-name,logical-id} on every
+ * resource it provisions; a runtime retained after its stack is deleted keeps them as
+ * orphans. They are not customer-writable, so persisting them into agentcore.json would
```
Hweinstock, line 22: nit: this comment is a little verbose. The first half feels useful, but the decision to not write to agentcore.json feels unnecessary.

Hweinstock (approved):
lgtm

PR #2293, opened 2026-09-14
feat(project): runtime-endpoint add/remove commands + surface in status

jariy17:
## What

Adds project-based management of **runtime endpoints** (named version aliases of a runtime) to the refactor CLI, and surfaces them in `project status`.

Endpoints live in a runtime's `endpoints` record in `agentcore.json` and are provisioned by the L3 CDK construct at deploy time — no direct SDK call, matching how gateways/policies/payment-connectors work.

## Commands

- `agentcore project add runtime-endpoint --runtime <r> --name <n> [--version <v>] [--description <d>]`
- `agentcore project remove runtime-endpoint --runtime <r> --name <n>`

## Changes

- **ProjectManager** `addResource`/`removeResource`: handle the nested `endpoints` record — duplicate-name and unknown-runtime checks; prune the record when the last endpoint is removed.
- **Types**: `runtime-endpoint` added to `AddResourceInput`, `RemoveResourceInput`, `DeployableResource`, and the mutation-output parent map.
- **Add handler**: new `project/add/runtime-endpoint/` (+ unit tests).
- **Remove handler**: `runtime-endpoint` resource type + `--runtime` parent flag.
- **Status**: resolver maps endpoints as runtime children and resolves each to its CFN export ARN, so they appear in CLI JSON and the TUI tree; selecting one opens its detail via `runtime endpoint get`.

## Testing

- `bun run typecheck` clean; `bun test` for project/status/cdk suites green.
- Verified end-to-end against a live deployment: add → deploy creates a real `AWS::BedrockAgentCore::RuntimeEndpoint`; remove → deploy deletes it; endpoint shows as a `deployed` child in `project status` (CLI + TUI) with a working detail page.

## Follow-ups (not in this PR)

- Bad endpoint names surface a raw zod `Invalid key in record` error; a friendly validation message in the add handler would be nicer.


src/core/project/backends/cdk.ts
```diff
@@ -538,7 +541,13 @@ export class CdkBackend implements ProjectBackend {
     };
 
     return [
-      ...spec.runtimes.map(({ name }) => resolve("runtime", name)),
+      ...spec.runtimes.map((runtime) =>
+        resolve("runtime", runtime.name, {
+          children: Object.keys(runtime.endpoints ?? {}).map((endpointName) =>
```
jariy17, line 546: Why can't you do this? 
```
children: (runtime.endpoints ?? []).map(({ name }) =>
            resolve("runtime-endpoint", name, { owner: runtime.name }),
          ),
```

src/handlers/project/remove/index.ts
```diff
@@ -27,6 +27,11 @@ export const createRemoveProjectHandler = (config: RemoveProjectResourceConfig)
         "name of the parent payment manager for a connector",
         z.string().min(1).optional(),
       ),
+      flag(
+        "runtime",
```
Hweinstock, line 31: i see this follows the pattern for payments, but why do we need this additional flag if the name is enforced to be globally unique? 

runtime endpoint: https://github.com/aws/agentcore-cli/blob/3de39dfdfa2e58556efabc5099d9b39f310ed5ea/src/core/project/manager.tsx#L332

payment manager: https://github.com/aws/agentcore-cli/blob/3de39dfdfa2e58556efabc5099d9b39f310ed5ea/src/core/project/manager.tsx#L318

src/core/project/manager.tsx
```diff
@@ -1219,6 +1267,8 @@ function toProjectSpecKey(resourceType: ProjectResource) {
     case "payment-manager":
     case "payment-connector":
       return "payments";
+    case "runtime-endpoint":
```
Hweinstock, line 1270: nit: should we move this next to runtime above?

```
case "runtime-endpoint":
case"runtime":
   return "runtimes";
```

src/core/project/manager.tsx
```diff
@@ -592,6 +620,26 @@ export class FsProjectManager implements ProjectManager {
       removed = connectors.length !== manager.connectors.length;
       payments[managerIndex] = { ...manager, connectors };
       newSpec = { ...existingProjectSpec, payments };
+    } else if (input.resourceType === "runtime-endpoint") {
```
Hweinstock, line 623: man this manager code is tough to follow (not your changes)

src/core/project/manager.tsx
```diff
@@ -317,6 +318,20 @@ export class FsProjectManager implements ProjectManager {
           `a payment connector with name '${input.resourceConfig.name}' already exists in manager '${input.managerName}'`,
         );
       }
+    } else if (input.resourceType === "runtime-endpoint") {
```
Hweinstock, line 321: is there any reason this logic can't live in the switch below? I'm hoping we can remove as much branching as possible.

Hweinstock (approved):
implementation fully makes sense and follows patterns, few comments on those patterns being problematic but maybe OOS.

src/core/project/manager.tsx
jariy17: Yeah we can fix it after the release.

src/core/project/manager.tsx
jariy17: Yeah we can.

src/handlers/project/remove/index.ts
jariy17: Line 332 only enforces runtime endpoint to be unique for a desired runtime. Therefore, we need this flag.

src/core/project/manager.tsx
jariy17: It can be put in the switch statement below but in this function every resource checks their name uniqueness here. Therefore, we should leave it here till we clean this up.

aidandaly24 (approved):

PR #2294, opened 2026-09-14
fix(online-eval): stop re-scoping the execution role on update

jariy17:
## What

Make `eval online-eval update` match the `harness update` pattern for execution roles: **only create the role (on `create`), never update/re-scope it on `update`.** If a customer wants their execution role to be fully managed, they should managed online-evals through the project based commands. 

## Changes

- **`core/eval.tsx`** `updateOnlineEvaluationConfig`: delete the role re-scope branch; the send path just forwards `evaluationExecutionRoleArn`. Return type drops `roleScopeWarning`. Unused role-scoping imports pruned (`grantOnlineEvalScope` kept — `create` still uses it).
- **`handlers/eval/online-eval/update`**: remove the `--update-role` flag and the stderr warning block. `--role-arn` pass-through kept (harness keeps its equivalent).
- **`handlers/eval/types.tsx`**: drop the `updateRole` input field and the now-unused `RoleScopeWarning` type; simplify the client return type.
- **test**: rewrite the "execution role scoping on update" case to assert the role is left untouched (no stderr advisory) when the data source moves.

Net −191 lines.

## Testing

- `bun test` — 3204 pass, 0 fail
- `bun run typecheck`, `oxlint`, `prettier --check` — all clean


nborges-aws (approved):
Clean LGTM

PR #2295, opened 2026-09-14
feat(wizard): interactive project add memory

notgitika:
A wizard for `project add memory`, on the shared shell: name, which long-term strategies to extract, how long raw events are kept, review.

Everything else the flags accept stays flag-only (description, indexed keys, stream delivery, KMS key, execution role, tags, and the JSON form of `--strategies`).

Two new fields the shell was missing:
- `MultiChoiceField` — a subset of its choices, over the existing `FormCheckboxMultiSelect`.
- `NumberField` — keeps the answer as typed, validates the number it parses to.

Both entry points now go through `toAddMemoryInput`, so the flags and the wizard cannot accept a memory the other would reject.

The help-scrolling test in `project.screen.test.tsx` moves to `add gateway-target`, since `add memory` is no longer command line only.


src/components/wizard/fields.tsx
```diff
@@ -21,14 +22,19 @@ interface ValidateOptions {
   label: string;
   required: boolean;
   schema?: z.ZodType;
+  // number validates the number the answer parses to rather than the text, so a
+  // numeric flag's own schema can bound the field.
+  number?: boolean;
```
jariy17, line 27: Wouldn't the zod schema verify if it's a whole number? Also, do we need `required` for the same reason.

src/handlers/project/add/memory/index.ts
```diff
@@ -7,16 +7,36 @@ import {
...
 import { TagsSchema } from "../../../../projectSchemas/tags";
+import type { AddResourceInput } from "../../types";
 import { addProjectResource } from "../shared";
 
 // The service default for raw event retention
-const DEFAULT_EVENT_EXPIRY_DURATION = 30;
+export const DEFAULT_EVENT_EXPIRY_DURATION = 30;
+
+/** The bounds the service puts on raw event retention, in days. */
+export const EventExpiryDurationSchema = z.number().int().min(3).max(365);
```
jariy17, line 24: Let's use this Schema within MemorySchema to prevent duplication.

jariy17 (approved):
Good just have follow up nits

nborges-aws (approved):
LGTM nice work!

PR #2297, opened 2026-09-14
feat: use YAML for harness project configuration

aidandaly24:
## Summary

- Replace project `harness.json` files with commented `harness.yaml` configuration.
- Generate the file from a readable Handlebars asset template using the existing renderer, with `yaml.stringify` for dynamic values. A complete default-output snapshot replaces fragment-level presentation assertions.
- Keep agent instructions in the conventional `system-prompt.md` file. Inline `systemPrompt` text overrides that file for deployment and CLI export; summary prompts are inline text.
- Newly scaffolded harnesses use managed memory instead of the prior omitted/disabled default because create_harness api creates memory by default. Explicit memory choices are preserved, existing configurations with omitted memory stay omitted, and tools remain opt-in.
- Preserve the thin generated CDK app introduced by #2249. Harness loading now lives in the library's existing project transform, with YAML support supplied by [CDK #383](https://github.com/aws/agentcore-l3-cdk-constructs/pull/383); CLI export continues reading locally.

<details>
<summary>Generated harness.yaml</summary>

```yaml
# Optional settings are shown with example values.

name: assistant

# Instructions come from system-prompt.md unless systemPrompt is set inline.
# systemPrompt: You are a helpful assistant.

# Model
model:
  provider: bedrock
  modelId: global.anthropic.claude-sonnet-4-6
  # Output tokens per model call, rather than across the whole invocation.
  # maxTokens: 4096

# Tools
# Code Interpreter and Browser use built-in resources when no ARN is specified.
# awsIam uses the harness execution role, which must allow Gateway invocation.
# https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/gateway.html
# tools:
#   - name: code_interpreter
#     type: agentcore_code_interpreter
#   - name: browser
#     type: agentcore_browser
#   - name: research
#     type: remote_mcp
#     config:
#       remoteMcp:
#         url: https://mcp.example.com/mcp
#   - name: company_tools
#     type: agentcore_gateway
#     config:
#       agentCoreGateway:
#         gatewayArn: arn:aws:bedrock-agentcore:us-west-2:123456789012:gateway/example-1234567890
#         outboundAuth:
#           awsIam: {}

# Tool patterns: @<server-name>/<tool-name> or @builtin.
# This controls agent tool selection, not IAM permissions.
# allowedTools:
#   - "@builtin"
#   - "@research/search"

# Skill path sources refer to files already present in the runtime container.
# skills:
#   - s3Uri: s3://your-skills-bucket/skills/research/
#   - gitUrl: https://github.com/your-org/agent-skills.git
#     path: skills/research
#   - path: /opt/skills/research

# Memory
# https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/memory.html
# Alternatives: existing (name or ARN), disabled.
memory:
  mode: managed
  # Managed-memory settings. Absent values use service defaults.
  # Event retention is in days.
  # strategies:
  #   - SEMANTIC
  #   - SUMMARIZATION
  # eventExpiryDuration: 30
  # Existing-memory reference: project resource name or external ARN.
  # name: ConversationMemory
  # arn: arn:aws:bedrock-agentcore:us-west-2:123456789012:memory/example-1234567890

# Execution limits apply per invocation, across all model calls.
# maxIterations: 15
# maxTokens: 20000
# timeoutSeconds: 300

# Truncation changes the context sent to the model, not the saved memory.
# truncation:
#   strategy: sliding_window
#   config:
#     slidingWindow:
#       messagesCount: 40

# dockerfile and containerUri are mutually exclusive. With neither set, the
# service-provided environment is used. Dockerfile paths are relative to this directory.
# dockerfile: Dockerfile
# containerUri: 123456789012.dkr.ecr.us-west-2.amazonaws.com/my-harness:latest

# Environment values are stored in plaintext.
# environmentVariables:
#   LOG_LEVEL: info

# Deployment creates a role when executionRoleArn is absent.
# executionRoleArn: arn:aws:iam::123456789012:role/MyHarnessRole

# Tags
# tags:
#   team: support
#   environment: development
```

</details>

**Breaking change:** Project harness configuration now uses `harness.yaml`; `harness.json` is no longer supported.

**Release dependency:** The thin app requires the CDK `refactor` API plus [CDK #383](https://github.com/aws/agentcore-l3-cdk-constructs/pull/383). Before shipping, publish that CDK build and sync the generated app's package pin to the release containing both; verification below uses a local refactor build, not the older CDK `main`.

## Testing

- Merged CLI `refactor` through `de54554996bbf7f7c0be948d162cb248d753049f`; **3,386 CLI tests passed**, with six snapshots. Typecheck, lint, formatting, secrets checks, and build passed.
- **960 CDK tests passed** on the matching refactor build with #383, including the existing inline/fixed-file tests and focused YAML syntax and prompt-content coverage.
- The built Node CLI scaffolded a fresh project, installed the locally built CDK package with normal peer dependencies, and successfully ran `agentcore project build`. Separate synthesis checks confirmed fixed-file loading and inline override.
- Installed CDK bytes match the local build. The generated app now uses `aws-cdk-lib ~2.269.0`, satisfying the CDK refactor's existing peer requirement.
- Upstream removed the generated Jest scaffolding; this merge preserves that removal and keeps the harness loading tests in the CDK library. No new AWS deployment was needed.

**Observed service limitation:** AWS runtime creation rejected a summary prompt with a trailing newline. The successful smoke run used a single-line summary; the CLI does not trim prompt contents.


nborges-aws (dismissed):

src/core/project/templates/harnessYaml.ts
```diff
@@ -0,0 +1,172 @@
...
+  {
+    name: "company_tools",
+    type: "agentcore_gateway",
+    config: {
+      agentCoreGateway: {
+        gatewayArn: "arn:aws:bedrock-agentcore:us-west-2:123456789012:gateway/example-1234567890",
+        outboundAuth: { awsIam: {} },
+      },
+    },
+  },
```
AlexanderRichey: Any reason why we're not using handlebars like the other templates? I also don't see a test asserting YAML output anywhere. Without either of these, I can't tell what the generated YAML would look like.

src/core/project/templates/harnessYaml.ts
```diff
@@ -0,0 +1,172 @@
...
+  {
+    comment: "Deployment creates a role when executionRoleArn is absent.",
+    values: { executionRoleArn: "arn:aws:iam::123456789012:role/MyHarnessRole" },
+  },
+  { comment: "Tags", values: { tags: { team: "support", environment: "development" } } },
+];
+
+/** Serializes supplied values separately from inactive examples, so examples cannot become defaults. */
+export class HarnessYamlRenderer {
```
jariy17: Can't we use jinja to generate the yaml file?

src/assets/cdk/io/harnessConfig.ts
```diff
@@ -0,0 +1,98 @@
...
+import { parseDocument } from "yaml";
+
+const MAX_PROMPT_FILE_SIZE = 1024 * 1024;
+const PROMPT_FIELDS = [
+  ["systemPrompt"],
+  ["truncation", "config", "summarization", "summarizationSystemPrompt"],
+] as const;
+
+/** Project-file I/O only; callers validate the resolved data with their HarnessSpecSchema. */
+export class HarnessConfigReader {
```
jariy17: Can't we use a library like 'yaml' to convert the yaml into a typescript object

src/io/harnessConfig.ts
```diff
@@ -0,0 +1,101 @@
...
+import { DeserializationError } from "../errors";
+
+const MAX_PROMPT_FILE_SIZE = 1024 * 1024;
+const PROMPT_FIELDS = [
+  ["systemPrompt"],
+  ["truncation", "config", "summarization", "summarizationSystemPrompt"],
+] as const;
+
+/** Project-file I/O only; callers validate the resolved data with their HarnessSpecSchema. */
+export class HarnessConfigReader {
```
jariy17: Same as the comments above.

jariy17 (commented):
Pretty good, see if we can use `yaml` and jinja to simplify the code here.

src/assets/cdk/io/harnessConfig.ts
jariy17: nevermind, I see we need to process another file for the system prompt.

src/assets/cdk/package.json
```diff
@@ -17,7 +17,7 @@
   },
   "dependencies": {
     "@aws/agentcore-cdk": "0.1.0-alpha.53",
-    "aws-cdk-lib": "~2.266.0",
+    "aws-cdk-lib": "~2.269.0",
```
jariy17, line 20: why you are bumping the aws-cdk-lib?

jariy17 (approved):

AlexanderRichey (approved):
This looks solid. One thing I noticed is that we removed the `systemPrompt: file://system-prompt.md` syntax, which I liked since it makes it explicit where the system prompt content is coming from. Can we restore that in a follow up?

Also wanted so show some examples of this kind of commented file pattern:
- https://github.com/ohmyzsh/ohmyzsh/blob/master/templates/zshrc.zsh-template
- https://gist.github.com/nishantmodak/d08aae033775cb1a0f8a

PR #2299, opened 2026-09-14
feat(eval): add --model-provider to imperative evaluator create/update

jariy17:
## What

Adds `--model-provider` (Bedrock default / OpenResponses) to the imperative `agentcore eval evaluator llm-as-a-judge create` and `update` commands, and makes `update` provider-safe.

## Verification

- `bun test .../sharedFlags.test.tsx .../core/evalEvaluatorProvider.test.tsx` — 15 pass (provider resolution, arm construction, and the update-safety matrix: model-only Bedrock preserves tuning; instructions-only OpenResponses stays on the responses arm; provider switch requires a model; switch selects the correct arm without leaking the old tuning).
- Existing fixture-backed `evaluator.test.tsx` — 32 pass (Bedrock path unchanged).
- `bun run typecheck` + `bun run lint:check` clean; create/update `--help` list both providers.

## Follow-ups

- OpenResponses lifecycle service fixtures for `evaluator.test.tsx` require recording against an account with OpenResponses access (bug-bash item).
- CDK synthesis of `modelProvider: OpenResponses` is handled by `@aws/agentcore-cdk` in the generated app; deploy-time verification in the bug-bash notes.

Bug-bash recording against the explore account to follow in a comment.


jariy17:
## Bug bash — imperative evaluator commands (live, sandbox account, us-west-2)

Recorded with the TUI harness against a live AgentCore account. All cases verified end-to-end; the created evaluators were deleted afterward (no residue).

| # | Case | Result |
|---|------|--------|
| 1 | `create --model-provider OpenAI` (unsupported) | ❌ fails fast, before any AWS call: `invalid --model-provider "OpenAI": expected Bedrock or OpenResponses` |
| 2 | `create --model-provider OpenResponses --model openai.gpt-5.4` | ✅ ACTIVE; service stores `responsesEvaluatorModelConfig` with `maxOutputTokens: 4096`, `temperature: 0`, **no** `topP` |
| 3 | instructions-only `update` on the OpenResponses evaluator | ✅ stays on the responses arm — not silently converted to Bedrock, tuning preserved |
| 4 | `update --model-provider Bedrock` (no new model) | ❌ rejected: changing provider requires a new `--model` (model ids are not portable across provider APIs) |
| 5 | provider switch **with** a new `--model` | ✅ selects the target arm and does not leak the previous provider's tuning |

Baseline Bedrock create/get/update/delete unchanged and green.

A screen recording of this run is archived internally (Artifactory); it is not linked here because this repository is public.

notgitika:
Failing CI test

notgitika (commented):
The description mentioned this test file .../core/evalEvaluatorProvider.test.tsx which is not here was it meant to be included?

src/handlers/eval/evaluator/llm-as-a-judge/create/index.tsx
```diff
@@ -15,7 +20,12 @@ export const createLlmAsAJudgeCreateHandler = (core: Core, io: AppIO) =>
     flags: [
       flag("name", "the name of the evaluator", z.string().min(1)),
       flag("level", `evaluation level (${LEVELS.join(" | ")})`, z.enum(LEVELS)),
-      flag("model", "the Bedrock model ID used to judge", z.string().min(1)),
+      modelProviderFlag,
+      flag(
+        "model",
+        "judge model: a Bedrock model ID / ARN, or an OpenResponses model ID",
+        z.string().min(1),
```
notgitika, line 28: can we reuse the provider-specific model validation from `project add evaluator` here and in the `update` path? 

string min 1 means at this point " " would be valid. not a blocker imo

src/handlers/eval/evaluator/llm-as-a-judge/update/index.tsx
```diff
@@ -17,7 +17,12 @@ export const createLlmAsAJudgeUpdateHandler = (core: Core, io: AppIO) =>
         "evaluation instructions (inline, file://<path>, or - for stdin)",
         z.string().optional(),
       ),
-      flag("model", "the Bedrock model ID used to judge", z.string().optional()),
+      modelProviderFlag,
+      flag(
+        "model",
+        "judge model: a Bedrock model ID / ARN, or an OpenResponses model ID (required when changing provider)",
+        z.string().optional(),
```
nborges-aws: should we be validating the input here with `isValidBedrockModelId` / `isValidOpenResponsesModelId`?

nborges-aws (commented):
Plus one on Gitika's comment on validation. 

Also noticed the description mentions `evalEvaluatorProvider.test.tsx` under the testing header, but I don't see that file in the PR. Maybe it didn't get committed?

src/handlers/eval/evaluator/llm-as-a-judge/create/index.tsx
nborges-aws: +1

jariy17:
I forgot to update the PR descriptions, `../core/evalEvaluatorProvider.test.tsx` was proposely removed because the golden tests cover it.

nborges-aws (approved):

notgitika (approved):
Thanks this is great!

PR #2310, opened 2026-09-15
feat(remove): add a TUI for removing resources from project. 

Hweinstock:
### Problem 
Remove is missing a TUI. 

### Solution 
- add a TUI that allows users to see the resources in their project, and remove them. 
- add a remove all option to empty the project. 
- show parent resources in the resource explorer table to disambiguate. 

### Verification

https://github.com/user-attachments/assets/30e8d277-cd9d-4378-860d-ce4c4211dd1f


src/handlers/project/remove/screen.tsx
```diff
@@ -0,0 +1,376 @@
...
+  };
+}
+
+const RESOURCE_PICKER_CONFIGS: RemovableResourcePickerConfig[] = [
+  rootResourceTypePickerConfig("runtime", (spec) => spec.runtimes),
+  rootResourceTypePickerConfig("harness", (spec) => spec.harnesses),
+  rootResourceTypePickerConfig("memory", (spec) => spec.memories),
+  rootResourceTypePickerConfig("credential", (spec) => spec.credentials),
+  rootResourceTypePickerConfig("config-bundle", (spec) => spec.configBundles),
+  rootResourceTypePickerConfig("online-eval", (spec) => spec.onlineEvalConfigs),
```
nborges-aws: The spec for `onlineEvalConfigs` will return both evals and insights here, so insights will be listed as evals. Think we should have a separate resource here for insights? Insights configs have a `insights` field, which we can filter on to differentiate the two types

src/handlers/project/remove/screen.tsx
```diff
@@ -0,0 +1,376 @@
...
+  | "credential"
+  | "config-bundle"
+  | "online-eval"
+  | "evaluator"
+  | "gateway"
+  | "policy-engine"
+  | "payment-manager";
+
+type RemovableResourceType =
+  RootResourceType | "gateway-target" | "gateway-connector" | "policy" | "payment-connector";
```
notgitika: FYI runtime endpoint was also just merged in, could it be added in rebase by either this PR or a subsequent one?

src/handlers/project/remove/screen.tsx
```diff
@@ -0,0 +1,376 @@
...
+];
+
+const KEY_HINTS = [
+  { key: "↑↓/jk", label: "navigate" },
+  { key: "/", label: "filter" },
+  { key: "enter", label: "select" },
+  ...STATIC_KEY_HINTS,
+];
+
+function resourceTypeCounts(
```
notgitika, line 171: Could remove all check every collection that removeAllResources clears, rather than only the picker configs? 

a project containing only a knowledge base currently says it has no resources and can’t use remove-all. With a runtime + a kb, the confirmation lists only the runtime even though it deletes both. The same seems true for AB tests, datasets, mcp runtime tools, and unassigned targets.

src/handlers/project/remove/screen.tsx
```diff
@@ -0,0 +1,376 @@
...
+
+  return (
+    <Layout
+      breadcrumb={["agentcore", "project", "remove", config.resourceType]}
+      description={`choose a ${config.resourceType} to remove`}
+      keyHints={rows.length > 0 ? KEY_HINTS : STATIC_KEY_HINTS}
+    >
+      <DataTable
+        borderStyle="none"
+        showFooter={false}
```
notgitika, line 314: Was playing around with this with multiple resources and saw that it cut off after 10 + because there was no footer, there was no indication that there is a second page. Could we retain the footer or add paging hints here?

src/handlers/project/remove/screen.tsx
Hweinstock: ahh, good catch, forgot those were modeled as the same resource. I can split them up.

src/handlers/project/remove/screen.tsx
Hweinstock: > a project containing only a knowledge base currently says it has no resources and can’t use remove-all. With a runtime + a kb, the confirmation lists only the runtime even though it deletes both. The same seems true for AB tests, datasets, mcp runtime tools, and unassigned targets.

I see, the spec contains some dead fields which are silently removed on remove all. Let me tweak it so that those are left behind.

Hweinstock: Oh I see `removeAllResources` was actually built this way intentionally. In that case, we can show remove all when one exists. I think this behavior is kind of strange IMO, but don't want to expand the scope of this PR. We can revisit in the future if we see fit.

Hweinstock:
re-verified after the changes. 
https://github.com/user-attachments/assets/a4fd9916-4cfd-4858-9546-e9ebe590c885

nborges-aws (approved):
LGTM thanks for fixes!

Hweinstock:
Merging since earlier comments were addressed, can take additional comments as a follow-up.

PR #2324, opened 2026-09-17
fix(handlers): swap to required flags. 

Hweinstock:
### Problem 
The framework now supports required flags, but we haven't migrated existing use cases. This results in excess code, inconsistent error messages, and incorrect telemetry (commanderError instead of InputValidationError on missing args).

### Solution 
- switch to required flags. 
- update tests to use new message format. 
- create a shared util for asserting this new error format. 

### Verification 
CI passing


nborges-aws (dismissed):

src/handlers/gateway/invoke/index.tsx
```diff
@@ -27,7 +27,7 @@ export const createInvokeGatewayHandler = (
     name: "invoke",
     description: "invoke an AgentCore Gateway",
     flags: [
-      flag("id", "the ID of the Gateway", gatewayIdSchema.optional()),
+      flag("id", "the ID of the Gateway", gatewayIdSchema),
```
notgitika, line 30: for these "id" flags, is there a way we can preserve the exit code as USAGE error that was done below? otherwise, it will now exit with code 1. found the same with runtime invoke

notgitika (dismissed):

src/handlers/gateway/invoke/index.tsx
Hweinstock: I see, it looks like it was a little inconsistent before. It sounds like exit code 2 is for syntax issues, so perhaps we can swap all input validation errors over to use that exit code nd make it consistent. I can take this as a follow-up.

nborges-aws (approved):

notgitika (approved):

PR #2328, opened 2026-09-18
fix: show root version option in help

### LINKED ISSUE #2327: Root --version option is missing from help output
## Summary

`agentcore --version` works and root help prose mentions it, but `agentcore --help` does not list `--version` under Options.

## Actual behavior

The router handles the exact root `--version` invocation manually and appends version text to the help footer. Because the flag is not registered with Commander, it is absent from the generated option list.

## Expected behavior

Root help should declare the supported option, for example:

```text
-V, --version  display the installed AgentCore CLI version
```

The nested `agentcore harness version get --version <number>` option must continue to work.

## Suggested direction

Register the root version option with Commander and enable positional option handling so the root option does not consume the nested harness `--version` flag. Add regression tests for root help, root version output, and the nested harness version option.

notgitika:
This PR shows the supported root `-V, --version` option in `agentcore --help`.
Keeps the existing exact-root version handling so nested options such as `agentcore harness version get --version <value>` are not shadowed.
Replace the less discoverable help footer with the standard option entry.

Fixes #2327.

Testing successful


aidandaly24 (approved):
LGTM

PR #2329, opened 2026-09-18
fix(eval): support online-eval sources for batch insights

### LINKED ISSUE #2326: batch-insights --online-eval sends a conflicting default insight and treats IDs as ARNs
## Summary

The documented command:

```bash
agentcore eval batch-insights run \
  --name MyInsightsRun \
  --online-eval <online-evaluation-config-id>
```

fails because the CLI unconditionally supplies the default insight `Builtin.Insight.FailureAnalysis`. The service does not allow an explicit insight or evaluator configuration when the data source is an existing online evaluation configuration.

There is a second issue in the same path: help describes `--online-eval` as an ID, but the core request inserts its value directly into `onlineEvaluationConfigArn`.

## Actual behavior

- `--insight` has a client-side default even when `--online-eval` is selected.
- The default is sent in `evaluationConfig`, which conflicts with `onlineEvaluationConfigSource`.
- A bare online evaluation configuration ID is copied directly into an ARN field and fails service validation.

## Expected behavior

- For `--online-eval`, omit `evaluationConfig` unless the API explicitly supports it.
- Reject explicit `--insight` or `--evaluators` with a clear local validation error when `--online-eval` is used.
- Accept either an online evaluation configuration ID or ARN, as the help contract states, and resolve an ID to its ARN before starting the batch job.
- Continue applying `Builtin.Insight.FailureAnalysis` by default for compatible agent/raw data sources.

## Suggested tests

- `--online-eval <id>` sends neither insights nor evaluators and resolves the ID to an ARN.
- `--online-eval <arn>` uses the ARN directly.
- `--online-eval` combined with `--insight` or `--evaluators` fails locally.
- An agent source without `--insight` still defaults to `Builtin.Insight.FailureAnalysis`.

notgitika:
## Summary

- Do not send the default failure-analysis insight when `--online-eval` is the batch-insights source.
- Reject explicit `--insight` or `--evaluators` options for an online-evaluation source instead of submitting a conflicting request.
- Resolve an online evaluation config ID to its ARN while continuing to accept a supplied ARN directly.
- Add handler and client tests for the online-evaluation paths.

Fixes #2326.

## Testing

- `bun test src/handlers/eval/batch-insights/batch-insights.test.tsx src/core/batchInsights.test.tsx`
- `bun test` (3,379 passed, 0 failed)
- `bun run typecheck`
- `bun run lint:check`
- `bun run format:check`
- `bun run build`


tejaskash (approved):
Reviewed: online-evaluation sources omit conflicting analysis configuration, IDs are resolved to ARNs, and the focused tests and type checking pass.

src/handlers/eval/batch-insights/run/index.tsx
```diff
@@ -44,11 +44,20 @@ export const createRunBatchInsightsHandler = (core: Core, io: AppIO) =>
...
+          "--insight and --evaluators cannot be used with --online-eval",
+        );
+      }
       const response = await core.eval.startBatchInsights(
         {
           name: flags["name"],
           description: flags["description"],
-          insightIds: flags["insight"],
+          insightIds:
+            source.origin === "online-eval" ? undefined : (flags["insight"] ?? [DEFAULT_INSIGHT]),
```
tejaskash, line 60: Could we check that the online eval config has insights before starting the job? Regular online-eval configs can be evaluator-only, which creates a job that won’t show up in get or list. This should also work when the config is passed as an ARN.

aidandaly24 (approved):

src/handlers/eval/batch-insights/run/index.tsx
notgitika: will be taking up in a follow up

PR #2330, opened 2026-09-18
feat(project): pass the deployment target to the vended CDK app and budget it in derived names

tejaskash:
## What

- The vended `bin/cdk.ts` passes `target: stack.target` to `transformAgentCoreJson`.
- `project create` shortens the default harness name so `<project>_default_<harness>` fits 40 characters.
- `project add policy-engine` and `project add gateway` check the deployed name against the service cap at add time, through one shared function, budgeting for the longest declared target (`default` when none is declared).
- Target names are letters and digits only and unique ignoring case, matching the library, so no two targets compose the same deployed name or the same lowercased ECR path.
- Region list gains `ap-south-2` and `us-west-1`, matching the library.
- Vended `package.json` pins `aws-cdk-lib ~2.269.0`, the library's peer range. Installs with the published alpha.53 and with the branch tarball.

## Why

aws/agentcore-l3-cdk-constructs#386 makes every resource name `<project>_<target>_<name>` so two targets in one account and region stop colliding. The CLI has to hand the target to the library and account for the extra segment where it derives names.

The `@aws/agentcore-cdk` pin is not bumped here. Owner coordinates the release and the pin.

## How tested

- `bun test`: 3374 tests pass.
- Live with this branch built and the library branch packed into the project: two targets in one account and region for a harness project and a runtime plus memory project. Deploy, status, invoke and teardown all work per target.
- Build only: 23-character project name scaffolds a shortened harness name that fits exactly 40. `add policy-engine` accepts a name that composes to 48 and rejects 49.


src/handlers/project/create/index.ts
```diff
@@ -187,17 +184,14 @@ export function resolveScaffoldHarnessInput(flags: HarnessPathFlagValues): Scaff
...
-// name is the project name. If the project name is 19 characters or less, the
-// joined name fits. If the project name is longer, this function shortens the
-// harness name so that the joined name is 40 characters. The shortened name
-// ends with a 5-character hash of the project name.
+/**
+ The deployed `<project>_default_<harness>` must fit CloudFormation's 40-character HarnessName cap, so a project name over 15 characters gets a truncated harness name ending in a 5-character hash.
+**/
 function defaultHarnessNameFor(projectName: string): string {
-  if (projectName.length <= 19) return projectName;
+  const budget = 40 - `_${DEFAULT_TARGET_NAME}_`.length;
```
notgitika, line 191: Would it make sense to account for named targets here, rather than only default? 
For eg, I could have a  16 char project name, the generated harness fits with default, but a valid target like "production" pushes the physical name to 43 chars and synth fails. 

I'm not sure what the best way to handle it is tbh

tejaskash: Agreed, budgeting for `default` only was a hole. The 40 cap for harness, 23 for project and 64 for target mean no add-time check can cover every legal target, so the check is a best-effort budget and synth stays the final gate (its message names the project, target and resource name as the three levers). In 7a128c4c the add-time pre-checks now budget for the longest declared target in `aws-targets.json`, falling back to `default` when none is declared, which catches your example whenever the target exists before the resource is added. `project create` still uses `default` because no target exists yet. The complete fix is raising the harness cap, which I have as a question for the harness team.

nborges-aws (approved):

PR #2332, opened 2026-09-18
feat(project): scope Identity credential providers by deployment target

tejaskash:
#2330 is merged. Merge after aws/agentcore-l3-cdk-constructs#391 is published and pinned, or agents receive the spec name and `requires_api_key` fails.

## What

- Every credential provisions an Identity provider named `<project>_<target>_<credential>`. Spec names, `.env.local` keys and the deployed-state key are unchanged.
- Teardown deletes every provider kind the target owns (API key, OAuth, payment). A credential removed from the spec has its provider deleted on the next deploy of each target.
- Scaffolded and exported agents read the provider name from `AGENTCORE_CREDENTIAL_<NAME>_NAME` with the spec name as fallback.
- A credential name ending in `-name` or `_name` is rejected, it would shadow that variable.
- A composed name over the 128 cap fails before the first Identity call.

## Why

Two targets in one account and region shared one provider named after the spec credential. The second deploy overwrote the first target's secret, and teardown of one target could delete a payment provider the other still used. Old bare providers from earlier versions are left in place, nothing deletes them.

## How tested

- `bun test`: 3377 tests pass.
- Live, one account and region, runtime plus memory project with an API key credential, targets `dev` and `staging`: two providers `TgtCredB_dev_openai` and `TgtCredB_staging_openai` with distinct ARNs, each recorded under the spec key `openai` for its target, each runtime's `AGENTCORE_CREDENTIAL_OPENAI_NAME` set to its own provider. Rotating the secret and redeploying `dev` updated only the `dev` provider's secret. Teardown removed both stacks and both providers.


src/core/project/backends/cdk.ts
```diff
@@ -367,9 +367,10 @@ export class CdkBackend implements ProjectBackend {
     yield { type: "step", message: `Removing stack ${artifact.stackName}` };
     yield* this.runCdk({ kind: "destroy", stackArtifactId: artifact.id }, options);
     // After the stack, since a resource in it may still be using the provider.
-    yield* this.removePaymentCredentials(project, {
+    yield* this.removeCredentials(project, {
```
nborges-aws, line 383: If we only run removeCredentials for a full project teardown, what happens if the user just removes the credential from spec? Does the provider end up orphaned?

nborges-aws (dismissed):

src/core/project/backends/cdk.ts
tejaskash: Yes, it did, and that was pre-existing: deploy overwrote the target's recorded credentials with the declared set and the provider stayed in Identity, forgotten. Before this PR that was on purpose, a bare name might belong to someone else. Now the name is ours, so e721cad3 removes it: deploy computes the providers recorded for the target that the spec no longer declares and deletes them after the stack update (same ordering reason as teardown, a stack resource may reference the provider until the update drops it). Teardown passes recorded plus declared through the same remover. Verified live: add credential, deploy, `project remove credential`, deploy again logs `Removing credential provider 'TgtOrphA_default_openai'` and `list-api-key-credential-providers` no longer lists it while the stack still exists.

nborges-aws (approved):

PR #2333, opened 2026-09-18
feat(project): pick the deployment target in the TUI, pin its region, pre-check every deployed name

tejaskash:
#2330 is merged.

## What

- One `DeploymentTargetPicker` component, used by the deploy, status and invoke screens. Zero or one declared target resolves without asking, several ask which. The choice lives in the route query string so escape from a detail page returns to the chosen target.
- The TUI pins the chosen target's region: `Root` renders every route with that region in the context, so detail pages and whatever they open fetch where the target deployed. `useCoreOpts` and the `?region=` links are gone. A pin lasts as long as the screen that set it is in the history: popping below it, or reaching a command menu, restores the launch region.
- Headless `project status` no longer refuses when `AWS_REGION` differs from the target's region.
- Every `project add` handler that names a physical resource checks the deployed name against the service cap at add time (harness 40, runtime, memory, evaluator, online-eval 48, config-bundle 100).

## Why

Status and invoke were hard-coded to `default`, so a named target could not be inspected or invoked from the TUI. The region guard existed because detail pages opened from status ran their follow-up actions in the launch region. Pinning the region removes the cause. The pre-checks fail at add time instead of at synth, minutes later.

## How tested

- `bun test`: 3384 tests pass.
- Live, harness project with `dev` and `staging` in us-west-2, `AWS_REGION=us-east-1` throughout: headless `status --target staging --json` succeeds and reports `us-west-2`. TUI status shows the target table, choosing `staging` lists the deployed harness and opens its detail page fetched in us-west-2. TUI invoke shows the picker, then the resource picker for `dev`, and the harness answers `pong`. `add harness` with a 24-character name is rejected at 41 characters and accepted at 23, `add runtime` is rejected at 49. Both stacks torn down.


src/components/Root.tsx
```diff
@@ -141,7 +143,26 @@ export interface RootProps {
 
 // Root is the top of the Ink React tree, rendered by the `agentcore` default
 // handler when the CLI is invoked without a subcommand.
-export function Root({ path, ctx, core, queryClient }: RootProps) {
+export function Root({ ctx, ...props }: RootProps) {
+  // A screen may pin the region the resource it shows lives in (see
+  // usePinRegion). The pin replaces the launch region on every route until a
+  // screen pins another.
```
nborges-aws: What if the user navigates back? The region would remained pinned to what was set here and wouldn't be restored, which could cause issues for resource fetching? I would expect some pin restoration here

tejaskash: Agreed, the sticky pin was the weak point. Fixed in 294c2a48. A pin now belongs to the history entry that set it: Root tracks the router depth (PUSH increments, POP decrements, REPLACE keeps) and drops every pin recorded above the depth a POP lands on, so escaping from a detail page back to status keeps the target's region and popping below the pinning screen restores the launch region. One more rule was needed because our screens escape to a parent by pushing its path rather than popping: a command menu drops every pin, since a menu is never about a resource. Two tests cover both paths. Verified in the TUI: status on a us-west-2 target with AWS_REGION=us-east-1, open the harness detail (loads from us-west-2), escape out to the root menu, then harness list shows the us-east-1 harnesses.

nborges-aws (approved):

PR #2336, opened 2026-09-18
ci(release): publish through the workflow name and environment npm trusts

tejaskash:
## What

- `release-publish.yml` becomes `release-main-and-preview.yml`, contents unchanged, and its publish job runs under the `npm-publish` environment.
- `release-prepare.yml` pins the vended `@aws/agentcore-cdk` to the library's `preview` dist-tag for `rc` releases and `latest` for stable.
- Workflow README updated.

## Why

The npm trusted publisher for `@aws/agentcore` is registered for `aws/agentcore-cli`, workflow `release-main-and-preview.yml`, environment `npm-publish`, and the account that could change it is locked. The rc.2 publish from `release-publish.yml` failed with `404 Not Found - PUT .../@aws%2fagentcore`, npm's answer when the OIDC claims do not match. Same filename on both branches is fine: each branch runs its own copy on its own pushes. The environment has required reviewers, so every rc publish now waits for one approval.

The library publishes from `refactor` under `preview`, so pinning to `latest` would re-pin the vended app to alpha.53, which lacks the exports the vended `bin/cdk.ts` imports.

## How tested

- Both workflow files parse. Prettier clean.
- Not testable before merge: the OIDC exchange only happens on a real release. The first rc through this file is the proof. If it 404s again, the remaining variable is the account itself.


.github/workflows/release-main-and-preview.yml
```diff
@@ -48,6 +50,9 @@ jobs:
     needs: verify
     # npm provenance is only issued from GitHub-hosted runners.
     runs-on: aws-release-4-core
+    environment:
```
Hweinstock, line 53: wait is this part necessary?

Hweinstock (approved):

nborges-aws (approved):

.github/workflows/release-main-and-preview.yml
tejaskash: Yeah the trusted publisher config has the env name in it

PR #2339, opened 2026-09-18
feat(e2e): add basic e2e test suite for runtime templates

Hweinstock:
### Problem 
We are missing integration tests. 

### Solution 
- add tests that trigger on push events only (we can add scoped down version to PRs soon). 
- tests leverage vitest (not bun build in) because it gives some nice features like tagging. (which we can leverage for canary). 
- port over stack clean up logic. 

Some patterns this PR attempts to establish:
- tests should read as a sequence of CLI commands. 
- configuration should be separated from the the execution, such that making configuration changes is as easy as possible. 
- parametrize tests when possible to make it easy to re-use logic for different variants. 
- tests should be concurrent when possible, otherwise explicitly serial. 

### Verification 
- Running e2e locally using instructions (note I had to patch my CDK due to breaking changes). 
- there are two other bugs/gaps for follow-up: dev race condition and local invoke support for other protocols.


.github/workflows/ci.yml
```diff
@@ -21,3 +21,13 @@ jobs:
       ref: ${{ github.event.pull_request.head.sha || github.sha }}
     secrets:
       CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+
+  e2e:
+    if: github.event_name == 'push'
```
Hweinstock, line 26: we skip for now for simplicity, but eventually I think we should support a separate tag to run a reduced set.

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,121 @@
+name: e2e-test
```
Hweinstock, line 1: separating this out into its own workflow with `workflow_call` means the canary will be a thin wrapper.

.github/workflows/ci.yml
notgitika: that is a good idea! when should that suite run and what would it contain?

notgitika: are you thinking of the idea of a tag that can "mark" certain suites as canary?

e2eTest/project/templates.test.ts
```diff
@@ -0,0 +1,386 @@
+import { afterAll, beforeAll, describe, expect, test } from "vitest";
```
jariy17, line 1: should we name the test folder to e2e-tests? It would be clear for agents to only add e2e tests in here.

e2eTest/project/templates.test.ts
```diff
@@ -0,0 +1,386 @@
...
+  };
+
+  if (runtime.protocol === "MCP") {
+    const response = parseResult(McpResponseSchema, result);
+    expect(response.id).toBe(runtime.payload.id);
+    return;
+  }
+
+  if (runtime.protocol === "A2A") {
+    const response = parseResult(A2aResponseSchema, result);
```
jariy17, line 376: Can you verify we got a 2xx response for A2A and http?

.github/workflows/ci.yml
Hweinstock: > are you thinking of the idea of a tag that can "mark" certain suites as canary?

yes exactly!

> that is a good idea! when should that suite run and what would it contain?

I have some ideas here. 
- One is that we should run the basic tests across functionality (effectively smoke test each feature)
- another is that we could maybe check the filepaths touched to guess some tags that should run. 
- another is that a reviewer or harness reviewer could manually/automatically ask for certain tags to be run. 

I'll likely start with the first one, and we can explore the others as we get more time. Overall, I want to keep the PR e2e tests short for faster feedback, and reserve all of them for the push event.

jariy17 (commented):
LGTM, just small nits.

.github/workflows/ci.yml
notgitika: 2 and 3 would be killer!

e2eTest/project/templates.test.ts
Hweinstock: sure, I can rename to `e2eTest` to match the camel case convention. I'm used to the `test/` convention from other languages, but not sure if this is a pattern in TS.

e2eTest/project/templates.test.ts
Hweinstock: I don't think we get a direct status code on non-http protocol runtimes, but I can add a check that there is not an error in the response.

jariy17 (approved):

